PR #18, opened 2024-09-11
Fix: remove authToken to build the ts-lib/linea-native-lib node package

VGau:
- Removed authToken to build the `ts-lib/linea-native-lib`
- Removed github access token env variable from CI workflows

### Checklist


ts-libs/linea-native-libs/src/scripts/build.ts
```diff
@@ -38,15 +38,15 @@ async function getReleaseAssetUrl(authToken: string, nativeLibReleaseTag: string
   }
 
   const asset = release.assets.find((asset: any) => asset.name.includes(nativeLibReleaseTag));
-  return `https://${authToken}:@api.github.com/repos/Consensys/zkevm-monorepo/releases/assets/${asset.id}`;
+  return `https://api.github.com/repos/Consensys/linea-monorepo/releases/assets/${asset.id}`;
```
jpnovais, line 41: Does this work? My understanding is that we need GITHUB token to access https://api.github.com/... releases/assets even if the repo is open source.

jpnovais (approved):

ts-libs/linea-native-libs/src/scripts/build.ts
VGau: yes it works without the github token :)

PR #31, opened 2024-09-12
Decompressor Library

Tabaie:
This PR implements creates a decompressor library capable of managing different blob versions and dictionaries. Its output consists of an RLP encoded list of blocks. Due to some information loss during compression, two pieces of information are represented "hackily":

* The block hash is in the `ParentHash` field, not the hash field itself, as `go-ethereum` computes that field itself, not allowing users to set it manually.
* The transaction `from` address is in the `signature.R` field.

**NOTE** Only the Go side has been tested. Need contribution from a Kotlin developer to make this merge-ready.

### Checklist


jpnovais:
Hi @Tabaie I made a release of the decompressor lib from this branch and created a basic test but it fails with error. 

Maybe I am initializing the decompressor wrongly. Please take a take a look at the test. You can run them with 

```
./gradlew jvm-libs:blob-decompressor:test
```

```
Error looking up function 'Error': dlsym(0x8b047c60, Error): symbol not found
java.lang.UnsatisfiedLinkError: Error looking up function 'Error': dlsym(0x8b047c60, Error): symbol not found
	at com.sun.jna.Function.<init>(Function.java:255)
```

Tabaie:
The Kotlin side is probably all wrong. I was hoping to hammer that part out together with yourself or @gauravahuja.

prover/lib/compressor/blob/blob.go
```diff
@@ -57,3 +48,34 @@ func GetDict() ([]byte, error) {
...
+	default:
+		return nil, errors.New("unrecognized blob version")
+	}
+
+	if err != nil {
+		return nil, err
+	}
+	blockObjs := make([]encode.DecodedBlockData, len(blocks))
+	for i, block := range blocks {
+		blockObjs[i], err = blockDecoder(bytes.NewReader(block))
```
gauravahuja: Should you convert the decoded block to the standard representation from go-ethereum types.Block
```suggestion
		decodedBlock, err := blockDecoder(bytes.NewReader(block))
		blockObjs[i] = decodedBlock.ToStd()
```

Tabaie:
Closed in favor of https://github.com/Consensys/linea-monorepo/pull/229 due to merge conflicts

PR #53, opened 2024-09-19
Cancel in progress workflows on new commits

bradbown:
This PR implements issue(s) #

### Checklist


.github/workflows/coordinator-build-and-publish.yml
```diff
@@ -27,6 +27,10 @@ on:
       DOCKERHUB_TOKEN:
         required: true
 
+concurrency:
+  group: coordinator-build-and-publish-${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
Filter94: For main though I'm not sure if we want to cancel them. I'd rather run everything that is supposed to be running for every single commit in case things are broken by some commit. It would be easier to identify the commit that breaks things

bradbown: I wasn't sure which we'd want either. I'll remove the if/else.

jpnovais (approved):

PR #80, opened 2024-09-23
Prover: fix the txnrlp duplicated txnrlp bug

AlexandreBelling:
This PR implements a fix for an outstanding integration issue between the prover and txnrlp. The issue is that the prover infers a transaction signature via a mapping between tx-hash and tx-signature but it is possible that two transactions have the same transaction hash (as in the hash for signing). In that case, the current approach will not work. This PR addresses the issue by mapping the signature by tx-id. 

The tx-hash is still used to sanity-check that the signature that is infered relates to the expected transaction. Thus, helping in the debugging process if any bugs stands out here.


prover/zkevm/witness.go
```diff
@@ -16,21 +17,33 @@ type Witness struct {
...
 
-	if !found {
-		return nil, nil, nil, fmt.Errorf("could not find signature for tx hash = 0x%x", txHash)
+	if i > len(w.TxHashes) {
+		return nil, nil, nil, fmt.Errorf("requested txID outgoes the total number of transactions we found in the conflation")
 	}
 
+	if utils.HexEncodeToString(txHash) != utils.HexEncodeToString(w.TxHashes[i][:]) {
+		return nil, nil, nil, fmt.Errorf(
+			"requested txID=%v while txnrlp expects it to have it for txhash=%v but the blocks transaction has hash=%v",
```
gusiri, line 40: There are some existing terms in the context of a `block` (not conflation).
`transactionIndex`: When a transaction is added to a block, it is assigned a numbered position within that block. The transactionIndex returns the position of the transaction in the block it has been added to.
(https://docs.alchemy.com/docs/understanding-the-transaction-object-on-ethereum#transactionindex)

It is referred to as `position in block` on Etherscan.

Here `txID` is something similar but within a conflation, correct?

Soleimani193 (approved):

arijitdutta67 (approved):

gusiri (approved):

prover/zkevm/witness.go
AlexandreBelling: Yes correct, that would be position in conflation

gusiri (approved):

PR #97, opened 2024-09-25
Coordinator metrics - blob and finalization submissions

bradbown:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/coordinator/app/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/MaxLongTracker.kt
+++ b/coordinator/app/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/MaxLongTracker.kt
@@ -43,7 +43,7 @@ class HighestProvenBlobTracker(initialProvenBlockNumber: ULong) :
   }
 }
 
-class HighestAggregationTracker(initialProvenBlockNumber: ULong) :
+open class HighestULongTracker(initialProvenBlockNumber: ULong) :
   MaxLongTracker<ULong>(initialProvenBlockNumber.toLong()) {
   override fun convertToLong(trackable: ULong): Long {
     return trackable.toLong()
--- a/coordinator/app/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/SubmittedEventTracker.kt
+++ b/coordinator/app/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/SubmittedEventTracker.kt
@@ -1,0 +1,36 @@
+package net.consensys.zkevm.ethereum.coordination
+
+import net.consensys.zkevm.domain.BlobSubmittedEvent
+import net.consensys.zkevm.domain.FinalizationSubmittedEvent
+import tech.pegasys.teku.infrastructure.async.SafeFuture
+import java.util.concurrent.atomic.AtomicReference
+import java.util.function.Supplier
+
+abstract class SubmittedEventTracker<T>(submission: T?) : (T?) -> SafeFuture<*>, Supplier<T?> {
+  private val submissionCache = AtomicReference(submission)
+
+  override fun invoke(trackable: T?): SafeFuture<*> {
+    submissionCache.set(trackable)
+    return SafeFuture.completedFuture(Unit)
+  }
+
+  override fun get(): T? {
+    return submissionCache.get()
+  }
+}
+
+class BlobSubmittedEventTracker(
+  blobSubmittedEvent: BlobSubmittedEvent? = null
+) : SubmittedEventTracker<BlobSubmittedEvent>(blobSubmittedEvent) {
+  fun getEndBlockNumber(): ULong {
+    return get()?.blobs?.last()?.endBlockNumber ?: 0UL
+  }
+}
+
+class FinalizationSubmittedEventTracker(
+  finalizationSubmittedEvent: FinalizationSubmittedEvent? = null
+) : SubmittedEventTracker<FinalizationSubmittedEvent>(finalizationSubmittedEvent) {
+  fun getEndBlockNumber(): ULong {
+    return get()?.aggregationProof?.finalBlockNumber?.toULong() ?: 0UL
+  }
+}
--- a/coordinator/core/build.gradle
+++ b/coordinator/core/build.gradle
@@ -36,6 +36,7 @@ dependencies {
   testImplementation(testFixtures(project(':jvm-libs:traces')))
   testImplementation(testFixtures(project(':jvm-libs:kotlin-extensions')))
   testImplementation("io.vertx:vertx-junit5")
+  testImplementation "org.apache.logging.log4j:log4j-core-test:${libs.versions.log4j.get()}"
 }
 
 test {
--- a/coordinator/core/src/main/kotlin/net/consensys/zkevm/domain/Aggregation.kt
+++ b/coordinator/core/src/main/kotlin/net/consensys/zkevm/domain/Aggregation.kt
@@ -132,3 +132,43 @@ data class Aggregation(
     Proving
   }
 }
+
+data class FinalizationSubmittedEvent(
+  val aggregationProof: ProofToFinalize,
+  val parentShnarf: ByteArray,
+  val parentL1RollingHash: ByteArray,
+  val parentL1RollingHashMessageNumber: Long,
+  val submissionTimestamp: Instant,
+  val transactionHash: String
+) : BlockInterval by aggregationProof {
+
+  override fun equals(other: Any?): Boolean {
+    if (this === other) return true
+    if (javaClass != other?.javaClass) return false
+
+    other as FinalizationSubmittedEvent
+
+    if (aggregationProof != other.aggregationProof) return false
+    if (!parentShnarf.contentEquals(other.parentShnarf)) return false
+    if (!parentL1RollingHash.contentEquals(other.parentL1RollingHash)) return false
+    if (parentL1RollingHashMessageNumber != other.parentL1RollingHashMessageNumber) return false
+    if (submissionTimestamp != other.submissionTimestamp) return false
+    if (transactionHash != other.transactionHash) return false
+
+    return true
+  }
+
+  override fun hashCode(): Int {
+    var result = aggregationProof.hashCode()
+    result = 31 * result + parentShnarf.contentHashCode()
+    result = 31 * result + parentL1RollingHash.contentHashCode()
+    result = 31 * result + parentL1RollingHashMessageNumber.hashCode()
+    result = 31 * result + submissionTimestamp.hashCode()
+    result = 31 * result + transactionHash.hashCode()
+    return result
+  }
+
+  fun getSubmissionDelay(): Long {
+    return submissionTimestamp.minus(aggregationProof.finalTimestamp).inWholeSeconds
+  }
+}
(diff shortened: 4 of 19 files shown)


jpnovais:
@bradbown sorry for the conflicts, can you please merge main?

bradbown:
> @bradbown sorry for the conflicts, can you please merge main?

@jpnovais All merged

jvm-libs/metrics/micrometer/src/test/kotlin/net/consensys/linea/metrics/micrometer/MicrometerMetricsFacadeTest.kt
```diff
@@ -65,4 +66,35 @@ class MicrometerMetricsFacadeTest {
...
+    val histogram = metricsFacade.createHistogram(
+      category = LineaMetricsCategory.BATCH,
+      name = "some.metric",
+      description = "This is a test metric",
+      tags = expectedTags,
+      baseUnit = "seconds"
+    )
+    val createdHistogram = meterRegistry.find("linea.test.batch.some.metric").summary()
+    assertThat(createdHistogram!!.count()).isEqualTo(0L)
+    assertThat(createdHistogram).isNotNull
```
jpnovais: I believe these lines are swapped.

jvm-libs/metrics/micrometer/src/main/kotlin/net/consensys/linea/metrics/micrometer/MicrometerMetricsFacade.kt
```diff
@@ -16,6 +18,14 @@ class MicrometerMetricsFacade(private val registry: MeterRegistry, private val m
         "$name must adhere to Micrometer naming convention!"
       }
     }
+    fun requireValidBaseUnit(baseUnit: String) {
+      val validBaseUnits = listOf(
+        "seconds",
+        "minutes",
+        "hours"
+      )
```
jpnovais: validBaseUnits is a static field, so I recommend to make it a companion instance.

coordinator/core/src/main/kotlin/net/consensys/zkevm/domain/Blob.kt
```diff
@@ -163,3 +163,39 @@ enum class BlobStatus {
   COMPRESSION_PROVING,
   COMPRESSION_PROVEN
 }
+
+data class BlobSubmittedEvent(
+  val blobs: List<BlockInterval>,
+  val endBlockTime: Instant,
+  val lastShnarf: ByteArray,
+  val submissionTimestamp: Instant,
+  val transactionHash: String
```
jpnovais: there is an inconsistency here: `transactionHash: String` `lastShnarf: ByteArray` even though they both are kind of hasshes. I suggested to make them all ByteArray.

coordinator/core/src/main/kotlin/net/consensys/zkevm/domain/Aggregation.kt
```diff
@@ -132,3 +132,43 @@ data class Aggregation(
...
   }
 }
+
+data class FinalizationSubmittedEvent(
+  val aggregationProof: ProofToFinalize,
+  val parentShnarf: ByteArray,
+  val parentL1RollingHash: ByteArray,
+  val parentL1RollingHashMessageNumber: Long,
+  val submissionTimestamp: Instant,
+  val transactionHash: String
```
jpnovais: same inconsistency here

coordinator/ethereum/blob-submitter/src/main/kotlin/net/consensys/zkevm/ethereum/finalization/AggregationFinalizationCoordinator.kt
```diff
@@ -219,7 +220,12 @@ class AggregationFinalizationCoordinator(
         lineaRollup = lineaRollup,
         aggregationsRepository = aggregationsRepository,
         blobsRepository = blobsRepository,
-        aggregationSubmitter = AggregationSubmitterImpl(lineaRollup, gasPriceCapProvider),
+        aggregationSubmitter = AggregationSubmitterImpl(
```
Filter94: Would it be better if we inject `AggregationSubmitter` into `AggregationFinalizationCoordinator`? 

`aggregationSubmittedEventConsumer` isn't used by `AggregationFinalizationCoordinator` at all, it's just used to create `AggregationSubmitterImpl` and this creates a redundant dependency

coordinator/app/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/MaxLongTracker.kt
```diff
@@ -43,7 +43,7 @@ class HighestProvenBlobTracker(initialProvenBlockNumber: ULong) :
   }
 }
 
-class HighestAggregationTracker(initialProvenBlockNumber: ULong) :
+open class HighestULongTracker(initialProvenBlockNumber: ULong) :
```
Filter94: It seems we still may leave it closed

coordinator/app/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/SubmittedEventTracker.kt
```diff
@@ -0,0 +1,36 @@
...
+
+class BlobSubmittedEventTracker(
+  blobSubmittedEvent: BlobSubmittedEvent? = null
+) : SubmittedEventTracker<BlobSubmittedEvent>(blobSubmittedEvent) {
+  fun getEndBlockNumber(): ULong {
+    return get()?.blobs?.last()?.endBlockNumber ?: 0UL
+  }
+}
+
+class FinalizationSubmittedEventTracker(
```
Filter94: I believe it's isomorphic to `HighestULongTracker` from Gauge's perspective. What I mean by this is that this class can compose an instance of `HighestULongTracker` and delegate `HighestULongTracker`'s functionality to its instance rather than defining new `getEndBlockNumber` method that doesn't belong to any interface, so we have to call it in a lambda 

Same applies to `BlobSubmittedEventTracker`

coordinator/app/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/SubmittedEventTracker.kt
```diff
@@ -0,0 +1,36 @@
+package net.consensys.zkevm.ethereum.coordination
+
+import net.consensys.zkevm.domain.BlobSubmittedEvent
+import net.consensys.zkevm.domain.FinalizationSubmittedEvent
+import tech.pegasys.teku.infrastructure.async.SafeFuture
+import java.util.concurrent.atomic.AtomicReference
+import java.util.function.Supplier
+
+abstract class SubmittedEventTracker<T>(submission: T?) : (T?) -> SafeFuture<*>, Supplier<T?> {
```
Filter94: I think we don't leverage `, Supplier<T?>` part anywhere

coordinator/app/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/SubmittedEventTracker.kt
Filter94: Or to take a step back. `HighestConflationTracker`, `HighestProvenBlobTracker` and so on seem to be solving the same problem. Why don't we add more implementations for new required mappings to `MaxLongTracker.kt` file?

coordinator/app/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/SubmittedEventTracker.kt
Filter94: And I don't see where we actually access the whole stored submission. In the two implementations we have, we only access the ULong, it means we don't need to store the whole T. And this means that existing `MaxLongTracker` should be sufficient

jvm-libs/metrics/micrometer/src/main/kotlin/net/consensys/linea/metrics/micrometer/MicrometerHistogramAdapter.kt
```diff
@@ -0,0 +1,26 @@
+package net.consensys.linea.metrics.micrometer
+
+import io.micrometer.core.instrument.DistributionSummary
+import net.consensys.linea.metrics.Histogram
+
+class MicrometerHistogramAdapter(private val adapter: DistributionSummary) : Histogram {
+  override fun record(var1: Double) {
```
Filter94: var1 is not the best name

jvm-libs/metrics/micrometer/src/main/kotlin/net/consensys/linea/metrics/micrometer/MicrometerHistogramAdapter.kt
```diff
@@ -0,0 +1,26 @@
+package net.consensys.linea.metrics.micrometer
+
+import io.micrometer.core.instrument.DistributionSummary
+import net.consensys.linea.metrics.Histogram
+
+class MicrometerHistogramAdapter(private val adapter: DistributionSummary) : Histogram {
```
Filter94: `record` is the only useful method here. Linea code shouldn't care about the rest, because it's taken care of by Micrometer. Look at the `Counter` for example. It has only methods to write data, not to read it. Because reading it is not a responsibility of the code we write

coordinator/core/build.gradle
```diff
@@ -36,6 +36,7 @@ dependencies {
   testImplementation(testFixtures(project(':jvm-libs:traces')))
   testImplementation(testFixtures(project(':jvm-libs:kotlin-extensions')))
   testImplementation("io.vertx:vertx-junit5")
+  testImplementation "org.apache.logging.log4j:log4j-core-test:${libs.versions.log4j.get()}"
```
Filter94: Why do we need log4j tests here?

Filter94 (approved):

jvm-libs/metrics/micrometer/src/main/kotlin/net/consensys/linea/metrics/micrometer/MicrometerMetricsFacade.kt
```diff
@@ -63,6 +74,28 @@ class MicrometerMetricsFacade(private val registry: MeterRegistry, private val m
...
+  override fun createHistogram(
+    category: LineaMetricsCategory,
+    name: String,
+    description: String,
+    tags: List<Tag>,
+    baseUnit: String
+  ): Histogram {
+    requireValidMicrometerName(category.toString())
+    requireValidMicrometerName(name)
+    requireValidBaseUnit(baseUnit)
```
Filter94: But where `baseUnit` is used?

coordinator/core/build.gradle
bradbown: I've removed this, I added this because I expected to require it for a unit test with a `ListAppender` but didn't use that in the end.

jvm-libs/metrics/micrometer/src/main/kotlin/net/consensys/linea/metrics/micrometer/MicrometerMetricsFacade.kt
bradbown: I'd missed it, the builders `baseUnit` is now being se.t

Filter94: Can you please adjust the test as well so we don't miss it?

bradbown: Added checks for the description and base unit

Filter94 (approved):

Filter94 (approved):

PR #98, opened 2024-09-25
Feat/94 emit version updated event

thedarkjester:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/contracts/contracts/LineaRollup.sol
+++ b/contracts/contracts/LineaRollup.sol
@@ -24,8 +24,7 @@ contract LineaRollup is
 {
   using Utils for *;
 
-  /// @dev These are the ABI versions and not the reinitialize version.
-  string public constant PREVIOUS_CONTRACT_VERSION = "5.0";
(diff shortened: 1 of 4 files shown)


contracts/contracts/LineaRollup.sol
```diff
@@ -24,7 +24,8 @@ contract LineaRollup is
 {
   using Utils for *;
 
-  /// @dev This is the ABI version and not the reinitialize version.
+  /// @dev These are the ABI versions and not the reinitialize version.
+  string public constant PREVIOUS_CONTRACT_VERSION = "5.0";
```
jpnovais: I believe this extra storage space does not deliver much value because in case we need it we can fetch the latest event and find out the previous version,

thedarkjester: it is a constant - no extra storage used

thedarkjester: removed it anyway

VGau (approved):

count-sum (approved):

PR #100, opened 2024-09-25
Coordinator: DRY refactor test helpers and data for future reuse

jpnovais:
This PR implements issue(s) #

### Checklist


jvm-libs/testing/l1-blob-and-proof-submission/build.gradle
```diff
@@ -0,0 +1,14 @@
+import org.gradle.api.tasks.testing.logging.TestExceptionFormat
+import org.gradle.api.tasks.testing.logging.TestLogEvent
+
+plugins {
+  id 'net.consensys.zkevm.kotlin-library-conventions'
+}
+
+dependencies {
+  api(project(':jvm-libs:core:domain-models'))
```
jonesho, line 6: why not just `implemenetation` here?

jpnovais: Because the client will need access to it's domain models.

bradbown (approved):

jvm-libs/testing/l1-blob-and-proof-submission/build.gradle
jonesho: Okay but what about the rest? just to check if we could use `implementation` instead of `api` for some of the dependencies here to avoid unnecessary transitive module access, for instance does the client really needs access to ':coordinator:clients:prover-client:serialization'?

jpnovais: @jonesho reduced the transitive dependencies. Please take a look.

jonesho (approved):

PR #101, opened 2024-09-25
3822 rejected transaction api service

jonesho:
This PR implements issue(s) #

### Checklist


.github/workflows/transaction-exclusion-api-testing.yml
```diff
@@ -0,0 +1,51 @@
...
+    inputs:
+      coverage:
+        description: To generate test report
+        required: false
+        type: boolean
+        default: false
+
+jobs:
+  run-tests:
+    runs-on: ubuntu-latest
```
jpnovais:
```suggestion
    runs-on: ubuntu-24.04
```

we should use fixed versions. Otherise, when GitHub updates the `lastest` we may have issues.

.github/workflows/transaction-exclusion-api-testing.yml
```diff
@@ -0,0 +1,51 @@
...
+          fetch-depth: 0
+          ref: ${{ github.head_ref }}
+      - uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: 21
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v3
+      - name: Run tests with coverage
+        uses: nick-fields/retry@v2
```
jpnovais: why retry? it's a new codebase. We must avoid "flacky" tests. If get some, let fix them.

.github/workflows/transaction-exclusion-api-testing.yml
```diff
@@ -0,0 +1,51 @@
...
+        if: ${{ inputs.coverage }}
+        with:
+          max_attempts: 2
+          retry_on: error
+          timeout_minutes: 20
+          command: |
+            ./gradlew transaction-exclusion-api:app:buildNeeded jacocoRootReport
+      - name: Run tests without coverage
```
jpnovais, line 38: why are we running 2x, 1 with coverage and then 1 without coverage?

e2e/src/transaction-exclusion.spec.ts
```diff
@@ -0,0 +1,129 @@
+import { beforeAll, describe, expect, it } from "@jest/globals";
+import { TransactionExclusionClient } from "./utils/utils";
+
+const transactionExclusionTestSuite = (title: string) => {
```
jpnovais: As already mentioned before, this shall be an integration test inside `transaction-exclusion-api:app` module, not an end-to-end tests. This test is a sort of integration/contract test, not end2end.


We shall have, however, a true e2e test: send a Transaction to the sequencer that cannot be proven, and verify that it was delivered and available through `linea_getTransactionExclusionStatusV1 `. This will ensure that the whole e2e flow works.

jpnovais (changes requested):

.github/workflows/transaction-exclusion-api-testing.yml
jonesho: In our CI workflow, only the one without coverage would run, and the one with coverage could be ran with the coverage by checking the `To generate test report` checkbox on manual run, similar to `traces-api-facade-testing`

e2e/src/transaction-exclusion.spec.ts
jonesho: Yes, I'm opted for a true e2e test too, but we need to wait for the release of the `linea-sequencer` plugin with this [PR](https://github.com/Consensys/linea-sequencer/pull/85) which would support reporting rejected transactions on tx simulation, but more importantly, the fixes of some blocking issues such as: missing `overflows` field in the save rejected transaction request and more descriptive reason messages. I'd update the e2e tests once the release is ready from besu team

jpnovais: > As already mentioned before, this shall be an integration test inside transaction-exclusion-api:app module.

Still, the end2end test you will write does not invalidate the integration test inside `transaction-exclusion-api:app` they serve different purposes.

jonesho: Yes, integration tests inside `transaction-exclusion-api:app` will be added

transaction-exclusion-api/app/src/integrationTest/kotlin/net/consensys/linea/transactionexclusion/TransactionExclusionAppTest.kt
```diff
@@ -0,0 +1,428 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.extension.ExtendWith
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.ValueSource
+import java.time.Duration
+import kotlin.random.Random
+
+@ExtendWith(VertxExtension::class)
+class TransactionExclusionAppTest : CleanDbTestSuiteParallel() {
```
jonesho, line 31: Here, I'm making this integration test to be able to run each test case in parallel with no order restriction instead of in sequential order (which was the initial design), the tradeoff is more databases to create (each per test case) and more code for initial setup in each test case.

.github/workflows/main.yml
```diff
@@ -202,9 +202,39 @@ jobs:
       e2e-tests-logs-dump: true
     secrets: inherit
 
+  run-e2e-tests-tx-exclusion-geth-tracing:
```
jonesho: Here, I separate the transaction exclusion e2e tests from the existing e2e tests and let them run in different runners to achieve parallelism, the reason is not to increase the total time for all e2e tests given that transaction exclusion e2e tests are not related to any of the existing e2e tests, and also given that they would use a sequencer and l2 besu node with reduced trace limits (otherwise if using the default trace limits, it would take quite some time or even timeout for the nodes to reject an over traces-limit transaction), so it'd be best to separate them from the existing e2e tests (where the nodes are shared across each e2e test)

e2e/src/transaction-exclusion.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+      } while (!getResponse?.result);
+
+      expect(getResponse.result.txHash).toStrictEqual(rejectedTxHash);
+      expect(getResponse.result.txRejectionStage).toStrictEqual("RPC");
+      expect(getResponse.result.from.toLowerCase()).toStrictEqual(l2AccountLocal.address.toLowerCase());
+    },
+    120_000,
+  );
+
+  it.skip("Should get the status of the rejected transaction reported from Besu SEQUENCER node", async () => {
```
jonesho, line 51: Skipping the sequencer test for now, as there's an issue of "wrongly rejecting normal transactions" still being investigating by besu team which would fail the e2e test occasionally.

jonesho:
@jpnovais updated after rebased on latest main, pls help review again, especially the integration test, thx!

build.gradle
```diff
@@ -147,6 +147,20 @@ dockerCompose {
     projectName = "docker"
     environment.put("L1_GENESIS_TIME", "${Instant.now().plusSeconds(3).getEpochSecond()}")
   }
+
+  localStackPostgresDbOnly {
+    startedServices = [
+      "postgres"
+    ]
+    composeAdditionalArgs = ["--profile", "l2",  "--profile", "l1"]
```
jpnovais: do we need this if are only spinning the DB?

build.gradle
```diff
@@ -147,6 +147,20 @@ dockerCompose {
...
+    useComposeFiles = ["${project.rootDir.path}/docker/compose.yml"]
+    waitForHealthyStateTimeout = Duration.ofMinutes(3)
+    waitForTcpPorts = false
+    removeOrphans = true
+    noRecreate = true
+    projectName = "docker"
+    environment.put("L1_GENESIS_TIME", "${Instant.now().plusSeconds(3).getEpochSecond()}")
```
jpnovais: Also this?

config/transaction-exclusion-api/transaction-exclusion-app-docker.config.toml
```diff
@@ -0,0 +1,30 @@
...
+port="5432"
+username="postgres"
+password="postgres"
+[database.write]
+host="postgres"
+port="5432"
+username="postgres"
+password="postgres"
+[database.cleanup]
+polling-interval="PT60S"
```
jpnovais, line 19: Note: when deploying to prod I recommend a polling time 10+min. So it's not cleaning all the time.

e2e/src/config/jest/global-setup.ts
```diff
@@ -16,10 +16,12 @@ export default async (): Promise<void> => {
     deployContract(new DummyContract__factory(), account),
     deployContract(new DummyContract__factory(), l2Account),
   ]);
-
   console.log(`L1 Dummy contract deployed at address: ${await dummyContract.getAddress()}`);
   console.log(`L2 Dummy contract deployed at address: ${await l2DummyContract.getAddress()}`);
 
+  const l2TestContract = await deployContract(new TestContract__factory(), l2Account);
+  console.log(`L2 Test contract deployed at address: ${await l2TestContract.getAddress()}`);
```
jpnovais, line 23: What is the impact of having this deploy here? 

This seems to be for the txExclusion API, so IMO it shall be there and not add delay to the remaining testsuite.

e2e/src/transaction-exclusion.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+
+const l2AccountManager = config.getL2AccountManager();
+
+describe("Transaction exclusion test suite", () => {
+  it.concurrent(
+    "Should get the status of the rejected transaction reported from Besu RPC node",
+    async () => {
+      if (!config.getTransactionExclusionEndpoint()) {
+        // Skip this test if transaction exclusion endpoint is not defined
+        return;
```
jpnovais: this is bad practice, Tests shall not be short-circuit like this! The report will say the test passed and it did not even if the endpoint is undefined.

So, either:
a) when no endpoint is defined we report as skipped;
b) we make the endpoint mandatory and the test fails if not defined;

e2e/src/transaction-exclusion.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+          to: await testContract.getAddress(),
+          data: testContract.interface.encodeFunctionData("testAddmod", [13000, 31]),
+          maxPriorityFeePerGas: etherToWei("0.000000001"), // 1 Gwei
+          maxFeePerGas: etherToWei("0.00000001"), // 10 Gwei
+        };
+        rejectedTxHash = await getTransactionHash(txRequest, l2AccountLocal);
+        await l2AccountLocal.sendTransaction(txRequest);
+      } catch (err) {
+        // This shall return error with traces limit overflow
+        console.log(`sendTransaction expected err: ${JSON.stringify(err)}`);
```
jpnovais: IMO, If the error is expected, we should not log it, otherwise it's entropy. when we have logging maybe log it in DEBUG/TRACE.

e2e/src/transaction-exclusion.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+      }
+
+      expect(!!rejectedTxHash).toBeTruthy();
+      console.log(`rejectedTxHash (RPC): ${rejectedTxHash}`);
+
+      let getResponse;
+      do {
+        await wait(5_000);
```
jpnovais: why wait 5s when our block time is 2s? Can we decrease this to 1s?

transaction-exclusion-api/app/src/integrationTest/kotlin/net/consensys/linea/transactionexclusion/TransactionExclusionAppTest.kt
```diff
@@ -0,0 +1,291 @@
...
+        "transactionRLP": "0x02f8388204d2648203e88203e88203e8941195cf65f83b3a5768f3c496d3a05ad6412c64b38203e88c666d93e9cc5f73748162cea9c0017b8201c8",
+        "blockNumber": "10000",
+        "reasonMessage": "Transaction line count for module ADD=402 is above the limit 70 (from Sequencer)",
+        "overflows": [
+          { "module": "ADD", "count": 402, "limit": 70 },
+          { "module": "MUL", "count": 587, "limit": 401 }
+        ]
+      }]
+    }
+    """.trimIndent()
```
jpnovais, line 160: this is a bit of repeated code. Why not make a `saveRejectedTx` which gets timestamp, txRejectionStage, etc fields as argument with defaults? this saveRejectedTx would only save and return the response (no assertions)

jpnovais (commented):
Please addess short-circut in the e2e test and deploy the contract shall be inside rejected-tx-api suite in IMO. 

Other comments are up to your discretion

e2e/src/transaction-exclusion.spec.ts
jonesho: I will add an assertion to check if config.getTransactionExclusionEndpoint() is defined

e2e/src/config/jest/global-setup.ts
jonesho: I'd prefer to leave it here in the global-setup.ts, because:
- I tried to put them in the beforeAll() of txExclusion test suite but noticed it was not a good idea as it's no longer guaranteed to run before all tests in the test suite when `log.concurrent` is used (jest [issue](https://github.com/jestjs/jest/issues/4281) here)
- when put in `global-setup.ts`, it will only deploy once for the whole e2e test (i.e. all contracts will be deployed before it starts running the test suites in parallel), so we didn't deploy more than it shouldn't
wdyt @jpnovais ?

jonesho: Btw, to play safe, I removed `beforeAll` in all e2e test suites where i`t.concurrent` is used due to the [jest](https://github.com/jestjs/jest/issues/4281) issue mentioned above

transaction-exclusion-api/app/src/integrationTest/kotlin/net/consensys/linea/transactionexclusion/TransactionExclusionAppTest.kt
jonesho: imo, the way it is now is better for readability, it clearly shows the request and response in json plain text in each test case, I'd prefer to keep it this way

jpnovais (approved):

jpnovais (approved):

PR #113, opened 2024-09-27
Allow Main branch workflow concurrency

bradbown:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/.github/workflows/coordinator-build-and-publish.yml
+++ b/.github/workflows/coordinator-build-and-publish.yml
@@ -29,7 +29,7 @@ on:
 
 concurrency:
   group: coordinator-build-and-publish-${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: ${{ github.ref_name != 'main' }}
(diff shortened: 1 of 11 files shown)


.github/workflows/coordinator-build-and-publish.yml
```diff
@@ -29,7 +29,7 @@ on:
 
 concurrency:
   group: coordinator-build-and-publish-${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: ${{ github.ref_name != 'main' }}
+  cancel-in-progress: ${{ github.ref == 'refs/heads/main' }}
```
jonesho: What confuses me now is what if the ref branch is not main, then `cancel-in-progress` would be false now, whereas previous it would be true, is it expected? I lack the context here...

bradbown: You're correct, I've amended this. We want false on main and true when not on main.

jpnovais (approved):

Filter94 (approved):

PR #114, opened 2024-09-27
fix revert in accordance with comments

Tabaie:
As per auditor feedback https://consensys.slack.com/archives/C07771UFHUG/p1725977141736399, even if unreachable, the code section should match its comments and reset.


gbotrel (approved):

prover/lib/compressor/blob/v1/blob_maker.go
```diff
@@ -165,6 +165,10 @@ func (bm *BlobMaker) Write(rlpBlock []byte, forceReset bool) (ok bool, err error
 		// 1. underlying writer error (shouldn't happen we use a simple in memory buffer)
 		// 2. we exceed the maximum input size of 2Mb (shouldn't happen either)
 		// In both cases, we can't do anything, so we reset the state.
+		if innerErr := bm.compressor.Revert(); innerErr != nil {
```
AlexandreBelling, line 168: Do we know if there are implications with the coordinator if we change that?

AlexandreBelling: And what are the differences between `compressor.Revert()` and `header.removeLastBlock()`

Tabaie: This was already the expected behavior of the compressor. But in any case I think it's not expected to ever happen right @gbotrel?

Tabaie: `bm.compressor` only handles the "payload" part of the blob, but every block has some representation in the block header as well (how many blocks for each conflated batch) so if we want to remove a block we have to revert on both header and payload.

AlexandreBelling (approved):

PR #118, opened 2024-09-30
Use multi-stage builds to unzip and copy jars

bradbown:
This PR implements issue(s) #134 

Multi stage reduces size of image by 110MB.
Removes unnecessary libs depending on architecture.

### Checklist


coordinator/Dockerfile
```diff
@@ -1,20 +1,32 @@
-FROM openjdk:22-ea-17-slim-bookworm
+FROM ubuntu as jar
```
jonesho: Why uses `ubuntu` as the base image? I thought the PR would be using the same image as the [one](https://github.com/Consensys/linea-monorepo/blob/43782deca24f9fc90dda8e3e5dafa8fd64d39b33/transaction-exclusion-api/Dockerfile#L2) used in transaction-exclusion-api dockerfile, and we should also avoid using `openjdk:22-ea-17-slim-bookworm` as it has vulnerbilities listed in docker hub.

julien-marchand: Additionally, please use images with versions

coordinator/Dockerfile
```diff
@@ -1,20 +1,32 @@
...
 
 RUN apt-get update \
 	&& apt-get install curl -y \
   && apt-get install unzip -y \
 	&& apt-get clean \
 	&& rm -rf /var/lib/apt/lists/*
 
+RUN mkdir -p src \
+  && mkdir -p libs
```
julien-marchand:
```suggestion
RUN mkdir -p src libs
```

coordinator/Dockerfile
bradbown: @jonesho I tried the image you mentioned and am getting errors with the native libraries for Go it seems. On going issue described here https://github.com/golang/go/issues/13492

coordinator/Dockerfile
```diff
@@ -1,20 +1,30 @@
-FROM openjdk:22-ea-17-slim-bookworm
+FROM openjdk:21-jdk-slim AS builder
 
 RUN apt-get update \
-	&& apt-get install curl -y \
   && apt-get install unzip -y \
 	&& apt-get clean \
```
jonesho: We could probably get rid of these two lines for the builder stage:
```
&& apt-get clean \
	&& rm -rf /var/lib/apt/lists/*
```
wdyt?

coordinator/Dockerfile
```diff
@@ -1,20 +1,30 @@
...
 	&& rm -rf /var/lib/apt/lists/*
 
+RUN mkdir -p src libs
```
jonesho: This would create dir under the root `/` ?

julien-marchand: Wherever WORKDIR is defined

coordinator/Dockerfile
```diff
@@ -1,20 +1,30 @@
...
+    && rm -rf /var/lib/apt/lists/* \
+    && mkdir -p /opt/consensys/linea/coordinator/libs/ \
+      /opt/consensys/linea/coordinator/logs \
+      /opt/consensys/linea/coordinator/tmp/prover/request \
+      /opt/consensys/linea/coordinator/tmp/prover/response
 
-RUN mkdir -p /opt/consensys/linea/coordinator/logs
-RUN mkdir -p /opt/consensys/linea/coordinator/tmp/prover/request
-RUN mkdir -p /opt/consensys/linea/coordinator/tmp/prover/response
+COPY --from=builder libs/coordinator/lib/** libs/
```
jonesho: More explicitly as
```
COPY --from=builder /libs/coordinator/lib/** libs/
```

coordinator/Dockerfile
```diff
@@ -1,20 +1,30 @@
...
-COPY --from=zip ./coordinator.zip /opt/consensys/linea/coordinator/libs/
-RUN unzip libs/coordinator.zip -d libs/ && mv libs/coordinator/lib/** libs/ && rm -R libs/coordinator/
+RUN apt-get update \
+    && apt-get install curl -y \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/* \
+    && mkdir -p /opt/consensys/linea/coordinator/libs/ \
+      /opt/consensys/linea/coordinator/logs \
+      /opt/consensys/linea/coordinator/tmp/prover/request \
+      /opt/consensys/linea/coordinator/tmp/prover/response
```
jonesho: I believe we could remove all the `/opt/consensys/linea/coordinator/` here as the WORKDIR is defined

coordinator/Dockerfile
```diff
@@ -1,20 +1,30 @@
...
-	&& apt-get install curl -y \
   && apt-get install unzip -y \
 	&& apt-get clean \
 	&& rm -rf /var/lib/apt/lists/*
 
+RUN mkdir -p src libs
+
+COPY --from=zip ./coordinator.zip /src
+
+RUN unzip /src/coordinator.zip -d libs/
```
jonesho:
```
RUN unzip /src/coordinator.zip -d /libs
```

coordinator/Dockerfile
jonesho: I know but if WORKDIR is not specified it will be on the root `/`, so to play safe can we update the above to be more explicit as
```
RUN mkdir -p /src /libs
```

coordinator/Dockerfile
```diff
@@ -1,20 +1,47 @@
...
+
+RUN unzip /src/coordinator.zip -d /libs
+
+RUN mkdir -p /libs/coordinator/lib/unpacked-blob-compressor /libs/coordinator/lib/unpacked-blob-shnarf-calculator && \
+  cd /libs/coordinator/lib/unpacked-blob-compressor/ && jar -xf ../blob-compressor-0.0.4.jar && \
+  cd /libs/coordinator/lib/unpacked-blob-shnarf-calculator/ && jar -xf /libs/coordinator/lib/blob-shnarf-calculator-0.0.4.jar
+
+RUN case $(uname -m) in \
+    x86_64) \
+      rm -rf /libs/coordinator/lib/unpacked-blob-compressor/darwin-**; \
```
jonesho: Can this be moved out from the switch case here? Seems they would be removed regardless of the case, or can we use a local label to replace the the switch case to avoid repeated commands? sth like:
```
...
rm -rf /libs/coordinator/lib/unpacked-blob-compressor/linux-$MACHINE_TYPE/; \
...
rm -rf /libs/coordinator/lib/unpacked-blob-shnarf-calculator/linux-$MACHINE_TYPE/;
```

coordinator/Dockerfile
```diff
@@ -1,20 +1,47 @@
...
-	&& apt-get clean \
-	&& rm -rf /var/lib/apt/lists/*
+RUN apt-get update && apt-get install unzip -y
+
+RUN mkdir -p /src /libs
+
+COPY --from=zip ./coordinator.zip /src
+
+RUN unzip /src/coordinator.zip -d /libs
+
```
jonesho: I'd suggest to add the below line here to avoid the repeated prefix in paths:
```
WORKDIR /libs/coordinator/lib/
```

coordinator/Dockerfile
```diff
@@ -1,20 +1,47 @@
...
 
 WORKDIR /opt/consensys/linea/coordinator
 
-# copy application
-COPY --from=zip ./coordinator.zip /opt/consensys/linea/coordinator/libs/
-RUN unzip libs/coordinator.zip -d libs/ && mv libs/coordinator/lib/** libs/ && rm -R libs/coordinator/
+RUN apt-get update \
+    && apt-get install curl -y \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/* \
```
jonesho: I dun think we even need these `apt-get` related commands here, right?

.github/workflows/coordinator-build-and-publish.yml
```diff
@@ -76,7 +76,7 @@ jobs:
         uses: gradle/actions/setup-gradle@v4
       - name: Build dist
         run: |
-          ./gradlew coordinator:app:distZip --no-daemon
+          ./gradlew coordinator:app:installDist --no-daemon
```
jonesho, line 79: use `installDist` instead of `distZip` to avoid redundant zip and unzip procedures, from my understanding, `installDist` shall be the same as `distZip` without zipping, can someone second here? @jpnovais

coordinator/Dockerfile
```diff
@@ -1,20 +1,38 @@
-FROM openjdk:22-ea-17-slim-bookworm
+# BUILDER image
+FROM eclipse-temurin:21-jdk-noble AS builder
```
jonesho, line 2: An image with JDK is needed here as we're using jar command to extract and recreate archive below, welcome any suggestion for a good way to use jar command without JDK

coordinator/Dockerfile
```diff
@@ -1,20 +1,38 @@
...
+      rm -rf unpacked-blob-compressor/linux-x86-64/; \
+      rm -rf unpacked-blob-shnarf-calculator/linux-x86-64/; \
+    ;; \
+    esac && \
+    jar -cf $bcJar -C unpacked-blob-compressor . && \
+    jar -cf $bscJar -C unpacked-blob-shnarf-calculator .
+RUN rm -rf unpacked-blob-compressor unpacked-blob-shnarf-calculator
+
+# FINAL image
+FROM eclipse-temurin:21-jre-noble
```
jonesho, line 29: We can't use `eclipse-temurin:21.0.4_7-jre-alpine` or `eclipse-temurin:21-jre-alpine` here because we'd get the following error at runtime:
```
2024-11-22 14:55:03 Native library (root/.cache/JNA/temp/jna3653017674932247023.tmp) not found in resource path (libs/coordinator.jar)
2024-11-22 14:55:03     at com.sun.jna.NativeLibrary.loadLibrary(NativeLibrary.java:325)
2024-11-22 14:55:03     at com.sun.jna.NativeLibrary.getInstance(NativeLibrary.java:481)
2024-11-22 14:55:03     at com.sun.jna.Library$Handler.<init>(Library.java:197)
2024-11-22 14:55:03     at com.sun.jna.Native.load(Native.java:618)
2024-11-22 14:55:03     at com.sun.jna.Native.load(Native.java:592)
2024-11-22 14:55:03     at net.consensys.linea.blob.GoNativeBlobCompressorFactory$Companion.getInstance(GoNativeBlobCompressor.kt:118)
```
which could be caused by the known issue that Alpine is using `muslibc` instead of `glibc` as mentioned [here](https://github.com/iegomez/mosquitto-go-auth/issues/20#issuecomment-497017101), I tried adding the apk package `gcompat` as suggested by others but no help, or it might be related to how we build the jna libs, but instead of spending more time into this issue, I'd recommend to keep using `eclipse-temurin:21-jre-noble` for now, and if it's really the requirement for us to support alpine image for coordinator, we could add a ticket to following this up in greater length

jonesho:
@jpnovais can you help review again?

coordinator/Dockerfile
jpnovais, line 2: @jonesho IIRC from our last our meeting with @julien-marchand we agreed on not having a builder image because the step commands bellow in `RUN` can be executed in the host machine.?

What is the motivation and value of having a builder image?

jonesho: @jpnovais Yes, but as mentioned above, the `jar` command was not available in JRE image, and hence we need to use JDK image here to use the `jar` command to extract and recreate the JAR archive based on the machine type (from `uname -m`). I was thinking about moving this "extract and recreate JAR" step to pipeline job but then we'd need separate "extract + build" jobs for `arm64` and `amd64` image respectively, which would take up more pipeline resources and runtime imo, and complicate the build process. Welcome for any better approach :)

jpnovais (approved):

PR #119, opened 2024-09-30
perf(prover): preallocate slices

estensen:
If we know the size a slice will have we can preallocate the slice to avoid [memory pressure](https://github.com/golang/go/blob/87e48c5afdcf5e01bb2b7f51b7643e8901f4b7f9/src/runtime/slice.go#L100-L112).

This PR also adds [prealloc](https://github.com/alexkohler/prealloc) via golangci-lint to avoid regressions.


estensen:
It's possible to preallocate more, but this is a first pass.

prover/zkevm/prover/ecdsa/adress.go
```diff
@@ -236,11 +236,13 @@ func (addr *Addresses) assignMainColumns(
 	split := splitAt(nbEcRecover)
 	n := nbRowsPerEcRec
 
-	var (
-		hashHi, hashLo, isHash, trimmedHi []field.Element
-	)
-
 	permTrace := keccak.GenerateTrace(pkModule.Data.ScanStreams(run))
+
```
AlexandreBelling, line 243: Given the number of consecutive allocations, it would make sense to wrap all that in a var statement

prover/zkevm/prover/ecdsa/utils_for_test.go
```diff
@@ -103,14 +103,12 @@ func (td *txnData) assignTxnDataFromPK(
...
 ) {
-
-	// compute the hash of public keys
 	var (
-		pkHash  [][]byte
 		hasher  = sha3.NewLegacyKeccak256()
 		maxNbTx = ac.Inputs.settings.MaxNbTx
 	)
-
+	// compute the hash of public keys
```
AlexandreBelling, line 110: Same comment here. Why remove this statement from the "var"?

AlexandreBelling (approved):
Thanks a lot for the PR

LGTM for merging

Just a few cosmetic comments

prover/zkevm/prover/ecdsa/adress.go
estensen: I don't think I understand this. Now `:=` shorthand is used for declaring and initializing. Why is `var` necessary`

AlexandreBelling: It's not. It just looks nicer as it indents everything. I'll pass on that one as it's not super important either

PR #131, opened 2024-10-02
Switch workflows to self hosted runners

bradbown:
This PR implements issue(s) #

### Checklist


bradbown:
> @bradbown can you please clarify what `runs-on: [self-hosted, Linux, X64, small]` mean?
> 
> Are those labels specifying a static immutable runner image, or are those filters where we can get different runners as long as they match the filtering criteria?
> 
> We must ensure that the runners are static on operating system versions/packages, otherwise we may run on "flacky" ci. The `Linux` there does not even specify the distribution, we shall have something specific as `ubuntu-22.04`

@jpnovais The labels are what are defined per each runner here: https://github.com/Consensys/linea-monorepo/actions/runners?tab=self-hosted
The labels can be set to anything to be more specific, I'm sure we can request changes to these labels.

jpnovais (approved):

jpnovais (approved):

PR #139, opened 2024-10-04
Updating tracer

Filter94:
This PR implements issue(s) #

### Checklist


.github/workflows/postman-testing.yml
```diff
@@ -14,7 +14,7 @@ concurrency:
 
 jobs:
   run-tests:
-    runs-on: [self-hosted, ubuntu-22.04, X64, medium]
```
bradbown, line 17: These shouldn't work, they large runners are missing the `ubuntu-22.04` still so need `Linux` for now temporarily, same for the other large runners below.

Filter94: Do we know the exact version of linux they have available?

bradbown: It's ubuntu-22.04 but the labels are based on what's [here](https://github.com/Consensys/linea-monorepo/actions/runners?tab=self-hosted).

bradbown (approved):

.github/workflows/postman-testing.yml
Filter94: I reverted this change

diana-borbe-consensys (approved):

PR #146, opened 2024-10-07
Adding more constraints for IsEmptyLeaf

arijitdutta67:
This PR adds the following constraints on `IsEmptyLeaf`:
1. It is one only at the third row of INSERT operation and the fourth row of DELETE operation
2. It is zero elsewhere


prover/zkevm/prover/statemanager/accumulator/define.go
```diff
@@ -516,6 +516,26 @@ func (am *Module) checkLeafHashes() {
 	expr4 := symbolic.Sub(symbolic.Square(cols.IsEmptyLeaf), cols.IsEmptyLeaf)
 	expr4 = symbolic.Mul(expr4, cols.IsActiveAccumulator)
 	am.comp.InsertGlobal(am.Round, am.qname("IS_EMPTY_LEAF_BOOLEANITY"), expr4)
+	// IsEmptyLeaf is set to true only for the third row for INSERT
```
AlexandreBelling: The constraint does not cover the cases where the operation is an insertion/deletions but the row position is not 3/4. You could solve the issue by using a unified constraint of the form:

isEmptyLeaf = isFirst[-2] * isInsert[-2] + isFirst[-3] * isDelete[-3]

Also, with the immutable API for symbolic variables, you don't need to use  `ifaces.ColumnAsVariable`

AlexandreBelling: (I am not sure if you need to multiply the right-hand side by IsActive), but you get the picture

arijitdutta67: Done.

AlexandreBelling (approved):

bogdanbear (approved):

PR #147, opened 2024-10-07
Prover/Constraint for IsInsertRow3

arijitdutta67:
This PR adds the following constraint on `IsInsertRow3`
1. If `IsEmptyLeaf` is one for INSERT, then it is set to true


prover/zkevm/prover/statemanager/accumulator/define.go
```diff
@@ -541,6 +541,15 @@ func (am *Module) checkNextFreeNode() {
 		cols.IsInsertRow3,
 		symbolic.Sub(cols.NextFreeNode, cols.InsertionPath, symbolic.NewConstant(1)))
 	am.comp.InsertGlobal(am.Round, am.qname("NEXT_FREE_NODE_CONSISTENCY_2"), expr4)
+
+	// If it is an empty leaf for the INSERT operation, then it must be at row 3
+	// and IsInsertRow3 is true i.e.,
+	// IsActive[i] * IsInsert[i] * IsEmptyLeaf[i] * (1-IsInsertRow3[i])
+	expr5 := symbolic.Mul(cols.IsActiveAccumulator,
```
AlexandreBelling, line 604: Same point as for the issue M, the constraint would more safely be expressed as,

isInsertRow3 = isActive * isInsert * isEmptyLeaf

Note, this will only work thanks to the constraint you added for "M" as it ensures that isEmptyLeaf is correctly set and you should emphase that out in the comment as it is a little hard to grasp.

arijitdutta67: Done.

bogdanbear (approved):

PR #149, opened 2024-10-07
Prover/constraint for the update operation

arijitdutta67:
This PR adds the following constraint for the update operation:
1. For Update operation, `HKey` remains the same


prover/zkevm/prover/statemanager/accumulator/define.go
```diff
@@ -373,6 +376,17 @@ func (am *Module) checkDelete() {
 	am.comp.InsertGlobal(am.Round, am.qname("IS_DELETE_BOOLEAN"), expr3)
 }
 
+func (am *Module) checkUpdate() {
```
AlexandreBelling, line 381: Again, `ifaces.ColumnAsVariable` should not be needed here as `symbolic.Sub` accepts columns directly

AlexandreBelling: But the constraint itself is ok

arijitdutta67: Thanks, I removed `ifaces.ColumnAsVariable` from everywhere. To avoid merge conflict, I suggest to merge this PR at the last.

AlexandreBelling (approved):

bogdanbear (approved):

PR #157, opened 2024-10-07
feat: rewrite error logs from blob submission eth_call due to insuffi…

jonesho:
…cient gas prices

This PR implements issue(s) #

### Checklist

### DIFF
--- a/config/coordinator/log4j2-dev.xml
+++ b/config/coordinator/log4j2-dev.xml
@@ -27,8 +27,8 @@
           <KnownError logLevel="INFO"  message="^.*?\bNonce too low\b.*?$" />
           <KnownError logLevel="DEBUG" message="^.*?\bConnection reset\b.*?$" stackTrace="true" />
           <KnownError logLevel="INFO"  message="^.*?\beth_call\b.*?\bContract Call has been reverted by the EVM with the reason\b.*?$"/>
-          <KnownError logLevel="INFO"  message="^.*?\beth_call for blob submission failed\b.*?$"/>
           <KnownError logLevel="INFO"  message="^.*?\brequest beyond head block\b.*?$"/>
+          <KnownError logLevel="INFO"  message="^.*?\beth_call for blob submission failed\b.*?\bmax fee per (blob )?gas less than block (blob gas|base) fee\b.*?$"/>
         </KnownErrors>
       </Log4jLineaRewriter>
     </Rewrite>
@@ -97,15 +97,15 @@
     <!--    <Logger name="net.consensys.zkevm.ethereum.coordination.blob" level="TRACE" additivity="false">-->
     <!--      <appender-ref ref="console"/>-->
     <!--    </Logger>-->
-    <Logger name="net.consensys.zkevm.ethereum.submission" level="INFO" additivity="false">
+    <Logger name="net.consensys.zkevm.ethereum.submission" level="DEBUG" additivity="false">
       <DebouncingFilter/>
       <appender-ref ref="rewrite"/>
     </Logger>
-    <Logger name="net.consensys.zkevm.ethereum.finalization" level="INFO" additivity="false">
+    <Logger name="net.consensys.zkevm.ethereum.finalization" level="DEBUG" additivity="false">
       <DebouncingFilter/>
       <appender-ref ref="rewrite"/>
     </Logger>
-    <Logger name="net.consensys.linea.ethereum.gaspricing.dynamiccap" level="INFO" additivity="false">
+    <Logger name="net.consensys.linea.ethereum.gaspricing.dynamiccap" level="DEBUG" additivity="false">
       <DebouncingFilter/>
       <appender-ref ref="rewrite"/>
     </Logger>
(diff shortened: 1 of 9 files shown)


coordinator/ethereum/blob-submitter/src/test/kotlin/net/consensys/zkevm/ethereum/submission/LoggingHelperTest.kt
```diff
@@ -0,0 +1,146 @@
...
+    val error = Error(insufficientMaxFeePerGasErrMsg)
+    logSubmissionError(
+      log = logger,
+      logMessage = blobSubmissionFailedLogMsg,
+      intervalString = blobSubmissionFailedIntervalStr,
+      error = error,
+      isEthCall = true
+    )
+    val expectedErrorMessage =
+      "max fee per gas less than block base fee:" +
```
jpnovais: we should keep the original terminology `maxFeePerGas`. Likewise in the other places as well.

config/coordinator/log4j2-dev.xml
```diff
@@ -28,6 +28,7 @@
           <KnownError logLevel="DEBUG" message="^.*?\bConnection reset\b.*?$" stackTrace="true" />
           <KnownError logLevel="INFO"  message="^.*?\beth_call\b.*?\bContract Call has been reverted by the EVM with the reason\b.*?$"/>
           <KnownError logLevel="INFO"  message="^.*?\brequest beyond head block\b.*?$"/>
+          <KnownError logLevel="INFO"  message="^.*?\beth_call for blob submission failed\b.*?\bmax fee per (blob )?gas less than block (blob gas|base) fee\b.*?$"/>
```
jpnovais: This log level shall be addressed in the code IMO. Not, by configuration.

The configuration shall be used as a "quick fix/workaround" in prod until the log is is fixed in the code.

jonesho: This is to let the log rewriter to set the log level to INFO instead of ERROR for the reformatted error message given that the one in line 29 won't be working for insufficent gas fee error after the error message get reformatted. 

My initial thought is to align the use of log rewriter to re-set the log level, but I also think, for simplicity, we could just call `log.info()` in `LoggingHelper` if eth_call blob submission insufficent gas fee error is detected. If no objection, I'd change that to be handled in `LoggingHelper`

coordinator/ethereum/blob-submitter/src/test/kotlin/net/consensys/zkevm/ethereum/submission/LoggingHelperTest.kt
jonesho: If we'd like to keep the original terminology `maxFeePerGas`, I'd also do the same for `maxFeePerBlobGas`

jonesho:
@jpnovais see if you could review this again for the changes based on your previous comments

coordinator/ethereum/blob-submitter/src/main/kotlin/net/consensys/zkevm/ethereum/submission/LoggingHelper.kt
```diff
@@ -2,6 +2,17 @@ package net.consensys.zkevm.ethereum.submission
...
 
+private const val insufficientGasFeeRegexStr =
+  "(max fee per (blob )?gas less than block (blob gas|base) fee:" +
+    " address 0x[a-fA-F0-9]{40},? (blobGasFeeCap|maxFeePerGas): [0-9]+," +
+    " (blobBaseFee|baseFee): [0-9]+ \\(supplied gas [0-9]+\\))"
+
+private fun rewriteInsufficientGasFeeErrorMessage(errorMessage: String): String? {
+  return Regex(insufficientGasFeeRegexStr).find(errorMessage)?.groupValues?.first()
+    ?.replace("max fee per gas", "maxFeePerGas")
+    ?.replace(Regex("max fee per blob gas|blobGasFeeCap"), "maxFeePerBlobGas")
```
jpnovais: can we please cache the Regex compilation? Regex compilation can be performance consuming and we can reuse them even their patterns are static.

coordinator/ethereum/blob-submitter/src/test/kotlin/net/consensys/zkevm/ethereum/submission/LoggingHelperTest.kt
```diff
@@ -0,0 +1,144 @@
...
+    " address 0x47C63d1E391FcB3dCdC40C4d7fA58ADb172f8c38'. revertReason=UNKNOWN errorData=null"
+
+  @BeforeEach
+  fun setUp() {
+    logger = mock()
+  }
+
+  @Test
+  fun `insufficient max fee per gas with isEthCall is true triggers rewrite info message`() {
+    val error = Error(insufficientMaxFeePerGasErrMsg)
```
jpnovais: Please use a RuntimeException as example.

We don't - and generally - shall not intercept Errors. explanation [here](https://stackoverflow.com/questions/912334/differences-between-exception-and-error)

coordinator/ethereum/blob-submitter/build.gradle
```diff
@@ -34,6 +34,7 @@ dependencies {
   testImplementation("com.fasterxml.jackson.core:jackson-annotations:${libs.versions.jackson.get()}")
   testImplementation("com.fasterxml.jackson.module:jackson-module-kotlin:${libs.versions.jackson.get()}")
   testImplementation("com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${libs.versions.jackson.get()}")
+  testImplementation("org.apache.logging.log4j:log4j-core-test:${libs.versions.log4j.get()}")
```
jpnovais: curiosity: what is this for?

jonesho: This is not needed, it was added for the removed test cases from previous commit, thanks!

jpnovais (approved):

bradbown (approved):

PR #159, opened 2024-10-08
Prover/Hash Key Consistency

arijitdutta67:
This PR adds constraints to verify that the hash keys those are used in the sandwitch check are consistent with that of the corresponding leaf opening values.


prover/zkevm/prover/statemanager/accumulator/define.go
```diff
@@ -458,6 +462,49 @@ func (am *Module) checkEmptyLeaf() {
 	am.comp.InsertGlobal(am.Round, am.qname("EMPTY_LEAVES_FOR_DELETE"), expr2)
 }
 
+func (am *Module) checkHKeyConsistency() {
```
AlexandreBelling: These constraints themselves seem OK, but why do you only cover the READZERO and the insert cases?

arijitdutta67: This is because, the columns `HKey`, `HKeyMinus`, and `HKeyPlus` are non-zero only when the sandwitch check is enabled, which is essentially for `INSERT` and `READ-ZERO`.

arijitdutta67: I think it makes more sense to add these constraints inside sandwitch check. Moved them. Please review @AlexandreBelling .

AlexandreBelling (approved):

bogdanbear (approved):

Issue #133, opened 2024-10-02
E2E tests - Parallel Run and Improve test Isolation


jpnovais:
## Description
End2End tests shall refactored to:
- allow them to run in parallel and 
- have better isolation

### Refactor Suggestions
One pinpoint at the moment is that we need to configure loads of environment variables in multiple files and know them by heart. This does not scale well and is error-prone. 

Alternatively, we can he a global setup helper(s) for the test setup, that is responsible:
 - provide the overall configs: l1-endpoint, l2-endpoint, smart contract address etc
 - create l1/l2 accounts on the fly and fund them from whale accounts in the genesis

## Motivation
- Simulate Real-World Scenarios: e.g Coordinator & Message anchoring will benefit from being restarted when other things are happening. This is like in production
- Speed up CI times
- Shift-Left - Faster feedback loop - by reducing e2e times developers are more likely to run them locally.

## Acceptance criteria
- [ ] All the tests suites run in parallel, including the coordinator-restart

Filter94:
I think there are a few things to improve still:
* NodeJs installation alone takes 24 seconds which becomes a considerable share of total E2E time. Can we speed it up with cache?
* Env spin up takes 1m 30 seconds. I suspect most of the time is taken by contracts deployments. I know we had some effort in the past to speed deployments up, but I'd just like to revisit this, if we make it in an optimal way
* E2E testsuite itself takes 2 minutes, longest of which is `src/submission-finalization.spec.ts (106.366 s)`. We should see if there is a way to reduce the latency between L2 transaction sending to L1 finalization. As a compromise, we should consider using `latest` tag instead of `finalize`. Another thing we should revisist is Coordinator configuration. Some polling intervals and delays might be slowing us down.


@jonesho I think it would be good for you to take a look at that as you're proficient in the stack as a whole, as well as the typescript part of the E2E tests

PR #164, opened 2024-10-09
Feat(#133): run e2e tests in parallel

VGau:
This PR implements issue(s) #133 

### Checklist


e2e/src/common/utils.ts
```diff
@@ -170,34 +171,63 @@ export async function sendXTransactions(
...
+
+      if (isRunning) {
+        timeoutId = setTimeout(sendTransaction, pollingInterval);
+      }
+    } catch (error) {
+      console.error("Error sending transaction:", error);
+      if (isRunning) {
+        timeoutId = setTimeout(sendTransaction, pollingInterval);
+      }
+    }
```
jonesho, line 186: Maybe we can put the following line in a finally clause:
```
if (isRunning) {
    timeoutId = setTimeout(sendTransaction, pollingInterval);
}
```

e2e/src/config/tests-config/environments/local.ts
```diff
@@ -0,0 +1,44 @@
...
+  L2: {
+    rpcUrl: L2_RPC_URL,
+    chainId: 1337,
+    l2MessageServiceAddress: "0xe537D669CA013d86EBeF1D64e40fC74CADC91987",
+    dummyContractAddress: "0x2f6dAaF8A81AB675fbD37Ca6Ed5b72cf86237453",
+    accountManager: new GenesisBasedAccountManager(
+      new ethers.JsonRpcProvider(L2_RPC_URL.toString()),
+      path.resolve(
+        process.env.LOCAL_L2_GENESIS ||
+          path.resolve(__dirname, "../../../../..", "docker/config", "linea-local-dev-genesis.json"),
```
jonesho: Any reason why we're not using `linea-local-dev-genesis-PoA-besu.json` for L2 network, it's the default genesis file used by sequencer, besu nodes, etc.. in our local stack

e2e/src/common/utils.ts
VGau: yes you're right thanks Jones :)

e2e/src/l2.spec.ts
```diff
@@ -1,339 +1,321 @@
...
-        });
+    await expect(
+      dummyContract.connect(account).setPayload(ethers.randomBytes(TRANSACTION_CALLDATA_LIMIT)),
+    ).rejects.toThrow("missing revert data");
+  });
 
-        const receipt = await tx.wait();
-        expect(receipt.status).toEqual(1);
-      });
+  it.concurrent("Should succeed if transaction data size is below the limit", async () => {
```
jonesho, line 25: I think this test can be combined with the next test below: `Should succeed in conflating transactions with large calldata with low gas into multiple L1 blocks`, they're basically the same with different transactions, and I don't see much benefit of separating them, anyway up to you, the gain of combining might not be much as well

e2e/src/restart.spec.ts
```diff
@@ -1,51 +1,77 @@
...
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+let coordinatorRestartPromise: any = null;
+let testsWaitingForRestart = 0;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+let resolveCoordinatorRestartPromise: any = null;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+let rejectCoordinatorRestartPromise: any = null;
+const TOTAL_TESTS_WAITING = 2;
+
+function waitForCoordinatorRestart() {
```
jonesho: This function is a bit hard to read, just to see if you have any idea in mind that could improve, otherwise we could leave it as is.

jonesho: Sth like:
```
let hasRestarted = false;
...
async function waitForCoordinatorRestartV2() {
  testsWaitingForRestart += 1;
  while (testsWaitingForRestart < TOTAL_TESTS_WAITING ) {
    await wait(1000);
    if (!hasRestarted) {
      hasRestarted = true;
      try {
        await execDockerCommand("restart", "coordinator");
        console.log("Coordinator restarted.");
      } catch (error) {
        console.error("Failed to restart coordinator:", error);
        throw error
      }
    }
  }
}
```
Just a quick draft of suggestion, didn't test it, not sure if it makes sense

e2e/src/l2.spec.ts
Filter94: Or we can remove them at all, because they don't actually test conflation much nor we can get a deterministic conflation behaviour in E2E test

jonesho (approved):

Filter94 (approved):

PR #183, opened 2024-10-14
Fix: optimize postman docker build

VGau:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/sdk/Dockerfile
+++ b/sdk/Dockerfile
@@ -5,6 +5,8 @@ ENV PATH="$PNPM_HOME:$PATH"
 
 RUN corepack enable
 
+RUN apt-get update && apt-get install -y --no-install-recommends python3 ca-certificates bash curl make g++
+
 FROM base AS builder
 
 WORKDIR /usr/src/app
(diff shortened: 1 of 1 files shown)


.github/workflows/postman-build-and-publish.yml
```diff
@@ -71,7 +71,11 @@ jobs:
           push: true
           tags: |
             ${{ env.IMAGE_NAME }}:${{ env.COMMIT_TAG }}-${{ env.UNTESTED_TAG_SUFFIX }}
-          cache-from: type=registry,ref=${{ env.IMAGE_NAME }}:buildcache
-          cache-to: type=registry,ref=${{ env.IMAGE_NAME }}:buildcache,mode=max
+          cache-from: |
+            type=registry,ref=${{ env.IMAGE_NAME }}:buildcache-amd64,platform=linux/amd64
```
jonesho, line 75: What was the benefit of specifying the platform for the cache? reduced the cache size? Also as of now (before the changes here), the cache is working for both platforms, right?

VGau: The cache was not working properly for multi platforms. We were using the same cache for both platform. Apparently the cache is computed differently on each platform. So I added a cache for each platform

jonesho (approved):

bradbown (approved):

bradbown (approved):

jonesho (approved):

PR #185, opened 2024-10-15
Prover: apply the AVX optimizations for math/vector

AlexandreBelling:
This PRs uses the gnark function for vector field operations. The gain is that these operations are taking advantage of AVX.

### Checklist


gbotrel (dismissed):
lgtm; we can also kill the duplicate len checks on the vectors I believe (already done on gnark-crypto side)

gusiri (dismissed):

PR #186, opened 2024-10-15
Coordinator - Extra tests for findConsecutiveProvenBlobs

bradbown:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/coordinator/persistence/aggregation/src/integrationTest/kotlin/net/consensys/zkevm/persistence/dao/aggregation/AggregationsPostgresDaoTest.kt
+++ b/coordinator/persistence/aggregation/src/integrationTest/kotlin/net/consensys/zkevm/persistence/dao/aggregation/AggregationsPostgresDaoTest.kt
@@ -294,6 +294,90 @@ class AggregationsPostgresDaoTest : CleanDbTestSuiteParallel() {
   }
 
   @Test
+  fun findConsecutiveProvenBlobsReturnsFullListOfProvenBlobsWithProvenBatches() {
+    val batch1 = createBatch(0L, 10L)
+    val batch2 = createBatch(11L, 20L)
+    val batch3 = createBatch(21L, 30L)
+
+    val blob1 = createBlobRecordFromBatches(listOf(batch1))
+    val blob2 = createBlobRecordFromBatches(listOf(batch2, batch3))
+
+    val batches = listOf(batch1, batch2, batch3)
+    val blobs = listOf(blob1, blob2)
+
+    batches.forEach { insertBatch(it).get() }
+    blobs.forEach { insertBlob(it).get() }
+
+    aggregationsPostgresDaoImpl.findConsecutiveProvenBlobs(0).get().also { blobCounters ->
+      assertThat(blobCounters).hasSameElementsAs(
+        listOf(
+          getBlobAndBatchCounters(listOf(batch1), blob1),
+          getBlobAndBatchCounters(listOf(batch2, batch3), blob2)
+        )
+      )
+    }
+  }
+
+  @Test
+  fun findConsecutiveProvenBlobsWhenDbIsEmpty() {
+    aggregationsPostgresDaoImpl.findConsecutiveProvenBlobs(0).get().also { blobCounters ->
+      assertThat(blobCounters).hasSameElementsAs(emptyList())
+    }
+  }
+
+  @Test
+  fun findConsecutiveProvenBlobsUnprovenBlobAndBatchAtDifferentIntervals() {
+    // Batch2 proof is missing
+    // Blob for batch3 is missing
+    val batch1 = createBatch(0L, 10L)
+    val batch2 = createBatch(11L, 20L, status = Batch.Status.Finalized)
+    val batch3 = createBatch(21L, 30L)
+    val batch4 = createBatch(31L, 40L)
+
+    val blob1 = createBlobRecordFromBatches(listOf(batch1, batch2))
+    val blob2 = createBlobRecordFromBatches(listOf(batch3), status = BlobStatus.COMPRESSION_PROVING)
+    val blob3 = createBlobRecordFromBatches(listOf(batch4))
+
+    val batches = listOf(batch1, batch2, batch3, batch4)
+    val blobs = listOf(blob1, blob2, blob3)
+
+    batches.forEach { insertBatch(it).get() }
+    blobs.forEach { insertBlob(it).get() }
+
+    aggregationsPostgresDaoImpl.findConsecutiveProvenBlobs(0).get().also { blobCounters ->
+      assertThat(blobCounters).hasSameElementsAs(emptyList())
+    }
+  }
+
+  @Test
+  fun findConsecutiveProvenBlobsUnprovenBlobAndBatchInTheMiddle() {
+    val batch1 = createBatch(1L, 10L)
+    val batch2 = createBatch(11L, 20L)
+    val batch3 = createBatch(21L, 30L)
+    val batch4 = createBatch(31L, 40L)
+    val batch6 = createBatch(51L, 60L)
+    val batches = listOf(batch1, batch2, batch3, batch4, batch6)
+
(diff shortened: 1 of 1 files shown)


coordinator/persistence/aggregation/src/integrationTest/kotlin/net/consensys/zkevm/persistence/dao/aggregation/AggregationsPostgresDaoTest.kt
```diff
@@ -293,6 +293,90 @@ class AggregationsPostgresDaoTest : CleanDbTestSuiteParallel() {
...
+    val batch3 = createBatch(21L, 30L)
+
+    val blob1 = createBlobRecordFromBatches(listOf(batch1))
+    val blob2 = createBlobRecordFromBatches(listOf(batch2, batch3))
+
+    val batches = listOf(batch1, batch2, batch3)
+    val blobs = listOf(blob1, blob2)
+
+    batches.forEach { insertBatch(it).get() }
+    blobs.forEach { insertBlob(it).get() }
```
jpnovais: this is slow. can't they be inserted in parallel ino the Db with `SafeFuture.collectAll`? 

I keep on calling on this :(

coordinator/persistence/aggregation/src/integrationTest/kotlin/net/consensys/zkevm/persistence/dao/aggregation/AggregationsPostgresDaoTest.kt
```diff
@@ -293,6 +293,90 @@ class AggregationsPostgresDaoTest : CleanDbTestSuiteParallel() {
...
+          getBlobAndBatchCounters(listOf(batch2, batch3), blob2)
+        )
+      )
+    }
+  }
+
+  @Test
+  fun findConsecutiveProvenBlobsWhenDbIsEmpty() {
+    aggregationsPostgresDaoImpl.findConsecutiveProvenBlobs(0).get().also { blobCounters ->
+      assertThat(blobCounters).hasSameElementsAs(emptyList())
```
jpnovais:
```suggestion
      assertThat(blobCounters).isEmpty()
```

coordinator/persistence/aggregation/src/integrationTest/kotlin/net/consensys/zkevm/persistence/dao/aggregation/AggregationsPostgresDaoTest.kt
```diff
@@ -293,6 +293,90 @@ class AggregationsPostgresDaoTest : CleanDbTestSuiteParallel() {
...
+    }
+  }
+
+  @Test
+  fun findConsecutiveProvenBlobsUnprovenBlobAndBatchAtDifferentIntervals() {
```
jpnovais: I thought we were not supposed to have unproven records in the DB. Are we trying to represent any real scenario here?

However, if the DAO/db was designed for such scenarios, then it shall test it

coordinator/persistence/aggregation/src/integrationTest/kotlin/net/consensys/zkevm/persistence/dao/aggregation/AggregationsPostgresDaoTest.kt
```diff
@@ -293,6 +293,90 @@ class AggregationsPostgresDaoTest : CleanDbTestSuiteParallel() {
...
+    val blob3 = createBlobRecordFromBatches(listOf(batch4))
+
+    val batches = listOf(batch1, batch2, batch3, batch4)
+    val blobs = listOf(blob1, blob2, blob3)
+
+    batches.forEach { insertBatch(it).get() }
+    blobs.forEach { insertBlob(it).get() }
+
+    aggregationsPostgresDaoImpl.findConsecutiveProvenBlobs(0).get().also { blobCounters ->
+      assertThat(blobCounters).hasSameElementsAs(emptyList())
```
jpnovais: I still don't understand well what this test is trying to ensure. Can you please explain?

`findConsecutiveProvenBlobs(startBlockNumber)` is supposed to return result if there is a blob starting at `startBlockNumber` right? If so, shall we call it with 1 instead and return blob1 [0..20]?

coordinator/persistence/aggregation/src/integrationTest/kotlin/net/consensys/zkevm/persistence/dao/aggregation/AggregationsPostgresDaoTest.kt
jonesho: Yes, the Dao and db were designed at the beginning to support unproven records

coordinator/persistence/aggregation/src/integrationTest/kotlin/net/consensys/zkevm/persistence/dao/aggregation/AggregationsPostgresDaoTest.kt
jonesho: > I still don't understand well what this test is trying to ensure. Can you please explain?

I think this test case is to test if the `findConsecutiveProvenBlobs()` would return the result correctly when some batches or blobs in the middle are unproven, even though unproven records will not exist in real scenario, but the dao/db was supporting them at the beginning, so this test is needed to cover that.

jpnovais: For simplicity, I suggest to just remove the support in the DAO for unproven stuff:

- 1. we took the decision that we won't have unproven elements in the DB;
- 2. it's unnecessary complexity that we keep all over :(

jonesho:
Hey @jpnovais , I wanna see your input here, as we agree not to support unproven status in batch, blob, aggregation dao, I'd removed the related codes in dao classes that allow status to be inserted or updated, and ensure status would always be proven, and hence to ensure that dao couldn't be used for inserting records with status other than "proven". To keep the change simple and minimize the breaking changes, the rest (including status enums, mapping functions between status enum and integer value, sql query, db columns) are all left intact, do u think this change of scope is good for this ticket?

coordinator/persistence/aggregation/src/integrationTest/kotlin/net/consensys/zkevm/persistence/dao/aggregation/AggregationsPostgresDaoTest.kt
```diff
@@ -205,8 +162,10 @@ class AggregationsPostgresDaoTest : CleanDbTestSuiteParallel() {
     val blob2 = createBlobRecordFromBatches(listOf(batch3, batch4))
     val blobs = listOf(blob1, blob2)
 
-    batches.forEach { insertBatch(it).get() }
-    blobs.forEach { insertBlob(it).get() }
+    SafeFuture.allOf(
```
jpnovais, line 165: Nice optimization: :+1:

coordinator/persistence/batch/src/main/kotlin/net/consensys/zkevm/persistence/dao/batch/persistence/BatchesPostgresDao.kt
```diff
@@ -19,9 +19,27 @@ class BatchesPostgresDao(
...
+    fun batchStatusToDbValue(status: Batch.Status): Int {
+      // using manual mapping to catch errors at compile time instead of runtime
+      return when (status) {
+        Batch.Status.Finalized -> 1
+        Batch.Status.Proven -> 2
+      }
+    }
+
+    @JvmStatic
+    val batchesDaoTableName = "batches"
```
jpnovais:
```suggestion
    val batchesTableName = "batches"
```

jpnovais (dismissed):

jonesho (approved):

jpnovais (approved):

PR #187, opened 2024-10-15
Decompression Prover input padding and error handling

Tabaie:
During integration testing some proving requests were failing with an error claiming the blob checksum didn't match the one provided by the request.

This error was incorrectly reported because in fact the `assign` function in the prover was failing. However before checking for a returned error, the backend first compared the returned "snark hash" (which was empty) with the given, correct one.

The cause of the actual error was that the data fed into the prover was not the right length (128KB). The fix is for the prover to accept shorter blobs than the full size and pad them with zeros as needed.


gusiri (approved):

gusiri:
@AlexandreBelling Could you please review this?###

PR #195, opened 2024-10-16
JSON-RPC client improvements 

jpnovais:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/JsonRpcRequest.kt
+++ b/jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/JsonRpcRequest.kt
@@ -20,6 +20,13 @@ interface JsonRpcRequest {
     }
 }
 
+internal data class JsonRpcRequestData(
+  @JsonProperty("jsonrpc") override val jsonrpc: String,
+  @JsonProperty("id") override val id: Any,
+  @JsonProperty("method") override val method: String,
+  @JsonProperty("params") override val params: Any
+) : JsonRpcRequest
+
 data class JsonRpcRequestListParams(
   @JsonProperty("jsonrpc") override val jsonrpc: String,
   @JsonProperty("id") override val id: Any,
--- a/jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/JsonRpcRequest.kt
+++ b/jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/JsonRpcRequest.kt
@@ -1,7 +1,6 @@
 package net.consensys.linea.jsonrpc
 
 import com.fasterxml.jackson.annotation.JsonIgnore
-import com.fasterxml.jackson.annotation.JsonProperty
 import java.util.StringJoiner
 
 interface JsonRpcRequest {
@@ -21,17 +20,17 @@ interface JsonRpcRequest {
 }
 
 internal data class JsonRpcRequestData(
-  @JsonProperty("jsonrpc") override val jsonrpc: String,
-  @JsonProperty("id") override val id: Any,
-  @JsonProperty("method") override val method: String,
-  @JsonProperty("params") override val params: Any
+  override val jsonrpc: String,
+  override val id: Any,
+  override val method: String,
+  override val params: Any
 ) : JsonRpcRequest
 
 data class JsonRpcRequestListParams(
-  @JsonProperty("jsonrpc") override val jsonrpc: String,
-  @JsonProperty("id") override val id: Any,
-  @JsonProperty("method") override val method: String,
-  @JsonProperty("params") override val params: List<Any?>
+  override val jsonrpc: String,
+  override val id: Any,
+  override val method: String,
+  override val params: List<Any?>
 ) : JsonRpcRequest {
   override fun toString(): String {
     return StringJoiner(", ", JsonRpcRequestListParams::class.java.simpleName + "[", "]")
@@ -44,10 +43,10 @@ data class JsonRpcRequestListParams(
 }
 
 data class JsonRpcRequestMapParams(
-  @JsonProperty("jsonrpc") override val jsonrpc: String,
-  @JsonProperty("id") override val id: Any,
-  @JsonProperty("method") override val method: String,
-  @JsonProperty("params") override val params: Map<String, *>
+  override val jsonrpc: String,
+  override val id: Any,
+  override val method: String,
+  override val params: Map<String, *>
 ) : JsonRpcRequest {
   override fun toString(): String {
     return StringJoiner(", ", JsonRpcRequestMapParams::class.java.simpleName + "[", "]")
--- a/jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/client/JsonRpcRequestRetryerV2.kt
+++ b/jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/client/JsonRpcRequestRetryerV2.kt
@@ -7,7 +7,6 @@ import com.github.michaelbull.result.Result
 import com.github.michaelbull.result.map
 import com.github.michaelbull.result.mapError
 import com.github.michaelbull.result.onFailure
-import com.github.michaelbull.result.onSuccess
 import io.vertx.core.Vertx
 import net.consensys.linea.async.AsyncRetryer
 import net.consensys.linea.async.RetriedExecutionException
@@ -40,15 +39,21 @@ class JsonRpcRequestRetryerV2(
     return makeRequestWithRetryer(request, resultMapper, shallRetryRequestPredicate)
   }
 
+  private fun shallWarnFailureRetries(retries: Int): Boolean {
+    return requestRetry.failuresWarningThreshold > 0u &&
+      retries > 0 &&
+      (retries % requestRetry.failuresWarningThreshold.toInt()) == 0
+  }
+
   private fun <T> makeRequestWithRetryer(
     request: JsonRpcRequest,
     resultMapper: (Any?) -> T,
     shallRetryRequestPredicate: Predicate<Result<T, Throwable>>
   ): SafeFuture<T> {
-    val lastResult = AtomicReference<T>()
     val lastException = AtomicReference<Throwable>()
     val retriesCount = AtomicInteger(0)
     val requestPredicate = Predicate<Result<T, Throwable>> { result ->
+      log.info("result: {}", result)
       shallRetryRequestsClientBasePredicate.test(result) || shallRetryRequestPredicate.test(result)
     }
 
@@ -58,16 +63,11 @@ class JsonRpcRequestRetryerV2(
       maxRetries = requestRetry.maxRetries?.toInt(),
       timeout = requestRetry.timeout,
       stopRetriesPredicate = { result: Result<T, Throwable> ->
-        result
-          .onSuccess { lastResult.set(it) }
-          .onFailure { lastException.set(it) }
+        result.onFailure(lastException::set)
         !requestPredicate.test(result)
       }
     ) {
-      if (requestRetry.failuresWarningThreshold > 0u &&
-        retriesCount.get() > 0 &&
-        (retriesCount.get() % requestRetry.failuresWarningThreshold.toInt()) == 0
-      ) {
+      if (shallWarnFailureRetries(retriesCount.get())) {
         log.log(
           failuresLogLevel,
           "Request '{}' already retried {} times. lastError={}",
@@ -77,9 +77,7 @@ class JsonRpcRequestRetryerV2(
         )
       }
       retriesCount.incrementAndGet()
-      delegate.makeRequest(request, resultMapper)
-        .toSafeFuture()
-        .thenApply { unfoldResultValueOrException<T>(it) }
+      delegate.makeRequest(request, resultMapper).toSafeFuture().thenApply { unfoldResultValueOrException<T>(it) }
         .exceptionally { th ->
           if (th is Error || th.cause is Error) {
             // Very serious JVM error, we should stop retrying anyway
@@ -88,15 +86,14 @@ class JsonRpcRequestRetryerV2(
             Err(th.cause ?: th)
           }
         }
-    }
-      .handleComposed { result, throwable ->
-        when {
-          result is Ok -> SafeFuture.completedFuture(result.value)
-          result is Err -> SafeFuture.failedFuture<T>(result.error)
-          throwable != null && throwable is RetriedExecutionException -> SafeFuture.failedFuture(lastException.get())
-          else -> SafeFuture.failedFuture(throwable)
-        }
+    }.handleComposed { result, throwable ->
+      when {
+        result is Ok -> SafeFuture.completedFuture(result.value)
+        result is Err -> SafeFuture.failedFuture<T>(result.error)
+        throwable != null && throwable is RetriedExecutionException -> SafeFuture.failedFuture(lastException.get())
+        else -> SafeFuture.failedFuture(throwable)
       }
+    }
   }
 
   companion object {
--- a/jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/client/JsonRpcRequestRetryerV2.kt
+++ b/jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/client/JsonRpcRequestRetryerV2.kt
@@ -1,0 +1,113 @@
+package net.consensys.linea.jsonrpc.client
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.github.michaelbull.result.Err
+import com.github.michaelbull.result.Ok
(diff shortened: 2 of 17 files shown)


jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/JsonRpcRequest.kt
```diff
@@ -20,6 +20,13 @@ interface JsonRpcRequest {
     }
 }
 
+internal data class JsonRpcRequestData(
```
Filter94, line 22: Can't we activate `KotlinModule` so we don't set those `JsonProperty` explicitly?

jpnovais: Kotlin Modile shall be enabled, otherwise it will fail for simple Kotlin types I think. 
Anyway, I don't have a strong preference, I just did it this way to be aligned with `JsonRpcRequest{List/Map}Params`.

jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/client/JsonRpcRequestRetryerV2.kt
```diff
@@ -0,0 +1,113 @@
...
+        .thenApply { unfoldResultValueOrException<T>(it) }
+        .exceptionally { th ->
+          if (th is Error || th.cause is Error) {
+            // Very serious JVM error, we should stop retrying anyway
+            throw th
+          } else {
+            Err(th.cause ?: th)
+          }
+        }
+      future
```
jonesho:
```suggestion
      delegate.makeRequest(request, resultMapper)
        .toSafeFuture()
        .thenApply { unfoldResultValueOrException<T>(it) }
        .exceptionally { th ->
          if (th is Error || th.cause is Error) {
            // Very serious JVM error, we should stop retrying anyway
            throw th
          } else {
            Err(th.cause ?: th)
          }
        }
```
Is it possible to simplify a bit as above?

jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/JsonRpcRequest.kt
Filter94: I created a [branch](https://github.com/Consensys/linea-monorepo/pull/206/files) to test if we can remove them. Let me run an E2E tests there

Filter94: We're dealing with Vertx's Json, not Jackson directly in this case. I had to explicitly register KotlinModule on the codec in order to make it work in that PR

jpnovais: Vertx uses [jackson under the hood](https://vertx.io/docs/vertx-core/java/#_mapping_between_json_objects_and_java_objects). 

Unfortunately, that change breaks tests elsewhere. I suggest to leave that improvement to another time as it's a minor thing and move forward with this.

Filter94: I fixed that as well. Will see if my fix works

jpnovais: Ok, great! thanks for the contribution. 
FYI, I rebased with main.

Filter94: Tests passed

jonesho (dismissed):
Approved as the feature implemetation looks good, but just the improvement on syntax stuff was left

jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/client/JsonRpcRequestRetryerV2.kt
```diff
@@ -0,0 +1,112 @@
...
+      timeout = requestRetry.timeout,
+      stopRetriesPredicate = { result: Result<T, Throwable> ->
+        result
+          .onSuccess { lastResult.set(it) }
+          .onFailure { lastException.set(it) }
+        !requestPredicate.test(result)
+      }
+    ) {
+      if (requestRetry.failuresWarningThreshold > 0u &&
+        retriesCount.get() > 0 &&
```
Filter94: We can put this `retriesCount.get()` into a variable to reduce verbosity

jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/client/JsonRpcRequestRetryerV2.kt
```diff
@@ -0,0 +1,112 @@
...
+  ): SafeFuture<T> {
+    return makeRequestWithRetryer(request, resultMapper, shallRetryRequestPredicate)
+  }
+
+  private fun <T> makeRequestWithRetryer(
+    request: JsonRpcRequest,
+    resultMapper: (Any?) -> T,
+    shallRetryRequestPredicate: Predicate<Result<T, Throwable>>
+  ): SafeFuture<T> {
+    val lastResult = AtomicReference<T>()
```
Filter94: I can see the set for it, but it seems we never use it. Can we remove it?

jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/client/JsonRpcRequestRetryerV2.kt
```diff
@@ -0,0 +1,112 @@
...
+          .onSuccess { lastResult.set(it) }
+          .onFailure { lastException.set(it) }
+        !requestPredicate.test(result)
+      }
+    ) {
+      if (requestRetry.failuresWarningThreshold > 0u &&
+        retriesCount.get() > 0 &&
+        (retriesCount.get() % requestRetry.failuresWarningThreshold.toInt()) == 0
+      ) {
+        log.log(
```
Filter94, line 71: If we move this logging inside `stopRetriesPredicate` can we get rid of atomic wrappers?

jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/client/SequentialIdSupplier.kt
```diff
@@ -0,0 +1,21 @@
+package net.consensys.linea.jsonrpc.client
+
+import java.math.BigInteger
+import java.util.function.Supplier
+import javax.annotation.concurrent.ThreadSafe
+
+@ThreadSafe
+class SequentialIdSupplier : Supplier<Any> {
```
Filter94, line 8: Can we turn this into an object? Or you want a class for some tests?

jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/client/JsonRpcV2ClientImpl.kt
```diff
@@ -0,0 +1,28 @@
...
+
+import com.github.michaelbull.result.Result
+import net.consensys.linea.jsonrpc.JsonRpcRequestData
+import tech.pegasys.teku.infrastructure.async.SafeFuture
+import java.util.function.Predicate
+import java.util.function.Supplier
+
+internal class JsonRpcV2ClientImpl(
+  private val delegate: JsonRpcRequestRetryerV2,
+  private val idSupplier: Supplier<Any>
```
Filter94, line 11: I know that Id can be a string, but here I think we would be just fine with a `Supplier<ULong>`. And I think restricting it to be a numeric type both improves clarity and may prevent some bugs.

jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/JsonRpcRequest.kt
jpnovais: Hi, I was keen on onboarding the changes. However, after seeing this [fix](https://github.com/Consensys/linea-monorepo/pull/206/files#diff-5b3914b01126db2aa8b6e484a6a0687af392d993970a7aedc20425fe1c3955fdR1164-R1167) I don't think it's a good idea:

- As things are (in JsonMessageProcessor, etc..), removing the annotations will require a non-obvious side-effect of setting up Vertx's Jackson codec. IMO, this is worse than having the @JsonProperty explicitly defined.
- That fix is using a deprecated API, we shall not do that - I dint't check if there is replacement. 

So, I suggest to keep the annotations for now.

Filter94: * Now we're only adding KotlinModule to Vertx's Object Mapper in Coordinator

* I removed usage of deprecated API and re-configuration. We're using a separate ObjectMapper properly configured by builder for that sole failing conversion now

jvm-libs/json-rpc/src/test/kotlin/net/consensys/linea/jsonrpc/client/JsonRpcV2ClientImplTest.kt
```diff
@@ -0,0 +1,671 @@
...
+    wiremock.stop()
+  }
+
+  private fun WireMockServer.jsonRequest(requestIndex: Int = 0): String {
+    return this.serveEvents.serveEvents[requestIndex].request.bodyAsString
+  }
+
+  @Test
+  fun `when request is a list of params shall serialize to json array`() {
+    replyRequestWith(200, """{"jsonrpc": "2.0", "id": 1, "result": "OK"}""")
```
Filter94: We can reduce this copypaste a bit by either extracting `"""{"jsonrpc": "2.0", "id": 1, "result": "OK"}"""` into a constant or maybe even set these values as default

jvm-libs/json-rpc/src/test/kotlin/net/consensys/linea/jsonrpc/client/JsonRpcV2ClientImplTest.kt
```diff
@@ -0,0 +1,671 @@
...
+            "key3" to listOf(1, 2, 3),
+            "key4" to null
+          )
+        )
+        assertThat(cause.rpcErrorData).isEqualTo(expectedData)
+      }
+  }
+
+  @Test
+  fun `when it gets an error propagates to shallRetryRequestPredicate  and retries while is true`() {
```
Filter94:
```suggestion
  fun `when it gets an error propagates to shallRetryRequestPredicate and retries while is true`() {
```

jvm-libs/json-rpc/src/test/kotlin/net/consensys/linea/jsonrpc/client/JsonRpcV2ClientImplTest.kt
```diff
@@ -0,0 +1,671 @@
...
+          retryPredicateCalls.add(it)
+          it != Ok("EXPECTED RESULT")
+        },
+        resultMapper = {
+          it as String?
+          it?.uppercase()
+        }
+      ).get()
+
+      assertThat(wiremock.serveEvents.serveEvents.size).isEqualTo(6)
```
Filter94: Small suggestion. We can put list of requests into a separate variable and use its size instead of 6, so reader doesn't have to think about how many elements are in that list

jvm-libs/json-rpc/src/test/kotlin/net/consensys/linea/jsonrpc/client/JsonRpcV2ClientImplTest.kt
```diff
@@ -0,0 +1,671 @@
...
+          assertThat(it.cause).isInstanceOfSatisfying(ConnectException::class.java) {
+            assertThat(it.message).contains("Connection refused: localhost/127.0.0.1:")
+          }
+        }
+      assertThat(retryPredicateCalls).hasSizeBetween(1, 3)
+    }
+  }
+
+  @Test
+  fun `when shared predicate is defined shall retry when any of then returns true`() {
```
Filter94:
```suggestion
  fun `when shared predicate is defined shall retry when any of them returns true`() {
```

Filter94: I don't understand purpose of this test TBH. What is a shared predicate? I can only see 1 usage of it, so to me it's exclusive, not shared

Filter94 (dismissed):

jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/client/SequentialIdSupplier.kt
jpnovais: It was initially an object, but then I thought for testing isolation and use inside the factory a class would be more suitable. So I will keep it as a class for now.

jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/client/JsonRpcV2ClientImpl.kt
jpnovais: I also took that into consideration initially. The reasons I went with Any, aka Support for String was: 
- For applications with long uptime and many requests we may run out of space with Int/Longs, hence I use a BigInteger in SequentialIdSupplier;
- Allowing strings, allow for later having prefixed Id in the requests e.g `coordiantor-123`
So I will leave it as is

Filter94: If it's running long enough I'm afraid we'd get into memory consumption issue by `BigInteger`. We can solve it by resetting the ULong. We need id that is unique for inflight requests after all, not unique ever.

Not that it's a legit performance concern, but I'm just saying that ULong fits here as well

jvm-libs/json-rpc/src/test/kotlin/net/consensys/linea/jsonrpc/client/JsonRpcV2ClientImplTest.kt
jpnovais: "shared" its an **optional predicate** that's is passed to the constructor that is applied to every request before the predicate  passed in `makeRequest`. The intent is to pass generic predicates in the constructor for generic errors as Connection error, etc..

jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/client/JsonRpcV2ClientImpl.kt
jpnovais: > If it's running long enough I'm afraid we'd get into memory consumption issue by BigInteger. We can solve it by resetting the ULong. We need id that is unique for inflight requests after all, not unique ever.

That makes me want to change the internal implementation to use ULong, however, I still prefer keeping the interface generic.

jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/client/JsonRpcRequestRetryerV2.kt
jpnovais: Good catch, thanks :+1:

jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/client/JsonRpcRequestRetryerV2.kt
jpnovais: Although atm it's running in the same thread an seems safe. In the event of refactor of retrier on delegate client that may change and leave door to a bug. I feel more comfortable to keep the wrappers

jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/client/JsonRpcRequestRetryerV2.kt
jpnovais: did a small refactor, looks better now.

jvm-libs/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/client/JsonRpcV2ClientImpl.kt
jpnovais: Changed to use atomic long

julien-marchand (approved):

Filter94 (approved):

Issue #181, opened 2024-10-14
Refactor CI to push Docker images  from main


julien-marchand:
## Description
Refactor CI to push Docker images from main

## Motivation
Have more transparency on where images have been built. Ensure all deployed software has been built from main. Streamline the release process. Ensure no docker image is pushed from branches and forks.

## Tasks
- [ ] Reuse locally built images when running e2e tests
- [ ] Don't push docker images on branches/PR
- [ ] Push images at the end of the workflow on main

## Risks
- [ ] Make it harder to debug failures in e2e related to runners/CI

PR #200, opened 2024-10-16
GHA - Only publish docker builds on main branch

bradbown:
This PR implements issue(s) #181

### Checklist


.github/actions/check-image-tags-exist/action.yml
```diff
@@ -49,4 +43,4 @@ runs:
       shell: bash
       run: |
         echo "last_commit_tag_exists: ${{ steps.last_commit_image_exists.outputs.last_commit_tag_exists }}"
-        echo "common_ancestor_commit_tag_exists: ${{ steps.ancestor_commit_image_exists.outputs.common_ancestor_commit_tag_exists }}"
\ No newline at end of file
+        echo "common_ancestor_commit_tag_exists: ${{ steps.ancestor_commit_image_exists.outputs.common_ancestor_commit_tag_exists }}"
```
jonesho: Why we still need this? can remove this echo command as `common_ancestor_commit` will not be used

jonesho: Can we remove the above step with id = "ancestor_commit_image_exists" as well?

.github/actions/image-tag-and-push/action.yml
```diff
@@ -50,29 +41,21 @@ runs:
...
-      if: ${{ inputs.last_commit_tag == '0000000' && inputs.common_ancestor_commit_tag_exists == '0' }}
-      run: |
-        docker buildx imagetools create --tag ${{ inputs.image_name }}:${{ inputs.commit_tag }} ${{ inputs.image_name }}:${{ inputs.common_ancestor_tag }}
-        docker buildx imagetools create --tag ${{ inputs.image_name }}:${{ inputs.commit_tag }}-${{ inputs.untested_tag_suffix }} ${{ inputs.image_name }}:${{ inputs.common_ancestor_tag }}
         echo IMAGE_TAGGED=true >> $GITHUB_ENV
     - name: Tag Docker image with develop if on main branch
       shell: bash
       if: ${{ github.ref == 'refs/heads/main' && inputs.last_commit_tag_exists == '0' }}
       run: |
         docker buildx imagetools create --tag ${{ inputs.image_name }}:${{ inputs.develop_tag }} ${{ inputs.image_name }}:${{ inputs.last_commit_tag }}
```
jonesho, line 49: Here we probably could add the below command to tag the image as current commit tag if on main:
```
docker buildx imagetools create --tag ${{ inputs.image_name }}:${{ inputs.commit_tag }} ${{ inputs.image_name }}:${{ inputs.last_commit_tag }}
```
and we could remove the step of "Tag Docker image with last commit tag with the commit hash plus" above, given that there's no sense to re-tag the image as current commit tag on feature branch, given that the image from feature branch would no longer push into docker hub, wdyt?

.github/actions/image-tag-and-push/action.yml
```diff
@@ -50,29 +41,21 @@ runs:
       shell: bash
       run: |
         echo IMAGE_TAGGED=false >> $GITHUB_ENV
-    - name: Tag Docker image with last commit tag with the commit hash plus w/o "untested" suffix
+    - name: Tag Docker image with last commit tag with the commit hash plus
```
jonesho: We might be able to remove this entire step, see below comment

.github/workflows/coordinator-build-and-publish.yml
```diff
@@ -38,10 +40,14 @@ jobs:
     env:
       COMMIT_TAG: ${{ inputs.commit_tag }}
       DEVELOP_TAG: ${{ inputs.develop_tag }}
```
jonesho, line 42: Can we remove `DEVELOP_TAG` as it's not used anywhere?

.github/workflows/postman-build-and-publish.yml
```diff
@@ -38,9 +40,14 @@ jobs:
     env:
       COMMIT_TAG: ${{ inputs.commit_tag }}
       DEVELOP_TAG: ${{ inputs.develop_tag }}
```
jonesho, line 42: Can we remove DEVELOP_TAG as not used?

.github/workflows/prover-build-and-publish.yml
```diff
@@ -41,9 +43,14 @@ jobs:
     env:
       COMMIT_TAG: ${{ inputs.commit_tag }}
       DEVELOP_TAG: ${{ inputs.develop_tag }}
```
jonesho, line 45: Can we remove DEVELOP_TAG as not used?

.github/workflows/traces-api-facade-build-and-publish.yml
```diff
@@ -38,9 +40,14 @@ jobs:
     env:
       COMMIT_TAG: ${{ inputs.commit_tag }}
       DEVELOP_TAG: ${{ inputs.develop_tag }}
```
jonesho, line 42: DEVELOP_TAG can be removed as not used

.github/actions/image-tag-and-push/action.yml
jpnovais: I think we may optimize this a step further: I am not confident on how buildx imagetools  works, but if it creates/build the image everytime we run `docker buildx imagetools create`, we shall call it only once and then just retag it.

.github/actions/check-image-tags-exist/action.yml
jonesho: Why do we still need to echo `common_ancestor_commit_tag_exists`? in fact we could remove the above step `ancestor_commit_image_exists`, right?

.github/workflows/coordinator-build-and-publish.yml
bradbown: Instead of deleting the env var, I replaced the usage of inputs with the appropriate env var.

.github/actions/image-tag-and-push/action.yml
jonesho: `docker buildx imagetools create -tag` is just for tagging the given image, we won't build any image from this command.

PR #214, opened 2024-10-21
fixed the bug in PolyEval

### LINKED ISSUE #192: bug in genrating Random Linear Combination of expressions and extracting the assignment.
## Impact
__environment [Testnet/Mainnet]__

__an user is impacted [Y/N], if yes type of user: [External/Internal]__

## Describe the bug 
for special edge cases, one of the following functions is not working correctly.

`func RandLinCombColSymbolic(x coin.Info, hs []ifaces.Column) *symbolic.Expression {}`

` func EvalExprColumn(run *wizard.ProverRuntime, board symbolic.ExpressionBoard) smartvectors.SmartVector {}`

### Steps to Reproduce
for the moment, we are not able to reproduce the bug independently via, e.g., test cases. it shows itself when running
`func TestSelfRecursionMultiLayered(t *testing.T) {}`.

This bug is temporality ignored by replacing the original function

```
func RandLinCombColSymbolic(x coin.Info, hs []ifaces.Column) *symbolic.Expression {
	cols := make([]*symbolic.Expression, len(hs))
	for c := range cols {
		cols[c] = ifaces.ColumnAsVariable(hs[c])
	}
	expr := symbolic.NewPolyEval(x.AsVariable(), cols)
	return expr
}
```
With what is supposed to be equivalent as
```
func RandLinCombColSymbolic(x coin.Info, hs []ifaces.Column) *symbolic.Expression {
	res := symbolic.NewConstant(0)
	for i := len(hs) - 1; i >= 0; i-- {
		res = symbolic.Mul(res, x)
		res = symbolic.Add(res, hs[i])
	}
	return res
}
```

### more cases
 also, check the test  `TestInnerProduct()` that is not working alongside symbolic.NewPolyEval()
 see  https://github.com/Consensys/linea-monorepo/blob/main/prover/protocol/compiler/innerproduct/context.go

### Actual result/Expected result
swapping the mentioned function should not affect the tests, as they are expected to be equivalent. However, some tests do not pass the original function.

Soleimani193:
This PR  solves issue #192 for the case of PolyEval over Product and also  PolyEval over PolyEval happening in the inner-product compiler and selfrecursion. For the product operator Product.Evaluate()  outputs *Pooled type that was not supported as input to PolyEval. 

close #192 

### Checklist

### DIFF
--- a/prover/symbolic/polyeval.go
+++ b/prover/symbolic/polyeval.go
@@ -52,6 +52,11 @@ func NewPolyEval(x *Expression, coeffs []*Expression) *Expression {
 	}
 }
 
+// Name outputs the name for the PolyEval [Operator].
+func (PolyEval) Name() string {
+	return "PolyEval"
+}
+
 /*
 Returns the degree of the operation given, as input, the degree of the children
 */
--- a/prover/symbolic/polyeval.go
+++ b/prover/symbolic/polyeval.go
@@ -52,11 +52,6 @@ func NewPolyEval(x *Expression, coeffs []*Expression) *Expression {
 	}
 }
 
-// Name outputs the name for the PolyEval [Operator].
-func (PolyEval) Name() string {
-	return "PolyEval"
-}
-
 /*
 Returns the degree of the operation given, as input, the degree of the children
 */
--- a/prover/maths/common/smartvectors/arithmetic_basic.go
+++ b/prover/maths/common/smartvectors/arithmetic_basic.go
@@ -117,6 +117,10 @@ func PolyEval(vecs []SmartVector, x field.Element, p ...mempool.MemPool) (result
 			anyReg = true
 			v := *casted
 			accumulateReg(resReg, v, xPow)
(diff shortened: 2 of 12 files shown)


prover/symbolic/polyeval.go
```diff
@@ -52,6 +52,11 @@ func NewPolyEval(x *Expression, coeffs []*Expression) *Expression {
 	}
 }
 
+// Name outputs the name for the PolyEval [Operator].
+func (PolyEval) Name() string {
```
AlexandreBelling: You can use `reflect.TypeOf(o).String()` or `fmt.Printf("typeName=%T", o)`

AlexandreBelling (dismissed):

gusiri (dismissed):

gusiri (approved):

arijitdutta67 (approved):

AlexandreBelling (approved):

Issue #128, opened 2024-10-02
Improve the tx history in the Bridge UI


VGau:
## Description
The transaction history in the Bridge UI is not updating correctly sometime and it's taking quite some time. We need to ensure accurate and timely updates without further degrading performance.

## Motivation
Improving the transaction history will provide users with a better experience.

## Tasks
- [ ] Identify performance bottlenecks related to transaction history loading and updates
- [ ] Implement efficient solutions to load the transaction history
- [ ] Optimize performance to minimize UI slowdowns

## Acceptance criteria
- [ ] Transaction history loads promptly and accurately
- [ ] Performance is improved
- [ ] No major slowdowns in the UI after the fix

## Risks
- [ ] Further performance issues if not optimized properly
- [ ] Risk of incomplete or delayed transaction history updates

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #216, opened 2024-10-21
Fix(128): improve bridge UI history

VGau:
This PR implements issue(s) #128 

### Checklist


bridge-ui/src/services/tokenService.ts
```diff
@@ -170,3 +171,58 @@ export async function fetchTokenPrices(
...
+          symbol: token.symbol,
+          decimals: token.decimals,
+          type: tokenType,
+          L1: token?.extension?.rootAddress ?? null,
+          L2: token.address,
+          UNKNOWN: null,
+          image: token.logoURI,
+          isDefault: true,
+        };
+      }),
```
thedarkjester: the object construction is the same as that of Sepolia - you could build a generic object builder from token object and return that instead of two places doing exactly the same thing.

thedarkjester (approved):

PR #230, opened 2024-10-22
Blob Decompressor and Some Helpers

jpnovais:
This PR implements issue(s) #

### Checklist


jvm-libs/linea/blob-compressor/src/main/kotlin/net/consensys/linea/blob/GoNativeBlobCompressor.kt
```diff
@@ -1,8 +1,8 @@
 package net.consensys.linea.blob
 
+import build.linea.jvm.ResourcesUtil.copyResourceToTmpDir
 import com.sun.jna.Library
 import com.sun.jna.Native
-import net.consensys.jvm.ResourcesUtil.copyResourceToTmpDir
 
 interface GoNativeBlobCompressor {
```
jonesho, line 7: Can we name this as `BlobCompressor` to align with `BlobDecompressor`?

jvm-libs/linea/blob-decompressor/src/main/kotlin/net/consensys/linea/blob/GoNativeBlobDecompressor.kt
```diff
@@ -0,0 +1,105 @@
...
+
+internal interface GoNativeBlobDecompressorJnaLib : GoNativeBlobDecompressorJnaBinding, Library
+
+enum class BlobDecompressorVersion(val version: String) {
+  V1_1_0("v1.1.0")
+}
+
+class GoNativeBlobDecompressorFactory {
+  companion object {
+    private const val DICTIONARY_NAME = "compressor_dict.bin"
```
jonesho, line 89: Is this "compressor_dict.bin" required to be the same as the one used by `BlobCompressor`? If so, should we make sure they're referring to the same file instead of having two same copies for `BlobCompressor` and `BlobDecompressor` in the repo?

jonesho (commented):
Just two quesstions from my side, other looks good to me

jvm-libs/linea/blob-compressor/src/main/kotlin/net/consensys/linea/blob/GoNativeBlobCompressor.kt
jpnovais: That would require a deeper refactor to be properly done.  
`BlobDecompressor` is a high level abstraction, whereas `GoNativeBlobCompressor` is mapping/binding to the Go interface. Therefore the names are correct. We just don't have that abstraction for the compressor here and it's out of scope of this PR.

jvm-libs/linea/blob-decompressor/src/main/kotlin/net/consensys/linea/blob/GoNativeBlobDecompressor.kt
jpnovais: It is for now, by coincidence. 
However, if tomorrow we have a new compressor with a new dictionary, then the decompressor will need to have both files because it needs to be backward compatible.

jpnovais:
@jonesho questions addressed. Please take a 2nd look.

jvm-libs/linea/blob-compressor/src/main/kotlin/net/consensys/linea/blob/GoNativeBlobCompressor.kt
jonesho: If that's the case, why do we name the new interface as `BlobDecompressor` instead of `GoNativeBlobDecompressor`?

jpnovais: Read the code carefully, as I explained `BlobDecompressor` is a higher level abstraction. 

Ther is an other interface `GoNativeBlobDecompressorJnaBinding` which is the equivalent of `GoNativeBlobCompressor `. The problem is the Compressor poor decoupling, not the decompressor

jonesho (dismissed):

jonesho (approved):

jvm-libs/linea/blob-decompressor/src/test/kotlin/net/consensys/linea/blob/RLPHelper.kt
```diff
@@ -0,0 +1,26 @@
...
+
+internal fun rlpEncode(list: List<ByteArray>): ByteArray {
+  return RLP.encode { rlpWriter ->
+    rlpWriter.startList()
+    list.forEach { bytes ->
+      rlpWriter.writeBytes(Bytes.wrap(bytes))
+    }
+    rlpWriter.endList()
+  }.toArray()
+}
```
jonesho, line 14: @jpnovais A late question, not a blocker to this PR, why not put them in `ByteArrayExtension`?

jpnovais: Because generic/extensions/XYZ, must be vanila packages that do not bring extra dependencies. To do that, I would need to add Besu libraries to `generic/extensions/kotlin`.

jvm-libs/linea/blob-decompressor/src/test/kotlin/net/consensys/linea/blob/GoNativeBlobDecompressorTest.kt
```diff
@@ -0,0 +1,42 @@
...
+    assertTrue(compressor.Write(blocks[0], blocks[0].size))
+    assertTrue(compressor.Write(blocks[1], blocks[1].size))
+
+    val compressedData = ByteArray(compressor.Len())
+    compressor.Bytes(compressedData)
+
+    val decompressedBlob = decompressor.decompress(compressedData)
+    assertThat(decompressedBlob.size).isGreaterThan(compressedData.size)
+    val decompressedBlocks: List<ByteArray> = rlpDecodeAsListOfBytes(decompressedBlob)
+    assertThat(decompressedBlocks).hasSize(2)
```
Filter94, line 40: Why don't we compare the decompressed items to original ones to check if compression - decompression mutates the original data?

jvm-libs/linea/blob-decompressor/src/test/kotlin/net/consensys/linea/blob/BlobDecompressorDataDecodingTest.kt
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  @Test
+  fun `can deserialize native lib testdata blobs`() {
+    val blob = findPathTo("prover")!!
+      .resolve("lib/compressor/blob/testdata/v0/sample-blob-0151eda71505187b5.bin")
+      .readBytes()
+    val decompressedBlob = decompressor.decompress(blob)
+    val blocksRlpEncoded = rlpDecodeAsListOfBytes(decompressedBlob)
+    assertThat(blocksRlpEncoded).hasSize(254)
```
Filter94, line 29: Here as well, why don't we check the data itself?

jvm-libs/linea/blob-decompressor/build.gradle
```diff
@@ -0,0 +1,59 @@
...
+compileKotlin {
+  dependsOn tasks.downloadNativeLibs
+}
+
+task cleanResources(type: Delete) {
+  fileTree(project.layout.projectDirectory.dir('src/main/resources'))
+    .filter {
+      it.name.endsWith(".so") || it.name.endsWith(".dll") || it.name.endsWith(".dylib")
+    }.each {
+      println("Deleting: ${it}")
```
Filter94, line 54: I believe it's a debug leftover. Can we remove it?

jvm-libs/linea/blob-decompressor/src/test/kotlin/net/consensys/linea/blob/GoNativeBlobDecompressorTest.kt
jpnovais: > to check if compression - decompression mutates the original data?

Our compression is lossy, therefore that is not possible to recover fully the same data.

Want later on to extend the test with deserialization of the RLP and compare if the date is correct, but for that need changes in the Besu classes to allow decoding without signature verification.

jvm-libs/linea/blob-decompressor/src/test/kotlin/net/consensys/linea/blob/BlobDecompressorDataDecodingTest.kt
jpnovais: See my comment [here](https://github.com/Consensys/linea-monorepo/pull/230#discussion_r1812261637)

jvm-libs/linea/blob-decompressor/src/test/kotlin/net/consensys/linea/blob/GoNativeBlobDecompressorTest.kt
Filter94: Makes sense, thanks

jvm-libs/linea/blob-decompressor/build.gradle
jpnovais: Yes, I will. Thanks 👍

jvm-libs/generic/extensions/kotlin/src/main/kotlin/build/linea/URIExtensions.kt
```diff
@@ -0,0 +1,16 @@
+package build.linea
+
+import java.net.URI
+
+fun URI.getPortWithSchemeDefaults(): Int {
```
Filter94, line 5: Typo, `Scheme` -> `Schema`

jpnovais: It was actually on purpose. My initial implementation was `Shema`, but then I saw that Java uses `scheme` and googled and the are synonyms, so I went to the [RFC](https://datatracker.ietf.org/doc/html/rfc3986) to check the official term and it's scheme. So I made the change to [scheme](https://github.com/Consensys/linea-monorepo/pull/230/commits/8a01ae95681ba22288ce9cba6cd36b6dc6ee27cf).

Filter94: Ok, good to know

PR #234, opened 2024-10-23
Fix(133): Improve e2e tests performance

VGau:
This PR implements issue(s) #133 

### Checklist


.github/actions/setup-nodejs/action.yml
```diff
@@ -9,7 +9,7 @@ inputs:
   pnpm-version:
     description: 'The version of pnpm to use'
     required: true
-    default: '9.1.1'
+    default: '9.12.2'
```
thedarkjester, line 12: is there documentation that needs to be updated to match this? readmes etc.

.github/workflows/reuse-run-e2e-tests.yml
```diff
@@ -151,7 +151,7 @@ jobs:
           docker container ls -a
       - name: Run e2e tests
         id: run_e2e_tests
-        timeout-minutes: 25
+        timeout-minutes: 4
```
thedarkjester, line 154: 😀

Makefile
```diff
@@ -68,17 +68,19 @@ compile-contracts-no-cache:
 deploy-linea-rollup:
```
thedarkjester: should this be deploy-linea-rollupV5

contracts/local-deployments-artifacts/deployL2MessageService.ts
```diff
@@ -0,0 +1,85 @@
...
+    TransparentUpgradeableProxyAbi,
+    TransparentUpgradeableProxyBytecode,
+    wallet,
+    l2MessageServiceImplementationAddress,
+    proxyAdminAddress,
+    initializer,
+  );
+
+  console.log(
+    `${messageServiceName} deployed: address=${await proxyContract.getAddress()} blockNumber=${(await proxyContract.deploymentTransaction()?.wait())?.blockNumber}`,
```
thedarkjester: if logging doesn't happen / is disabled, the deployment will fail, ideally it is safer and cleaner to strip it out vs. inline

e2e/src/config/jest/global-setup.ts
```diff
@@ -11,31 +11,41 @@ declare global {
...
+    ).wait(),
+  ]);
+
+  console.log(`L1 Dummy contract deployed at address: ${await dummyContract.getAddress()}`);
+  console.log(`L2 Dummy contract deployed at address: ${await l2DummyContract.getAddress()}`);
+  console.log(`L2 Test contract deployed at address: ${await l2TestContract.getAddress()}`);
 
   console.log("Generating L2 traffic...");
-  const stopPolling = await sendTransactionsToGenerateTrafficWithInterval(l2Account, 2_000);
+  const pollingAccount = await config.getL2AccountManager().generateAccount(etherToWei("200"));
```
jonesho, line 47: Why not just use `l2Account` here but instead spending time to create a new account? Any particular reason?

e2e/src/config/tests-config/accounts/account-manager.ts
```diff
@@ -48,72 +53,61 @@ abstract class AccountManager implements IAccountManager {
     return { account: whaleAccount, accountWallet: whaleTxManager };
   }
```
jonesho, line 55: Just to confirm, each of the parellelly running e2e tests will have its own `JEST_WORKER_ID`, right?

e2e/src/config/tests-config/accounts/account-manager.ts
```diff
@@ -48,72 +53,61 @@ abstract class AccountManager implements IAccountManager {
...
-  ): Promise<Wallet[]> {
+  async generateAccounts(numberOfAccounts: number, initialBalanceWei = etherToWei("10")): Promise<Wallet[]> {
     const { account: whaleAccount, accountWallet: whaleAccountWallet } = this.selectWhaleAccount();
 
     console.log(
       `Generating accounts: chainId=${this.chainId} numberOfAccounts=${numberOfAccounts} whaleAccount=${whaleAccount.address}`,
     );
 
     const accounts: Account[] = [];
+    const promises: TransactionResponse[] = [];
```
jonesho: Chore: can we name this as just `transactionResponses` ? it was not a list of `Promise` at all but just a list of object having a field of `Promise` type

e2e/src/submission-finalization.spec.ts
```diff
@@ -152,21 +129,20 @@ describe("Submission and finalization test suite", () => {
...
 
-      let safeL2BlockNumber = -1,
-        finalizedL2BlockNumber = -1;
+      let safeL2BlockNumber = 0,
+        finalizedL2BlockNumber = 0;
       while (
-        safeL2BlockNumber < parseInt(lastFinalizedL2BlockNumberOnL1) ||
-        finalizedL2BlockNumber < parseInt(lastFinalizedL2BlockNumberOnL1)
+        safeL2BlockNumber < lastFinalizedL2BlockNumberOnL1 ||
+        finalizedL2BlockNumber < lastFinalizedL2BlockNumberOnL1
```
jonesho, line 145: Can we revert the above changes? Cause essentially what we're doing here is just bypassing the while loop every time and check if `safeL2BlockNumber` and `safeL2BlockNumber` are equal to 0

jonesho (commented):
LGTM, but just that the `submission-finalization` e2e test needs to amend

e2e/src/submission-finalization.spec.ts
VGau: @jpnovais Any though on this ?

e2e/src/config/jest/global-setup.ts
VGau: to avoid potential nonce issue with tests suites

e2e/src/config/tests-config/accounts/account-manager.ts
VGau: yes each tests suite will have its own JEST_WORKER_ID. So if 5 tests suites are running in parallel, JEST_WORKER_IDs will be 1, 2, 3, 4, 5. So each test suite will use a different whale account

thedarkjester (approved):

jonesho (approved):

PR #238, opened 2024-10-23
State recover: Blobscan and Execution Layer client

jpnovais:
### Note
Need to define
```
group = 'build.linea.staterecover'
```

because of this [issue](https://discuss.gradle.org/t/multi-module-project-with-sub-modules-with-same-name/31928)

Otherwise, if a module depends on `jvm-libs:linea:core:domain-models` and `staterecover:appcore:domain-models` it will have issues, only one gets imported because both have same end module `domain-models`

### Checklist


state-recover/appcore/clients-interfaces/build.gradle
```diff
@@ -0,0 +1,11 @@
+plugins {
+  id 'net.consensys.zkevm.kotlin-library-conventions'
+}
+
+group = 'build.linea.staterecover'
```
jpnovais, line 5: need to specify this because if this gradle [issue](https://discuss.gradle.org/t/multi-module-project-with-sub-modules-with-same-name/31928)

state-recover/clients/execution-layer-json-rpc-client/src/main/kotlin/build/linea/staterecover/clients/el/ELJsonRpcClient.kt
```diff
@@ -0,0 +1,72 @@
...
+import net.consensys.fromHexString
+import net.consensys.linea.BlockNumberAndHash
+import net.consensys.linea.BlockParameter
+import net.consensys.linea.jsonrpc.client.JsonRpcClientFactory
+import net.consensys.linea.jsonrpc.client.JsonRpcV2Client
+import net.consensys.linea.jsonrpc.client.RequestRetryConfig
+import tech.pegasys.teku.infrastructure.async.SafeFuture
+import java.net.URI
+
+class ELJsonRpcClient internal constructor(
```
jonesho: I'd prefer a more descriptive class name like ExecutionLayerJsonRpcClient

jvm-libs/generic/extensions/kotlin/src/test/kotlin/net/consensys/ByteArrayExtensionsTest.kt
```diff
@@ -111,4 +111,12 @@ class ByteArrayExtensionsTest {
       .isInstanceOf(AssertionError::class.java)
       .hasMessage("slice 64..95 is out of array size=80")
   }
+
+  @Test
+  fun toULongFromLast8Bytes() {
+    assertThat(byteArrayOf(0x00).toULongFromLast8Bytes(lenient = true)).isEqualTo(0uL)
+    assertThat(byteArrayOf(0x01).toULongFromLast8Bytes(lenient = true)).isEqualTo(1uL)
+    val max = ByteArray(32) { 0xff.toByte() }
+    assertThat(max.toULongFromLast8Bytes()).isEqualTo(ULong.MAX_VALUE)
```
jonesho, line 120: Can we improve the test a bit by testing that it actually grabs the last 8 bytes, sth like:
```
val max = ByteArray(32) { 0xff.toByte() }.apply { set(31, 0xfe.toByte()) }
assertThat(max.toULongFromLast8Bytes()).isEqualTo(ULong.MAX_VALUE - 1UL)
```

jonesho (dismissed):
LGTM, with minor suggestions

jvm-libs/generic/extensions/kotlin/src/test/kotlin/net/consensys/ByteArrayExtensionsTest.kt
jpnovais: That's a good suggestion. Thanks 👍

jpnovais:
@jonesho addressed your feedback. Please take a look, thanks.

jonesho (approved):

PR #247, opened 2024-10-25
[Fix] G-2,3,4 and 5 - Use return vars on function and optimize reverts

thedarkjester:
This PR addresses issues G-2,3,4 and 5 where the calldata array length is no longer cached, named variables are used for most of the function returns, and some cached storage variables on the initializer.

There are also fast returns in place for blob submission and calldata submission. Note: `eth_call` is generally performed and would fail anyway, but this simplifies it.

### Checklist


devdacian:
Fixes look good to me.

VGau (approved):

VGau (approved):

VGau (approved):

VGau (approved):

julien-marchand (approved):

PR #260, opened 2024-10-28
Reduce memory usage of different pods

Filter94:
Signed-off-by: Ameziane H <ameziane.hamlat@consensys.net>

This PR implements issue(s) #

### Checklist


julien-marchand (approved):

ahamlat (approved):
Let's start with this configuration and see if the suggested -Xmx are sufficient. We can also add -XX:+HeapDumpOnOutOfMemoryError `-XX:HeapDumpPath=/path-to-store-heap-dumps/heapdump-{name-of-the-pod}.hprof`. This will generate heap dumps on OOM errors.

docker/compose.yml
```diff
@@ -141,11 +141,12 @@ services:
     restart: "no"
     environment:
       LOG4J_CONFIGURATION_FILE: /var/lib/besu/log4j.xml
+      JAVA_OPTS: -Xmx1536m -XX:+UnlockExperimentalVMOptions -XX:-UseG1GC -XX:+UseZGC
```
jonesho, line 144: @Filter94 reduced this from 2GB to 1.5GB, tested it locally and should be safe for max heap size of l2-node-besu

Filter94: Thanks! Let's see if it passes

jonesho (commented):
@Filter94 I think we'd need to remove the `JAVA_OPTS` overrides in `docker/compose-local-dev.overrides.yml` as well

Filter94:
@jonesho Done, thanks!

jonesho (approved):

docker/compose-local-dev.overrides.yml
```diff
@@ -36,19 +36,7 @@ services:
...
-      JAVA_OPTS: -XX:+UnlockExperimentalVMOptions -XX:-UseG1GC -XX:+UseZGC
-
-  l1-el-node:
-    environment:
-      JAVA_OPTS: -XX:+UnlockExperimentalVMOptions -XX:-UseG1GC -XX:+UseZGC
   
   l2-node-besu:
     environment:
-      JAVA_OPTS: -XX:+UnlockExperimentalVMOptions -XX:-UseG1GC -XX:+UseZGC
+      JAVA_OPTS: -XX:+UnlockExperimentalVMOptions -XX:-UseG1GC -XX:+UseZGC
```
jonesho, line 42: @Filter94 I think we might need to keep the above overrides for `l2-node-besu` for now given that the pipeline was failing at the e2e geth test and I see local e2e geth test would still fail with OOM for max heap size set as `-Xmx2g`, provided that e2e geth test would be removed in the future, maybe we could keep this overrides for now, and in parallel, I'd search for a less heap size consumption alternative for transaction exclusion e2e test

jpnovais (approved):

VGau (approved):

PR #263, opened 2024-10-28
feat: separate tx exclusion api integration tests from coordinator

jonesho:
This PR implements issue(s) #

### Checklist


transaction-exclusion-api/app/build.gradle
```diff
@@ -94,7 +94,7 @@ sourceSets {
   }
 }
 
-task integrationTest(type: Test) {
+task transactionExclusionApi_integrationTest(type: Test) {
```
jpnovais: @jonesho what options have you considered for solving this issue?

prefixing the task with the application's name does not scale and can become messy pretty quickly. I didn't ave time to investigate but I believe we can do better with gradle... 


We should have something like

- `./gradlew integrationTest` - runs all the integration test, like we have at the moment
- `./gradlew coordinator:app:intergrationTestAllNeeded` - would run the integration tests of all the dependencies of the coordinator
- `./gradlew transaction-exclusion-api:app:intergrationTestAllNeeded` - would run the integration tests of all the dependencies of the tx exclusion API;

Something in lines of
```
task intergrationTestAllNeeded {
    group = 'verification'
    description = 'Runs integration tests of all dependencies of MY_APP '

    // Find all project dependencies
    def dependencies = configurations.compileClasspath.resolvedConfiguration.resolvedArtifacts.collect { it.moduleVersion.id }

    // Add dependencies on the integrationTest tasks of each dependent project
    dependencies.each { dependency ->
        def projectPath = ":${dependency.group}:${dependency.name}"
        def integrationTestTask = "${projectPath}:integrationTest"
        dependsOn integrationTestTask
    }
}
```

jonesho: @jpnovais thanks for the suggestion, yes I did think about the other way to achieve via gradle but was heading the wrong direction ( I'm not versed in gradle :( ) and hence resorted to this simplified approach, but your suggestion above looks viable and I'd give it a try.

jpnovais (approved):
Shall we do the same for the finalized-tag plugin?

PR #265, opened 2024-10-29
Extend MetricsFacade with createSimpleTimer() factory

bradbown:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/blob/BlobCompressor.kt
+++ b/coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/blob/BlobCompressor.kt
@@ -4,6 +4,9 @@ import net.consensys.encodeHex
 import net.consensys.linea.blob.BlobCompressorVersion
 import net.consensys.linea.blob.GoNativeBlobCompressor
 import net.consensys.linea.blob.GoNativeBlobCompressorFactory
+import net.consensys.linea.metrics.LineaMetricsCategory
+import net.consensys.linea.metrics.MetricsFacade
+import net.consensys.linea.metrics.Timer
 import org.apache.logging.log4j.LogManager
 import kotlin.random.Random
 
@@ -35,7 +38,8 @@ interface BlobCompressor {
 }
 
 class GoBackedBlobCompressor private constructor(
-  internal val goNativeBlobCompressor: GoNativeBlobCompressor
+  internal val goNativeBlobCompressor: GoNativeBlobCompressor,
+  private val metricsFacade: MetricsFacade
 ) : BlobCompressor {
 
   companion object {
@@ -44,7 +48,8 @@ class GoBackedBlobCompressor private constructor(
 
     fun getInstance(
       compressorVersion: BlobCompressorVersion = BlobCompressorVersion.V0_1_0,
-      dataLimit: UInt
+      dataLimit: UInt,
+      metricsFacade: MetricsFacade
     ): GoBackedBlobCompressor {
       if (instance == null) {
         synchronized(this) {
@@ -57,7 +62,7 @@ class GoBackedBlobCompressor private constructor(
             if (!initialized) {
               throw InstantiationException(goNativeBlobCompressor.Error())
             }
-            instance = GoBackedBlobCompressor(goNativeBlobCompressor)
+            instance = GoBackedBlobCompressor(goNativeBlobCompressor, metricsFacade)
           } else {
             throw IllegalStateException("Compressor singleton instance already created")
           }
@@ -69,29 +74,45 @@ class GoBackedBlobCompressor private constructor(
     }
   }
 
+  private val canAppendBlockTimer: Timer<Boolean> = metricsFacade.createSimpleTimer(
+    category = LineaMetricsCategory.BLOB,
+    name = "go.backed.blob.compressor.can.append.block",
+    description = "Time taken to run CanWrite method"
+  )
+  private val appendBlockTimer: Timer<BlobCompressor.AppendResult> = metricsFacade.createSimpleTimer(
+    category = LineaMetricsCategory.BLOB,
+    name = "go.backed.blob.compressor.append.block",
+    description = "Time taken to run AppendResult method"
+  )
+
   private val log = LogManager.getLogger(GoBackedBlobCompressor::class.java)
 
   override fun canAppendBlock(blockRLPEncoded: ByteArray): Boolean {
-    return goNativeBlobCompressor.CanWrite(blockRLPEncoded, blockRLPEncoded.size)
+    return canAppendBlockTimer.captureTime {
+      goNativeBlobCompressor.CanWrite(blockRLPEncoded, blockRLPEncoded.size)
+    }
   }
 
   override fun appendBlock(blockRLPEncoded: ByteArray): BlobCompressor.AppendResult {
     val compressionSizeBefore = goNativeBlobCompressor.Len()
     val appended = goNativeBlobCompressor.Write(blockRLPEncoded, blockRLPEncoded.size)
     val compressedSizeAfter = goNativeBlobCompressor.Len()
+    val compressionRatio = 1.0 - ((compressedSizeAfter - compressionSizeBefore).toDouble() / blockRLPEncoded.size)
     log.trace(
       "block compressed: blockRlpSize={} compressionDataBefore={} compressionDataAfter={} compressionRatio={}",
       blockRLPEncoded.size,
       compressionSizeBefore,
       compressedSizeAfter,
-      1.0 - ((compressedSizeAfter - compressionSizeBefore).toDouble() / blockRLPEncoded.size)
+      compressionRatio
     )
     val error = goNativeBlobCompressor.Error()
     if (error != null) {
       log.error("Failure while writing the following RLP encoded block: {}", blockRLPEncoded.encodeHex())
       throw BlobCompressionException(error)
     }
-    return BlobCompressor.AppendResult(appended, compressionSizeBefore, compressedSizeAfter)
+    return appendBlockTimer.captureTime {
+      BlobCompressor.AppendResult(appended, compressionSizeBefore, compressedSizeAfter)
+    }
   }
 
   override fun startNewBatch() {
--- a/jvm-libs/linea/core/metrics/src/main/kotlin/net/consensys/linea/metrics/MetricsFacade.kt
+++ b/jvm-libs/linea/core/metrics/src/main/kotlin/net/consensys/linea/metrics/MetricsFacade.kt
@@ -1,5 +1,7 @@
 package net.consensys.linea.metrics
 
+import java.util.concurrent.Callable
+import java.util.concurrent.CompletableFuture
 import java.util.function.Supplier
 
 data class Tag(val key: String, val value: String)
@@ -27,6 +29,11 @@ interface Histogram {
   fun record(data: Double)
 }
 
+interface Timer<T> {
+  fun captureTime(f: CompletableFuture<T>): CompletableFuture<T>
+  fun captureTime(f: Callable<T>): T
+}
+
 interface MetricsFacade {
   fun createGauge(
     category: LineaMetricsCategory,
@@ -50,4 +57,11 @@ interface MetricsFacade {
     tags: List<Tag> = emptyList(),
     baseUnit: String
   ): Histogram
+
+  fun <T> createSimpleTimer(
+    category: LineaMetricsCategory,
+    name: String,
+    description: String,
+    tags: List<Tag> = emptyList()
+  ): Timer<T>
 }
--- a/jvm-libs/linea/metrics/micrometer/src/main/kotlin/net/consensys/linea/metrics/micrometer/MicrometerMetricsFacade.kt
+++ b/jvm-libs/linea/metrics/micrometer/src/main/kotlin/net/consensys/linea/metrics/micrometer/MicrometerMetricsFacade.kt
@@ -8,8 +8,10 @@ import net.consensys.linea.metrics.Histogram
 import net.consensys.linea.metrics.LineaMetricsCategory
 import net.consensys.linea.metrics.MetricsFacade
 import net.consensys.linea.metrics.Tag
+import net.consensys.linea.metrics.Timer
 import java.util.function.Supplier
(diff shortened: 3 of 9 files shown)


jvm-libs/linea/core/metrics/src/main/kotlin/net/consensys/linea/metrics/MetricsFacade.kt
```diff
@@ -27,6 +29,11 @@ interface Histogram {
   fun record(data: Double)
 }
 
+interface TimerCapture<T> {
+  fun captureTime(f: CompletableFuture<T>): CompletableFuture<T>
+  fun captureTime(f: Callable<T>): T
```
jpnovais: suggestion: call it `action`

jvm-libs/linea/core/metrics/src/main/kotlin/net/consensys/linea/metrics/MetricsFacade.kt
```diff
@@ -50,4 +57,10 @@ interface MetricsFacade {
     tags: List<Tag> = emptyList(),
     baseUnit: String
   ): Histogram
+
+  fun <T> createSimpleTimer(
+    name: String,
```
jpnovais, line 65: why doesn't this have a category?

jvm-libs/generic/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/client/VertxHttpJsonRpcClient.kt
```diff
@@ -85,14 +85,14 @@ class VertxHttpJsonRpcClient(
...
-        .setTag("method", request.method)
-        .captureTime(requestFuture)
+      metricsFacade.createSimpleTimer<Future<Result<JsonRpcSuccessResponse, JsonRpcErrorResponse>>>(
+        name = "jsonrpc.request",
+        description = "Time of Upstream API JsonRpc Requests",
+        tags = listOf(
+          Tag("endpoint", endpoint.host),
+          Tag("method", request.method)
+        )
+      ).captureTime { requestFuture }
```
jpnovais: why a lambda instead of just passing the function reference? 
Every unnecessary lamba is extra memory consumption

jvm-libs/linea/metrics/micrometer/src/main/kotlin/net/consensys/linea/metrics/micrometer/MicrometerMetricsFacade.kt
```diff
@@ -8,8 +8,10 @@ import net.consensys.linea.metrics.Histogram
 import net.consensys.linea.metrics.LineaMetricsCategory
 import net.consensys.linea.metrics.MetricsFacade
 import net.consensys.linea.metrics.Tag
+import net.consensys.linea.metrics.TimerCapture
 import java.util.function.Supplier
 import io.micrometer.core.instrument.Counter as MicrometerCounter
+import io.micrometer.core.instrument.Timer as MicrometerTimer
 
 class MicrometerMetricsFacade(private val registry: MeterRegistry, private val metricsPrefix: String) : MetricsFacade {
```
jpnovais: I am ware that it's part of your PR, but maybe we should make `metricsPrefix` and optional field and add only when defined.

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/CoordinatorApp.kt
```diff
@@ -57,9 +57,10 @@ class CoordinatorApp(private val configs: CoordinatorConfig) {
     Vertx.vertx(vertxConfig)
   }
   private val meterRegistry: MeterRegistry = BackendRegistries.getDefaultNow()
+  private val micrometerMetricsFacade = MicrometerMetricsFacade(meterRegistry, "linea")
```
jpnovais, line 50: please use named arguments.

Why `linea` and not `linea.coordinator` or just empty?

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/blob/BlobCompressor.kt
```diff
@@ -69,29 +73,43 @@ class GoBackedBlobCompressor private constructor(
     }
   }
 
+  private val canAppendBlockTimer: TimerCapture<Boolean> = metricsFacade.createSimpleTimer(
+    name = "go.backed.blob.compressor.can.append.block",
```
jpnovais:
```suggestion
    name = "blob.compressor.canappendblock",
```

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/blob/BlobCompressor.kt
```diff
@@ -69,29 +73,43 @@ class GoBackedBlobCompressor private constructor(
...
+    description = "Time taken to run CanWrite method"
```
jpnovais:
```suggestion
    description = "Time taken to check if block fits in current blob or shall be moved to new one"
```

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/blob/BlobCompressor.kt
```diff
@@ -69,29 +73,43 @@ class GoBackedBlobCompressor private constructor(
...
+  )
+  private val appendBlockTimer: TimerCapture<BlobCompressor.AppendResult> = metricsFacade.createSimpleTimer(
+    name = "go.backed.blob.compressor.append.block",
```
jpnovais:
```suggestion
    name = "blob.compressor.apppendblock",
```

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/blob/BlobCompressor.kt
```diff
@@ -69,29 +73,43 @@ class GoBackedBlobCompressor private constructor(
...
+    description = "Time taken to run AppendResult method"
```
jpnovais:
```suggestion
    description = "Time taken to compress block into the blob"
```

jvm-libs/generic/json-rpc/src/test/kotlin/net/consensys/linea/jsonrpc/client/VertxHttpJsonRpcClientTest.kt
```diff
@@ -352,7 +356,7 @@ class VertxHttpJsonRpcClientTest {
 
     val timer =
       meterRegistry.timer(
-        "jsonrpc.request",
+        "linea.jsonrpc.request",
```
jpnovais: What is the value of `linea` prefix here? This JSON rpc client is agnostic an could be used outside linea apps. Furthermore, adding the prefix now is a breaking change, so all the dashboards using this will break. 

We need to carefully access these changes.

IIRC, The metrics are already under a component/cluster which shall provide enough scope, I don't see the need for `linea` here.

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/CoordinatorApp.kt
jonesho: If we replace it with `linea.coordinator` or `empty` it would break the existing exported metrics, I would keep it as is, and given that we can now pass `metricsPrefix` as optional in `MicrometerMetricsFacade` constructor, we could pass a separate `metricsFacade` to class that don't need `metricsPrefix` (e.g. `VertxHttpJsonRpcClientFactory` below at line 63 and `JsonRpcMessageProcessor`)

jvm-libs/generic/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/JsonRpcMessageProcessor.kt
```diff
@@ -50,24 +46,18 @@ private data class RequestContext(
  */
 class JsonRpcMessageProcessor(
   private val requestsHandler: JsonRpcRequestHandler,
-  private val meterRegistry: MeterRegistry,
+  private val metricsFacade: MetricsFacade,
```
jonesho, line 53: From Bradley's previous changes, `JsonRpcMessageProcessor` needed to be passed in both `metricsFacade` and `metricsRegistry`, and here I further reduced it to just pass in `metricsFacade` and added `createDynamicTagTimer` in  `MetricsFacade` class

jpnovais:
I don't have bandwidth to review this pr atm. But please don't merge for now as is believe it will introduce breaking changes to the metrics.

jonesho:
> I don't have bandwidth to review this pr atm. But please don't merge for now as is believe it will introduce breaking changes to the metrics.

@jpnovais the changes here should have taken care of not introducing breaking changes to existing metrics

jpnovais:
@jonesho I still see a lot of diff changes in the JSON-RPC area, It's hard by looking at the code to tell if there are breaking changes. 

Can you please run coordinator and share the content of  `/metrics` A) before the changes and B) now after these changes; ?

jvm-libs/generic/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/JsonRpcMessageProcessor.kt
```diff
@@ -64,17 +60,10 @@ class JsonRpcMessageProcessor(
...
   override fun invoke(user: User?, messageJsonStr: String): Future<String> =
     handleMessage(user, messageJsonStr)
 
   private fun handleMessage(user: User?, requestJsonStr: String): Future<String> {
-    val wholeRequestTimer =
-      Timer.builder("jsonrpc.processing.whole")
-        .description(
-          "Processing of JSON-RPC message: Deserialization + Business Logic + Serialization"
-        )
-    val timerSample = Timer.start(Clock.SYSTEM)
```
jpnovais, line 77: The time sample was taking the time here right at method beginning. Now it's move elsewhere. That will influence time measurements, no?

If read the code correctly, for instance now the time doing `decodeMessage(requestJsonStr)`

jonesho: Yes, I revised it a bit and the `jsonrpc.processing.whole` metric should now also capture the time spent for decoding and parsing of the request messages.

jvm-libs/generic/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/JsonRpcMessageProcessor.kt
```diff
@@ -61,46 +61,74 @@ class JsonRpcMessageProcessor(
...
+        requestParsingResults.first()
           .unwrap().first.method
       }
 
     // all or nothing: if any of the requests has a parsing error, return before execution
-    parsingResults.forEach {
+    requestParsingResults.forEach {
       when (it) {
-        is Err -> return Future.succeededFuture(Json.encode(it.error))
+        is Err -> return Future.failedFuture(Json.encode(it.error))
```
jpnovais: Why change this to an error? Same thing above at line 100. 

won't this be a behavior regression?

jvm-libs/generic/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/JsonRpcMessageProcessor.kt
```diff
@@ -61,46 +61,74 @@ class JsonRpcMessageProcessor(
...
+          handleMessage(
+            user = user,
+            requestJsonStr = requestJsonStr
+          ).toCompletionStage().toCompletableFuture()
+        )
+        .thenApply {
+          logResponse(it.third, it.second, requestJsonStr)
+          it.second
+        }
+        .exceptionally {
```
jpnovais: why this. It was not there before...

jvm-libs/generic/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/JsonRpcMessageProcessor.kt
jpnovais: We need to be very careful with these kinds of changes/regressions. Metrics are very important; we make improvement/performance and troubleshooting decisions based on them. If the values are not accurate, it will lead to wrong decisions and time waste.

jvm-libs/generic/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/JsonRpcMessageProcessor.kt
jonesho: This is to catch the `Future.failedFuture`(s) returned from `handleMessage` (i.e. line 100 and line 119) and return the json encoded string to the caller. Imo this is more semantically correct to set those failed cases as `failedFuture` (rather than `succeededFuture`) in `handleMessage` given that they are not supposed to be handled the same way as valid cases, e.g. `tagValueExtractorOnError` should be called and skipped calling `logResponse` in line 86.

Behavior-wise, they should be the same, as only `Future.failedFurture` returned by `handleMessage` would be caught by this exceptionally clause

jpnovais: I disagree, a failed future should be for an exceptional behavior. Receiving an Invalid JSON is valid use case that this class shall handle - which it does, and is part of JSON-RPC spec - and return parsing error. 

By returning a failed future we are creating an exception and filling a stacktrace unneccerarly, which is much slower. 

It would prefer to ever to successful futures, from the PoV me message processor he could successfully handle this. This client will get an error because sent a bad JSON.

jonesho: Fair point, switched back to use `Future.succeededFuture` for errors when decoding and parsing the request message, removed the `exceptionally` clause as well.

jpnovais (approved):

PR #275, opened 2024-10-31
feat: increase max size of reasonMessage to 512 and allow sender addr…

jonesho:
…ess to be nullable in db and data structure

This PR implements issue(s) #

### Checklist


transaction-exclusion-api/core/src/testFixtures/kotlin/net.consensys.linea.transactionexclusion.test/Common.kt
```diff
@@ -41,3 +41,26 @@ val defaultRejectedTransaction = RejectedTransaction(
...
+  timestamp = Instant.parse("2024-10-31T09:18:51Z"),
+  blockNumber = null,
+  transactionRLP = (
+    "0xb8d602f8d382e708018403780fc08403780fca830118fd8080b87960566023600b82828239805160001" +
+      "a607314601657fe5b30600052607381538281f3fe73000000000000000000000000000000000000000030" +
+      "146080604052600080fdfea2646970667358221220c399228be3777f235999f42c0d0a666dc163e66f600" +
+      "fdbfcc0a0500a4683db1064736f6c63430007060033c001a0a905ece7f3784afa2130063e332899fa60eb" +
+      "13863d96cea29810808c7d5a18eea0685b5237be1e44ccf7d4a9da4410a48cab5a23ba51e23fe3598294c7d34108c1"
+    ).decodeHex(),
+  reasonMessage = "Reverted transaction: 0x08c379a000000000000000000000000000000000000000000000000000000000000000" +
```
jpnovais: Why are we adding this test case?

IIRC from our discussions, we agreed that this service **won't save reverted transactions**. If so, we should not have such a test which can be misleading of the intent of the service.

jonesho: Yes agree, to avoid confusion, I'll keep this rejected transaction instance but change the `reasonMessage` to be relevant to "traces-limit overflow", as I'd still like to have test cases for rejected contract deployment transactions which have no sender address in the decoded the RLP data, will push that soon

transaction-exclusion-api/app/src/integrationTest/kotlin/net/consensys/linea/transactionexclusion/TransactionExclusionAppTest.kt
```diff
@@ -197,6 +200,68 @@ class TransactionExclusionAppTest : CleanDbTestSuiteParallel() {
       )
   }
 
+  @Test
+  fun `Should save the rejected contract deployment tx from RPC`() {
+    // Save the rejected contract deployment tx from RPC (without sender address)
+    val rejectionTimeStamp = Clock.System.now()
+      .trimToMillisecondPrecision()
+      .toString()
```
jpnovais: Note: it makes the test more readable if we use the real string that will be passed along e.g "2024-11-04T09:10:39:123Z"

With this `rejectionTimeStamp` I don't know exactly what is the expected String there...

transaction-exclusion-api/app/src/integrationTest/kotlin/net/consensys/linea/transactionexclusion/TransactionExclusionAppTest.kt
```diff
@@ -174,19 +177,19 @@ class TransactionExclusionAppTest : CleanDbTestSuiteParallel() {
...
     assertThatJson(makeRequestJsonResponse(getTxJsonRequest))
       .isEqualTo(
         """{
           "jsonrpc": "2.0",
           "id": 125,
           "result": {
-            "txHash": "0x526e56101cf39c1e717cef9cedf6fdddb42684711abda35bae51136dbb350ad7",
-            "from": "0x4d144d7b9c96b26361d6ac74dd1d8267edca4fc2",
+            "txHash": "${defaultRejectedTransaction.transactionInfo.hash.encodeHex()}",
+            "from": "${defaultRejectedTransaction.transactionInfo.from.encodeHex()}",
```
jpnovais: I know this is to avoid duplication, but this case duplication here helped to read the test better. PLease see [DAMP vs DRY](https://abseil.io/resources/swe-book/html/ch12.html#tests_and_code_sharing_dampcomma_not_dr)

In the sample below, I can't see how monces are serialized, numeric or hexadecimal?

transaction-exclusion-api/app/src/integrationTest/kotlin/net/consensys/linea/transactionexclusion/TransactionExclusionAppTest.kt
```diff
@@ -197,6 +200,68 @@ class TransactionExclusionAppTest : CleanDbTestSuiteParallel() {
...
+    // Check the get response is corresponding to the rejected contract deployment tx from RPC
+    assertThatJson(makeRequestJsonResponse(getTxJsonRequest))
+      .isEqualTo(
+        """{
+          "jsonrpc": "2.0",
+          "id": 125,
+          "result": {
+            "txHash": "${rejectedContractDeploymentTransaction.transactionInfo.hash.encodeHex()}",
+            "from": "${rejectedContractDeploymentTransaction.transactionInfo.from.encodeHex()}",
+            "nonce": "${rejectedContractDeploymentTransaction.transactionInfo.nonce.toHexString()}",
```
jpnovais: as mentioned above, in this case I think I would prefer DAMP over DRY

jpnovais (approved):

transaction-exclusion-api/app/src/integrationTest/kotlin/net/consensys/linea/transactionexclusion/TransactionExclusionAppTest.kt
jonesho: It'd not be straightforward to replace `rejectionTimeStamp` with static value here because the `transaction-exclusion-api` service running in the integration test would not retrieve rejected transaction with rejected timestamp older than 7 days

jpnovais (approved):

Issue #215, opened 2024-10-21
Coordinator Integration Test - Account Manager helper


jpnovais:
## Impact
Environments: CI/Local

## Describe the bug
Sometimes coordinator's integration test fail with 

```
Response.Error(code=-32000, message=Replacement transaction underpriced)
```

```
Response.Error(code=-32001, message=Nonce too low)
```

This comes from `WhaleBasedAccountManager`. It must handle and retry the request better.

```
    java.lang.RuntimeException: Failed to send funds from accAddress=9d14ef14d4c90436a062efd7dc8354a7ded9712e, accBalance=0x130d48919d5226beb809, accPrivKey=0x...f152866a, error: Response.Error(code=-32001, message=Nonce too low)
        at net.consensys.zkevm.ethereum.WhaleBasedAccountManager.generateAccounts(AccountManager.kt:149)
        at net.consensys.zkevm.ethereum.L2AccountManager.generateAccounts(AccountManager.kt)
```
### Steps to Reproduce
No deterministic way to reproduce.

jpnovais:
Closing as it was already solved.

PR #278, opened 2024-11-04
WhaleBasedAccountManager with Retry Mechanism 

nadeemb53:
This PR closes [#215](https://github.com/Consensys/linea-monorepo/issues/215), which reported intermittent failures in the coordinator's integration tests due to nonce-related and gas price errors. The errors included "Replacement transaction underpriced" and "Nonce too low," originating from the WhaleBasedAccountManager

### Checklist

Note: Changes are not tested


thedarkjester:
It looks like the coordinator tests are failing now, please have a look and adjust accordingly.

see: https://github.com/Consensys/linea-monorepo/actions/runs/11687214785/job/32693999670?pr=278

jonesho:
The coordinator test is failed [again](https://github.com/Consensys/linea-monorepo/actions/runs/11798313620/job/33180449283?pr=278) and should be easily fixed by replacing `sendWithRetry()` return type with `EthSendTransaction`, btw for running the unit tests locally, you could open the repo in your IDE (preferably IntelliJ) and then run `./gradlew -V coordinator:app:buildNeeded`, just like what we did in [this](https://github.com/Consensys/linea-monorepo/blob/docker-coordinator-image-size-reduction/.github/workflows/coordinator-testing.yml#L55) github workflow file. Similarily, in the same file, you could also see how we run coordinator integration tests.

coordinator/ethereum/test-utils/src/main/kotlin/net/consensys/zkevm/ethereum/AccountManager.kt
```diff
@@ -133,30 +128,16 @@ private open class WhaleBasedAccountManager(
...
-        }
-        newAccount to transferResult
+        val transferResult = sendWithRetry(whaleTxManager, newAccount.address, initialBalanceWei)
+        Pair(newAccount, transferResult)
       }
     }
-    result.forEach { (account, transferTx) ->
+
+    result.forEach { pair ->
+      val (account, transferTx) = pair
```
jonesho, line 140: I think this can be simplified as:
```
result.forEach { (account, transferTx) ->
```

jonesho:
@nadeemb53 FYI, we have a class called [AsyncRetryer](https://github.com/Consensys/linea-monorepo/blob/docker-coordinator-image-size-reduction/jvm-libs/generic/extensions/futures/src/main/kotlin/net/consensys/linea/async/AsyncRetryer.kt#L25) which could potentially be used for this case, the caveat is that it assumes the action return type is async and needs to pass in an vertx instance.

jpnovais:
Hi @nadeemb53 my apologies for the conflicts, I was not aware of this PR so in one of my PRs I fixed the issue because it was breaking the CI on my  PR. 

The issue was that when we ran all tests in parallel (can be even different JVM processes), some tests may randomly be assigned the same whale account: Imagine that we have 4 tests sharing the whale account, and they send 4 competing transactions with the same nonce, 3 of them will get an error. These 3 remaining tests just need to retry sending the same transaction with a new nonce until it succeeds (the least lucky one will retry 3 times).  Although the proposed solution would work just fine, the current solution with a simple retry works just fine, no need for extra logic for transaction replacement with gasPrice increase.

Since the problem seems to be solved I think we can close this PR?

nadeemb53:
makes sense, i'll close this

Issue #273, opened 2024-10-31
Use winston logger for the e2e tests logging


VGau:
## Description
Currently, our end-to-end (E2E) tests use simple console logs for output, which lacks flexibility and advanced features. We propose integrating the Winston logger into the E2E tests to enhance logging capabilities with features like log levels, formatting.

## Motivation
Implementing Winston logger will improve debugging and monitoring of E2E tests by providing structured and configurable logs.

## Tasks
- [ ] Integrate Winston logger into the E2E testing framework (Jest)
- [ ] Replace all console logs with Winston logger methods
- [ ] Configure appropriate log levels (e.g., info, error) and formatting
- [ ] Test logging output to ensure correctness

## Acceptance criteria
- [ ] All console logs are replaced with Winston logger calls
- [ ] Logs are properly formatted and include necessary details
- [ ] Logging does not negatively impact test performance
- [ ] E2E tests run successfully with the new logging system

## Risks
- [ ] Potential impact on test execution time due to logging overhead

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #282, opened 2024-11-05
feat/273 Use Winston logger for E2E Tests

count-sum:
This PR implements issue(s) #273 

### Checklist

* [X] I wrote new tests for my new core changes.
* [X] I have successfully ran tests, style checker and build against my new changes locally.
* [X] I have informed the team of any breaking changes if there are any.


e2e/src/common/logger.ts
```diff
@@ -0,0 +1,27 @@
...
+    new transports.Console({
+      level: "error",
+      format: format.combine(
+        format.colorize(),
+        format.printf(({ timestamp, level, message }) => `${timestamp} ${level}: ${message}`),
+      ),
+    }),
+  ],
+});
+
```
thedarkjester: @VGau do you think we want thread/process information here as well?

VGau: Would be nice to have the test name as well :)

PR #286, opened 2024-11-12
Feature: Optimizations by DistributedLab

olegfomenko:
# Feature: Optimizations by DistributedLab

## Optimizing Keccak 
- Replaced loops with pre-calculated slice indexes
- Using assembler xor for amd64 and arm64 platforms (taken from golang/x/crypto)

## Optimizing FFT
- Taken parallel approach from gnark-crypto
- Adding BitReverse from gnark-crypto that should be faster for non-arm platforms

## Optimizing util/parallel
- Use atomic uint instead of the chan of uints
- Adding thread pool based on chan that executes tasks with different execution time

## Optimizing exponent
-  Optimized exp function to user int power instead of big.Int

### Checklist


prover/maths/fft/fft.go
```diff
@@ -14,19 +16,26 @@ const (
 	DIF
 )
```
AlexandreBelling, line 18: What are the diffs between this version of the fft and gnark's ? It may be interesting to merge the two I believe

prover/crypto/keccak/keccakf.go
```diff
@@ -5,7 +5,6 @@ package keccak
 const Dim = 5           // size of the keccak state matrix
 const NumRound int = 24 // number of rounds in the keccak permutation
 
-// LR collects the triangular numbers for the \pi and \rho step
```
AlexandreBelling: Are there performances gains that are registered from these changes? The implementation is a toy implementation that is designed to make it, easier to test and debug the keccak arithmetization. But AFAICT, this had not been flagged as a performance critical piece of code (this is still used in production to produce the witness).

AlexandreBelling: If the gains are negligible, I would rather ditch those changes

prover/crypto/keccak/subtle/xor_generic.go
```diff
@@ -0,0 +1,64 @@
+// Copyright 2013 The Go Authors. All rights reserved.
```
AlexandreBelling: The header should be removed

prover/crypto/keccak/subtle/xor_arm64.go
```diff
@@ -0,0 +1,10 @@
+// Copyright 2020 The Go Authors. All rights reserved.
```
AlexandreBelling: Likewise, we should remove the header or avoid these optimization if we copy non-apache code.

prover/crypto/keccak/subtle/xor_amd64.s
```diff
@@ -0,0 +1,58 @@
+// Copyright 2018 The Go Authors. All rights reserved.
```
AlexandreBelling: Likewise, we should remove the header or avoid these optimization if we copy non-apache code.

prover/crypto/keccak/subtle/xor_amd64.go
```diff
@@ -0,0 +1,10 @@
+// Copyright 2018 The Go Authors. All rights reserved.
```
AlexandreBelling: Likewise, we should remove the header or avoid these optimization if we copy non-apache code.

prover/crypto/keccak/subtle/xor.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func XorIn(state *[25]uint64, buf [17]uint64) {
+	x := (*[25 * 64 / 8]byte)(unsafe.Pointer(state))
+	y := (*[17 * 64 / 8]byte)(unsafe.Pointer(&buf))
+	XORBytes(x[:], x[:], y[:])
+}
+
+//
+// xorIn xors the bytes in buf into the state.
```
AlexandreBelling: Commented code to remove

prover/crypto/keccak/subtle/xor.go
```diff
@@ -0,0 +1,46 @@
+// Copyright 2022 The Go Authors. All rights reserved.
```
AlexandreBelling: Likewise, we should remove the header or avoid these optimization if we copy non-apache code.

prover/crypto/keccak/subtle/xor_arm64.s
```diff
@@ -0,0 +1,69 @@
+// Copyright 2020 The Go Authors. All rights reserved.
```
AlexandreBelling: Likewise, we should remove the header or avoid these optimization if we copy non-apache code.

prover/crypto/ringsis/ringsis_64_8/transversal_hash.go
```diff
@@ -51,91 +49,86 @@ func TransversalHash(
 		constResults = make([]field.Element, 64)
 	)
 
-	parallel.ExecuteChunky(numJobs, func(start, stop int) {
+	ppool.ExecutePoolChunky(numJobs, func(i int) {
```
AlexandreBelling, line 52: Any hindsight what the gain was here?

prover/maths/fft/fft.go
olegfomenko: Presented improvements mostly migrates all logic from gnrak-crypto, except of some tricks with twiddles that will require changes in the twiddles generation that may not be suitable for your in-place implementation (imho).

prover/crypto/keccak/keccakf.go
olegfomenko: Got it, let me perform an additional benches to check the performance of this feature separately. I will share the results after

prover/crypto/ringsis/ringsis_64_8/transversal_hash.go
olegfomenko: We have searched the most critical performance places and realized that thread pools is one of them. So we've tested the approach, which can be found in ppool.ExecutePoolChunky. During the tests we replaced methods from the parallel package with ExecutePoolChunky one by one and saw if it gave us a gain. if yes - we kept it.

olegfomenko: Regarding the question why ExecutePoolChunky can be better then existing ExecuteChunky - we not sure that we have full answer. We assume that Go channels are faster than atomics when the number of CPU is large.

prover/crypto/keccak/hash.go
```diff
@@ -20,7 +23,9 @@ const (
 // https://keccak.team/keccak_specs_summary.html
 //
 // Namely, State[X][Y] gives the S[X, Y] in the spec.
-type State [5][5]uint64
+//type State [5][5]uint64
```
AlexandreBelling: Commented code to remove

olegfomenko:
Reverted changes to keccak due to lack of impact on the result

gusiri (approved):
LGTM

AlexandreBelling:
Closing as this was merged in another PR (to add some extra fixes requested by the CI)

PR #289, opened 2024-11-14
perf load setup in parallel to compilation and assignment

Tabaie:
PI proof generation involves two independent, time consuming tasks: loading the circuit setup (SRS in particular) and recompiling the circuit (the wizard Keccak prover in particular). This PI speeds the process up a bit by performing those two tasks in parallel.


gbotrel (dismissed):
fine with me, but CI seems to fail?

AlexandreBelling (approved):
Set on auto-merge

gbotrel (approved):

PR #296, opened 2024-11-19
Verifying Keys in aggregation manifest

Tabaie:
For debugging, the `ProveCheck` method runs the gnark test engine if proof generation fails. This does not currently work well for aggregation, as the verifying keys are embedded in the constraints and are not an input to the circuit. Thus to reconstruct the circuit object correctly for the test engine, we need to store the entirety of the aggregating verifying keys, not only their checksums.

### DIFF
--- a/prover/backend/aggregation/prove.go
+++ b/prover/backend/aggregation/prove.go
@@ -1,6 +1,9 @@
 package aggregation
 
 import (
+	"bytes"
+	"crypto/sha256"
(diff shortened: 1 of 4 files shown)


gbotrel:
Not sure this is needed. All srs in test and in prod are meant to be the same ; that is we should have a 1-1 mapping with commit id and the verifying key here. If we can read the manifest why can't we read the key ?

PR #297, opened 2024-11-19
test cache proofs for tests/debugging

Tabaie:
Option `WithCachedProof` to `circuits.ProveCheck` saves/reads proofs to enable faster debugging


gbotrel:
A `ProveCheck` method that takes as "optional" parameter a proof is a bit of a weird naming. 
Shouldn't this package just expose a `Verify` (or `Check` to stay on the nomenclature) function that's can be call directly and internally by `ProveCheck` ? then caller just read from cache for debugging purposes ? (or does that bring spaghetti somewhere else ?)

Tabaie:
I find `ProveCheck` as it is very convenient and what it does certainly makes sense as "one unit". The "contract" is basically:
1. Try and produce a proof based on the given assignment
2. See if the assignment/proof is valid
3. If it isn't, tell me why

So the purpose is just proof generation, the rest is just double checking and debugging tools. In that vein I think it makes sense to add another debugging convenience option to it, updating the flowchart to:
1. Try and load the proof from disk, if failed, go to 3
2. Verify that the proof passes with the given assignment (public input). If so, return and bypass the time-consuming rest of the function
3. Try and produce a proof based on the given assignment
4. See if the assignment/proof is valid
5. If it isn't, tell me why

How I would defend the option is that it involves very minimal modification to the backend code calling `ProveCheck`. For example
```go
return circuits.ProveCheck(
	setup,
	assignment,
	emPlonk.GetNativeProverOptions(ecc.BN254.ScalarField(), setup.Circuit.Field()),
	emPlonk.GetNativeVerifierOptions(ecc.BN254.ScalarField(), setup.Circuit.Field()),
)
```
simply becomes
```go
return circuits.ProveCheck(
	setup,
	assignment,
	emPlonk.GetNativeProverOptions(ecc.BN254.ScalarField(), setup.Circuit.Field()),
	emPlonk.GetNativeVerifierOptions(ecc.BN254.ScalarField(), setup.Circuit.Field()),
        circuits.WithCachedProof(".tmp/decomp.pf"),
)
```
which makes a debugging session easy and quick to set up and to tear down (the latter is especially important in my experience: if there are many modifications one might forget to remove them all, or that the process of removal might introduce new faults _after_ a fix is developed.)

AlexandreBelling (approved):

PR #298, opened 2024-11-19
[Fix] M-01 Prevent fallback operator calling renounceRole

thedarkjester:
This PR implements issue(s) #

### Checklist


contracts/contracts/LineaRollup.sol
```diff
@@ -150,6 +150,20 @@ contract LineaRollup is
     emit LineaRollupVersionChanged(bytes8("5.0"), bytes8("6.0"));
   }
 
+  /**
+   * @notice Revokes `role` from the calling account.
+   * @dev Fallback operator cannot renounce role. Reverts with OnlyNonFallbackOperator.
+   * @param _role The role to renounce.
+   * @param _account The account to renounce - can only be the _msgSender().
+   */
+  function renounceRole(bytes32 _role, address _account) public virtual override {
```
VGau: Do we want to put virtual here ?

thedarkjester: removed - was keeping the original signature for clarity, but we don't really need it.

VGau: perfect thanks :)

VGau (approved):

VGau (approved):

PR #309, opened 2024-11-19
Chore: fix staticcheck `prealloc` complaints

Tabaie:
golangci-lint now doesn't like the following pattern:
```go
var slice2 []T

for i := range slice1 {
  slice2 = append(slice2, something)
}
```
and prefers
```go
slice2 := make([]T, len(slice1))

for i := range slice1 {
  slice2[i] = something
}
```


Tabaie:
Issue already addressed in https://github.com/Consensys/linea-monorepo/pull/308

gbotrel (commented):
lgtm

PR #313, opened 2024-11-20
Chore/prover/adjust compression max

Tabaie:
Recent addition of a check to the decompression circuit necessitates a slight reduction of the maximum blob size. This PR encodes the new maximum as well as updating the tool used to obtain the bound.


jpnovais:
I have a few questions:
- 1. Is this 100% backwards compatible?
- 2. release: when does this have to be released? does it patch compressor V0_1_0 and V1_0_1?

Tabaie:
> I have a few questions:
> 
> * 1. Is this 100% backwards compatible?
> * 2. release: when does this have to be released? does it patch compressor V0_1_0 and V1_0_1?

1. Yes it is backwards compatible. The only difference is when the compressor decides that a blob is full enough.
2. My understanding is that we will release a new version after the merge, is that correct?

gbotrel (approved):

jpnovais:
>2.  My understanding is that we will release a new version after the merge, is that correct?

Atm we have 2 versions for blob-compressor:
- v0.1.0 - current version in mainnet/prod
- v1.0.1 - current in sepolia, which as the new compressor for the new release of the 100% proving release;

My question is if this well be need to be a new release for both V0 and V1, or just for V1.

jpnovais (approved):

Tabaie:
@jpnovais I see, it's v1, i.e. the one on Sepolia

AlexandreBelling:
Correct me if I am wrong, but I assume this is necessary before deploying the full decompressor

Tabaie:
> Correct me if I am wrong, but I assume this is necessary before deploying the full decompressor

I think you're correct

PR #321, opened 2024-11-21
Minor improvements to the SRS store

Tabaie:
1. The store attempts to load the Lagrange SRS as it is loading the canonical
2. Both are loaded directly from file instead of being loaded into bytes at first. This is doubly advantageous if the canonical SRS found is much larger than what we need.
3. ~If the Lagrange SRS has to be computed, it will be saved (attributed to the same organization as the canonical SRS used to produce it)~
4. In-memory cache (useful when running multiple commands in the same run)


prover/circuits/srs_store.go
```diff
@@ -24,12 +24,14 @@ import (
...
 type SRSStore struct {
 	entries map[ecc.ID][]fsEntry
+	rootDir string
 }
 
 type fsEntry struct {
 	isCanonical bool
 	size        int
 	path        string
+	org         string
```
gbotrel: what does "org" stand for ??

prover/circuits/srs_store.go
```diff
@@ -102,18 +106,52 @@ func (store *SRSStore) GetSRS(ctx context.Context, ccs constraint.ConstraintSyst
 	sizeCanonical, sizeLagrange := plonk.SRSSize(ccs)
 	curveID := fieldToCurve(ccs.Field())
 
+	var lagrangeSRS kzg.SRS
+	loadLagrangeErr := make(chan error, 2)
+	go func() { // attempt to find lagrange SRS
```
gbotrel, line 107: do we really win much by loading iin parallel? Makes sense to do it, but just curious, last I checked the readDump was pretty fast

prover/circuits/srs_store.go
```diff
@@ -122,33 +160,27 @@ func (store *SRSStore) GetSRS(ctx context.Context, ccs constraint.ConstraintSyst
...
 			panic("canonical SRS is smaller than lagrange SRS")
 		}
 		logrus.Debugf("computing lagrange SRS from canonical SRS %d -> %d\n", sizeCanonical, sizeLagrange)
 		var err error
-		lagrangeSRS, err = toLagrange(canonicalSRS, sizeLagrange)
-		if err != nil {
+		if lagrangeSRS, err = toLagrange(canonicalSRS, sizeLagrange); err != nil {
 			return nil, nil, err
 		}
+		// save the lagrange SRS to disk
```
gbotrel: while it looks convenient for dev, I would not do it here, since in production many machine can launch and use a shared directory for the srs, it might create corrupted files. 

i.e. if caller wants to serialize new files, he does it on its own, we should not do it on the prover runtime path.

prover/circuits/srs_store.go
Tabaie: Just a few seconds, mostly a win when debugging.

prover/circuits/srs_store.go
Tabaie: I see your point. Imo we could either
1. Put a lock on the file being written so a prover instance would give up trying to write if it couldn't obtain the lock
2. Let the function fail if the corresponding Lagrange was not found.
@AlexandreBelling what do you think?

prover/circuits/srs_store.go
Tabaie: "Organization": aleo, aztec, or celo

prover/circuits/srs_store.go
AlexandreBelling: In prod, we don't touch the SRS. Do we? Or if we do, we only read so I'd assume this is fine

Tabaie: The only downside I see for option 2 is that we would have to add a few more files to `prover/prover-assets` for our tests to pass and I know we don't like committing too many files into the repo.

prover/circuits/srs_store.go
AlexandreBelling: Not solely related to that but the structs would deserve some more doc

prover/circuits/srs_store.go
AlexandreBelling: Ah but this is on prover runtime? Then no, we should not write this to disk. This makes the whole system more complicated than needed especially if it happens on the distributed file system in prod. We don't want to deal with the ramification of that.

The cleaner approach would be;

a. Try loading it
b. If it does not, try recomputing it locally but without writing it on disk
c. If not possible, throw a human-readable error explaining what is going on

The persons that are going to see issues related to finding the SRS are the SRE so better make clean error message as it saves time in case of failed deployment.

Tabaie: Okay so definitely writing at prover runtime is off the table. No to option 1.
How about not generating locally at all? Since this is such a big performance penalty it should not be "swept under the rug" imo. We can set an upper bound, that up to say 1024 we can generate Lagrange on the fly (so that our tests still pass without storing more stuff on github) but that it also doesn't enable huge setup/prover delays.

AlexandreBelling: Something you could do is to enable this writing-to-disk behaviour thanks to a global variable.

Tabaie:
@gbotrel @AlexandreBelling Removed disk writes.

AlexandreBelling (approved):

prover/circuits/srs_store.go
0xtinkle: @gbotrel raised a good point earlier — did you run benchmarks comparing sequential vs. parallel loading?
If readDump is already fast, could the overhead of goroutines introduce additional complexity without significant gains?

0xy0k:
> @gbotrel @AlexandreBelling Removed disk writes.

removing disk writes should help reduce I/O overhead, but does it impact SRS reuse across runs?
Would it be beneficial to cache in-memory SRS snapshots for debugging purposes?

PR #322, opened 2024-11-21
statereover: update web3j

jpnovais:
This PR implements issue(s) #

### Checklist


jvm-libs/linea/besu-libs/build.gradle
Filter94, line 1: Where are we planning to reuse this exact set of dependencies?

jpnovais: Were they are being used now, the coordinator and test utils. 

Also, they will be necessary for the state-recover plugin.

This is to solve transitive dependency polling of web3j with different implementations kzg commitments

Filter94 (approved):

PR #324, opened 2024-11-21
State recover part 4

jpnovais:
This PR implements issue(s) #

### Checklist


config/common/traces-limits-v2.toml
jpnovais, line 1: These limits were extracted from Sepolia. Previous values were not allowing a conflation on more than 1 block

config/coordinator/log4j2-dev.xml
```diff
@@ -100,6 +100,10 @@
       <DebouncingFilter/>
       <appender-ref ref="rewrite"/>
     </Logger>
+    <Logger name="net.consensys.linea.ethereum.gaspricing.staticcap" level="ERROR" additivity="false">
```
jpnovais: This is because localy it loggin warning that can be dismissed localy

jvm-libs/generic/logging/src/main/kotlin/net/consensys/linea/logging/JsonRpcRequestResponseLogger.kt
jpnovais, line 1: This was to fix log4j warnings that number of arguments don't match placeholders {}

jvm-libs/linea/blob-compressor/src/test/kotlin/linea/blob/GoBackedBlobCompressorTest.kt
```diff
@@ -41,14 +41,20 @@ class GoBackedBlobCompressorTest {
...
-    while (result.blockAppended && blocks.hasNext()) {
+    // at least one block should be appended
+    assertThat(result.blockAppended).isTrue()
+    while (result.blockAppended) {
       val blockRlp = blocks.next()
       val canAppend = compressor.canAppendBlock(blockRlp)
       result = compressor.appendBlock(blockRlp)
       // assert consistency between canAppendBlock and appendBlock
       assertThat(canAppend).isEqualTo(result.blockAppended)
+      if (!blocks.hasNext()) {
```
jpnovais, line 54: this is not the ideal, but don't have time to fix all of it now.

Ideally should create static blocks in the test that fill the limit, instead of relying on the ones in the sample.

config/coordinator/coordinator-docker.config.toml
```diff
@@ -267,8 +266,8 @@ num-of-blocks-before-latest=4
 storage-period="PT2M"
 
 [conflation]
-blocks-limit=2
-conflation-deadline="PT6S" # =3*l2_block_time
+blocks-limit=3
+conflation-deadline="PT60S" # =3*l2_block_time
```
jonesho: The comment is misleading, `l2_block_time` in local stack is one second iirc

docker/compose.yml
```diff
@@ -684,6 +691,96 @@ services:
...
+        --plugin-staterecovery-l1-rpc-endpoint=http://l1-el-node:8545 \
+        --plugin-staterecovery-shomei-endpoint=http://shomei-sr:8888 \
+        --plugin-staterecovery-blobscan-endpoint=http://blobscan-api:4001 \
+        --plugin-staterecovery-overriding-recovery-start-block-number=1
+    volumes:
+      - ./config/zkbesu-shomei/all-permissions.policy:/var/lib/besu/all-permissions.policy:ro
+      - ./config/zkbesu-shomei/zkbesu-config.toml:/var/lib/besu/zkbesu-config.toml:ro
+      - ./config/zkbesu-shomei/log4j-staterecovery.xml:/var/lib/besu/log4j.xml:ro
+      - ./config/linea-local-dev-genesis-PoA-besu.json/:/var/lib/besu/genesis.json:ro
+      - ../state-recover/besu-plugin/build/libs/linea-staterecover-plugin-0.0.1-rc2.jar:/opt/besu/lib/linea-staterecover-plugin-0.0.1-rc2.jar
```
jonesho, line 752: Is it a temporary`linea-staterecover-plugin-0.0.1-rc2.jar` workaround? This plugin would be put in `linea-besu-package` in the future, right?

settings.gradle
```diff
@@ -66,9 +66,12 @@ include 'transaction-exclusion-api:app'
 include 'transaction-exclusion-api:core'
 include 'transaction-exclusion-api:persistence:rejectedtransaction'
 
-include 'state-recover:appcore:logic'
-include 'state-recover:appcore:domain-models'
 include 'state-recover:appcore:clients-interfaces'
+include 'state-recover:appcore:domain-models'
+include 'state-recover:appcore:logic'
+//include 'state-recover:besu-plugin'
```
jonesho: can remove this commented line?

state-recover/test-cases/src/integrationTest/kotlin/linea/staterecover/StateRecoveryManualReplayToLocalStackIntTest.kt
```diff
@@ -0,0 +1,134 @@
...
+
+  @Test
+  fun setupDeployContractForL2L1StateReplay() {
+    this.rollupDeploymentResult = ContractsManager.get()
+      .deployLineaRollup(numberOfOperators = 2, contractVersion = LineaContractVersion.V6).get()
+    log.info("""LineaRollup address=${rollupDeploymentResult.contractAddress}""")
+    log.info(
+      """
+      Start state recovery besu and shomei with the following configuration:
+       ./gradlew state-recover:besu-plugin:shadowJar \
```
Filter94, line 88: I think it would be better for the entrypoint to be just `make X` command instead

state-recover/test-cases/src/integrationTest/kotlin/linea/staterecover/StateRecoveryManualReplayToLocalStackIntTest.kt
```diff
@@ -0,0 +1,134 @@
...
+    val lastAggregation = lastAggregationAndBlobs.aggregation!!
+
+    Web3jClientManager.l1Client.waitForTxReceipt(
+      txHash = submissionTxHashes.aggregationTxHashes.last(),
+      timeout = 2.minutes
+    ).also {
+      assertThat(it.status).isEqualTo("0x1")
+        .withFailMessage(
+          "finalization=${lastAggregation.intervalString()} tx failed! " +
+            "replay data is not consistent with L1 state, potential cause: " +
```
Filter94, line 127: I don't quite understand what is L1 in this case. We've got data from the `testdata` which we submit to local stack's L1. I assume the data in the files is from some L1 somewhere, but since I don't have the context, it's confusing to me

config/common/traces-limits-v2.toml
```diff
@@ -10,34 +10,34 @@ BLOCK_HASH = 512
...
 TXN_DATA = 8192
 WCP = 262144
 #
-# Reference table limits
+# Reference table limits, set to UInt.MAX_VALUE
 #
-BIN_REFERENCE_TABLE = 196864
-SHF_REFERENCE_TABLE = 4096
-INSTRUCTION_DECODER = 512
+BIN_REFERENCE_TABLE = 4294967295
```
Filter94: In prod envs there were issues with this value used in Besu, because the Int is signed there. We had to change it to `2147483647` instead

jvm-libs/generic/logging/src/main/kotlin/linea/logging/TimeMeasureLogger.kt
```diff
@@ -0,0 +1,38 @@
...
+
+import org.apache.logging.log4j.Level
+import org.apache.logging.log4j.Logger
+import kotlin.time.measureTime
+
+/**
+ * This is more for debugging purposes, to measure the time taken by a certain action and log it.
+ * For production code, consider using metrics.
+ */
+fun <T> measureTimeAndLog(
```
Filter94, line 11: Since it's not a production code, maybe it's better to move it to `:linea:testing`? Otherwise it will just take up a little space in production jars

jvm-libs/linea/besu-libs/build.gradle
```diff
@@ -3,27 +3,30 @@ plugins {
   id 'java-library'
 }
 
-def besuArtifactGroup="org.hyperledger.besu"
+//def besuArtifactGroup="org.hyperledger.besu"
+//def besuVersion=libs.versions.besu.get()
+def besuArtifactGroup="io.consensys.linea-besu"
+def besuVersion="24.12-develop-2098dd2"
```
Filter94, line 9: Would it be harmful if we set this version in `libs.versions.toml` temporarily instead?

jvm-libs/linea/besu-libs/build.gradle
```diff
@@ -34,4 +37,11 @@ dependencies {
   api("io.tmio:tuweni-units:${libs.versions.tuweni.get()}") {
     transitive = false
   }
+
+  implementation("org.bouncycastle:bcpkix-jdk18on:1.79") {
```
Filter94: I see you've added bouncycastle to `libs.versions.toml`, so you can use it here

jvm-libs/linea/clients/interfaces/src/main/kotlin/linea/EthLogsSearcher.kt
```diff
@@ -0,0 +1,32 @@
...
+   * if fromBlock..toBlock range is too large, it shall break into smaller chunks
+   * and perform a binary search;
+   */
+  fun findLog(
+    fromBlock: BlockParameter,
+    toBlock: BlockParameter,
+    chunkSize: Int = 1000,
+    address: String,
+    topics: List<String>,
+    shallContinueToSearchPredicate: (build.linea.domain.EthLog) -> SearchDirection? // null means stop searching
```
Filter94: This is not a predicate, I think name Comparator fits here. I think if you add binary search to the name of this interface and name this function something like Comparator or Order, it will be more accurate and conventional.

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/Web3JLogsSearcher.kt
```diff
@@ -0,0 +1,214 @@
...
+import org.apache.logging.log4j.Logger
+import org.web3j.protocol.Web3j
+import org.web3j.protocol.core.methods.request.EthFilter
+import org.web3j.protocol.core.methods.response.EthLog
+import org.web3j.protocol.core.methods.response.Log
+import tech.pegasys.teku.infrastructure.async.SafeFuture
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.milliseconds
+
+sealed interface SearchResultT<T> {
```
Filter94: Is this a leftover?

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/SearchCursor.kt
```diff
@@ -0,0 +1,76 @@
...
+ */
+class SearchCursor(
+  val from: ULong,
+  val to: ULong,
+  val chunkSize: Int
+) {
+  private data class Chunk(val interval: Pair<ULong, ULong>, var searched: Boolean = false)
+
+  private val searchChunks = (from..to)
+    .chunked(chunkSize)
```
Filter94: This `chunked` call is unnecessarily heavy since it's going to instantiate a 2d list. Even though it's unlikely to cause any huge memory usage in practice, I think `ULongProgressionIterator` will suite prfectly what you're trying to do, up to you

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/SearchCursor.kt
```diff
@@ -0,0 +1,76 @@
...
+    .map { Chunk(it.first() to it.last(), searched = false) }
+    .toMutableList()
```
Filter94: I believe it's immutable, only the `Chunk.searched` is mutable

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/Web3JLogsSearcher.kt
```diff
@@ -0,0 +1,214 @@
...
+    toBlock: ULong,
+    chunkSize: Int,
+    address: String,
+    topics: List<String>,
+    shallContinueToSearchPredicate: (build.linea.domain.EthLog) -> SearchDirection?
+  ): SafeFuture<build.linea.domain.EthLog?> {
+    val cursor = SearchCursor(fromBlock, toBlock, chunkSize)
+    log.trace("searching between blocks={}", CommonDomainFunctions.blockIntervalString(fromBlock, toBlock))
+
+    var nexChunkToSearch: Pair<ULong, ULong>? = cursor.next(searchDirection = SearchDirection.FORWARD)
```
Filter94: Typo, `nexChunkToSearch` -> `nextChunkToSearch`

jvm-libs/linea/web3j-extensions/src/test/kotlin/linea/web3j/SearchCursorTest.kt
```diff
@@ -0,0 +1,146 @@
...
+      ).next(searchDirection = SearchDirection.BACKWARD)
+    ).isEqualTo(41UL to 50UL)
+  }
+
+  @Test
+  fun `next follows binary search when direction is provided`() {
+    val searchCursor = SearchCursor(from = 1uL, to = 100uL, chunkSize = 10)
+
+    assertThat(searchCursor.next(SearchDirection.FORWARD)).isEqualTo(41UL to 50UL)
+    assertThat(searchCursor.next(SearchDirection.FORWARD)).isEqualTo(71UL to 80UL)
```
Filter94, line 39: Why (50, 71) range is skipped here?

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/SearchCursor.kt
```diff
@@ -0,0 +1,76 @@
...
+        }
+      }?.let { (chunk, index) ->
+        prevCursor = index
+        chunk.searched = true
+        chunk.interval
+      }
+    }
+  }
+
+  private fun findNextUnsearchedChunk(): Pair<Chunk, Int>? {
```
Filter94: This causes the algorithm to become linear on chunks if the data is sparse. I think better strategy is to pick next unsearched chunk to be the closest to the middle

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/SearchCursor.kt
```diff
@@ -0,0 +1,76 @@
...
+            null
+          } else {
+            val mid = left + (right - left) / 2
+            val chunk = searchChunks[mid]
+            if (chunk.searched) {
+              // we have already searched this chunk, lets find next unsearched
+              findNextUnsearchedChunk()
+            } else {
+              chunk to mid
+            }
```
Filter94: I think this whole chunk of code can be reused in the `prevCursor == null` condition. Even though check `chunk.searched` is not needed in the first iteration, reusing this bit may make this function easier to read.

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/SearchCursor.kt
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * SearchCursor is a helper class to iterate over a range of ULong values in a binary search manner.
+ * When search is not provided returns the next unsearched chunk without moving left or right cursors because
+ * no direction is provided, but flags the chunk as searched to not go over again.
+ *
+ * This caveat is because when searching for EthLogs on L1, a blockInterval (represend by chunk here) may not have logs
+ * so the search predicate cannot tell what direction to go next, so we need to flag the chunk as searched and try the
+ * next chunk.
+ */
+class SearchCursor(
```
Filter94: The logic between `Web3JLogsSearcher` and `SearchCursor` is so intertwined, so I'd consider making `SearchCursor` a private class inside `Web3JLogsSearcher`

jvm-libs/linea/web3j-extensions/src/test/kotlin/linea/web3j/Web3JLogsSearcherIntTest.kt
```diff
@@ -0,0 +1,473 @@
...
+    )
+
+    logsClient.findLog(
+      fromBlock = 100UL.toBlockParameter(),
+      toBlock = 200UL.toBlockParameter(),
+      address = address,
+      topics = listOf("0xffaabbcc"),
+      chunkSize = 10,
+      shallContinueToSearchPredicate = { ethLog ->
+        shallContinueToSearch(ethLog, targetNumber = 89UL)
```
Filter94, line 323: I think `targetNumber` was meant to be in `[110, 149]` here

jvm-libs/linea/web3j-extensions/src/test/kotlin/linea/web3j/Web3JLogsSearcherIntTest.kt
```diff
@@ -0,0 +1,473 @@
...
+    )
+      .get()
+      .also { log ->
+        assertThat(log).isNull()
+        assertThat(TestingJsonRpcServer.callCountByMethod("eth_getLogs")).isBetween(1, 4)
+      }
+  }
+
+  @Test
+  fun `findLogs searches L1 and returns null when range has no logs`() {
```
Filter94, line 353: What is the difference between `findLogs searches L1 and returns null when no logs in blockRange` and this scenario?

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/SearchCursor.kt
```diff
@@ -0,0 +1,76 @@
...
+      }?.let { (chunk, index) ->
+        prevCursor = index
+        chunk.searched = true
+        chunk.interval
+      }
+    }
+  }
+
+  private fun findNextUnsearchedChunk(): Pair<Chunk, Int>? {
+    for (i in left..right) {
```
jonesho, line 79: Why not setting `left  = i` here to advance the left pointer in each iteration of this loop? such that the next `findNextUnsearchedChunk` would not start again at the visited index

config/common/traces-limits-v2.toml
Filter94: Actually it doesn't matter here. It would be important only for `traces-limits-besu-v2.toml` file. This one is for Coordinator and it's Ok to have it set to `4294967295` here

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/SearchCursor.kt
```diff
@@ -0,0 +1,76 @@
...
+  private val searchChunks = (from..to)
+    .chunked(chunkSize)
+    .map { Chunk(it.first() to it.last(), searched = false) }
+    .toMutableList()
+  private var left = 0
+  private var right = searchChunks.size - 1
+  private var prevCursor: Int? = null
+
+  @Synchronized
+  fun next(searchDirection: SearchDirection?): Pair<ULong, ULong>? {
```
jonesho, line 40:
```suggestion
  fun next(searchDirection: SearchDirection?): Pair<ULong, ULong>? {
    return if (prevCursor == null || searchDirection != null) {
      if (prevCursor != null) {
        if (searchDirection == SearchDirection.FORWARD) {
          left = min(right, prevCursor!! + 1)
        } else {
          right = max(left, prevCursor!! - 1)
        }
      }
      val mid = left + (right - left) / 2
      val chunk = searchChunks[mid]
      if (chunk.searched) {
        // we have already searched this chunk, lets find next unsearched
        findNextUnsearchedChunk()
      } else {
        chunk to mid
      }
    } else { //prevCursor != null && searchDirection == null
      findNextUnsearchedChunk()
    }?.let { (chunk, index) ->
      prevCursor = index
      chunk.searched = true
      chunk.interval
    }
  }
  private fun findNextUnsearchedChunk(): Pair<Chunk, Int>? {
    for (i in left..right) {
      left = i
      if (!searchChunks[i].searched) {
        return searchChunks[i] to i
      }
    }
    return null
  }
```

jonesho: Suggestion to revise the code above which passed the unit tests

jvm-libs/linea/web3j-extensions/src/test/kotlin/linea/web3j/SearchCursorTest.kt
```diff
@@ -0,0 +1,146 @@
...
+    // but it is a valid use case for this cursor because when searching for logs
+    // the predicate can try to go back but search chunks are exauhsted
+    assertThat(searchCursor.next(searchDirection = SearchDirection.BACKWARD)).isNull()
+  }
+
+  @Test
+  fun `next iterates follows direction without repeating - forward 3`() {
+    val searchCursor = SearchCursor(from = 0uL, to = 100uL, chunkSize = 5)
+    // chunks:
+    // 0..4,   5..9,   10..14, 15..19, 20..24,  25..29, 30..34, 35..39, 40..44, 45..49,
```
jonesho, line 176:
```suggestion
    // 0..4, 5..9, 10..14, 15..19, 20..24, 25..29, 30..34, 35..39, 40..44, 45..49,
```

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/SearchCursor.kt
```diff
@@ -0,0 +1,76 @@
+package linea.web3j
+
+import linea.SearchDirection
+
```
jonesho, line 4:
```suggestion
import kotlin.math.max
import kotlin.math.min
```

jonesho: For the code suggestion below

jvm-libs/linea/web3j-extensions/src/test/resources/log4j2.xml
```diff
@@ -2,12 +2,15 @@
 <Configuration status="warn">
   <Appenders>
     <Console name="console" target="SYSTEM_OUT">
-      <PatternLayout pattern="[%-5level] %d{yyyy-MM-dd HH:mm:ss.SSS} [%t] %c{1} - %msg%n"/>
+      <PatternLayout pattern="[%-5level] %d{yyyy-MM-dd HH:mm:ss.SSS} %t %c{1} - %msg%n"/>
+<!--      <PatternLayout pattern="%msg%n"/>-->
```
jonesho: Can the above commented code be removed?

config/coordinator/coordinator-docker.config.toml
```diff
@@ -126,12 +126,11 @@ fee-history-reward-percentile=15
 last-hash-search-window=25
 anchoring-receipt-polling-interval="PT01S"
 max-receipt-retries=120
-# Number of children blocks to wait before considering a block "finalized"
-# and elegible for conflation and
+# Number of children blocks to wait before considering a won't be reverted and elegible for conflation.
 # this a workaround to mitigate Geth fork issues with Clique PoA
```
jonesho:
```suggestion
# this is a workaround to mitigate Geth fork issues with Clique PoA
```

config/coordinator/coordinator-docker.config.toml
```diff
@@ -126,12 +126,11 @@ fee-history-reward-percentile=15
 last-hash-search-window=25
 anchoring-receipt-polling-interval="PT01S"
 max-receipt-retries=120
-# Number of children blocks to wait before considering a block "finalized"
-# and elegible for conflation and
+# Number of children blocks to wait before considering a won't be reverted and elegible for conflation.
```
jonesho:
```suggestion
# Number of children blocks to wait before considering they won't be reverted and elegible for conflation.
```

jvm-libs/linea/web3j-extensions/src/test/kotlin/linea/web3j/SearchCursorTest.kt
```diff
@@ -0,0 +1,146 @@
...
+  @Test
+  fun `next follows binary search when direction is provided`() {
+    val searchCursor = SearchCursor(from = 1uL, to = 100uL, chunkSize = 10)
+
+    assertThat(searchCursor.next(SearchDirection.FORWARD)).isEqualTo(41UL to 50UL)
+    assertThat(searchCursor.next(SearchDirection.FORWARD)).isEqualTo(71UL to 80UL)
+    assertThat(searchCursor.next(SearchDirection.FORWARD)).isEqualTo(81UL to 90UL)
+    assertThat(searchCursor.next(SearchDirection.FORWARD)).isEqualTo(91UL to 100UL)
+  }
+
```
jonesho, line 63:
```suggestion
  @Test
  fun `next follows binary search when direction BACKWARD is provided`() {
    val searchCursor = SearchCursor(from = 1uL, to = 100uL, chunkSize = 10)
    
    assertThat(searchCursor.next(SearchDirection.BACKWARD)).isEqualTo(41UL to 50UL)
    assertThat(searchCursor.next(SearchDirection.BACKWARD)).isEqualTo(11UL to 20UL)
    assertThat(searchCursor.next(SearchDirection.BACKWARD)).isEqualTo(1UL to 10UL)
  
```

jonesho: Suggested to add all BACKWARD direction test case

jonesho (commented):
Just halfway through the review, will continue tmr :)

state-recover/test-cases/src/integrationTest/kotlin/linea/staterecover/StateRecoveryManualReplayToLocalStackIntTest.kt
jpnovais: This may go away. It's meant to handover a reproducible away for Besu team to debug issues. 

Also, make is already so polluted so I may leave it here for now at least.

jvm-libs/generic/logging/src/main/kotlin/linea/logging/TimeMeasureLogger.kt
jpnovais: It's not advised for prod but it was meant to be used everywhere. This is extremely useful wen we want o measure things quickly in prod code, take a decision, and them implement metrics in final solution.

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/Web3JLogsSearcher.kt
jpnovais: Yes, it was thanks.

jvm-libs/linea/clients/interfaces/src/main/kotlin/linea/EthLogsSearcher.kt
jpnovais: I want interface name to be generic (Like a sort(), find(predicate)), not linked to the underlying implementation.

I thought about comparator, but did not use it because it does not really compare Log A with Log B. Hence I felt predicate would be more appropriate because it's more indicating if wants to keep searching based on any arbitrary logic.

jvm-libs/linea/besu-libs/build.gradle
jpnovais: that was my first try, it broke the build because this version is not availe in default hyperledger besu group. It would require more build refactors and effort without knowing it works. So leaving as is for now.

docker/compose.yml
jpnovais: Yes, it's a tempoeary dev workaround until the plugin is working. Once working and release is done we will rely on that one.

However, for CI purposes, this may actually be a good approach because we can check if localchanges are still compatible.

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/SearchCursor.kt
jpnovais: will make it internal, open for the for the tests.

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/SearchCursor.kt
jpnovais: Good suggestion, thanks

jvm-libs/linea/web3j-extensions/src/test/kotlin/linea/web3j/SearchCursorTest.kt
```diff
@@ -0,0 +1,166 @@
...
+    val searchCursor = SearchCursor(from = 1uL, to = 200uL, chunkSize = 10)
+
+    assertThat(searchCursor.next(searchDirection = null)).isEqualTo(91UL to 100UL)
+
+    assertThat(searchCursor.next(searchDirection = SearchDirection.FORWARD)).isEqualTo(141UL to 150UL)
+    assertThat(searchCursor.next(searchDirection = SearchDirection.FORWARD)).isEqualTo(171UL to 180UL)
+    assertThat(searchCursor.next(searchDirection = null)).isEqualTo(151UL to 160UL)
+    // This backward is not technically valid in a typical binary search,
+    // but it is a valid use case for this cursor because when searching for logs
+    // the predicate can try to go back but search chunks are exauhsted
```
jonesho:
```suggestion
    // the predicate can try to go back but search chunks are exhausted
```

jonesho: Same for the other "exhausted" typo in the same file

jvm-libs/linea/web3j-extensions/src/test/kotlin/linea/web3j/SearchCursorTest.kt
```diff
@@ -0,0 +1,166 @@
...
+    assertThat(searchCursor.next(searchDirection = null)).isEqualTo(91UL to 100UL)
+
+    assertThat(searchCursor.next(searchDirection = null)).isEqualTo(1UL to 10UL)
+    assertThat(searchCursor.next(searchDirection = SearchDirection.FORWARD)).isEqualTo(101UL to 110UL)
+    assertThat(searchCursor.next(searchDirection = SearchDirection.BACKWARD)).isEqualTo(51UL to 60UL)
+    assertThat(searchCursor.next(searchDirection = null)).isEqualTo(11UL to 20UL)
+    // This backward is not technically valid in a typical binary search,
+    // but it is a valid use case for this cursor because when searching for logs
+    // the predicate can try to go back but search chunks are exauhsted
+    assertThat(searchCursor.next(searchDirection = SearchDirection.BACKWARD)).isNull()
```
jonesho, line 137: I don't get how this test case could be possible: in `1..10` we were instructed to search forward but in `11..20` we were instructed to search backward? This seems contradicting, unless here we assume the given `searchDirection` can be arbitrary, and if that's the case, we should put comment to indicate that in this test case, otherwise, reader would be confused

state-recover/appcore/logic/src/main/kotlin/linea/staterecover/BlobDecompressorAndDeserializer.kt
```diff
@@ -0,0 +1,147 @@
...
+    var blockNumber = startBlockNumber
+    val startTime = Clock.System.now()
+    logger.debug("start decompressing blobs: startBlockNumber={} {} blobs", startBlockNumber, blobs.size)
+    val decompressedBlobs = blobs.map { decompressor.decompress(it) }
+    return SafeFuture
+      .collectAll(decompressedBlobs.map(::decodeBlocksAsync).stream())
+      .thenApply { blobsBlocks: List<List<Block>> ->
+        blobsBlocks.flatten().map { block ->
+          val header = BlockHeaderFromL1RecoveredData(
+            blockNumber = blockNumber++,
```
jonesho, line 93: Why not using `block.header.number` directly?

state-recover/appcore/logic/src/main/kotlin/linea/staterecover/LineaSubmissionEventsClientImpl.kt
```diff
@@ -1,40 +1,47 @@
...
+          l2BlockNumber < event.startBlockNumber -> SearchDirection.BACKWARD
+          l2BlockNumber > event.endBlockNumber -> SearchDirection.FORWARD
+          else -> null
+        }
+      }
+    ).thenApply { it?.let { DataFinalizedV3.fromEthLog(it) } }
   }
 
   override fun findDataFinalizedEventByStartBlockNumber(
     l2BlockNumber: ULong
```
jonesho: suggest to name this `l2StartBlockNumberInclusive`

jvm-libs/linea/clients/interfaces/src/main/kotlin/linea/EthLogsSearcher.kt
Filter94: You're right, that strictly speaking unary function isn't a comparator, but a function that returns non-boolean result isn't a predicate either.

In practice what happens here is you compare block number internally with a closure and this is where the ternary result comes from (-1, 0, 1). Having order defined is a fundamental property required for the binary search and you're using it.  So even if your comparator is a closure rather than a pure function, I still see it as a comparator. And that's the main reason I'm keen on this suggestion, because binary search doesn't work on predicates, by using a ternary "predicate" we're inevitably hiding a comparison.

Alternative way of arranging the code is to:
1. Use a comparator with 2 arguments
2. One of the arguments would be a generic target value T we're searching for.
3. The second can be `EthLog`. Mapping logic will be a part of the comparator logic
4. Comparator will compare `T` to `EthLog`, defining the order, making binary search possible.

Even cleaner way would be to divide the current single function into a mapping function and a comparator function of 2 arguments of the same type, but it's not a worthy investment in our case

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/SearchCursor.kt
Filter94: Just a note that regardless of the approach, in our case the worst case for our search is linear, so my suggestion is based on intuition and it's to improve "average" case

state-recover/besu-plugin/src/main/kotlin/linea/staterecover/plugin/PluginOptions.kt
```diff
@@ -0,0 +1,96 @@
...
+  @CommandLine.Option(
+    names = ["--plugin-$cliOptionsPrefix-l1-rpc-endpoint"],
+    description = ["L1 RPC endpoint"],
+    required = true
+  )
+  lateinit var l1RpcEndpoint: URI
+
+  @CommandLine.Option(
+    names = ["--plugin-$cliOptionsPrefix-shomei-endpoint"],
+    description = ["L1 RPC endpoint"],
```
jonesho: wrong description

state-recover/besu-plugin/src/main/kotlin/linea/staterecover/plugin/PluginOptions.kt
```diff
@@ -0,0 +1,96 @@
...
+    required = true
+  )
+  lateinit var shomeiEndpoint: URI
+
+  @CommandLine.Option(
+    names = ["--plugin-$cliOptionsPrefix-blobscan-endpoint"],
+    description = ["L1 RPC endpoint"],
```
jonesho: wrong description

state-recover/appcore/clients-interfaces/src/main/kotlin/linea/staterecover/RecoveryStatusPersistence.kt
```diff
@@ -0,0 +1,81 @@
...
+        RecoveryStatusEnvelopeDto(FileVersion.V1, status)
+      )
+    )
+  }
+
+  private fun loadStatusFromFileOrCreateIfDoesNotExist(): RecoveryStatusV1Dto? {
+    // eager file write because if we cannot read/write the file, then application shall throw and fail early
+    return if (!file.exists()) {
+      // Create the file with an empty status
+      saveToFile(status = null)
```
Filter94, line 58: Won't `file.canWrite()` be enough?

jpnovais: I tried to save because if it fails throw the error with the underlying reason, Otherwise need if returns false wont know exactly why...

state-recover/appcore/logic/src/main/kotlin/linea/staterecover/BlobDecompressorAndDeserializer.kt
jpnovais: because blobs don't have the block number. So we need to calculate is here.

jvm-libs/linea/web3j-extensions/src/test/kotlin/linea/web3j/SearchCursorTest.kt
jpnovais: I don't understand the contradiction. In a binary search we can jump forward/backwardm (aka left/right) depending on currently evaluated elements. 

Also, bare in mind that the predicate does not have visibly if there are elements between chunk 1..10 and 11..20.

jvm-libs/linea/web3j-extensions/src/test/kotlin/linea/web3j/SearchCursorTest.kt
jpnovais: this padding was on purpose for easier information scan by visual vertical alignment. I will keep it as is ;)

state-recover/appcore/clients-interfaces/src/test/kotlin/linea/staterecover/FileRecoveryStatusPersistenceTest.kt
```diff
@@ -0,0 +1,88 @@
...
+        assertThat(persistence.saveRecoveryStartBlockNumber(11U))
+        assertThat(persistence.getRecoveryStartBlockNumber()).isEqualTo(11UL)
+      }
+
+    // simulate application restart, the saved recovery start block number should be loaded from file OK
+    FileBasedRecoveryStatusPersistence(tempDir.resolve("recovery-status.json"))
+      .also { persistence ->
+        assertThat(persistence.getRecoveryStartBlockNumber()).isEqualTo(11UL)
+      }
+    println(Files.readString(tempDir.resolve("recovery-status.json")))
```
Filter94: I think this is a leftover

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/SearchCursor.kt
jpnovais: I was aiming to avoid side-effects on findNextUnsearchedChunk. But it's more efficient and I updated it, thanks 👍

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/SearchCursor.kt
jpnovais: Your suggested implementation has less lines of code and similar to my 1st implementation, but I find the current one easier to follow/read because it has less nested `if`s. I will favor Readbily (subjective I know) and keep the current implementation.

jvm-libs/linea/clients/interfaces/src/main/kotlin/linea/EthLogsSearcher.kt
jpnovais: Ok, I will rename to comparator.

jpnovais: How would you rename the `shallContinueToSearchPredicate` to?

Filter94: A few ideas:
* logsOrdering
* compareEthLogToTarget / targetComparator

`Closure` may come as a suffix

state-recover/test-cases/src/integrationTest/kotlin/linea/staterecover/StateRecoveryManualReplayToLocalStackIntTest.kt
jpnovais: this can happen when the replayed data has L1 -> L2 anchored messages. But in the reply stage we deploy a new contract that does not them and the verifications fail on L1. This has happen to me when I tried to use dataset generated from running thee e2e tests, which have anchoring messages.

jvm-libs/linea/clients/interfaces/src/main/kotlin/linea/EthLogsSearcher.kt
jpnovais: I don't find those more suggestive of the function purpose: indicate if it shall continue the search, if so which direction next. 

what about:
- `shallContinueToSearch`
- `nextSearchDirection`

state-recover/besu-plugin/build.gradle
```diff
@@ -0,0 +1,71 @@
...
+  id 'net.consensys.zkevm.kotlin-library-conventions'
+  id 'com.gradleup.shadow' version '8.3.5'
+}
+
+group = 'build.linea.staterecover'
+archivesBaseName = 'linea-staterecover-plugin'
+version = '0.0.1-rc2'
+
+dependencies {
+  compileOnly("info.picocli:picocli:4.7.6") {
```
Filter94: Can we use libs.versions.toml for version management here?

state-recover/appcore/domain-models/src/main/kotlin/build/linea/staterecover/TransactionFromL1RecoveredData.kt
```diff
@@ -0,0 +1,103 @@
+package build.linea.staterecover
+
+import net.consensys.encodeHex
+import java.math.BigInteger
+
+data class TransactionFromL1RecoveredData(
```
Filter94: Can we get away of our own transaction type? Maybe if we parse RLP Besu's Transaction directly, when transaction object is extended (with Prague fork, for example), it would be just a matter of updating Besu dependency on State recovery side. What benefit are we getting from defining our own type?

jvm-libs/linea/clients/interfaces/src/main/kotlin/linea/EthLogsSearcher.kt
Filter94: If we stick to `SearchDirection`, `shallContinueToSearch` `nextSearchDirection` are indeed more descriptive

state-recover/besu-plugin/src/main/kotlin/linea/staterecover/plugin/PluginOptions.kt
```diff
@@ -0,0 +1,107 @@
...
+  val overridingRecoveryStartBlockNumber: ULong? = null
+) {
+  init {
+    require(l1PollingInterval >= 1.seconds) { "Polling interval=$l1PollingInterval must be greater that 1s." }
+  }
+}
+
+class PluginCliOptions {
+  companion object {
+    const val cliOptionsPrefix = "staterecovery"
```
Filter94: Could as well put `--plugin-` here

state-recover/appcore/domain-models/src/main/kotlin/build/linea/staterecover/TransactionFromL1RecoveredData.kt
jpnovais: Maybe, but rather keep it like this because:
- this expresses better exactly what can be recovered from L1 and what not. 
- core logic is not coupled to Besu domain model

this way, `importBlocks` method has a very clear boundary if shall be provided. Using besu domain model we would  need to send Fake signatures and so on and this would make behaviour harder to understand.

state-recover/besu-plugin/src/main/kotlin/linea/staterecover/plugin/RecoveryModeManager.kt
```diff
@@ -0,0 +1,103 @@
...
+  val headBlockNumber: ULong
+    get() = currentBlockNumber
+
+  // Enum representing the states of recovery mode
+  private enum class RecoveryModeState {
+    NORMAL_MODE,
+    RECOVERY_MODE
+  }
+
+  private var recoveryModeState = RecoveryModeState.NORMAL_MODE
```
Filter94, line 35: I couldn't find any meaningful usage of it, other than switching the mode. What is the difference between recovery mode and normal mode?

state-recover/besu-plugin/src/main/kotlin/linea/staterecover/plugin/RecoveryModeManager.kt
```diff
@@ -0,0 +1,103 @@
...
+
+class RecoveryModeManager(
+  private val synchronizationService: SynchronizationService,
+  private val p2pService: P2PService,
+  private val miningService: MiningService,
+  private val recoveryStatePersistence: RecoveryStatusPersistence
+) :
+  BesuEvents.BlockAddedListener {
+  private val log: Logger = LogManager.getLogger(RecoveryModeManager::class.java.name)
+  private val triggered = AtomicBoolean(false)
```
Filter94, line 21: Why do we need a separate `triggered` on top of `recoveryModeState`? Can't we use either of the two?

state-recover/besu-plugin/src/main/kotlin/linea/staterecover/plugin/RecoveryModeManager.kt
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.Logger
+import org.hyperledger.besu.plugin.data.AddedBlockContext
+import org.hyperledger.besu.plugin.services.BesuEvents
+import org.hyperledger.besu.plugin.services.mining.MiningService
+import org.hyperledger.besu.plugin.services.p2p.P2PService
+import org.hyperledger.besu.plugin.services.sync.SynchronizationService
+import java.util.concurrent.atomic.AtomicBoolean
+
+class RecoveryModeManager(
```
Filter94, line 13: Just to make sure I understand, from Genesis block until block `targetBlockNumber` the node is expected to sync as usual, from the peers, without L1 data. Is this correct?

state-recover/clients/blobscan-client/src/testFixtures/kotlin/linea/staterecover/clients/blobscan/FakeBlobFetcher.kt
```diff
@@ -0,0 +1,36 @@
...
+import linea.staterecover.BlobFetcher
+import tech.pegasys.teku.infrastructure.async.SafeFuture
+
+data class BlobInfo(
+  val startBlockNumber: ULong,
+  val endBlockNumber: ULong,
+  val blob: ByteArray
+)
+
+class FakeBlobFetcher(
```
Filter94: This one doesn't seem to be used anywhere

state-recover/clients/execution-layer-json-rpc-client/src/main/kotlin/build/linea/staterecover/clients/el/ExecutionLayerJsonRpcClient.kt
```diff
@@ -33,38 +40,65 @@ class ExecutionLayerJsonRpcClient internal constructor(
       }
   }
 
-  override fun lineaEngineImportBlocksFromBlob(blocks: List<BlockL1RecoveredData>): SafeFuture<Unit> {
+  override fun lineaEngineImportBlocksFromBlob(blocks: List<BlockFromL1RecoveredData>): SafeFuture<Unit> {
     return rpcClient
       .makeRequest(
-        method = "linea_engine_importBlocksFromBlob",
+        method = "linea_importBlocksFromBlob",
```
Filter94, line 46: Do we actually need json RPC state recovery API or in process API is sufficient?

state-recover/test-cases/src/integrationTest/kotlin/linea/staterecover/StateRecoverAppWithFakeExecutionClientIntTest.kt
```diff
@@ -0,0 +1,299 @@
...
+      aggregationsAndBlobs = finalizationsAfterCutOff
+    )
+
+    await()
+      .atMost(1.minutes.toJavaDuration())
+      .pollInterval(1.seconds.toJavaDuration())
+      .untilAsserted {
+        assertThat(
+          executionLayerClient.lineaGetStateRecoveryStatus().get()
+            .also { println(it) }
```
Filter94: Do we want to leave this here?

state-recover/test-cases/src/integrationTest/kotlin/linea/staterecover/StateRecoveryManualReplayToLocalStackIntTest.kt
Filter94: * We deploy a new rollup
* we try to replay blob and finalization transactions to it
* we fail to do it because of the anchoring
* then finalizedL2BlockNumber() is @ the latest aggregation parsed from the files somehow, despite the failure
* zk state root hash matches what's on the contract

What I don't understand:
* If it fails because of the anchoring, how come, finalization is @ `lastAggregation` level?
* What are we testing here?

state-recover/test-cases/src/integrationTest/kotlin/linea/staterecover/StateRecoverAppWithFakeExecutionClientIntTest.kt
```diff
@@ -0,0 +1,299 @@
...
+class StateRecoverAppWithFakeExecutionClientIntTest {
+  private val log = LogManager.getLogger("test.case.StateRecoverAppWithFakeExecutionClientIntTest")
+  private lateinit var stateRecoverApp: StateRecoverApp
+  private lateinit var aggregationsAndBlobs: List<AggregationAndBlobs>
+  private lateinit var executionLayerClient: FakeExecutionLayerClient
+  private lateinit var fakeStateManagerClient: FakeStateManagerClient
+  private lateinit var transactionDetailsClient: TransactionDetailsClient
+  private lateinit var lineaContractClient: LineaRollupSmartContractClientReadOnly
+
+  private lateinit var contractClientForSubmittions: LineaRollupSmartContractClient
```
Filter94: Typo, `contractClientForSubmittions`  ->`contractClientForSubmissions`

Filter94 (approved):

state-recover/test-cases/src/integrationTest/kotlin/linea/staterecover/StateRecoverAppIntTest.kt
```diff
@@ -0,0 +1,207 @@
...
+class StateRecoverAppIntTest {
+  private val log = LogManager.getLogger("test.case.StateRecoverAppIntTest")
+  private lateinit var stateRecoverApp: StateRecoverApp
+  private lateinit var aggregationsAndBlobs: List<AggregationAndBlobs>
+  private lateinit var executionLayerClient: ExecutionLayerClient
+  private lateinit var stateManagerClient: StateManagerClientV1
+  private lateinit var transactionDetailsClient: TransactionDetailsClient
+  private lateinit var lineaContractClient: LineaRollupSmartContractClientReadOnly
+
+  private lateinit var contractClientForSubmittions: LineaRollupSmartContractClient
```
jonesho: typo `contractClientForSubmissions`

state-recover/test-cases/build.gradle
```diff
@@ -0,0 +1,83 @@
...
+      .toAbsolutePath().toString()
+
+  description = "Runs integration tests."
+  group = "verification"
+  useJUnitPlatform()
+
+  classpath = sourceSets.integrationTest.runtimeClasspath
+  testClassesDirs = sourceSets.integrationTest.output.classesDirs
+
+  //    dependsOn(":localStackForStateRecoverComposeUp")
```
jonesho, line 67: Why this is commented out? I tried running the integration test locally but failed, and also shouldn't `zkbesu-shomei-sr` and `shomei-sr` be included in `localStackForStateRecover`?

state-recover/appcore/logic/src/main/kotlin/linea/staterecover/StateSynchronizerService.kt
```diff
@@ -0,0 +1,164 @@
...
+    dataFinalizedV3: DataFinalizedV3
+  ): SafeFuture<BlockNumberAndHash> {
+    return blobDecompressor
+      .decompress(
+        startBlockNumber = dataFinalizedV3.startBlockNumber,
+        blobs = dataSubmissions.flatMap { it.blobs }
+      ).thenCompose { decompressedBlocks: List<BlockFromL1RecoveredData> ->
+        log.debug("importing blocks={}", dataFinalizedV3.intervalString())
+        blockImporterAndStateVerifier
+          .importBlocks(decompressedBlocks)
```
jonesho, line 123: Just curious, when we import blocks to besu, what happens if some of the importing blocks was already present in besu world state? E.g. if besu was stopped importing blocks from peers after block#100, and then later it resumes importing blocks from state recovery for finalized event of block#71..200, I guess re-importing block#71..100 would be handled smoothly by besu without any issue, right?

state-recover/test-cases/build.gradle
jpnovais: I will fix those details once I get all finished and CI done.

state-recover/appcore/logic/src/main/kotlin/linea/staterecover/StateSynchronizerService.kt
jpnovais: >block#70..200, I guess re-importing block#70..100 would be handled smoothly by besu without any issue, right?

I think it would not work. However, it's not the intended design because Shomei cannot revert state, so we cannot reimport/override the chain. For this reason, did not even test that scenario as it's not possible.

state-recover/besu-plugin/src/main/kotlin/linea/staterecover/plugin/RecoveryModeManager.kt
jpnovais: Yes, that is correct.

state-recover/clients/execution-layer-json-rpc-client/src/main/kotlin/build/linea/staterecover/clients/el/ExecutionLayerJsonRpcClient.kt
jpnovais: It was used when Gabriel was exposing this in a [Plugin here](https://github.com/Consensys/linea-state-recovery-plugin) before I migrated it to this to have just 1 plugin. 

Very likely will remove it. I just didn't do it because I was wondering if would be useful for testing to implement the JsonRpcExecutionLayerClient and run the StateRecoverApp as a stand alone. But likely I will remove this.

state-recover/appcore/logic/src/main/kotlin/linea/staterecover/StateSynchronizerService.kt
jonesho: ic, for my understanding, I'm more concern, in real-world, is it possible for the above scenario? That is the state recovery mode was enabled when the besu chain head was at #100, but the last finalized L2 block number at that moment was #70, and hence the besu would stop syncing blocks with peers after #100, and then when the next finalized event came #71..200, the besu would be re-importing blocks #71..100 (#101..200 should be fine as they were new to besu)

The above scenario essentially represents the case that `headBlockNumber > lastFinalizedBlock` [(mentioned here)](https://github.com/Consensys/linea-monorepo/blob/5d738e476262e2543dae93713fafc3a0bca4c2b1/state-recover/test-cases/src/integrationTest/kotlin/linea/staterecover/StateRecoverAppWithFakeExecutionClientIntTest.kt#L171), which iiuc, means importing blocks that were already been pulled seems very likely to happen, or do I miss something here?

> For this reason, did not even test that scenario as it's not possible

So that's why this test case `when recovery is disabled and headBlock is after lastFinalizedBlock resumes from headBlock+1` is not present in the integration test?

jpnovais: Oh, I see your point now. In that scenario where the head is at 100, the finalization is 70..200 then the recovery app will fetch the blobs uncompress them skip blocks 70..100, and resume synching from 101..200.

jonesho: @jpnovais A late comment after I realized this was already merged :)
> the recovery app will fetch the blobs uncompress them skip blocks 70..100, and resume synching from 101..200.

Thanks for the reply, but regarding the above, I couldn't find where in the code that would explicitly skip importing the already-imported blocks (i.e. 71..100), what I understand is that the whole chunk of the decompressed aggregation blocks (i.e. 71..200) would be imported as shown at line 124 above

Btw if it's a valid scenario, should we also add a test case for `when recovery is disabled and headBlock is after lastFinalizedBlock resumes from headBlock+1` in **StateRecoverAppWithFakeExecutionClientIntTest**, wdyt?

jpnovais: I will look into adding that test case scenario. thanks

jpnovais: @jonesho I added the test [here](https://github.com/Consensys/linea-monorepo/pull/565), good call I actually found a bug

state-recover/besu-plugin/src/main/kotlin/linea/staterecover/plugin/RecoveryModeManager.kt
jpnovais: This class was not implemented by me, I just translated to Kotlin to get things working ASAP. Will look into now to see if can remove the eventual redundancy.

jpnovais: Addressed [here](https://github.com/Consensys/linea-monorepo/pull/565/commits/b95cbf739127161de59dc1aac7788dd78bb9d390)

state-recover/test-cases/src/integrationTest/kotlin/linea/staterecover/StateRecoveryManualReplayToLocalStackIntTest.kt
jpnovais: > then finalizedL2BlockNumber() is @ the latest aggregation parsed from the files somehow, despite the failure

In case if inconsistency (Anchoring) that won't be true. The tx will be executed but reverted therefore the stateroot won't match.

jvm-libs/linea/web3j-extensions/src/test/kotlin/linea/web3j/Web3JLogsSearcherIntTest.kt
jpnovais: yeah, the names are not great. 

the tests do 3 scenarios:
- target is before the range
- target is after the range
- target is middle of the range

I will rename for a more intuitive name

jpnovais: Yeah there was some redundancy. I fixed it followup [PR](https://github.com/Consensys/linea-monorepo/pull/565). thank

PR #327, opened 2024-11-22
feat: improve tx-exclusion-api performance and increase size of reaso…

jonesho:
…n message column

This PR implements issue(s) #

### Checklist

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -162,7 +162,7 @@ dockerCompose {
     ]
     useComposeFiles = ["${project.rootDir.path}/docker/compose.yml"]
     waitForHealthyStateTimeout = Duration.ofMinutes(3)
-    waitForTcpPorts = false
+    waitForTcpPorts = true
     removeOrphans = true
     noRecreate = true
     projectName = "docker"
--- a/docker/compose-local-dev-traces-v2.overrides.yml
+++ b/docker/compose-local-dev-traces-v2.overrides.yml
@@ -68,7 +68,7 @@ services:
   prover-v3: # prover compatible with the traces from zkbesu
     container_name: prover-v3
     hostname: prover-v3
-    image: consensys/linea-prover:${PROVER_TAG:-8544ea1}
+    image: consensys/linea-prover:${PROVER_TAG:-c4abdd9}
     platform: linux/amd64
     # to avoid spinning up on CI for now
     profiles: [ "l2" ]
--- a/docker/compose-local-dev-traces-v2.overrides.yml
+++ b/docker/compose-local-dev-traces-v2.overrides.yml
@@ -68,7 +68,7 @@ services:
   prover-v3: # prover compatible with the traces from zkbesu
     container_name: prover-v3
     hostname: prover-v3
-    image: consensys/linea-prover:${PROVER_TAG:-bba9677}
+    image: consensys/linea-prover:${PROVER_TAG:-8544ea1}
     platform: linux/amd64
     # to avoid spinning up on CI for now
     profiles: [ "l2" ]
--- a/jvm-libs/generic/persistence/db/src/testFixtures/kotlin/net/consensys/zkevm/persistence/db/test/CleanDbTestSuiteParallel.kt
+++ b/jvm-libs/generic/persistence/db/src/testFixtures/kotlin/net/consensys/zkevm/persistence/db/test/CleanDbTestSuiteParallel.kt
@@ -26,8 +26,6 @@ abstract class CleanDbTestSuiteParallel {
 
   private fun createDataSource(databaseName: String): DataSource {
     return PGSimpleDataSource().also {
-      it.ssl = false
-      it.gssEncMode = "disable"
       it.serverNames = arrayOf(host)
       it.portNumbers = intArrayOf(port)
       it.databaseName = databaseName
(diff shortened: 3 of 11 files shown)


transaction-exclusion-api/persistence/rejectedtransaction/src/main/kotlin/net/consensys/zkevm/persistence/dao/rejectedtransaction/RejectedTransactionsPostgresDao.kt
```diff
@@ -123,7 +123,7 @@ class RejectedTransactionsPostgresDao(
   private val deleteFullTransactionsSql =
     """
       delete from $fullTransactionsTable
-      where tx_hash not in (select x.tx_hash from $rejectedTransactionsTable x)
+      f where not exists (select null from $rejectedTransactionsTable x where f.tx_hash = x.tx_hash)
```
jonesho, line 126: "not exists" would generally perform better than "not in": https://stackoverflow.com/questions/52638557/postgresql-huge-performance-difference-when-using-in-vs-not-in and https://explainextended.com/2009/09/16/not-in-vs-not-exists-vs-left-join-is-null-postgresql/

jvm-libs/generic/persistence/db/src/testFixtures/kotlin/net/consensys/zkevm/persistence/db/test/CleanDbTestSuiteParallel.kt
```diff
@@ -26,6 +26,8 @@ abstract class CleanDbTestSuiteParallel {
 
   private fun createDataSource(databaseName: String): DataSource {
     return PGSimpleDataSource().also {
+      it.ssl = false
+      it.gssEncMode = "disable"
```
jonesho: These are to address the random failures of integration tests in GHA pipeline

transaction-exclusion-api/persistence/rejectedtransaction/src/main/resources/db/V003__increase_reject_reason_size_and_add_index_on_tx_hash_schema.sql
```diff
@@ -0,0 +1,6 @@
+-- =======================================================
+-- rejected_transactions table
+-- =======================================================
+alter table if exists rejected_transactions
+  alter column reject_reason type varchar(1024);
+create index if not exists tx_hash_idx on rejected_transactions using btree (tx_hash);
```
jonesho, line 6: Add index on tx_hash to speed up the delete SQL query

transaction-exclusion-api/persistence/rejectedtransaction/src/main/resources/db/V003__increase_reject_reason_size_and_add_index_on_tx_hash_schema.sql
```diff
@@ -0,0 +1,6 @@
+-- =======================================================
+-- rejected_transactions table
+-- =======================================================
+alter table if exists rejected_transactions
+  alter column reject_reason type varchar(1024);
```
jonesho, line 5: increase to 1024 which would greatly remove the error logs from tx-exclusion-api as we're still seeing a few errors from mainnet that has reason message over 512 (mostly 522 characters long)

jonesho:
Related issue: https://github.com/Consensys/zkevm-monorepo/issues/4019

jpnovais (approved):

build.gradle
```diff
@@ -162,7 +162,7 @@ dockerCompose {
     ]
     useComposeFiles = ["${project.rootDir.path}/docker/compose.yml"]
     waitForHealthyStateTimeout = Duration.ofMinutes(3)
-    waitForTcpPorts = false
+    waitForTcpPorts = true
```
jonesho, line 165: Setting this to true should avoid race condition issue when postgres container starts up and immediately followed by tx-exclusion-api integration test, given the healthy state check might not be good enough to ensure that the postgres container is ready to accept connection, especially when the tx-exclusion-api integration test only requires postgres container alone.

docker/compose-local-dev-traces-v2.overrides.yml
```diff
@@ -68,7 +68,7 @@ services:
   prover-v3: # prover compatible with the traces from zkbesu
     container_name: prover-v3
     hostname: prover-v3
-    image: consensys/linea-prover:${PROVER_TAG:-bba9677}
+    image: consensys/linea-prover:${PROVER_TAG:-8544ea1}
```
jonesho: update the image tag of prover-v3 to cope with the recent change of prover config

jvm-libs/generic/persistence/db/src/testFixtures/kotlin/net/consensys/zkevm/persistence/db/test/CleanDbTestSuiteParallel.kt
```diff
@@ -26,8 +26,6 @@ abstract class CleanDbTestSuiteParallel {
 
   private fun createDataSource(databaseName: String): DataSource {
     return PGSimpleDataSource().also {
-      it.ssl = false
-      it.gssEncMode = "disable"
       it.serverNames = arrayOf(host)
```
jonesho, line 29: Reverted the above changes as it didn't fix the tx-exclusion-api integration test random failure issue

PR #334, opened 2024-11-22
Feat: Dummy PI Circuit

Tabaie:
This PR introduces a dummy circuit for public input interconnection. As with other circuits, the `dev` mode in the config will result in creating/proving with this small dummy circuit and `full` will do the same with the full circuit.
A third mode `light` is available for a real circuit that doesn't prove/check the correctness of its Keccak256 hashes.

Tests for the aggregation circuit are also fixed and refactored.

Also, the tests revealed that complete arithmetic is sometimes necessary for the verification of the PI circuit. The option is added back in.


AlexandreBelling:
@Tabaie Do we still want to push this PR forward?

Tabaie:
Yes I'll work on it next week

PR #352, opened 2024-11-28
Prover: implements the top-level of the limitless prover

AlexandreBelling:
This PR implements issue(s) #

### Checklist


prover/protocol/distributed/distributed.go
```diff
@@ -0,0 +1,97 @@
...
+func prepare(comp *wizard.CompiledIOP) {
+
+	mimc.CompileMiMC(comp)
+	specialqueries.RangeProof(comp)
+	specialqueries.CompileFixedPermutations(comp)
+
+	inclusion.IntoLogDerivativeSum(comp)
+	permutation.IntoGrandProduct(comp)
+	projection.IntoGrandSum(comp)
+	local.IntoDistributedLocal(comp)
```
AlexandreBelling: What do you do with these two functions?

prover/protocol/distributed/distributed.go
```diff
@@ -0,0 +1,97 @@
...
+
+func addSplittingStep(comp *wizard.CompiledIOP, disc ModuleDiscoverer) {
+	panic("unimplemented")
+}
+
+func extractDistModule(comp *wizard.CompiledIOP, disc ModuleDiscoverer, moduleName moduleName) DistributedModule {
+	panic("unimplemented")
+}
+
+func aggregator(n int, idsModules []DistributedModule, moduleNames []string) *wizard.CompiledIOP {
```
AlexandreBelling: Please give self-contained names `n` is super obscure and does not help understanding what the parameter is for

prover/protocol/distributed/distributed.go
```diff
@@ -0,0 +1,97 @@
...
+	for _, modName := range moduleLs {
+		// Segment Compilation;
+		// Compile every dist module with the same sequence of compilation steps for uniformity
+		distMod := extractDistModule(initialWizard, disc, modName)
+		distModules = append(distModules, distMod)
+	}
+
+	// for each [DistributedModule] it checks the consistency among
+	// its replications where the number of replications is maxNbSegments.
+	aggr := aggregator(maxNbSegments, distModules, moduleLs)
```
AlexandreBelling: `distModules, moduleLs)` -> That's redundant

prover/protocol/distributed/distributed.go
Soleimani193: for global it adds the boundary columns. for local it is probably useless.

prover/protocol/distributed/distributed.go
Soleimani193: agree. Also should we not add the list of Public inputs as a field to DistributedModule. so that the aggregator can check the consistency among them? Or maybe we can add it to the copmiledIOP objects in the DistibutedModule.

prover/protocol/distributed/distributed.go
```diff
@@ -0,0 +1,79 @@
...
+	LookupPermProj *wizard.CompiledIOP
+	GlobalLocal    *wizard.CompiledIOP
+}
+
+// ModuleDiscoverer a set of methods responsible for the horizontal splittings (i.e., splitting to modules)
+type ModuleDiscoverer interface {
+	// Analyze is responsible for letting the module discoverer compute how to
+	// group best the columns into modules.
+	Analyze(comp *wizard.CompiledIOP)
+	ModuleList(comp *wizard.CompiledIOP) []string
```
arijitdutta67: []string -> []moduleName

prover/protocol/distributed/distributed.go
```diff
@@ -0,0 +1,79 @@
...
+	FindModule(col ifaces.Column) moduleName
+	FindNbSegments(moduleName string, SegParam map[string]int) int
```
arijitdutta67: A comment describing SegParam is helpful. Considering it gives the max segment size for a given module,  consider `map[string]int` -> `map[moduleName]int`.

AlexandreBelling: Why is that a member of the module discovery interface? And what are the parameters? From the code I understand that the function returns `Segparams[moduleName]`, you don't need a function to do that

prover/protocol/compiler/lookup/compiler.go
```diff
@@ -108,6 +114,10 @@ func CompileLogDerivative(comp *wizard.CompiledIOP) {
 		va.Name = zC.Name
 	}
 
+	if comp.HasSeed {
```
AlexandreBelling: I don't understand what comp.HasSeed has to do with how the lookup compiler works. The only important thing is how the coin are sampled but then this logic should be implemented in the `coin` package.

prover/protocol/compiler/lookup/prover.go
```diff
@@ -330,6 +331,11 @@ func (z zAssignmentTask) run(run *wizard.ProverRuntime) {
 
 			run.AssignColumn(z.Zs[frag].GetColID(), sv.NewRegular(packedZ))
 			run.AssignLocalPoint(z.ZOpenings[frag].ID, packedZ[len(packedZ)-1])
+
+			// compute the global sum of all ZOpenings
+			zSum.Add(&zSum, &packedZ[len(packedZ)-1])
 		}
 	})
+	// assign the public input
+	run.AssignColumn(z.PI.GetColID(), sv.NewRegular([]field.Element{zSum}))
```
AlexandreBelling: This should be in a dedicated logDerivative compiler object.

prover/protocol/distributed/distributed.go
```diff
@@ -69,10 +76,56 @@ func extractDistModule(
...
+		LookupPermProj: &wizard.CompiledIOP{},
+		GlobalLocal:    &wizard.CompiledIOP{},
+	}
+
+	for _, qName := range comp.QueriesNoParams.AllUnignoredKeys() {
+
+		// Filter LPP queries
+		q := comp.QueriesNoParams.Data(qName)
+
+		switch v := q.(type) {
```
AlexandreBelling: Since you have a switch, it would be simpler to have one function for global, one for local and one for lookup, and one for projection and one for permutation. That way you don't have to do a switch in a switch, it's nicer to read that way.

You may want to move the `if disc.QueryIsInModule(v, moduleName)` our of the switch as it happens in all situations

AlexandreBelling:
Closing as the branch was rebased and another PR was opened: #813

PR #358, opened 2024-11-28
Fixed sigterm propagation, making geth containers faster to stop and …

Filter94:
…restart

Removed unused Geth scripts

This PR implements issue(s) #

### Checklist


docker/geth/scripts/node.sh
```diff
@@ -56,7 +56,7 @@ else
     echo "Starting geth VALIDATOR with BOOTNODE enabled on port '$BOOTNODE_PORT' with '/boot.key'"
     mkdir -p $DATA_DIR/keystore
     cp /keystore/$etherbase.json $DATA_DIR/keystore
-    geth --datadir $DATA_DIR \
+    exec geth --datadir $DATA_DIR \
```
jonesho, line 59: @Filter94 just to understand a bit more the context, are all these changes for ensuring the geth command would run in the same shell process such that sigterm would propagate properly to the process? and previously the sigterm would go to the parent shell process instead and hence slow down the termination of the child process which actually runs the geth command?

docker/compose.yml
```diff
@@ -108,8 +108,7 @@ services:
       MAX_TXDATA_BYTES: 30000 #30 KBytes
       ZKGETH_LOWER_GAS_MARGIN_PERCENTS: 120
       ZKGETH_UPPER_GAS_MARGIN_PERCENTS: 300
-    entrypoint: ""
-    command: [ "sh", "/scripts/node.sh", "1337", "/genesis.json", "0x0", "0x1C9C380", "0xF4240" ]
```
gauravahuja, line 112: What is the benefit of removing command?

jpnovais (commented):
Thanks for the fix, very much appreciated! those 10 stop were really annoying

docker/compose.yml
Filter94: Just 1 less key to define, no functional benefit

docker/geth/scripts/node.sh
Filter94: I think your explanation is accurate. I did it by [this](https://cloud.theodo.com/en/blog/docker-processes-container) guide

docker/compose.yml
jonesho: iiuc, by removing the `command` and just use `entrypoint` could replace the default entrypoint `/bin/sh -c` with the node.sh command and hence avoid running the node.sh command under a parent shell process, @Filter94 correct?

Filter94: I'm not sure about that, I'd need to check

PR #360, opened 2024-11-28
Fix PI Initial Rolling Hash Update

Tabaie:
Previously the PI circuit considered the initial rolling hash values (and message numbers) to be the previous values (those being changed).

Among less significant changes, this PR changes those values to be the rolling hash and message number associated with the first rolling hash update event in the execution.


gusiri (commented):
We might want to consider changing the package names `pi_interconnection` and `public_input` to better align with Go's naming conventions.
https://google.github.io/styleguide/go/decisions.html#package-names

Tabaie:
> We might want to consider changing the package names `pi_interconnection` and `public_input` to better align with Go's naming conventions. https://google.github.io/styleguide/go/decisions.html#package-names

How's `publicinputinterconnection` or some abbreviation thereof?

AlexandreBelling:
> > We might want to consider changing the package names `pi_interconnection` and `public_input` to better align with Go's naming conventions. https://google.github.io/styleguide/go/decisions.html#package-names
> 
> How's `publicinputinterconnection` or some abbreviation thereof?

I'd leave them as they are over replacing them with obscure `pii` or overly long `publicinputinterconnection`

prover/circuits/execution/pi.go
```diff
@@ -85,145 +96,78 @@ func (l *L2MessageHashes) CheckSumMiMC(api frontend.API) frontend.Variable {
...
 		for i := range v {
 			rc.Check(v[i], 8)
 		}
 	}
-	for i := range pi.FinalRollingHash {
-		rc.Check(pi.FinalRollingHash[i], 8)
+	for i := range spiq.FinalRollingHashUpdate {
+		rc.Check(spiq.FinalRollingHashUpdate[i], 8)
 	}
+	rc.Check(spiq.FinalBlockNumber, 64)
```
AlexandreBelling, line 123: I am fine leaving theses constraints, but just know that this is range-checked by the wizard already. We can leave it for the sake of carefulness.

Tabaie: We can keep in mind that we can drop these checks if the PI circuit ever became a bottleneck.

AlexandreBelling (approved):

gusiri (approved):

Issue #366, opened 2024-11-29
"Wholesale" blob compression


Tabaie:
## Problem Statement

Currently the blob compressor works on its payload block by block. This robs the underlying lzss compressor of some opportunities to find longer back references. When recompressing the entire blob payload on sample Sepolia blobs, we get the following results:
```
recompressing  4454961-4457351-e1e96ae51f742703127930ac10b1271bba431670cb58df245d447fec45b5abb0-getZkBlobCompressionProof.json
recompressed size 130016 -> 121312
Utilization 99.19% -> 92.55%

recompressing  4457352-4459439-11b64ec1986a1b3f91b2682f863ba295174141bdafa43991384e18f9cecb4424-getZkBlobCompressionProof.json
recompressed size 130048 -> 122496
Utilization 99.22% -> 93.46%

recompressing  4459440-4461292-033bc5d3158db24465a9320fdae89118d95e1f6d0845d6e6d274ee093458256f-getZkBlobCompressionProof.json
recompressed size 130016 -> 123296
Utilization 99.19% -> 94.07%

recompressing  4461293-4463201-a53da3be9d3a80b5ffe7f69942bfdd75d9728d11d00d0f46d6ef564f9f6b03c3-getZkBlobCompressionProof.json
recompressed size 130048 -> 123104
Utilization 99.22% -> 93.92%

recompressing  4463202-4464942-ae8a1b736e364c93e76034d42560ba2ac4937c3b13c3df49cc64f231acbd3231-getZkBlobCompressionProof.json
recompressed size 130048 -> 123712
Utilization 99.22% -> 94.38%

recompressing  4464943-4466962-e747d64b41e98ba6211a9af05971beee141fab4e4a4e3b7a9ed3ba0db297ea51-getZkBlobCompressionProof.json
recompressed size 129952 -> 122592
Utilization 99.15% -> 93.53%

recompressing  4466963-4469060-e33f117a3638c9d15d6f9295f4b69632cf5b48558ada42536287abca693c8e97-getZkBlobCompressionProof.json
recompressed size 130016 -> 122368
Utilization 99.19% -> 93.36%

recompressing  4469061-4471076-6d2296765837ff017e6a6fb9043bf95c8cac0d5ac9cb6c15818f1821f9ef9ed5-getZkBlobCompressionProof.json
recompressed size 129824 -> 122496
Utilization 99.05% -> 93.46%

[...]

PR #367, opened 2024-11-30
Compressor Optimizer

Tabaie:
This PR implements issue https://github.com/Consensys/linea-monorepo/issues/366.

The compressor now attempts to recompress everything from scratch upon encountering a full blob, before attempting to bypass compression altogether.

### Checklist

* [x] I have successfully run tests, style checker and build against my new changes locally.


gbotrel:
So to sum up, this is a non-breaking change, we could compress better (by ~7%) if we re-compress the full blob each time we attempt to append a block; but doing so is too slow so you propose to keep the original method, that would give a "preliminary result", and change the result behind the scenes between calls?

Wouldn't this have some side effects @jpnovais ? (i.e basically the compressor could say "we compressed block 1, it takes N Bytes, current blob is now at 100kB", then recompress it with more context and update the internal state to "current blob is 98kB" without notifying the coordinator. ). 

Re implementation, before introducing an async optimizer, I'ld prefer to understand perf constraints better; i.e. how long it takes now, how long it would take if we recompress all the blob at each append, and within what limit we need operate . i.e. if we say the compressor could take as much as XXXms , then we may just want to have a simpler cleaner code and kill this async pattern.

jpnovais:
My input on this optimization is:

Context:
- The coordinator does not actively keep track of the current blob size, that is the responsibility of the compressor, which is aware of blob limit. 
- Coordonator calls `fun CanWrite(data: ByteArray, data_len: Int): Boolean` before calling Write, if returns false then coordinator knows the blob is full and starts a new one

My take based on the above:
- As mentioned, Asyc can be problematic so I would avoid it. Also, I think we don't need it here.
- My suggestion: when coordinator calls `CanWrite` before returning false, the compressor would try to perform the **full compression** an see if that extra blob fits. 
  - **pros**: 
    -  lazy computation - it only does the "full re-compression" when it reaches the limit;
    - 100% transparent to the coordinator and keeps same API. This leaves room for further internal code performance optimizations ;

@Tabaie @gbotrel WDYT?

gbotrel:
Yep that would make less CPU use to do it only when full . But this last call to "CanWrite" may be 10x slower than the previous calls.
So what are the bounds perf-wise we need to operate in? If a call to CanWrite takes 500ms is that acceptable?

jpnovais:
> So what are the bounds perf-wise we need to operate in? If a call to CanWrite takes 500ms is that acceptable? (not saying it will, just want a rough order of magnitude of bound)

It's ok to have a call to CanWrite that takes 500ms **at the of the blob**, as long as the preceding calls are not affected timewise.

Tabaie:
I agree, doing it synchronously at the end is a good idea. In fact it's similar to the "no compress" logic we already have.

gbotrel (dismissed):
LGTM 👍 I would probably just add one or two test to ensure this is correctly triggered and that the internal state is correctly reset after

Tabaie:
> > The difference, from the user's point of view, is output nondeterminism, as it will differ based on how much time the compressor has had to optimize.
> 
> What does this mean, "_based on how much time the compressor has had to optimize_" from the coordinator's PoV?

This was for the parallel optimizer so it no longer applies. Removing from description.

Tabaie:
> > > The difference, from the user's point of view, is output nondeterminism, as it will differ based on how much time the compressor has had to optimize.
> > 
> > 
> > What does this mean, "_based on how much time the compressor has had to optimize_" from the coordinator's PoV?
> 
> This was for the parallel optimizer so it no longer applies. Removing from description.

@jpnovais Does it look good now?

jpnovais (approved):

jpnovais:
@Tabaie sure. Sorry for the delay and not reverting on the 1st comment.

prover/lib/compressor/blob/v1/blob_maker.go
```diff
@@ -410,3 +443,10 @@ func (bm *BlobMaker) RawCompressedSize(data []byte) (int, error) {
 
 	return n, nil
 }
+
+func wrapError(err error, format string, args ...any) error {
```
ivokub, line 447: Seems the method is only used once. I would remove it for now.

prover/lib/compressor/blob/v1/blob_maker.go
```diff
@@ -191,45 +189,80 @@ func (bm *BlobMaker) Write(rlpBlock []byte, forceReset bool) (ok bool, err error
...
+	recompressionAttempted := false
+	revert := func() error { // from this point on, we may have recompressed the entire payload in one go
+		// that makes the compressor's own Revert method unusable.
+		bm.header.removeLastBlock()
+		if !recompressionAttempted { // fast path for most "CanWrite" calls
+			return bm.compressor.Revert()
+		}
+		// we can't use the compressor's own Revert method because we tried to compress in one go.
+		bm.compressor.Reset()
+		_, err := bm.compressor.Write(payload[:prevLen])
```
ivokub, line 210: Cannot we have issues here? payload refers to the internal buf of compressor and we write it back to the compressor. Couldn't we potentially mess up the buffer due to overwrites?

prover/lib/compressor/blob/v1/blob_maker.go
```diff
@@ -191,45 +189,80 @@ func (bm *BlobMaker) Write(rlpBlock []byte, forceReset bool) (ok bool, err error
...
 	// check that the header + the compressed data fits in the blob
-	fitsInBlob := encode.PackAlignSize(bm.buf.Len()+bm.compressor.Len(), fr381.Bits-1) <= bm.Limit
-	if !fitsInBlob {
-		// first thing to check is if we bypass compression, would that fit?
+	if !fitsInBlob() {
+		recompressionAttempted = true
+
+		// first thing to check is whether we can fit the block if we recompress everything in one go, known to achieve a higher ratio.
+		bm.compressor.Reset()
+		if _, err = bm.compressor.Write(payload); err != nil {
```
ivokub, line 220: Similarly here, we write back the internal buf of compressor to compressor. I wonder if we may have overwrites here.

prover/lib/compressor/blob/v1/blob_maker.go
```diff
@@ -191,45 +189,80 @@ func (bm *BlobMaker) Write(rlpBlock []byte, forceReset bool) (ok bool, err error
...
+			err = fmt.Errorf("when recompressing the blob: %w", err)
+
+			if innerErr := revert(); innerErr != nil {
+				err = fmt.Errorf("%w\n\tto recover from write failure: %w", innerErr, err)
+			}
+
+			return false, err
```
ivokub, line 227: Could simplify error logic using [`errors.Join`](https://pkg.go.dev/errors#Join). In this case we could do:
```go
innerErr := revert()
return false, errors.Join(err, innerErr)
```
as `errors.Join` discards nil errors.

prover/lib/compressor/blob/v1/blob_maker.go
```diff
@@ -169,17 +168,16 @@ func (bm *BlobMaker) Write(rlpBlock []byte, forceReset bool) (ok bool, err error
 		if innerErr := bm.compressor.Revert(); innerErr != nil {
 			return false, fmt.Errorf("when reverting compressor because writing failed: %w\noriginal error: %w", innerErr, err)
```
ivokub, line 169: Can use `errors.Join` for simplifying error handling.

prover/lib/compressor/blob/v1/blob_maker.go
```diff
@@ -191,45 +189,80 @@ func (bm *BlobMaker) Write(rlpBlock []byte, forceReset bool) (ok bool, err error
...
 	}
 
 	// copy the compressed data to the blob
 	bm.packBuffer.Reset()
 	n2, err := encode.PackAlign(&bm.packBuffer, bm.buf.Bytes(), fr381.Bits-1, encode.WithAdditionalInput(bm.compressor.Bytes()))
 	if err != nil {
-		bm.compressor.Revert()
-		bm.Header.removeLastBlock()
+		err = fmt.Errorf("when packing blob: %w", err)
+		innerErr := revert()
```
ivokub, line 262: Can use `errors.Join`

prover/lib/compressor/blob/v1/blob_maker.go
```diff
@@ -191,45 +189,80 @@ func (bm *BlobMaker) Write(rlpBlock []byte, forceReset bool) (ok bool, err error
...
+			err = fmt.Errorf("when recompressing the blob: %w", err)
+
+			if innerErr := revert(); innerErr != nil {
+				err = fmt.Errorf("%w\n\tto recover from write failure: %w", innerErr, err)
+			}
+
+			return false, err
+		}
+		if fitsInBlob() {
+			goto bypass
```
ivokub, line 230: This path doesn't seem to be tested - e.g. first when compressing single block doesn't fit but recompressing full payload then fits.

ivokub (approved):
Seems good to me. Before merging can you confirm that by referencing to the same buffer when recompressing we don't have some internal overwrites of bytes.Buffer which could cause issues?

Also some comments about simplifying error handling.

That said - imo the current approach may be error prone when wanting to make changes in the future. It heavily relies on reverting the compressor in case the block doesnt fit (or when we only run CanWrite). And the state logic imo is complex enough that it is diffcult to follow and test.

What I would recommend is to add a method to clone the compressor using a pool and then perform operations on the cloned compressor. If write succeeds then we update the reference in `BlobMaker`, otherwise we discard the cloned compressor (by resetting and putting it back into the pool). Internally it would imo also require using a pool of `bytes.Buffer`.

Imo trying to optimize for memory here doesn't make a lot of sense as the blocks/blobs in general are small (imo blob maximally 2MB and maximum uncompressed data assuming 5x compression 10MB) relative to the rest of memory usage.

From my side this is not blocker for merging the PR for now, but would definitely simplify the implementation and guard against bugs in the future.

Issue #364, opened 2024-11-29
Exclude Build and test CI actions on purely document changes.


thedarkjester:
## Description
Exclude Build and test CI actions on purely document changes. Also try exclude gated e2e tests.

## Motivation
It delays merging needlessly and wastes resourced when the change is just documentation.

## Tasks
- [ ] Adjust Github actions/workflows to exclude certain tasks such as building and testing
- [ ] Adjust Github actions/workflows to not require the e2e tests for documentation

## Acceptance criteria
- [ ] No tests are run or compilations are run for documentation changes
- [ ] If documentation is included in code changes, tests and building etc. are run + gate check required

## Risks
- [ ] Required building and testing gets skipped

## Remember to
 - [x] Add `priority` and `team` labels

PR #370, opened 2024-12-02
[Feat] - Exclude *.md files from triggering CI workflow

kyzooghost:
This PR implements issue(s) #364 

### Checklist


.github/workflows/codeql.yml
```diff
@@ -14,6 +14,10 @@ name: "CodeQL"
 on:
   push:
     branches: [ "main" ]
+    paths:
+      # '!**.md' will not work by itself to exclude *.md files. See https://github.com/orgs/community/discussions/25369
```
VGau, line 18: Did you check this ?

.github/workflows/bridge-ui-e2e-tests.yml
```diff
@@ -6,11 +6,13 @@ on:
       - main
     paths:
       - 'bridge-ui/**'
+      - '!**.md'
```
VGau: I think it should be `!bridge-ui/**/*.md`. WDYT ?

.github/workflows/bridge-ui-e2e-tests.yml
```diff
@@ -6,11 +6,13 @@ on:
...
   push:
     branches:
       - main
     paths:
       - 'bridge-ui/**'
+      - '!**.md'
```
VGau: I think it should be `!bridge-ui/**/*.md`. WDYT ?

.github/workflows/bridge-ui-publish.yml
```diff
@@ -6,11 +6,13 @@ on:
       - main
     paths:
       - 'bridge-ui/**'
+      - '!**.md'
```
VGau: Same as previous comment

.github/workflows/bridge-ui-publish.yml
```diff
@@ -6,11 +6,13 @@ on:
...
   push:
     branches:
       - main
     paths:
       - 'bridge-ui/**'
+      - '!**.md'
```
VGau: Same as previous comment

.github/workflows/run-smc-tests.yml
```diff
@@ -6,13 +6,15 @@ on:
       - 'contracts/**'
       - 'testdata/**'
       - 'prover/**'
+      - '!**.md'
```
VGau: I think it should be `!contracts/**/*.md`. WDYT ?

VGau (changes requested):

.github/workflows/bridge-ui-e2e-tests.yml
kyzooghost: good pickup

.github/workflows/codeql.yml
kyzooghost: yep

Didn't test on codeql.yml, but tested using `!**.yml` on the main.yml workflow

.github/workflows/run-smc-tests.yml
kyzooghost: fixed

.github/workflows/bridge-ui-publish.yml
kyzooghost: fixed

.github/workflows/bridge-ui-publish.yml
kyzooghost: fixed

.github/workflows/bridge-ui-e2e-tests.yml
kyzooghost: fixed

.github/workflows/bridge-ui-e2e-tests.yml
kyzooghost: fixed

VGau (dismissed):

alainncls (dismissed):

.github/workflows/main.yml
```diff
@@ -20,6 +20,8 @@ jobs:
       traces-api-facade: ${{ steps.filter.outputs.traces-api-facade }}
       transaction-exclusion-api: ${{ steps.filter.outputs.transaction-exclusion-api }}
       finalized-tag-updater: ${{ steps.filter.outputs.finalized-tag-updater }}
+      # If 'false' => all changed files are *.md - https://stackoverflow.com/a/77068717
+      not-markdown-only: ${{ steps.filter-out.outputs.not-markdown-only }}
```
Filter94: What about "has-non-markdown-changes"?

kyzooghost: agreed it's less of a mindbend than 'not-markdown-only'

kyzooghost: Most explicit I can think of atm is `has-changes-other-than-markdown`

jpnovais: In the meetings we mentioned "markdown" as a example of a change that shall not trigger the pipeline. What we want to achieve is to a "label/term" that represent changes that shall be skipped, e.g what about a `jpeg` inside /doc folder? so I think we should find more suggestive name that does not stick to markdown.

Suggestions:
- has-code-changes
- has-changes-to-{build/test}
- ??? 

PS: I noticed that `filter-out` step allow to extend with other extensions. Shall we add the whole `/doc` folder there?

thedarkjester: how about `has-code-changes`

thedarkjester: I should have scrolled down - `has-code-changes` works for me. Skipping whole doc folders is good too.

kyzooghost: Good points

I'm thinking `has-changes-requiring-tests`

I will add the image file extensions to the list
- .svg
- .png
- .jpeg
- .jpg
- .gif
- .webp

I don't think we can add an entire folder to the list because `predicate-quantifier: 'every'` means that every filter has to be matched (default for dorny/paths-filter@v3 is that just one filter has to match).
- `/docs` currently has only *.md, *,svg and *.png files in it. So adding these file extensions should cover `/docs` changes for now

--
EDIT `has-code-changes` it is

EDIT 2: Never mind the last paragraph, can just add `!docs/**` as a one liner and it will work

Filter94: Yes. Or `requires-ci` to be super direct 🙂

jpnovais: I like 
`has-changes-requiring-build`

why: because if we put .jpg/.txt as static asset that shall be bundled, that should be build and published.
Example: UI application may have static assets in it.

kyzooghost: Good point re: static assets for building hmm...will need to combine a 'file extension blacklist' + 'folder blacklist' then

Actually nvm putting it that way, take back what I said re: not being able to add entire folder to the list

jpnovais (approved):

PR #377, opened 2024-12-03
Prover: couple of fixes betav1

AlexandreBelling:
This PR implements issue(s) #

### Checklist


prover/Makefile
```diff
@@ -57,7 +57,7 @@ zkevm/define/define.go: corset
 ## Generate the setup for the execution prover (to be run with S3 access)
 ##
 setup: bin/prover
-	bin/prover setup --config ./config/config-integration-full.toml --dict ./lib/compressor/compressor_dict.bin --assets-dir ./prover-assets
+	bin/prover setup --config ./config/config-sepolia-full.toml --dict ./lib/compressor/compressor_dict.bin --assets-dir ./prover-assets
```
AlexandreBelling, line 60: We will have to pay attention to which config we are using. Because for mainnet we don't want to use the configuration of sepolia

prover/circuits/execution/pi.go
```diff
@@ -14,16 +15,16 @@ import (
...
-	FinalRollingHashUpdate      [32]frontend.Variable
-	FinalRollingHashMsgNumber   frontend.Variable
+	DataChecksum                 frontend.Variable
+	L2MessageHashes              L2MessageHashes
+	InitialBlockTimestamp        frontend.Variable
+	FinalStateRootHash           frontend.Variable
+	FinalBlockNumber             frontend.Variable
+	FinalBlockTimestamp          frontend.Variable
+	InitialRollingHashUpdate     [32]frontend.Variable
+	FirstRollingHashUpdateNumber frontend.Variable
```
AlexandreBelling, line 25: That's a renaming collectively done with Arya to clarify what the value actually store (the first occurence of a rolling hash update)

prover/circuits/aggregation/circuit.go
```diff
@@ -173,7 +173,6 @@ func verifyClaimBatch(api frontend.API, vks []emVkey, claims []proofClaim) error
 	}
 
 	for i := 1; i < len(vks)-1; i++ { // TODO @Tabaie make sure these don't generate any constraints
-		fmt.Printf("checking base vk #%d/%d\n", i+1, len(vks)-1)
```
AlexandreBelling, line 176: I figured we could remove that printf

prover/circuits/pi-interconnection/assign.go
```diff
@@ -23,6 +24,9 @@ type Request struct {
 	Decompressions []blobsubmission.Response
 	Executions     []public_input.Execution
 	Aggregation    public_input.Aggregation
+	// Path to the compression dictionary. Used to extract the execution data
+	// for each execution.
+	DictPath string
```
AlexandreBelling, line 29: That was needed because otherwise, the code won't work if not run from the linea-monorepo folder

prover/config/config_test.go
```diff
@@ -43,8 +44,15 @@ func TestEnvironment(t *testing.T) {
 			config, err := NewConfigFromFile(file.Name())
 			assert.NoError(err, "when processing %s", file.Name())
 
+			// take the first word of both the match and the environment
```
AlexandreBelling, line 47: Without this change, the test tried to compare sepolia and sepolia-full

prover/lib/compressor/blob/blob.go
```diff
@@ -25,27 +23,7 @@ func GetVersion(blob []byte) uint16 {
...
-	return wd[:i], nil
-}
-
-func GetDict() ([]byte, error) {
-	repoRoot, err := GetRepoRootPath()
-	if err != nil {
-		return nil, err
-	}
-	dictPath := filepath.Join(repoRoot, "prover/lib/compressor/compressor_dict.bin")
+func GetDict(dictPath string) ([]byte, error) {
```
AlexandreBelling, line 26: The goal is here to confine any function with a dependency on [GetRepoRootPath] to testing packages only

prover/protocol/wizard/gnark_verifier.go
```diff
@@ -179,39 +179,42 @@ func (c *WizardVerifierCircuit) generateAllRandomCoins(_ frontend.API) {
...
-			/*
-				Also include the prover's allegations for all evaluations
-			*/
-			queries := c.Spec.QueriesParams.AllKeysAt(currRound - 1)
-			for _, qName := range queries {
-				// Implicitly, this will panic whenever we start supporting
-				// a new type of query params
-				params := c.GetParams(qName)
-				params.UpdateFS(c.FS)
+			if !c.Spec.DummyCompiled {
```
AlexandreBelling, line 182: That condition had been removed from verifier.go and it was there only to ensure circuits are not too big when using the dummy compiler. That had an impact when testing, but we figured it is better to just not test circuit generation uusing dummy compilers anyway.

prover/public-input/aggregation.go
```diff
@@ -105,7 +106,6 @@ type AggregationFPI struct {
 	LastFinalizedRollingHashMsgNumber uint64
 	ChainID                           uint64 // for now we're forcing all executions to have the same chain ID
 	L2MessageServiceAddr              types.EthAddress
-	NbL2Messages                      uint64 // TODO not used in hash. delete if not necessary
```
AlexandreBelling, line 108: That field was unused

gusiri (approved):

Issue #272, opened 2024-10-31
Move the Postman outside of the linea sdk package


VGau:
## Description
The Linea SDK package currently includes the Postman service. We propose removing this service to streamline the SDK and eliminate unnecessary components.

## Motivation
Removing the Postman service will reduce the package size and simplify the SDK. This enhances maintainability and aligns with best practices by keeping the package focused on essential functionalities.

## Tasks
- [ ] Moving the Postman service code in another folder.
- [ ] Update any dependencies or references related to the Postman service.
- [ ] Test the SDK to ensure it functions correctly without the Postman service.
- [ ] Update documentation to reflect this change.

## Acceptance criteria
- [ ] The Postman service is completely removed from the SDK package.
- [ ] The SDK operates without errors after removal.
- [ ] All references to the Postman service are updated or removed.
- [ ] Documentation is updated accordingly.

## Risks
- [ ] Other components may depend on the Postman service.
- [ ] Potential for unexpected issues post-removal.

## Remember to
 - [X] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #381, opened 2024-12-03
Feat/272 split sdk and postman

VGau:
This PR implements issue(s) #272 

### Checklist


postman/src/core/clients/blockchain/IMessageRetriever.ts
```diff
@@ -0,0 +1,7 @@
+import { MessageSent } from "@consensys/linea-sdk";
```
jonesho: This interface is not being used in postman, so can this file be deleted?

postman/scripts/sendMessageOnL1.ts
```diff
@@ -1,17 +1,10 @@
 import { BytesLike, ContractTransactionReceipt, Overrides, Wallet, JsonRpcProvider } from "ethers";
```
jonesho, line 1: @VGau should we move this script to sdk? I dun see how it's relevant to postman except as a handy tool for testing local postman

postman/scripts/sendMessageOnL2.ts
```diff
@@ -2,9 +2,9 @@ import { ContractTransactionReceipt, Overrides, JsonRpcProvider, Wallet } from "
 import { config } from "dotenv";
```
jonesho, line 2: @VGau Same, should we move this script to sdk? I dun see how it's relevant to postman except as a handy tool for testing local postman

postman/src/core/errors/Base.ts
```diff
@@ -0,0 +1,11 @@
+export class BaseError extends Error {
```
jonesho: Can we name this file as `BaseError.ts` instead of `Base.ts`?

postman/scripts/sendMessageOnL1.ts
VGau: This script is specific to the Postman so I think we should not put it inside the SDK ? WDYT ?

postman/scripts/sendMessageOnL2.ts
VGau: This script is specific to the Postman so I think we should not put it inside the SDK ? WDYT ?

postman/src/core/errors/Base.ts
VGau: yes sure :)

sdk/src/clients/gas/DefaultGasProvider.ts
```diff
@@ -54,7 +50,7 @@ export class DefaultGasProvider implements IEthereumGasProvider<TransactionReque
       };
     }
```
jonesho, line 55: By revisiting this class, I now think we should change `config.maxFeePerGas` to `maxFeePerGasCap`, this would be aligned with the config meaning (also aligning with coordinator config naming) which is to set a cap for maxFeePerGas , which means we should also:
- `config.enforceMaxGasFee` to `enforceMaxFeePerGasCap`
- `if (maxPriorityFeePerGas > this.config.maxFeePerGas)` below should change to `if (feeHistory.baseFeePerGas + maxPriorityFeePerGas > this.config.maxFeePerGasCap)`
wdyt? Btw in our envs, for postman, `getGasFees` is used only in L1 to L2 claim and `baseFeePerGas` on L2 should be neglible as its always 7 wei, right?

sdk/src/clients/providers/provider.ts
```diff
@@ -0,0 +1,34 @@
...
+        throw new BaseError("Error getting fee data");
+      }
+
+      return { maxPriorityFeePerGas, maxFeePerGas };
+    }
+  };
+}
+
+export class Provider extends ProviderMixIn(ethers.JsonRpcProvider) {}
+export class BrowserProvider extends ProviderMixIn(ethers.BrowserProvider) {}
```
jonesho, line 34: Should we declare the following alias type:
```
export type MixedProvider = Provider | BrowserProvider
```
given that `Provider | BrowserProvider` has been used in a lot places, `MixedProvider` might be a bad name though :(

sdk/src/clients/providers/lineaProvider.ts
```diff
@@ -0,0 +1,82 @@
...
+     */
+    public isCacheValid(currentBlockNumber: number): boolean {
+      return this.cacheIsValidForBlockNumber >= BigInt(currentBlockNumber);
+    }
+  };
+}
+
+export class LineaProvider extends LineaProviderMixIn(ethers.JsonRpcProvider) {}
+
+export class LineaBrowserProvider extends LineaProviderMixIn(ethers.BrowserProvider) {}
```
jonesho, line 82: Should we declare the following alias type:
```
export type MixedLineaProvider = LineaProvider | LineaBrowserProvider
```
given that `LineaProvider | LineaBrowserProvider` has been used in a lot places, `MixedLineaProvider` might be a bad name though

postman/scripts/sendMessageOnL1.ts
jonesho: This script seems to me is agnostic to postman as it just sends messges on L1, so I'm thinking it should be moved to sdk for more generic usage, but up to you, I'm fine keeping it here, same for the script to send message  on L2

postman/src/core/clients/blockchain/IProvider.ts
```diff
@@ -1,16 +1,16 @@
 import { GasFees } from "./IGasProvider";
 
-export interface IChainQuerier<TransactionReceipt, Block, TransactionRequest, TransactionResponse, JsonRpcProvider> {
-  getCurrentNonce(accountAddress?: string): Promise<number>;
-  getCurrentBlockNumber(): Promise<number>;
+export interface IProvider<TransactionReceipt, Block, TransactionRequest, TransactionResponse, Provider> {
```
jonesho, line 3: @VGau as we discussed before, do u think we can get rid of these template class names `TransactionReceipt, Block, TransactionRequest, TransactionResponse, Provider` and use the `ether.js` class types directly? Same for other interfaces. These look verbose, and as you mentioned, we indeed would always use `ether.js` and there's no need for supporting other libraries, wdyt? We could have another PR to follow that up if u think its easier

sdk/src/clients/gas/DefaultGasProvider.ts
VGau: Fixed :)

sdk/src/clients/gas/DefaultGasProvider.ts
```diff
@@ -59,9 +59,10 @@ export class DefaultGasProvider implements IEthereumGasProvider<TransactionReque
...
 
     const feeHistory = await this.fetchFeeHistory();
+    const baseFeePerGas = this.calculateBaseFeePerGas(feeHistory.baseFeePerGas);
     const maxPriorityFeePerGas = this.calculateMaxPriorityFee(feeHistory.reward);
 
-    if (maxPriorityFeePerGas > this.config.maxFeePerGasCap) {
+    if (baseFeePerGas + maxPriorityFeePerGas > this.config.maxFeePerGasCap) {
```
jonesho: Why do we need `baseFeePerGas + maxPriorityFeePerGas` here instead of just `maxPriorityFeePerGas`? `baseFeePerGas` is calcuated as average from the `feeHistory.baseFeePerGas` but on line 133 we just use the latest feeHistory baseFeePerGas times 2 as `const maxFeePerGas = BigInt(baseFeePerGas[baseFeePerGas.length - 1]) * 2n + maxPriorityFeePerGas;`

What's the point of calculating the above `baseFeePerGas` as `feeHistory.baseFeePerGas` average and putting it with check above?

VGau: Yes that's a mistake thanks Jones :)

jonesho (approved):

PR #403, opened 2024-12-06
Prover: fix empty to address decoding

AlexandreBelling:
This PR fixes a decoding error in the decompression logic. The error is without any effect on the production as decoding does not happen on the prover front but it shows up when testing.

The prover would decode "0x0" when given "null" as an address.

### Checklist


jpnovais (dismissed):

jpnovais (approved):

jpnovais (approved):

jpnovais:
@AlexandreBelling will this require an upgrade of the compressor as well?

AlexandreBelling:
> @AlexandreBelling will this require an upgrade of the compressor as well?

No, as it only changes things for the decoding parts (so not the compression part). This is not used by the compressor or by the coordinator, this only used to test the encoding scheme to ensure this is decodable.

jpnovais (approved):

gusiri (approved):

Issue #344, opened 2024-11-27
Publish jvm libs as separate importable jars


Filter94:
## Description
In order to reuse utilities we have in jvm-libs, we need to publish them individually

## Motivation
This will improve reusability of the utilities we've built in linea-monorepo

## Tasks
- [ ] Define how package identifiers are going to look like
- [ ] Implement publishing logic from main

## Acceptance criteria
- [ ] jvm-libs packages are published and publically available 

## Risks
- [ ] We shouldn't allow publishing from forks to Linea jar repo
- [ ] We shouldn't publish anything from feature branches

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #404, opened 2024-12-06
Implemented publishing of Linea JVM libs by Gradle

Filter94:
This PR implements issue(s) #344

### Checklist


jvm-libs/build.gradle
```diff
@@ -0,0 +1,32 @@
...
+  dependencies {
+    classpath(libs.jreleaser)
+  }
+}
+
+plugins {
+  id 'net.consensys.zkevm.kotlin-library-minimal-conventions'
+}
+
+Set<String> externalLibs = [
```
jpnovais: suggestion: `publicApis`, I suggest to add a comment to what is the intention of these.

jvm-libs/build.gradle
```diff
@@ -0,0 +1,32 @@
...
+  "blob-decompressor",
+  "blob-shnarf-calculator",
+  "l1-rollup",
+  "l2-message-service"
+]
+
+subprojects {
+  // Note that artifacts with generated code only will be skipped if :publish is called without preceding :build
+  def hasGeneratedSources = it.fileTree(dir: 'build/generated', includes: ['**/*.java', '**/*.kt']).files.size() > 0
+  def hasSourceFiles = it.fileTree(dir: 'src', includes: ['**/*.java', '**/*.kt']).files.size() > 0
```
jpnovais, line 23: the `subprojects {` with scans to the file system that run in the configuration phase can make gradle slower, did you measure the impact of this?

Even to run simple commands like `spotless, tasks, etc` we will pay the price... 

Also, I think checking just for `src` folder is good enough and may be lighter.

jpnovais (dismissed):

jvm-libs/build.gradle
Filter94: > did you measure the impact of this?

I tested it as is and the way you suggest, it doesn't seem to have a noticeable impact

jvm-libs/build.gradle
Filter94: I'll change it to `publicApiLibs` and will add a comment

jvm-libs/build.gradle
jpnovais: Can you define noticeable? 
Unless gradle eagerly loads the file tree regardless, a deep scanning of all project's files must be slower...  Also, not everyone works on top of the range laptop with fast SSDs like us.

jpnovais: To be precise, if takes more than 500ms IMO is worthy the optimization.
Do as you deam best

jvm-libs/linea/testing/teku-helper/build.gradle
jpnovais, line 1: FYI, I removed this in MY WIP PR

jpnovais (approved):

jvm-libs/build.gradle
Filter94: > Can you define noticeable?

If there is any difference it's not noticeable with manual runs. In order to see the real difference I'd have to benchmark it with hundreds of runs and look at the average to measure the impact of this change

PR #406, opened 2024-12-06
Coordinator: replace teku ExecutionPlayloadV1 by Domain Block class  - fix block encoding

jpnovais:
This PR implements issue(s) #

### Checklist


build.gradle
```diff
@@ -23,6 +23,12 @@ allprojects {
   apply plugin: 'java' // do not add kotlin plugin here, it will add unnecessary Kotlin runtime dependencies
   apply plugin: 'jacoco'
 
+  configurations.all {
+    resolutionStrategy {
+      force "org.web3j:core:${libs.versions.web3j.get()}"
```
Filter94: I pulled your code, removed this addition, built the project and it went fine. What was breaking for you so you had to add it? I want to check if we can work around it some other way on my machine

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/blockcreation/BlockCreationMonitor.kt
```diff
@@ -128,29 +127,29 @@ class BlockCreationMonitor(
...
                     "Shooting down conflation poller, " +
                       "chain reorg detected: block { blockNumber={} hash={} parentHash={} } should have parentHash={}",
-                    payload.blockNumber.longValue(),
-                    payload.blockHash.toHexString().subSequence(0, 8),
-                    payload.parentHash.toHexString().subSequence(0, 8),
-                    expectedParentBlockHash.get().toHexString().subSequence(0, 8)
+                    block.number,
+                    block.hash.encodeHex(),
+                    block.parentHash.encodeHex(),
+                    expectedParentBlockHash.get().encodeHex().subSequence(0, 8)
```
Filter94:
```suggestion
                    expectedParentBlockHash.get().encodeHex()
```
Maybe we should remove the truncation here as well for consistency?

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/blockcreation/BlockCreationMonitor.kt
```diff
@@ -21,7 +24,7 @@ import kotlin.time.Duration.Companion.days
 
 class BlockCreationMonitor(
   private val vertx: Vertx,
-  private val extendedWeb3j: ExtendedWeb3J,
+  private val web3j: Web3j,
```
Filter94: Why are we going back to plain Web3j here and in other places?

coordinator/app/src/test/kotlin/net/consensys/zkevm/coordinator/blockcreation/BlockCreationMonitorTest.kt
```diff
@@ -1,537 +0,0 @@
...
-import java.math.BigInteger
-import java.util.concurrent.Executors
-import java.util.concurrent.TimeUnit
-import kotlin.time.Duration
-import kotlin.time.Duration.Companion.milliseconds
-import kotlin.time.Duration.Companion.seconds
-import kotlin.time.toJavaDuration
-
-@ExtendWith(VertxExtension::class)
-class BlockCreationMonitorTest {
```
Filter94, line 44: Why are we removing this?

jvm-libs/generic/extensions/kotlin/src/test/kotlin/net/consensys/StringExtensionsTest.kt
```diff
@@ -27,4 +27,30 @@ class StringExtensionsTest {
     assertThat("this includes lorem ipsum".containsAny(stringList, ignoreCase = true)).isTrue()
     assertThat("this string won't match".containsAny(stringList, ignoreCase = true)).isFalse()
   }
+
+  @OptIn(ExperimentalStdlibApi::class)
```
Filter94: I didn't catch what experimental features we're using here

jvm-libs/linea/besu-rlp-and-mappers/src/main/kotlin/linea/domain/TransactionTypeMapper.kt
```diff
@@ -0,0 +1,23 @@
+package linea.domain
+
+import org.hyperledger.besu.datatypes.TransactionType
+
+fun TransactionType.toDomain(): linea.domain.TransactionType {
+  return when (this) {
+    TransactionType.FRONTIER -> linea.domain.TransactionType.FRONTIER
+    TransactionType.EIP1559 -> linea.domain.TransactionType.EIP1559
+    TransactionType.ACCESS_LIST -> linea.domain.TransactionType.ACCESS_LIST
+    TransactionType.BLOB -> linea.domain.TransactionType.BLOB
```
Filter94, line 10: Blobs and delegate code isn't really a part of our domain. Isn't it better to throw right here in case we get it from Besu? I also think that we should remove them from `TransactionType` so compiler helps those who is less familiar with the code to know what's supported and what is not

build.gradle
jpnovais: Yes, it works. The reason I added this is because we have multiple versions of web3j, through transitive dependencies and when coding/debugging intellIJ was linking to the wrong one. So did this a quick workaround and ended up pushing by accident.

I am aware that this is not a solution that scales! We shall use proper management like spring plugin but I don't have time to address it now.
I am happy to revert or keep it since it does not cause any harm anyway.

coordinator/app/src/test/kotlin/net/consensys/zkevm/coordinator/blockcreation/BlockCreationMonitorTest.kt
jpnovais: Like I mentioned on the standup, this test still needs to be done. I just removed to skipp a bunch of compilation errors. 

I plan to redo the test from scratch, but also reasses if BlockCreationMonitor needs to be as complex as it is.

jvm-libs/linea/web3j-extensions/src/test/kotlin/net/consensys/linea/web3j/DomainObjectMappersTest.kt
```diff
@@ -1,5 +1,6 @@
 package net.consensys.linea.web3j
 
+/*
```
Filter94, line 3: Maybe we can just delete it right away?

build.gradle
Filter94: If it doesn't cause issues for you, I'd like to remove it, because it creates a bad precedent

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/blockcreation/BlockCreationMonitor.kt
jpnovais: Because it was simpler this way and this extension is not properly designed and refactoring adapting would be more work. We created this interface to basically facilitate mock<> and return responses, if we create a fake rpc client this won't be necessary, we can tests against state that concrete implementations. That's my pan for new tests of BlockCreationMonitor.

Filter94: `getNetNextSafeBlock` is more complex now. Is it possible to make this change later, at the same time with the new tests?

jvm-libs/linea/web3j-extensions/src/test/kotlin/net/consensys/linea/web3j/DomainObjectMappersTest.kt
jpnovais: Yes, I will! I just didn't do to it yet until I have the test cases I the other test class.

jvm-libs/generic/extensions/kotlin/src/test/kotlin/net/consensys/StringExtensionsTest.kt
jpnovais: It had before, then I forgot to remove, will clean it, thanks

jvm-libs/linea/besu-rlp-and-mappers/src/main/kotlin/linea/domain/TransactionTypeMapper.kt
jpnovais: I prefer to have them here and throw an exception on the block creation monitor because we cannot conflate them. If for any reason we can one sequenced we will see clearly why it failed. 

Although, I don't have the time to implement 100% completeness of all cases (tx serialization, etc), but having all transaction types in the enum is valuable IMO because it raises awaress about those cases.

Filter94: Ok, up to you

gradle/libs.versions.toml
```diff
@@ -17,6 +17,7 @@ junit = "5.10.1"
 kotlinxDatetime = "0.6.1"
 ktlint = "0.47.0"
 log4j = "2.20.0"
+slf4j = "1.7.30"
```
jonesho: Wondering where is it being used?

jvm-libs/linea/besu-rlp-and-mappers/src/main/kotlin/linea/domain/MapperBesuToLineaDomain.kt
```diff
@@ -0,0 +1,63 @@
...
+
+    return block
+  }
+
+  fun mapToDomain(transaction: Transaction): linea.domain.Transaction {
+    return Transaction(
+      nonce = transaction.nonce.toULong(),
+      gasPrice = transaction.getGasPrice().getOrNull()?.toBigInteger()?.toULong(),
+      gasLimit = transaction.gasLimit.toULong(),
+      to = transaction.to?.getOrNull()?.toArray(),
```
jonesho: Without the ? operator should be sufficient as `to = transaction.to.getOrNull()?.toArray()`, same for the ? operators before `getOrNull()` below

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/blockcreation/BlockCreationMonitor.kt
jpnovais: done

coordinator/app/src/test/kotlin/net/consensys/zkevm/coordinator/blockcreation/BlockCreationMonitorTest.kt
```diff
@@ -111,427 +114,259 @@ class BlockCreationMonitorTest {
...
-    monitor.start().thenApply {
-      await()
-        .timeout(1.seconds.toJavaDuration())
-        .untilAsserted {
-          verify(blockCreationListener, times(1)).acceptBlock(BlockCreated(payload))
-          verify(blockCreationListener, never()).acceptBlock(BlockCreated(payload2))
-          assertThat(monitor.nexBlockNumberToFetch).isEqualTo(startingBlockNumberInclusive + 1)
-        }
-      testContext.completeNow()
+  fun `should stop when reorg is detected above blocksToFinalization limit - manual intervention necessary`() {
```
Filter94, line 271: Maybe we can instead remove reorg logic completely?

jvm-libs/generic/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/httpserver/HttpJsonRpcServer.kt
```diff
@@ -14,11 +14,11 @@ import org.apache.logging.log4j.Logger
...
   private val port: UInt,
   private val path: String,
-  private val requestHandler: Handler<RoutingContext>
+  private val requestHandler: Handler<RoutingContext>,
+  val serverName: String = ""
 ) : AbstractVerticle() {
   private val log: Logger = LogManager.getLogger(this.javaClass)
   private lateinit var httpServer: HttpServer
-
   val bindedPort: Int
```
Filter94: Bound*

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/blockcreation/BlockCreationMonitor.kt
```diff
@@ -21,7 +20,7 @@ import kotlin.time.Duration.Companion.days
 
 class BlockCreationMonitor(
   private val vertx: Vertx,
-  private val extendedWeb3j: ExtendedWeb3J,
+  private val web3j: ExtendedWeb3J,
```
Filter94, line 23: Just to give more context. Once we switch to Beta V2 on Mainnet, there is no reason to monitor block creation, entrypoint for conflation would be trace counting API directly.

We didn't create new, traces-v2 alternative to BlockCreationMonitor to avoid excessive code branching in conflation flow, which is not trivial to follow already.

jvm-libs/generic/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/JsonRpcMessageProcessor.kt
```diff
@@ -174,7 +181,13 @@ class JsonRpcMessageProcessor(
     return SimpleTimerCapture<String>(meterRegistry, "jsonrpc.serialization.response")
       .setDescription("Time of json response serialization")
       .setTag("method", requestContext.method)
-      .captureTime { Json.encode(requestContext.result.merge()) }
+      .captureTime {
```
Filter94, line 184: I don't really understand this change. Can you please explain the difference?

jvm-libs/generic/json-rpc/src/testFixtures/kotlin/linea/jsonrpc/TestingJsonRpcServer.kt
```diff
@@ -0,0 +1,175 @@
...
+  val recordRequestsResponses: Boolean = false,
+  val serverName: String = "FakeJsonRpcServer",
+  loggerName: String = serverName,
+  val vertx: Vertx = Vertx.vertx(),
+  val responseObjectMapper: ObjectMapper = jacksonObjectMapper(),
+  responsesArtificialDelay: Duration? = null
+) {
+  val log: Logger = LogManager.getLogger(loggerName)
+  private var httpServer: HttpJsonRpcServer = createHttpServer(port)
+  val bindedPort: Int
```
Filter94: Bound

jvm-libs/linea/besu-rlp-and-mappers/src/main/kotlin/linea/rlp/BlockRlpEncoderInterfaces.kt
```diff
@@ -0,0 +1,30 @@
+package linea.rlp
+
+import tech.pegasys.teku.infrastructure.async.SafeFuture
+
+interface BlockRlpEncoder<T> {
```
Filter94: These `BlockRlp*` don't actually care what to encode/decode, so they aren't specific to blocks or RLP. They just encode/decode 1 thing and list of things

transaction-decoder-tool/src/test/kotlin/linea/test/MainRunner.kt
```diff
@@ -0,0 +1,61 @@
...
+    println("Unhandled exception: message=${error.message}")
+    LogManager.getLogger("vertx").error("Unhandled exception: message={}", error.message, error)
+  }
+  val fetcherAndValidate =
+    FetchAndValidationRunner(
+      rpcUrl = rplUrl,
+      vertx = vertx,
+      log = LogManager.getLogger("test.validator")
+    )
+  configureLoggers(
```
Filter94, line 32: Can't we do it with xml?

transaction-decoder-tool/src/test/kotlin/linea/test/MainRunner.kt
```diff
@@ -0,0 +1,61 @@
...
+
+fun configureLoggers(loggerConfigs: List<Pair<String, Level>>) {
+  loggerConfigs.forEach { (loggerName, level) ->
+    Configurator.setLevel(loggerName, level)
+  }
+}
+
+fun main() {
+  val rplUrl = run {
+    "https://linea-sepolia.infura.io/v3/${System.getenv("INFURA_PROJECT_ID")}"
```
Filter94, line 18: Maybe it's easier to put the whole endpoint into an env var and make it generic?

transaction-decoder-tool/src/test/kotlin/linea/test/BlocksFetcher.kt
```diff
@@ -0,0 +1,77 @@
...
+import java.util.concurrent.atomic.AtomicLong
+import kotlin.time.Duration.Companion.milliseconds
+
+class BlocksFetcher(
+  val web3j: Web3j,
+  val vertx: Vertx = Vertx.vertx(),
+  val pollingChuckSize: UInt = 100U,
+  val log: Logger = LogManager.getLogger(BlocksFetcher::class.java)
+) {
+  fun fetchBlocks(
```
Filter94, line 23: This could be private I believe

transaction-decoder-tool/src/test/kotlin/linea/test/BlocksFetcher.kt
```diff
@@ -0,0 +1,77 @@
...
+import net.consensys.linea.async.AsyncRetryer
+import net.consensys.linea.async.toSafeFuture
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.Logger
+import org.web3j.protocol.Web3j
+import tech.pegasys.teku.infrastructure.async.SafeFuture
+import java.util.concurrent.atomic.AtomicLong
+import kotlin.time.Duration.Companion.milliseconds
+
+class BlocksFetcher(
```
Filter94, line 17: IIUC purpose of this class is to fetch blocks in chunks and execute some user logic. So it's kind of `BlockHandlingRunner` or something, not only fetcher

transaction-decoder-tool/src/test/kotlin/linea/test/BlocksFetcher.kt
```diff
@@ -0,0 +1,77 @@
...
+              }
+          }
+      }
+      .let { SafeFuture.collectAll(it.stream()) }
+      .thenApply { blocks: List<Block?> ->
+        blocks.filterNotNull().sortedBy { it.number }
+      }
+  }
+
+  fun consumeBlocks(
```
Filter94, line 51: I'd name it `runBlockJob` or something around class'name, since this seems to be the main function of this class, not `fetchBlocks`

transaction-decoder-tool/src/test/kotlin/linea/test/BlocksFetcher.kt
```diff
@@ -0,0 +1,77 @@
...
+    startBlockNumber: ULong,
+    chunkSize: UInt = pollingChuckSize,
+    endBlockNumber: ULong? = null,
```
Filter94, line 53: I think it's better to move it above, after `startBlockNumber`

transaction-decoder-tool/src/test/kotlin/linea/test/BlockEncodingValidator.kt
```diff
@@ -0,0 +1,179 @@
...
+    val blocks = queueOfBlocksToValidate.pull(300)
+    if (blocks.isEmpty()) {
+      return SafeFuture.completedFuture(Unit)
+    }
+    log.info(
+      "compression validation blocks={} started",
+      CommonDomainFunctions.blockIntervalString(blocks.first().number, blocks.last().number)
+    )
+    val besuBlocks = blocks.map { it.toBesu() }
+    val originalBlockInterval = CommonDomainFunctions.blockIntervalString(blocks.first().number, blocks.last().number)
```
Filter94, line 88: nit: This could be moved closer to failure message and message itself can be made lazy

coordinator/app/src/test/kotlin/net/consensys/zkevm/coordinator/blockcreation/BlockCreationMonitorTest.kt
jpnovais: I thought about that, It would be less work for me. However, since it's already coded and I don't have 100% assurance it won't happen I left it and re-wrote the tests. I suggest when we have Beta V2 with new CL and instant finality.

jvm-libs/generic/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/JsonRpcMessageProcessor.kt
jpnovais: Without this change we cannot encode numbers e.g Int/ULong/etc without affecting the JSON-RPC envelope. 

The goal is to support custom objectMapper to the jsonrpc.**result**.

coordinator/app/src/test/kotlin/net/consensys/zkevm/coordinator/blockcreation/BlockCreationMonitorTest.kt
Filter94: What kind of assurance would be sufficient for you? It [was never reported](https://github.com/hyperledger/besu/issues?q=is%3Aissue%20%20clique%20reorg) for Besu, it's not supposed to happen by the protocol and we've never seen this happening with Besu. That doesn't mean that it will never happen, of course, but it's the same with many other assumptions in our code.

jvm-libs/generic/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/JsonRpcMessageProcessor.kt
Filter94: I see, so with this you use non-default object mapper with Kotlin support, is it correct?

transaction-decoder-tool/src/test/kotlin/linea/test/BlocksFetcher.kt
jpnovais: It was deliberately left public: it can be handy to just fetch some blocks and do/validate something ...

Filter94: But `consumeBlocks` allows exactly that

transaction-decoder-tool/src/test/kotlin/linea/test/BlocksFetcher.kt
jpnovais: I would rather keep the name, 90%+ of the code is logic to actually fetch the blocks, then the `consumer` is that actually runs that logic, which can be sync or deferred like in the validator for compression and decompression.

transaction-decoder-tool/src/test/kotlin/linea/test/MainRunner.kt
jpnovais: We can, but:
1. we are already full of XML files in the code base... (55 to be precise)
2. this is much easier and faster to fast change the log levels and iterate...
So I rather keep it this way

transaction-decoder-tool/src/test/kotlin/linea/test/MainRunner.kt
jpnovais: What would you can that env var?

Filter94: `RPC_URL` (typo in `rplUrl`, btw)

jvm-libs/linea/besu-rlp-and-mappers/src/main/kotlin/linea/rlp/BlockRlpEncoderInterfaces.kt
jpnovais: Created Generic interface on domain package

jvm-libs/generic/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/JsonRpcMessageProcessor.kt
jpnovais: With this you can override the default one, yes.

jpnovais: I mean, it is still backwards compatible

transaction-decoder-tool/src/test/kotlin/linea/test/BlockEncodingValidator.kt
jpnovais: I will keep it as it's used below as well (ti was meant to be before).

transaction-decoder-tool/src/test/kotlin/linea/test/MainRunner.kt
jpnovais: I remember now why I didn't do env var, because to read from ENV we need to restart IntellIJ, which only loads ENV once on start, very inconvenient to bounce from environments. Which I did quite often so I find this more convenient for prototyping. Will fix the typo though :)

Filter94 (approved):

coordinator/app/src/test/kotlin/net/consensys/zkevm/coordinator/blockcreation/BlockCreationMonitorTest.kt
jpnovais: I see it more this way:
- If it was "let's implement this safe guard now in case a reorg happens", I would not advocate for it because as you mentioned it seems pretty stable
- But now, effort was already spent, the code is super isolated in a single class and fairly easy to understand, and if it happens it can be very nasty to fix... Also, are changing our consensus mechanism which until it gets battled tested this can be useful IMO. So I feel more comfortable leaving for now it's hurting..

Filter94 (approved):

PR #412, opened 2024-12-10
[Feat] - GASLIMIT opcode test

kyzooghost:
This PR implements [#4035](https://github.com/Consensys/zkevm-monorepo/issues/4035)

- Add GasLimitTestContract to test using GASLIMIT opcode and store it in storage
- Add E2E test executing above contract

### Checklist


e2e/src/config/tests-config/environments/local.ts
```diff
@@ -33,7 +33,9 @@ const config: Config = {
     besuNodeRpcUrl: L2_BESU_NODE_RPC_URL,
     chainId: 1337,
     l2MessageServiceAddress: "0xe537D669CA013d86EBeF1D64e40fC74CADC91987",
+    // Nonce 10
```
jonesho, line 36: What does this comment mean?

kyzooghost: So in this file we have multiple hardcoded contract deployment addresses
- I was stuck for a little while on where these hardcoded contract deployment addresses came from.
- It turns out that the contract deployment address is a function of i.) The deployer address, and ii.) The deployer nonce
- These comments are referring to the nonce used to compute the hardcoded contract address

kyzooghost: See this code snippet for further reference on nonce ordering for these contracts - https://github.com/Consensys/linea-monorepo/blob/38ad5fc63b06fc7db1fcf8600e38b1da60620213/e2e/src/config/jest/global-setup.ts#L49-L52

contracts/contracts/test-contracts/GasLimitTestContract.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity 0.8.19;
+
+contract GasLimitTestContract {
+  uint256 internal _gasLimit;
+
+  function getGasLimit() external returns (uint256 gasLimit) {
+    return _gasLimit;
```
thedarkjester: you can assign `gasLimit = _gasLimit` vs. the return 

the current approach is mixing modes. Ideally just assign the return var.

thedarkjester (changes requested):
my only other comment is that we will probably expect to test other opcodes and this contract could be positioned/later extended as an "opcode" testing contract vs. just gasLimit - it might be prudent to name it that way so that anyone can use it for next steps

e2e/src/gas-limit.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+    const { maxPriorityFeePerGas, maxFeePerGas } = await l2Provider.getFeeData();
+
+    const tx = await gasLimitTestContract.connect(account).setGasLimit({
+      nonce: nonce,
+      maxPriorityFeePerGas: maxPriorityFeePerGas,
+      maxFeePerGas: maxFeePerGas,
+    });
+
+    const receipt = await tx.wait();
+    expect(receipt?.status).toEqual(1);
```
VGau, line 33: To properly test the expected behavior, you need to wait for a finalization on L1. WDYT ?

e2e/src/gas-limit.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+    const account = await l2AccountManager.generateAccount();
+    const lineaRollupV6 = config.getLineaRollupContract();
+
+    const txReceipt = await setGasLimit(account);
+    expect(txReceipt?.status).toEqual(1);
+    // Ok to type assertion here, because txReceipt won't be null if it passed above assertion.
+    const txBlockNumber = <number>txReceipt?.blockNumber;
+
+    console.log("Waiting for the first DataFinalizedV3 event for OpcodeTestContract.setGasLimit()...");
+    await waitForEvents(lineaRollupV6, lineaRollupV6.filters.DataFinalizedV3(txBlockNumber), 1_000);
```
jonesho: `DataFinalizedV3(startBlockNumber = txBlockNumber)` event might not be found if there were transactions from other test suites (i.e. the `startBlockNumber` from `DataFinalizedV3` could be lower than `txBlockNumber`)
We should poll the `lineaRollupV6.currentL2BlockNumber()` and check if the return number is higher or equal to `txBlockNumber`

e2e/src/gas-limit.spec.ts
kyzooghost: fixed

e2e/src/gas-limit.spec.ts
kyzooghost: good point, fixed

e2e/src/common/utils.ts
```diff
@@ -191,6 +197,54 @@ export async function waitForEvents<
...
+
+  while (!isExceedTimeOut) {
+    const returnValue = await method();
+    if (returnValue >= targetReturnValue) return true;
+    await wait(pollingInterval);
+  }
+
+  return false;
+}
+
```
jonesho, line 237: Can we generalize the above two functions into sth like the following:
```
// Currently only handle simple single return types - uint256 | bytesX | string | bool
export async function pollForContractMethodReturnValue<
  ExpectedReturnType extends bigint | string | boolean,
  R extends [ExpectedReturnType],
>(
  method: TypedContractMethod<[], R, "view">,
  expectedReturnValue: ExpectedReturnType,
  compareFunc: (a: ExpectedReturnType, b: ExpectedReturnType) => Boolean = (a,b) => {return a === b},
  pollingInterval: number = 500,
  timeout: number = 2 * 60 * 1000,
): Promise<boolean> {
  let isExceedTimeOut = false;
  setTimeout(() => {
    isExceedTimeOut = true;
  }, timeout);

  while (!isExceedTimeOut) {
    const returnValue = await method();
    if (compareFunc(returnValue, expectedReturnValue)) return true;
    await wait(pollingInterval);
  }

  return false;
}
```
such that we could inject the comparison function (default to equal) and hence can be used like the following:
```
const isBlockFinalized = await pollForContractMethodReturnValue(
      lineaRollupV6.currentL2BlockNumber,
      BigInt(blockNumberToCheckFinalization),
      (returnVal, expectedVal) => { return returnVal >= expectedVal}
    );
```

kyzooghost: good point, updated

thedarkjester (approved):

PR #437, opened 2024-12-11
Using linea besu package

Filter94:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/docker/compose-local-dev-traces-v2.overrides.yml
+++ b/docker/compose-local-dev-traces-v2.overrides.yml
@@ -3,15 +3,14 @@
 
 services:
   sequencer:
+    image: consensys/linea-besu-package:devnet-be4f01c
     volumes:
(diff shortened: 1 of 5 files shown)


docker/compose-local-dev-traces-v2.overrides.yml
```diff
@@ -22,7 +21,7 @@ services:
   traces-node-v2:
     hostname: traces-node-v2
     container_name: traces-node-v2
-    image: consensys/linea-besu:24.12-delivery40
+    image: consensys/linea-besu-package:devnet-0d2fbde
     profiles: [ "l2", "l2-bc", "debug", "external-to-monorepo" ]
     depends_on:
       traces-node-v2-plugin-downloader:
```
jonesho: This line above and the download service itself can be removed, right?

Filter94: Yes, I forgot removing it here. Will do, thanks!

jonesho (approved):

PR #441, opened 2024-12-12
feature:440 - update prover api specs

srinathln7:
This PR implements issue(s) #

Updated the Prover API specs in the Linea architecture docs.


docs/architecture-description.md
```diff
@@ -577,52 +580,41 @@ Compression proof request file format:
 
 ```
 BlobCompressionProofJsonRequest
```
FlorianHuc, line 582: what is in BlobJsonFileRequestResponse.kt is different from what is updated here

FlorianHuc: ok, the order is good as per the prover/backend/execution/response.go file. maybe we can add an hyperlink to the classes, wdyt?

srinathln7: yes, done!

FlorianHuc (approved):

PR #459, opened 2024-12-17
Finalized Updater - log the block number when it is skipped

macfarla:
Signed-off-by: Sally MacFarlane <macfarla.github@gmail.com>This PR implements issue(s) #

### Checklist


jpnovais (approved):

finalized-tag-updater/src/main/kotlin/net/consensys/linea/LineaL1FinalizationUpdaterService.kt
```diff
@@ -46,7 +46,7 @@ class LineaBesuEngineBlockTagUpdater(private val blockchainService: BlockchainSe
         throw e
       }
     } else {
-      log.warn("Linea safe/finalized block update: Skipped as getBlockByNumber returns empty result")
+      log.warn("Linea safe/finalized block update: Skipped as getBlockByNumber for block=$finalizedBlockNumber returns empty result")
```
jpnovais:
```suggestion
      log.warn("Linea safe/finalized block update: Skipped as getBlockByNumber for block={} returns empty result", finalizedBlockNumber)
```

we tend to favour argument placeholders over string interpolation as they tend to be more efficient. Although not that releval in this particular case.

PR #467, opened 2024-12-18
Jacoco integration test fix

Filter94:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -58,15 +58,6 @@ allprojects {
     toolVersion = '0.8.11'
   }
 
-  gradle.projectsEvaluated {
-    project.tasks.jacocoTestReport {
-      if (project.tasks.findByName('integrationTest')) {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -56,16 +56,6 @@ allprojects {
...
-    if (project.tasks.findByName('integrationTest')) {
-      applyTo integrationTest
-    }
-    if (project.tasks.findByName('acceptanceTest')) {
-      applyTo acceptanceTest
-    }
-  }
-
-  jacocoTestReport {
-    dependsOn test
```
Filter94, line 68: We should either set this dependency for both `integrationTest` and `test` or remove it at all. I decided to remove it, because in our GHA we're executing tests + generating report as 2 separate steps

Anon-im:
Thats cool

PR #475, opened 2024-12-20
coordinator: update besu dependency

jpnovais:
This PR implements issue(s) #

### Checklist


Anon-im:
Thats cool

coordinator/ethereum/test-utils/src/main/kotlin/net/consensys/zkevm/ethereum/AccountManager.kt
```diff
@@ -231,3 +231,32 @@ object L2AccountManager : AccountManager by WhaleBasedAccountManager(
   genesisFile = getPathTo(System.getProperty("L2_GENESIS", "docker/config/linea-local-dev-genesis.json")),
   log = LogManager.getLogger(L2AccountManager::class.java)
 )
+
+fun <R, T : Response<R>> retry(
```
Filter94, line 235: Can't we use async retryer here instead?

.github/workflows/coordinator-build-and-publish.yml
```diff
@@ -78,7 +78,7 @@ jobs:
         uses: gradle/actions/setup-gradle@cc4fc85e6b35bafd578d5ffbc76a5518407e1af0 #v4.2.1
```
Filter94, line 78: Not a focus of this PR, but I think it was released as a part of v4 release

Filter94 (approved):

coordinator/ethereum/test-utils/src/main/kotlin/net/consensys/zkevm/ethereum/AccountManager.kt
jpnovais: I considered that but would imply a even bigger refactor (pass vertx, etc..) and the code is sync. So the cost/benefit did not pay off.

.github/workflows/coordinator-build-and-publish.yml
jpnovais: I don't get your point/suggestion... can you elaborate please?

Filter94: To leverage changes in this commit we can just use a regular import like 'gradle/actions/setup-gradle@v4' instead of this cryptic one

jpnovais: That that was the approach we had before. It was suggested by ops team because the V4 is an alias to the latest v4 and had regressions. With static approach it we have confidence it won't have regressions

PR #490, opened 2025-01-06
feat: run gaslimit e2e test only and increased block gas limit to 2b

jonesho:
This PR implements issue(s) [#4035](https://github.com/Consensys/zkevm-monorepo/issues/4035)

### Checklist


e2e/jest.config.ts
```diff
@@ -4,7 +4,7 @@ const config: Config = {
   preset: "ts-jest",
   testEnvironment: "node",
   rootDir: "src",
-  testRegex: ".spec.ts$",
```
VGau, line 7: Could you remove this change please because all tests files need to be executed ? :)

jonesho: @VGau this PR is just for running the gaslimit test on a devnet machine, which is not targetted to be merged into main, that's why I marked it as draft

VGau: Okay got it :) Thanks for the explanation 😄

Issue #491, opened 2025-01-06
Finalization tag updater plugin should follow the pattern of System.exit if an exception is caught in register


jonesho:
## Description
Finalization tag updater plugin should follow the [pattern](https://github.com/Consensys/linea-sequencer/blob/be18e11a180a82137001483d91dcc5d6476dfaee/sequencer/src/main/java/net/consensys/linea/AbstractLineaRequiredPlugin.java#L41) of System.exit if an exception is caught in register function.

## Motivation
This section should indicate what value we are receiving from it.

## Tasks
- [ ] Medium-level task that needs to be implemented for this feature request to be considered complete.

## Acceptance criteria
- [ ] List tests that need to be implemented or checks to consider this task completed. 

## Risks
- [ ] These are risks that may prevent completion of the task as well as areas to pay special attention to

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #494, opened 2025-01-06
feat: add try catch exit pattern in register function of finalized ta…

jonesho:
…g updater

This PR implements issue(s) #491 

### Checklist

### DIFF
--- a/finalized-tag-updater/src/main/java/net/consensys/linea/consensus/L1FinalizationTagUpdaterPlugin.java
+++ b/finalized-tag-updater/src/main/java/net/consensys/linea/consensus/L1FinalizationTagUpdaterPlugin.java
@@ -14,7 +14,7 @@ import org.hyperledger.besu.plugin.services.PicoCLIOptions;
 
 @AutoService(BesuPlugin.class)
 public class L1FinalizationTagUpdaterPlugin implements BesuPlugin {
-  private final Logger log = LogManager.getLogger(this.getClass().getName());
+  private final Logger log = LogManager.getLogger(L1FinalizationTagUpdaterPlugin.class);
 	private static final String CLI_OPTIONS_PREFIX = "linea";
 	private PluginCliOptions cliOptions = new PluginCliOptions();
 	private Vertx vertx = Vertx.vertx();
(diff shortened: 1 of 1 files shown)


finalized-tag-updater/src/main/java/net/consensys/linea/consensus/L1FinalizationTagUpdaterPlugin.java
jpnovais, line 1: What is the motivation to do this PR now?

It seems very dangerous and antipattern for a plugin to do a `System.exit()` call killing the whole node. Why not throwing the exception and let Besu plugin system handle it and close the node?

jonesho: This pattern is suggested by Sarah from Besu team in a Slack [thread](https://consensys.slack.com/archives/C07QZJL82TD/p1734378875026499?thread_ts=1734376690.713699&cid=C07QZJL82TD):
```
The L1FinalizationTagUpdaterPlugin doesn't follow the [pattern](https://github.com/Consensys/linea-sequencer/blob/be18e11a180a82137001483d91dcc5d6476dfaee/sequencer/src/main/java/net/consensys/linea/AbstractLineaRequiredPlugin.java#L41) of System.exit if an exception is caught in register.
I think you would still see in the logs if it was the method signature breaking change issue.
However if there's a missing config required, that could be a silent error.
```
From Sarah, it's the common [pattern](https://github.com/Consensys/linea-sequencer/blob/be18e11a180a82137001483d91dcc5d6476dfaee/sequencer/src/main/java/net/consensys/linea/AbstractLineaRequiredPlugin.java#L41) for plugins to do the "try catch systemexit" in register function to ensure Besu would exit in case of plugin register errors to avoid them being silently ignored, I see that most plugins are doing the same thing, and it'd be good to align our plugin behavior to be the same as the rest.

> It seems very dangerous and antipattern for a plugin to do a System.exit() call killing the whole node. Why not throwing the exception and let Besu plugin system handle it and close the node?

@macfarla This point is also valid, maybe could you elaborate more on what's the impact of not doing "try catch systemexit" in register function if errors are thrown inside the function? Btw for local testing, I'd tried to manually throw exception inside the register function without the "try catch systemexit" and I could see besu was still able to be exited and printed the error log, so maybe this suggests it is not necessary for finalized-tag-updater to add the "try catch systemexit"?

jonesho: Given that since [24.10-delivery36](https://github.com/Consensys/linea-besu/releases/tag/24.10-delivery36) Besu will halt if any plugins fail during registration: https://github.com/hyperledger/besu/pull/7662, this PR might not be necessary anymore

jonesho:
Closed this issue given that since [24.10-delivery36](https://github.com/Consensys/linea-besu/releases/tag/24.10-delivery36) Besu will halt if any plugins fail during registration: https://github.com/hyperledger/besu/pull/7662, this PR might not be necessary anymore

Issue #474, opened 2024-12-20
Prover/Transition to Smaller Fields—Smartvectors


bogdanbear:
For the transition to smaller fields, we need to modify the smartvectors interface/its implementations to support both base fields and field extensions.

PR #499, opened 2025-01-07
Prover/feat/Adding field extension support to Smartvectors 

bogdanbear:
This PR completes the work concerning smartvectors and associated packages [Issue 474](https://github.com/Consensys/linea-monorepo/issues/474), concerning the prover transition to a mixture of base fields and field extensions. 

### Checklist

* [x] Updated the field extension mock.
* [x] Completed the smartvector functionality for the field extension case. 
* [x] Unit testing.


AlexandreBelling (approved):
There is really a lot so I really just skimmed through it, but that seems fine

PR #508, opened 2025-01-08
[Chore] - Add beta v1 proof verification test

thedarkjester:
This PR implements issue(s) #

### Checklist


alainncls (approved):

contracts/test/LineaRollup.ts
```diff
@@ -1537,6 +1552,132 @@ describe("Linea Rollup contract", () => {
...
+            finalizationData.l1RollingHashMessageNumber,
+            finalizationData.l1RollingHash,
+            finalizationData.finalTimestamp,
+          ],
+        ),
+      );
+    });
+
+    // Function to extract range from the file name
+    function extractBlockRangeFromFileName(fileName: string): [number, number] | null {
```
kyzooghost, line 1658: Question for thought - Do you think that adding more function definitions to this 2000+ line test file increases our refactoring/maintenance burden down the line?

kyzooghost: Also in reading a test case for this file currently, reading a utility function implementation involves the following dev flow:
1. Click on utility function reference to jump to implementation
2. Manually scroll back to the test case

I'm thinking that if the utility functions were in another file, step 2 would be changing VSCode tabs - which can be done with a keyboard shortcut

thedarkjester: I don't think it will increase the burden atm - this will be moved to somewhere yet to be decided - if it was in a file now, it may get moved later still.

PR #524, opened 2025-01-10
fix(ecpair): ensure last non-trivial input pair goes to ML+finalexp circuit

ivokub:
This PR fixes ECPAIR glue in case when the last input pair is not non-trivial (either (0,0) or (0, Q) or (P, 0)). Previously, two things happened when the last input was half-trivial (only G2 membership check i.e. (0, Q)) - it was sent for the G2 membership circuit for testing (as requested by the arithmetization) and we didn't mark the last non-trivial input to go to the ML+finalexp circuit (as it was decided by the input pair ID relative to the total number of pairs).

To overcome that, we now in the glue index the input pairs differently. Instead of following the input pair IDs coming from the arithmetization, we count it ourselves only for the inputs which go to the pairing circuits. And for the "total pairs" we count only non-trivial pairs (as the ML of trivial pairs is 1 and doesn't change the accumulator. What only matters is that they are valid i.e. belonging to the correct subgroups).

To make it work, we had to add a constraint to ensure the total number of pairs was correctly set. But the constraints to ensure sequential counting of the non-trivial pairs still follows from the previous constraints (we check that after every 60 rows the input ID increases).

Additionally, we had to ensure the accumulator consistency correctness. When we have only one valid pair, then we only call ML+finalexp circuit. In this case we cannot check the consistency of the accumulator, but instead have to check that the previous accumulator going to the ML+finalexp circuit is correctly initialized.

Another fix I had to implement is to allow to have pairing circuit inputs and G2 membership inputs in the same rows (see https://github.com/Consensys/linea-monorepo/commit/60ee51c0b21fb7e82def347e2c7432938573f112). Previously when a row was an input to the G2 membership circuit, then the index constraint was not enforce, but should have been.

Finally - added regression test for a transaction which incurred the bug.

Also did some cosmetic changes:
* generated testdata modules for some test cases
* unified the testdata modules to have same order of columns
* added option `InHex` to `csvTraces.FmtCSV` method to force dumping also small outputs in hex. This is mainly to have minimal diff of the current testcase modules.
* defined a utility method for dumping the testdata module. It is useful for creating more regression tests.

On top of that, I added a comprehensive test case generator which generates all possible 1-5 input pair combinations for the pairing check. Currently the tests are skipped as it generates about 65000 test cases, but can be manually called.

### Checklist


ivokub:
@AlexandreBelling - PR is now ready, I have run the test suite locally on my machine and passed cleanly.

gusiri:
@AlexandreBelling Enabled ecpair in zkevm.go

ivokub:
One more thing - we merged the optimization to gnark to assert G2 membership already inside the Miller loop computation. It saves (depending on the number of ML instances per circuit) 13-15% constraints. It requires updating gnark dependency and I haven't run all the test suite with it.

Should I do it in a separate PR (and waiting 24h for the test suite to run) or add the gnark dependency update here?

AlexandreBelling:
If it's only a performance thing we can wait a few days. The point for us here is to add back the pairings in prod quickly to validate beta-v1.

ivokub:
> If it's only a performance thing we can wait a few days. The point for us here is to add back the pairings in prod quickly to validate beta-v1.

Alright - then it is good from my side.

AlexandreBelling (approved):

Issue #482, opened 2024-12-24
Refactor: Remove reverts from compressor


Tabaie:
## Description
In accordance with @ivokub's feedback on https://github.com/Consensys/linea-monorepo/pull/367, change the compressor's flow into one that writes on a "scratch" and in case of success copies it into the current state. This would replace the current strategy of optimistically attempting to compress and reverting if something went wrong.

## Motivation
The fact that there are more ways of going wrong than going right (there are many control paths leading to a revert) has lead to complex code involving many goto-labels. The pessimistic strategy will result in smaller, more readable code.

@gbotrel thoughts?

## Tasks
- [ ] Medium-level task that needs to be implemented for this feature request to be considered complete.

## Acceptance criteria
- [ ] List tests that need to be implemented or checks to consider this task completed. 

## Risks
- [ ] These are risks that may prevent completion of the task as well as areas to pay special attention to

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #532, opened 2025-01-11
refactor: remove reverts from compressor using draft pool approach

crStiv:
This PR implements #482 by replacing the current revert-based approach with a draft pool strategy.

Key improvements:
- Removes complex goto-based logic with reverts
- Uses a pool of draft compressors for better isolation
- Thread-safe implementation with ConcurrentLinkedQueue
- Improved testability and maintainability

Implementation details:
- Each compression attempt uses a separate compressor from the pool
- Main state is only updated after successful compression
- Failed attempts don't affect the main state
- Efficient reuse of compressors through pooling

Based on feedback from @ivokub in #367 regarding state isolation and avoiding reverts.

Tests added:
- Basic compression scenarios
- Concurrent compression using pool
- Pool reuse and efficiency
- State isolation verification
- Error handling

All tests passing locally.


ivokub:
Thanks for the PR. Unfortunately at first look it doesn't work. In our approach we have the compression library written in Go where we expose some functions using CGO and then call them in Java using JNI.

In your current approach you pool the compression instances, but we want to pool the compression instance with a state. This imo can efficiently done on the Go side.

PR #546, opened 2025-01-14
chores(ci): deactivate the slack notification for prover-ci

AlexandreBelling:
This PR deactivate the slack notification so that we can onboard external PRs cleanly

### Checklist


julien-marchand (dismissed):

.github/workflows/prover-testing.yml
```diff
@@ -1,13 +1,5 @@
 name: Prover testing CI
```
kyzooghost, line 2:
```suggestion

on: workflow_call
```

Got "_No event triggers defined in `on`_" error otherwise :( See https://github.com/Consensys/linea-monorepo/actions/runs/12782331876

PR #547, opened 2025-01-14
Prover/vertical splitting for limitless prover

Soleimani193:
This PR implements the vertical splitting of CompiledIOP and its corresponding prover.

### Checklist


prover/protocol/wizard/builder.go
```diff
@@ -52,21 +52,27 @@ Compile an IOP from a protocol definition
 func Compile(define DefineFunc, compilers ...func(*CompiledIOP)) *CompiledIOP {
 	builder := newBuilder()
 	define(&builder)
+	comp := builder.CompiledIOP
+	return ContinueCompilation(comp, compilers...)
+}
+
+// ContinueCompilation continues a set of compilation steps over a initial CompiledIOP object.
```
AlexandreBelling, line 59: It's a good idea to add that IMO

AlexandreBelling (approved):

prover/protocol/distributed/comp_splitting.go
```diff
@@ -0,0 +1,92 @@
...
+	numSegments int
+}
+
+// It implements [wizard.ProverAction] for the module prover.
+func (p segmentModuleProver) Run(run *wizard.ProverRuntime) {
+
+	if run.ParentRuntime == nil {
+		utils.Panic("invalid call: the runtime does not have a [ParentRuntime]")
+	}
+	if run.ProverID > p.numSegments {
```
AlexandreBelling, line 76: You should make ProverID a field of `segmentModuleProver` and not `ProverRuntime`

Soleimani193: ModulCop for all segments should remain the same and proverID should only affect the prover. 

On the other hand the run time already has the parent run time and is natural to say what is the position of the prover among its siblings.

PR #550, opened 2025-01-14
CHANGED - pin actions, use updated self-hosted runners

waskow-consensys:
When using Third Party Actions, pinning allows for not only consistent CI Runs, as no underlying changes with the action are promoted unknowingly. 
https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

Additionally, use new config for updated self-hosted runners.


waskow-consensys:
FYI @kyzooghost

PR #553, opened 2025-01-14
Feat/multi dict prover

Tabaie:
This PR enables the prover to support multiple compression dictionaries at once.
Note that blob submission would still operate with a single dictionary and is thus not affected.

**BREAKING CHANGE:**
As shown in prover/config/config-sepolia-full.toml, the name of the property `dict_path` under `[blob_decompression]` has changed to `dict_paths` and its value is an array of strings, instead of a single string.


prover/lib/compressor/blob/v1/blob_maker.go
```diff
@@ -302,31 +302,30 @@ func (bm *BlobMaker) Equals(other *BlobMaker) bool {
 }
 
 // DecompressBlob decompresses a blob and returns the header and the blocks as they were compressed.
```
gbotrel, line 304: I think this could use an updated comment on what this function do in a bit more details;

i.e. `DecompressBlob` will parse the header, get a dict checksum, attempt to find the corresponding dict from the store, etc... landed here when trying to follow through the usage of this method with many returns from somewhere else it was not clear :)

prover/cmd/prover/cmd/setup.go
```diff
@@ -32,7 +33,8 @@ import (
 type SetupArgs struct {
 	Force      bool
 	Circuits   string
-	DictPath   string
+	DictPath   string // to be deprecated; only used for compiling v0 blob decompression circuit
```
gbotrel, line 36: call it `DictPathV0Deprecated` ?

gbotrel (approved):
lgtm

prover/lib/compressor/blob/v1/blob_maker.go
Tabaie: Would it be better to bundle all the non-error return values into a response struct?

prover/cmd/prover/cmd/setup.go
Tabaie: The thing is, it is technically not deprecated since it still has a legitimate use-case that the new alternative doesn't cover (v0 circuits).

prover/lib/compressor/dict/25-01-15-INCOMPLETE.bin
Tabaie, line 1: This file is 64KB - 60 B long and cannot be used as is. After three 20-byte addresses are appended to it, we can rename it to `25-01-15.bin`

gbotrel (approved):

julien-marchand (approved):

PR #571, opened 2025-01-20
3128 export metrics for conflation blob and aggregation size

jonesho:
This PR implements issue(s) #[3128](https://github.com/Consensys/zkevm-monorepo/issues/3128)

### Checklist

### DIFF
--- a/coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/aggregation/GlobalAggregationCalculator.kt
+++ b/coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/aggregation/GlobalAggregationCalculator.kt
@@ -8,7 +8,6 @@ import net.consensys.zkevm.domain.BlobsToAggregate
 import org.apache.logging.log4j.LogManager
 import org.apache.logging.log4j.Logger
 import java.util.PriorityQueue
-import java.util.concurrent.atomic.AtomicInteger
 
 class GlobalAggregationCalculator(
   private var lastBlockNumber: ULong,
@@ -24,9 +23,6 @@ class GlobalAggregationCalculator(
     b1.startBlockNumber.compareTo(b2.startBlockNumber)
   }
 
-  private val aggregationSizeInBlocks = AtomicInteger(0)
-  private val aggregationSizeInBatches = AtomicInteger(0)
-  private val aggregationSizeInBlobs = AtomicInteger(0)
   private val blobsCounter: Counter = metricsFacade.createCounter(
     category = LineaMetricsCategory.AGGREGATION,
     name = "calculator.blobs.accepted",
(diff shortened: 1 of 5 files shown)


Filter94:
@jonesho I see that you addressed part of metrics, mentioned in the desscription of the ticket, but do you plan to address stuff [mentioned](https://github.com/Consensys/zkevm-monorepo/issues/3128#issuecomment-2093218976) by Alex in the comments?

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/aggregation/GlobalAggregationCalculator.kt
```diff
@@ -7,8 +7,8 @@ import net.consensys.zkevm.domain.BlobCounters
 import net.consensys.zkevm.domain.BlobsToAggregate
 import org.apache.logging.log4j.LogManager
 import org.apache.logging.log4j.Logger
-import java.lang.IllegalStateException
 import java.util.PriorityQueue
+import java.util.concurrent.atomic.AtomicInteger
 
 class GlobalAggregationCalculator(
```
Filter94, line 12: Would it be easier to capture these metrics in `ProofAggregationCoordinatorService::onAggregation` instead?

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/aggregation/GlobalAggregationCalculator.kt
```diff
@@ -24,6 +24,9 @@ class GlobalAggregationCalculator(
     b1.startBlockNumber.compareTo(b2.startBlockNumber)
   }
 
+  private val aggregationSizeInBlocks = AtomicInteger(0)
```
Filter94: I think gauge doesn't fit well what we're trying to capture here. 

I think histogram is more suitable, because:
* We won't lose any data in case multiple aggregations are processed in a short period of time (less than scrape interval)
* It will give us much more information on the distribution of block / blob / block sizes per aggregation

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/aggregation/GlobalAggregationCalculator.kt
jonesho: Yes, indeed, given that aggregation interval in `onAggregation` is final.

jonesho:
> @jonesho I see that you addressed part of metrics, mentioned in the description of the ticket, but do you plan to address stuff [mentioned](https://github.com/Consensys/zkevm-monorepo/issues/3128#issuecomment-2093218976) by Alex in the comments?

I'm asking Alex to clarify if all of the suggested metrics are still valid, if yes, we might create another issue to track

Filter94 (approved):

PR #576, opened 2025-01-20
feat: increase max pool size to 100 in http client option in unit test

jonesho:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/jvm-libs/generic/json-rpc/src/test/kotlin/net/consensys/linea/jsonrpc/client/VertxHttpJsonRpcClientTest.kt
+++ b/jvm-libs/generic/json-rpc/src/test/kotlin/net/consensys/linea/jsonrpc/client/VertxHttpJsonRpcClientTest.kt
@@ -56,7 +56,7 @@ class VertxHttpJsonRpcClientTest {
     HttpClientOptions()
       .setKeepAlive(true)
       .setProtocolVersion(HttpVersion.HTTP_1_1)
-      .setMaxPoolSize(10)
+      .setMaxPoolSize(100)
   private lateinit var endpoint: URL
 
   @BeforeEach


jvm-libs/generic/json-rpc/src/test/kotlin/net/consensys/linea/jsonrpc/client/VertxHttpJsonRpcClientTest.kt
```diff
@@ -56,7 +56,7 @@ class VertxHttpJsonRpcClientTest {
     HttpClientOptions()
       .setKeepAlive(true)
       .setProtocolVersion(HttpVersion.HTTP_1_1)
-      .setMaxPoolSize(10)
+      .setMaxPoolSize(100)
```
jpnovais, line 59: What is the motivation for this change?

jonesho: My speculation is that the test is running parallel with many test cases and each of them is sharing the same http client, and the cause of getting this error:
```
VertxHttpJsonRpcClientTest > makesRequest_measuresRequest() FAILED
    org.opentest4j.AssertionFailedError: 
    expected: 3L
     but was: 2L
``` 
is due to this max pool size was not large enough (i.e. 10) for all the connections made in the parallel test cases and hence one of the requests in the test would sometimes get dropped which resulted in 2 instead of 3 for the metric counter. I couldn't reproduce the failure locally and by changing this the test didn't fail anymore when running multiple times in a row.

jonesho: Nvm, the above assumption was wrong, the http client is actually not shared with all test cases, each test cases is having its own instance of http client. https://github.com/Consensys/linea-monorepo/pull/577 is a more proper fix

PR #584, opened 2025-01-21
Feat/revise GitHub docker cache images

jonesho:
This PR implements issue(s) #589

- It fixes and refactors the uses of `actions/cache` for caching pulled docker images, previously the reload of cached docker images was missing and save cache jobs would run twice (both in `coordinator-testing` and `staterecovery-testing` )
- Due to performance issue, temporarily disables the caching for pulled docker images in our pipeline (see [comment](https://github.com/Consensys/linea-monorepo/pull/584#issuecomment-2609177799))

### Checklist


jonesho:
The idea of this PR is to cahce the pulled docker images (those defined compose files excluding images that would be built by pipeline such as coordinator, prover, transaction-exclusion-api, etc..) and retrieve the cache and load the docker images during integration and e2e tests, however, it turns out this doesn't lead to any performance gain and instead makes the tests run even longer due to the time spent for retrieving the cache and then `docker load` them back to docker images: it usually takes around 1m30s for "Restore cached images" (see [here](https://github.com/Consensys/linea-monorepo/actions/runs/12923908886/job/36042059078) for example) whereas "Pull all images with retry" without cache (see [here](https://github.com/Consensys/linea-monorepo/actions/runs/12915351744/job/36017628961)) usually takes just under 30 sec. This means the caching would instead introduce 1m more run time to the e2e tests, therefore for now, I would suggest to remove the use of GHA `actions/cache` for pulled docker images in this PR, and only enable when we could justify the use of them for pulled docker images.

kyzooghost (commented):
CI execution is looking good. Have one overall suggestion:

Should we delete the commented out lines that refer to `restore-docker-images` or `cache-docker-images`? Perhaps we could replace them with a one-line comment e.g. "Empirically found that caching Docker images actually increased test time-to-completion"

I'm thinking that in a few months when we have lost context, we may be looking at the commented out lines and asking 'why did we comment that out?' or 'why did we disable the cache?'

VGau (approved):

PR #585, opened 2025-01-21
Prover/Convert projection protocol to a query

arijitdutta67:
This PR converts the projection protocol to a query of the compiled IOP. This is a prerequisite for horizontal and vertical splitting for the limitless prover.

### Checklist


prover/backend/execution/prove.go
AlexandreBelling, line 1: Please don't add a nolint on top of a file. That's a security risk. Add it alongside the line that you want to ignore.

jvm-libs/blob-compressor/src/test/resources/net/consensys/linea/nativecompressor/rlp_blocks.bin
AlexandreBelling, line 1: Why do we commit a change for this file?

prover/protocol/dedicated/plonk/alignment.go
```diff
@@ -273,7 +273,7 @@ func (a *Alignment) csIsActive(comp *wizard.CompiledIOP) {
 // csProjection ensures the data in the [Alignment.Data] column is the same as
 // the data provided by the [Alignment.CircuitInput].
 func (a *Alignment) csProjection(comp *wizard.CompiledIOP) {
-	projection.InsertProjection(comp, ifaces.QueryIDf("%v_PROJECTION", a.Name), []ifaces.Column{a.DataToCircuit}, []ifaces.Column{a.CircuitInput}, a.DataToCircuitMask, a.ActualCircuitInputMask)
+	projection.RegisterProjection(comp, ifaces.QueryIDf("%v_PROJECTION", a.Name), []ifaces.Column{a.DataToCircuit}, []ifaces.Column{a.CircuitInput}, a.DataToCircuitMask, a.ActualCircuitInputMask)
```
AlexandreBelling: Shouldn't that be a method of 'comp' now?

prover/protocol/dedicated/projection/projection_query.go
```diff
@@ -0,0 +1,52 @@
...
+*/
+
+import (
+	"github.com/consensys/linea-monorepo/prover/protocol/ifaces"
+	"github.com/consensys/linea-monorepo/prover/protocol/query"
+	"github.com/consensys/linea-monorepo/prover/protocol/wizard"
+	"github.com/consensys/linea-monorepo/prover/protocol/wizardutils"
+)
+
+func RegisterProjection(
```
AlexandreBelling: Iiuc, this function does not do really anything on top of comp.InsertProverAction.

It would simplify the code if we removed it.

jvm-libs/blob-compressor/src/test/resources/net/consensys/linea/nativecompressor/rlp_blocks.bin
arijitdutta67: Thanks for the catch. removed.

prover/protocol/dedicated/plonk/alignment.go
arijitdutta67: yes. Resolved now.

prover/protocol/dedicated/projection/projection_query.go
arijitdutta67: Thanks for the suggestion. Addressed.

prover/maths/common/poly/poly.go
```diff
@@ -122,3 +122,31 @@ func EvaluateLagrangesAnyDomain(domain []field.Element, x field.Element) []field
 
 	return lagrange
 }
+
+// CmptHorner computes a random Horner accumulation of the filtered elements
+// starting from the last entry down to the first entry. The final value is
+// stored in the last entry of the returned slice.
+// Todo: send it to a common utility package
```
AlexandreBelling, line 129: remove this todo as I think you did what it says

prover/maths/common/poly/poly.go
```diff
@@ -122,3 +122,31 @@ func EvaluateLagrangesAnyDomain(domain []field.Element, x field.Element) []field
...
+func CmptHorner(c, fC []field.Element, x field.Element) []field.Element {
```
AlexandreBelling, line 130: How about `GetHornerTrace` for the name?

prover/protocol/compiler/projection/prover.go
```diff
@@ -0,0 +1,129 @@
...
+				row[c] = fmt.Sprintf("%v=%v", pa.ColB[c].GetColID(), fString)
+			}
+
+			rowsB = append(rowsB, row)
+			cntB++
+		}
+
+		larger := max(len(rowsA), len(rowsB))
+
+		for i := 0; i < larger; i++ {
```
AlexandreBelling, line 109: That seems like a debugging printf you want to remove I guess. Don't hesitate to use the search tool of VS code and hunt for calls to `fmt.Printf` it helps to avoid adding them.

arijitdutta67: To me it seems like a useful debugging tool (not added by me though, already there in the projection protocol). Also there are many `fmt.Printf` in the codebase. For now I am keeping them.

Issue #583, opened 2025-01-21
Github workflow runners - Remove warnings from GitHub UI for GitHub Actions


kyzooghost:
## Description
Where possible, remove warnings from GitHub Action runs in the GitHub UI
- e.g. `Unexpected input(s) 'predicate-quantifier', valid inputs are ['token', 'working-directory', 'ref', 'base', 'filters', 'list-files', 'initial-fetch-depth']`

## Motivation
- Remove tech debt

## Tasks
- [ ] Where possible, remove warnings from GitHub Action runs in the GitHub UI

## Acceptance criteria
- [ ] PR merged

## Risks
- [ ] May break existing CI workflows

## Remember to
 - [x] Add the `documentation` label in case there is an impact on the documentation
 - [x] Add `priority` and `team` labels
 - [x] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

Issue #589, opened 2025-01-22
Revise the use of GHA cache for pulling docker images in pipeline


jonesho:
## Description
The github actions job `pull-and-cache-images` doesn't work as intented:
- runs twice if compose files are modified
- only cache the docker images but doesn't seems to restore them correctly in other workflow, i.e. just restore the tar file but not doing `docker load` command to restore the docker images

Need to revise the github actions job to make sure it caches the docker images once if any images in the compose files are updated and properly restore the cache and load them as docker images in other workflows.
 
## Motivation
This section should indicate what value we are receiving from it.

## Tasks
- [ ] Medium-level task that needs to be implemented for this feature request to be considered complete.

## Acceptance criteria
- [ ] List tests that need to be implemented or checks to consider this task completed. 

## Risks
- [ ] These are risks that may prevent completion of the task as well as areas to pay special attention to

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #590, opened 2025-01-22
[Fix] CI warnings

kyzooghost:
This PR implements issue(s) #583 

- Updated `actions/cache` references to 4.2.0
- Use in-built caching for `setup-go`
- 'checkout' before 'setup-go' (https://github.com/actions/setup-go/issues/427#issuecomment-2273249463)

### Checklist

### DIFF
--- a/.github/workflows/prover-testing.yml
+++ b/.github/workflows/prover-testing.yml
@@ -21,6 +21,8 @@ jobs:
       uses: actions/setup-go@v5
       with:
         go-version: 1.23.x
+        cache-dependency-path: |
+              prover/go.sum
     - uses: actions/cache@v4.2.0
       with:
         path: |
@@ -63,6 +65,8 @@ jobs:
       uses: actions/setup-go@v5
       with:
         go-version: ${{ matrix.go-version }}
+        cache-dependency-path: |
+              prover/go.sum
     - uses: actions/cache@v4.2.0
       with:
         path: |
--- a/.github/workflows/prover-testing.yml
+++ b/.github/workflows/prover-testing.yml
@@ -21,7 +21,7 @@ jobs:
       uses: actions/checkout@v4
       with:
         fetch-depth: 0
-    - uses: actions/cache@v4
+    - uses: actions/cache@v4.2.0
       with:
         path: |
           ~/go/pkg/mod
@@ -63,7 +63,7 @@ jobs:
         go-version: ${{ matrix.go-version }}
     - name: checkout code
       uses: actions/checkout@v4
-    - uses: actions/cache@v4
+    - uses: actions/cache@v4.2.0
       with:
         path: |
           ~/go/pkg/mod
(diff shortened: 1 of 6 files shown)


.github/workflows/prover-testing.yml
```diff
@@ -13,18 +13,19 @@ jobs:
...
       with:
         fetch-depth: 0
-    - uses: actions/cache@v4
+    - name: install Go
+      uses: actions/setup-go@v5
+      with:
+        go-version: 1.23.x
+        cache-dependency-path: |
+              prover/go.sum
+    - uses: actions/cache@v4.2.0
```
jonesho, line 26: Why bump to v4.2.0? I think v4 already point to v4.2.0: https://github.com/actions/cache/tags, same for other places in this PR

.github/workflows/prover-testing.yml
```diff
@@ -13,18 +13,19 @@ jobs:
...
       uses: actions/checkout@v4
       with:
         fetch-depth: 0
-    - uses: actions/cache@v4
+    - name: install Go
+      uses: actions/setup-go@v5
+      with:
+        go-version: 1.23.x
+        cache-dependency-path: |
+              prover/go.sum
```
jonesho, line 25: What's this trick? Will use the cache if same `prover/go.sum` was found in caches?

kyzooghost: Currently we get the following warning on invoking **action/setup-go**:
```
Restore cache failed: Dependencies file is not found in /runner/_work/linea-monorepo/linea-monorepo. Supported file pattern: go.sum
```

See the most recent runs in main
- https://github.com/Consensys/linea-monorepo/actions/runs/12909079251/job/35996300589
- https://github.com/Consensys/linea-monorepo/actions/runs/12909079251/job/35996677420 

From the **action/setup-go** readme:
- The action defaults to search for the dependency file - go.sum in the repository root, and uses its hash as a part of the cache key.
- Use cache-dependency-path input for cases where the go.sum file is located in different subdirectories

We don't have a go.sum in the repository root - it is at `prover/go.sum`. So we were previously not using the built-in cache for setup-go.

kyzooghost: So same `prover/go.sum` = same cache key = cache hit

.github/workflows/prover-testing.yml
kyzooghost: In the #github-actions-self-hosted-runners Slack channel, there's a recommendation from DevOps to pin the Github Action version so that they are the same in every run.

They gave an example that `gradle/actions/setup-gradle@v4` should be `gradle/actions/setup-gradle@cc4fc85e6b35bafd578d5ffbc76a5518407e1af0 #v4.2.1` instead

jonesho: Actually I'm working on a [PR](https://github.com/Consensys/linea-monorepo/pull/584) for actions/cache, given that it was not used properly (see [issue](https://github.com/Consensys/linea-monorepo/issues/589)), and we might even drop it if there's no performance gain (i.e. it seems the integration and e2e tests would be running longer when cache is used for pulling docker images)

jonesho (approved):

PR #627, opened 2025-01-30
Prover/cross-checking among segments of limitless prover

Soleimani193:
This PR implements the crosse compiledIOP responsible for cross-checking among segments.

### Checklist


prover/protocol/column/store.go
```diff
@@ -263,7 +263,11 @@ func (r *Store) AllPrecomputed() []ifaces.ColID {
 // ordered by rounds and then by order of insertion.
 func (r *Store) AllVerifyingKey() []ifaces.ColID {
 	res := []ifaces.ColID{}
-	rnd := r.byRounds.MustGet(0) // precomputed are always at round zero
+
+	var rnd []*storedColumnInfo
+	if (r.byRounds.Len()) != 0 {
+		rnd = r.byRounds.MustGet(0) // precomputed are always at round zero
```
AlexandreBelling: If it's an assumption, it is best to panic and assume the user is never given the possibility to input a non-zero round number for the store.

prover/protocol/distributed/xcomp/xcomp.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/consensys/gnark/frontend"
+	"github.com/consensys/linea-monorepo/prover/maths/field"
+	"github.com/consensys/linea-monorepo/prover/protocol/accessors"
+	"github.com/consensys/linea-monorepo/prover/protocol/wizard"
+	"github.com/consensys/linea-monorepo/prover/utils/collection"
+)
+
+// GetCrossComp generates an (empty) compiledIOP object that is handling the crosse checks
+// for example, the global sum over logDerivativeSum is zero.
+func GetCrossComp(vRuntimes []*wizard.VerifierRuntime) *wizard.CompiledIOP {
```
AlexandreBelling, line 15: Yes, to the idea to construct it as a verifierAction as it is simple and small enough. But the API of the function cannot make sense as is. We will fix this when integrating with the conglomeration that I wrote.

AlexandreBelling (approved):

prover/protocol/column/store.go
Soleimani193: non indeed MustGet gives panic for the empty store. I wanted to prevent this case as the crosse CompiledIOP does not have any column and is tasked to do only some crosse-checks among segments/modules.

AlexandreBelling: In that case, it would be cleaner to implement it as:

```
// This supports the case where the compiled-IOP does not store any column.
if r.byRounds.Len() == 0 {
    return []ifaces.ColID{}
}
```

That way, the edge-case is explicitly handled and not the main-case

PR #642, opened 2025-01-31
Staterecovery adapt to new besu

jpnovais:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/state-recovery/appcore/logic/src/main/kotlin/linea/staterecovery/StateSynchronizerService.kt
+++ b/state-recovery/appcore/logic/src/main/kotlin/linea/staterecovery/StateSynchronizerService.kt
@@ -130,7 +130,7 @@ class StateSynchronizerService(
       .thenCompose { decompressedBlocksToImport: List<BlockFromL1RecoveredData> ->
         if (decompressedBlocksToImport.isEmpty()) {
           log.info(
-            "stopping recovery synch: imported all blocks up to debugForceSyncStopBlockNumber={} finalization={}",
+            "stopping recovery sync: imported all blocks up to debugForceSyncStopBlockNumber={} finalization={}",
             debugForceSyncStopBlockNumber,
             dataFinalizedV3.intervalString()
           )
--- a/jvm-libs/linea/besu-libs/build.gradle
+++ b/jvm-libs/linea/besu-libs/build.gradle
@@ -6,7 +6,7 @@ plugins {
 //def besuArtifactGroup="org.hyperledger.besu"
 //def besuVersion=libs.versions.besu.get()
 def besuArtifactGroup="io.consensys.linea-besu"
-def besuVersion="25.1-delivery45"
+def besuVersion="25.2-delivery46"
 
 dependencies {
   api("${besuArtifactGroup}:besu-datatypes:${besuVersion}") {


jvm-libs/linea/blob-decompressor/src/test/kotlin/net/consensys/linea/blob/GoNativeBlobDecompressorTest.kt
```diff
@@ -1,42 +1,144 @@
...
+
+    // Only BlockHash and Timestamp are compressed to the Blob
+    assertThat(decodedBlock.header.hash).isEqualTo(originalBesuBlock.header.hash)
+    assertThat(decodedBlock.header.timestamp).isEqualTo(Instant.parse("2025-01-02T12:23:45Z").epochSeconds)
+
+    assertThat(decodedBlock.body.transactions).hasSize(2)
+    val decompressedTx0 = decodedBlock.body.transactions[0]
+    val decompressedTx1 = decodedBlock.body.transactions[1]
+
+    assertThat(decompressedTx0.type).isEqualTo(tx0.type.toBesu())
```
Filter94, line 102: I suspect that we can't just use equals here, but from these assertions I'm missing which fields are there in the decompressed transaction and which are not verified. It would be better if we had a definition of equality defined somewhere as a function

Filter94 (approved):

jvm-libs/linea/blob-decompressor/src/test/kotlin/net/consensys/linea/blob/GoNativeBlobDecompressorTest.kt
jpnovais: > I suspect that we can't just use equals here,
Yes, we cannot.

The fields that are compressed/decompressed are all asserted here. The remaining fields are outside of the decompressor "contract", hence not asserted because they have dummy values. 

> It would be better if we had a definition of equality defined somewhere as a function
Not sure how to do that much better than having current fields flatten out as they are. If something comes up I will address it.

Filter94: > The remaining fields are outside of the decompressor "contract"

The contract is unclear, that's what I'm trying to say

jpnovais: I could come up with something like:
```fun assertDecompressedTransactionFields(
    decompressedTx: org.hyperledger.besu.ethereum.core.Transaction,
    expectedTx: linea.domain.Transaction
  ) {
    assertThat(decompressedTx.type).isEqualTo(expectedTx.type.toBesu())
    assertThat(decompressedTx.sender.toArray()).isEqualTo(expectedTx.toBesu().sender.toArray())
    assertThat(decompressedTx.nonce.toULong()).isEqualTo(expectedTx.nonce)
    assertThat(decompressedTx.gasLimit.toULong()).isEqualTo(expectedTx.gasLimit)
    expectedTx.gasPrice?.let {
      assertThat(decompressedTx.gasPrice.getOrNull()?.asBigInteger).isEqualTo(it.toBigInteger())
    }
    expectedTx.maxFeePerGas?.let {
      assertThat(decompressedTx.maxFeePerGas.getOrNull()?.asBigInteger).isEqualTo(it.toBigInteger())
    }
    expectedTx.maxPriorityFeePerGas?.let {
      assertThat(decompressedTx.maxPriorityFeePerGas.getOrNull()?.asBigInteger).isEqualTo(it.toBigInteger())
    }
    expectedTx.to.let {
      assertThat(decompressedTx.to.getOrNull()?.toArray()).isEqualTo(it)
    }

    assertThat(decompressedTx.value.asBigInteger).isEqualTo(expectedTx.value)
    assertThat(decompressedTx.payload.toArray()).isEqualTo(expectedTx.input)
    expectedTx.accessList?.also { accessList ->
      assertThat(decompressedTx.accessList.getOrNull()).isNotNull
      val decompressedAccList = decompressedTx.accessList.getOrNull()!!
      assertThat(decompressedAccList).hasSize(accessList.size)
      decompressedAccList.zip(accessList).forEach { (decompressedAcc, expectedAcc) ->
        assertThat(decompressedAcc!!.address.toArray()).isEqualTo(expectedAcc.address)
        decompressedAcc.storageKeys.zip(expectedAcc.storageKeys).forEach { (decompressedKey, expectedKey) ->
          assertThat(decompressedKey.toArray()).isEqualTo(expectedKey)
        }
      }
    }
    ```

buit IMO does more harm than good if it fails  to see exactly where the regression was

File: coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/CoordinatorApp.kt [hunk 2]
```diff
@@ -60,15 +59,13 @@ class CoordinatorApp(private val configs: CoordinatorConfig) {
...
-    Web3j.build(
-      HttpService(
-        configs.l2.rpcEndpoint.toString(),
-        okHttpClientBuilder(LogManager.getLogger("clients.l2")).build()
-      ),
-      1000,
-      Async.defaultExecutorService()
-    )
+  private val l2Web3jClient: Web3j = createWeb3jHttpClient(
+    rpcUrl = configs.zkTraces.ethApi.toString(),
```
jonesho, line 63: Why use `configs.zkTraces.ethApi` here instead of `configs.l2.rpcEndpoint`?

jvm-libs/linea/blob-decompressor/src/test/kotlin/net/consensys/linea/blob/GoNativeBlobDecompressorTest.kt
Filter94: I don't get your point. Since every field is asserted individually, it will be clear from the stacktrace where the mismatch is, but your suggestion doesn't solve the problem of clarity which fields remain after compression and which do not.

Something like this could make the contract very clear IMO:

```
assertThat(besuTransaction).isEqualTo(lineaTransaction.copy(r=null, s=null, v=null).toBesu())
```

Filter94 (approved):

state-recovery/appcore/logic/src/main/kotlin/linea/staterecovery/StateSynchronizerService.kt
```diff
@@ -120,40 +120,55 @@ class StateSynchronizerService(
...
-            BlockNumberAndHash(
-              number = decompressedBlocks.last().header.blockNumber,
-              hash = decompressedBlocks.last().header.blockHash
-            )
-          }
+      .thenCompose(this::filterOutBlocksAlreadyImportedAndBeyondStopSync)
+      .thenCompose { decompressedBlocksToImport: List<BlockFromL1RecoveredData> ->
+        if (decompressedBlocksToImport.isEmpty()) {
+          log.info(
+            "stopping recovery synch: imported all blocks up to debugForceSyncStopBlockNumber={} finalization={}",
```
jonesho: Stopping recovery `sync`

state-recovery/besu-plugin/src/main/kotlin/linea/staterecovery/plugin/RecoveryModeManager.kt
```diff
@@ -52,11 +53,26 @@ class RecoveryModeManager(
...
     headBlockNumber = blockNumber.toULong()
     if (!recoveryModeTriggered.get() && hasReachedTargetBlock()) {
+      log.info(
+        "Stopping synchronization services at block={} recoveryTargetBlockNumber={} was reached",
+        headBlockNumber,
+        targetBlockNumber
+      )
       switchToRecoveryMode()
+    } else if (debugForceSyncStopBlockNumber != null && headBlockNumber >= debugForceSyncStopBlockNumber) {
+      log.info(
```
jonesho, line 63: I'm wondering do we want keep printing this info log when `debugForceSyncStopBlockNumber` has reached

config/coordinator/log4j2-dev.xml
```diff
@@ -112,7 +112,7 @@
       <DebouncingFilter/>
       <appender-ref ref="rewrite"/>
     </Logger>
-    <Logger name="clients.l1" level="INFO" additivity="false">
+    <Logger name="clients.l2" level="DEBUG" additivity="false">
```
jonesho, line 115: Why stop printing`clients.l1` info log anymore?

jonesho (approved):

state-recovery/besu-plugin/src/main/kotlin/linea/staterecovery/plugin/RecoveryModeManager.kt
jonesho: I might be wrong, it would not be kept printing given that no block will arrive after `debugForceSyncStopBlockNumber` has reached

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/CoordinatorApp.kt
jpnovais: Because we cannot conflate the block until Geth traces has imported the block and generated the traces for it. Traces counters depend on this.

This will go away after we migrate to Besu.

state-recovery/besu-plugin/src/main/kotlin/linea/staterecovery/plugin/RecoveryModeManager.kt
jpnovais: No we won't because it stops the sync so it won't import more block therefore wont reach this code again.

config/coordinator/log4j2-dev.xml
jpnovais: root log level is in INFO, it will print it. I added DEBUG for L2 to log failures

Filter94 (approved):

PR #645, opened 2025-02-03
fix: temp fix for the corepack issue

jonesho:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/postman/Dockerfile
+++ b/postman/Dockerfile
@@ -3,6 +3,9 @@ FROM node:lts-slim AS base
 ENV PNPM_HOME="/pnpm"
 ENV PATH="$PNPM_HOME:$PATH"
 
+# Temp fix for the corepack issue described in https://github.com/pnpm/pnpm/issues/9029
+RUN npm i -g corepack@latest
+
 RUN corepack enable
 
 RUN apt-get update \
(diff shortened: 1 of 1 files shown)


postman/Dockerfile
```diff
@@ -23,6 +23,9 @@ COPY ./postman/package.json ./postman/package.json
 COPY ./sdk/package.json ./sdk/package.json
 COPY ./ts-libs/linea-native-libs/package.json ./ts-libs/linea-native-libs/package.json
 
+# Temp fix for the corepack issue described in https://github.com/pnpm/pnpm/issues/9029
+RUN npm i -g corepack@latest
```
thedarkjester: what is the difference between this and `RUN corepack enable` that we use in the operations folder? would it work and be faster?

jonesho: We also use `corepack enable` in this Dockerfile, its just a command to enable the corepack explicitly which I believe is the default package manager used by [pnpm](https://pnpm.io/installation#using-corepack). The recommended [workaround](https://github.com/pnpm/pnpm/issues/9029#issuecomment-2629866277) for now is to update the corepack version to latest, which means  `corepack enable` has nothing to do with the issue, anyway I'd moved the fix line above `corepack enable` to make it more clear.

kyzooghost: From working with the POH Docker image, I learned that `RUN corepack enable` doesn't download anything, whereas `npm i -g corepack@latest` or `pnpm i -g corepack@latest` will

PR #647, opened 2025-02-03
Prover/distribution of global queries

Soleimani193:
This PR implements the distribution of global constraints

### Checklist


prover/protocol/distributed/comp_splitting.go
```diff
@@ -90,3 +95,223 @@ func getSegmentFromWitness(wit ifaces.ColAssignment, numSegs, segID int) ifaces.
 	segSize := wit.Len() / numSegs
 	return wit.SubVector(segSize*segID, segSize*segID+segSize)
 }
+
+func GetFreshCompGL(in SegmentModuleInputs) *wizard.CompiledIOP {
```
AlexandreBelling, line 99: Document the function and explains everything that it does from an external point of view

Issue #604, opened 2025-01-27
Export extra metrics for conflation, blob and aggregation


jonesho:
## Description
These extra metrics for conflation, blob and aggregation were requested by the prover team in [#3128](https://github.com/Consensys/zkevm-monorepo/issues/3128) :
### For block

* Avg tx data on L1 `(size of the average uncompressed data of each tx: the size of the entire uncompressed L2 block data divided by the number of tx in the block)`
* Avg compressed tx data on L1 `(size of the average compressed data of each tx: the size of the entire compressed L2 block data divided by the number of tx in the block)`
* ~Avg compression ratio of tx data on L1 `(1.0 - "avg post-compressed tx data size" / "avg pre-compressed tx data size")`~ `Duplicated by the blob compression ratio metric (linea_blob_block_compression_ratio)`

### For conflation

* L1 data `(size of the uncompressed L2 data that would be posted on L1 via blob transaction)`
* L1 data compressed `(size of the compressed L2 data that would be posted on L1 via blob transaction)`
* ~The limits (which already include L1 data)~ `Can be skipped given that traces limits would be removed soon`
* Total gas used 

### For blobs

* ~Number of conflation~ `Done`
* ~Number of blocks~ `Done`
* Size of the uncompressed data
* Size of the compressed data
* Total gas used

### For aggregation

* Number of bridge messages (L1 and L2) `Not a priority, can be revisited later and it can be on prover side to avoid the use of coordinator database`
* Total gas used `Not a priority, can be revisited later`
* ~Number of conflation~ `Done`
* ~Number of block~ `Done`
* ~Number of blobs~ `Done`

## Motivation
Useful for prover and arithmetization team to perform data analysis 

## Tasks
- [x] Export histogram metric about total gas used in each conflation
- [x] Export histogram metric about size of the uncompressed data in each conflation
- [x] Export histogram metric about size of the compressed data in each conflation
- [x] Export histogram metric about average uncompressed transaction data size in each conflation
- [x] Export histogram metric about average compressed transaction data size in each conflation
- [x] Export histogram metric about total gas used in each blob
- [x] Export histogram metric about size of the uncompressed data in each blob
- [x] Export histogram metric about size of the compressed data in each blob
- [ ] ~Export metrics about total gas used in each aggregation~
- [ ] ~Export metrics about number of bridge messages in each aggregation~

## Acceptance criteria
- [x] All metrics above are exported

## Risks
- [ ] N/A

## Remember to
[...]

PR #654, opened 2025-02-05
feat: add metrics in GlobalBlobAwareConflationCalculator and move bri…

jonesho:
…dge logs and state root hash retrieval from execution prover client to coordinator

This PR implements issue(s) #604

### Checklist


coordinator/clients/prover-client/file-based-client/src/main/kotlin/net/consensys/zkevm/coordinator/clients/prover/FileBasedExecutionProverClientV2.kt
```diff
@@ -33,48 +26,6 @@ data class BatchExecutionProofRequestDto(
   val blocksData: List<RlpBridgeLogsData>
 )
 
-internal class ExecutionProofRequestDataDecorator(
```
jonesho, line 36: This class is deleted as I would like to move the responsibility of retrieving "L2 state root hash and bridge logs" from execution prover client to the execution proof coordinator, and hence the prover client would just be responsible for creating the proof request based on the data passed by the caller.

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/proofcreation/ZkProofCreationCoordinatorImpl.kt
```diff
@@ -1,19 +1,41 @@
...
+import org.web3j.protocol.Web3j
+import org.web3j.protocol.core.DefaultBlockParameter
 import tech.pegasys.teku.infrastructure.async.SafeFuture
 
 class ZkProofCreationCoordinatorImpl(
-  private val executionProverClient: ExecutionProverClientV2
+  private val executionProverClient: ExecutionProverClientV2,
+  private val l2MessageServiceLogsClient: L2MessageServiceLogsClient,
+  private val l2Web3jClient: Web3j,
+  private val encoder: BlockEncoder
```
jonesho: Move the responsibility of retrieving "L2 state root hash and bridge logs" from execution prover client to the execution proof coordinator.

jpnovais: I think it makes more sense. Thanks for the change.

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/proofcreation/ZkProofCreationCoordinatorImpl.kt
```diff
@@ -1,19 +1,41 @@
...
 ) : ZkProofCreationCoordinator {
   private val log: Logger = LogManager.getLogger(this::class.java)
 
+  private fun getBlockStateRootHash(blockNumber: ULong): SafeFuture<String> {
+    return l2Web3jClient
+      .ethGetBlockByNumber(
+        DefaultBlockParameter.valueOf(blockNumber.toBigInteger()),
```
jpnovais:
```suggestion
        blockNumber.toBlockParameter().toWeb3j(),
```

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/proofcreation/ZkProofCreationCoordinatorImpl.kt
```diff
@@ -22,16 +44,36 @@ class ZkProofCreationCoordinatorImpl(
...
+              val rlp = encoder.encode(block).encodeHex()
+              RlpBridgeLogsData(rlp, bridgeLogs)
+            }
+            executionProverClient.requestProof(
+              BatchExecutionProofRequestV1(
+                blocks = blocksConflation.blocks,
+                tracesResponse = traces.tracesResponse,
+                type2StateData = traces.zkStateTraces,
+                blocksData = blocksData,
+                keccakParentStateRootHash = previousKeccakStateRootHash
```
jpnovais: we should find a better name for this key:  `blocks` and `blocksData` are booth very generic, yet we have 2 keys for blocks...

IMO, it should not be the responsibility of this class to RLP encode the blocks. We pass the Domain Model block to the Prover. RLP is a contract of the prover API, therefore shall be handled there. If tomorrow Prover API changes and uses different encoding you will need to change this business logic classes, breaking encapsulation. 

So I think the new interface Should Be

```kotlin
data class BatchExecutionProofRequestV1(
  val blocks: List<Block>,
  val tracesResponse: GenerateTracesResponse,
  val type2StateData: GetZkEVMStateMerkleProofResponse,
  val blocksData: List<BridgeLogsData> //not RlpBridgeLogsData, this is internal of the prover client
  val keccakParentStateRootHash: ByteArray
) 
```

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/conflation/GlobalBlobAwareConflationCalculator.kt
```diff
@@ -32,10 +35,93 @@ class GlobalBlobAwareConflationCalculator(
   override val lastBlockNumber: ULong
     get() = conflationCalculator.lastBlockNumber
 
+  private val gasUsedInBlobHistogram = metricsFacade.createHistogram(
+    category = LineaMetricsCategory.BLOB,
+    name = "gas.used",
+    description = "Total gas used in each blob"
+  )
```
jpnovais: This metric's name is misleading IMO "gas used" suggests gas used to send blobs to L1. 

I think we want to measure how much "linea gas is compressed in a blob"; 

So my suggestions are:
- `gas` (without used) - my favourite
- `gas.compressed`

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/conflation/GlobalBlobAwareConflationCalculator.kt
```diff
@@ -32,10 +35,93 @@ class GlobalBlobAwareConflationCalculator(
...
+    description = "Compressed L2 data size in bytes of each blob"
+  )
+  private val uncompressedDataSizeInBlobHistogram = metricsFacade.createHistogram(
+    category = LineaMetricsCategory.BLOB,
+    name = "uncompressed.data.size",
+    description = "Uncompressed L2 data size in bytes of each blob"
+  )
+  private val gasUsedInBatchHistogram = metricsFacade.createHistogram(
+    category = LineaMetricsCategory.BATCH,
+    name = "gas.used",
```
jpnovais: Same note as the blob

coordinator/core/src/test/kotlin/net/consensys/zkevm/ethereum/coordination/aggregation/ProofAggregationCoordinatorServiceTest.kt
```diff
@@ -93,7 +93,7 @@ class ProofAggregationCoordinatorServiceTest {
       consecutiveProvenBlobsProvider = mockAggregationsRepository::findConsecutiveProvenBlobs,
       proofAggregationClient = mockProofAggregationClient,
       aggregationL2StateProvider = aggregationL2StateProvider,
-      metricsFacade = MicrometerMetricsFacade(registry = SimpleMeterRegistry()),
+      metricsFacade = mock<MetricsFacade>(defaultAnswer = Mockito.RETURNS_DEEP_STUBS),
```
jpnovais: What is your motivation to use a mock instead of the real implementation? 

I encourage you to read [this section](https://abseil.io/resources/swe-book/html/ch13.html#real_implementations) about favoring Real Implementation of mocks

coordinator/core/src/test/kotlin/net/consensys/zkevm/ethereum/coordination/conflation/GlobalBlobAwareConflationCalculatorTest.kt
```diff
@@ -86,7 +87,8 @@ class GlobalBlobAwareConflationCalculatorTest {
     calculator = GlobalBlobAwareConflationCalculator(
       conflationCalculator = globalCalculator,
       blobCalculator = calculatorByDataCompressed,
-      batchesLimit = defaultBatchesLimit
+      batchesLimit = defaultBatchesLimit,
+      metricsFacade = mock<MetricsFacade>(defaultAnswer = Mockito.RETURNS_DEEP_STUBS)
```
jpnovais: Same goes here.


Also, we should add tests to check that the metrics are correct.

jonesho: Agree to add tests for the metrics and would use mocks here for `metricsFacade` and histogram instances to check the metric values passed to the histogram `record` function

docker/compose-spec-l1-services.yml
```diff
@@ -78,9 +78,9 @@ services:
       - /bin/bash
       - -c
       - |
-        PRAGUE=${PRAGUE} bash /generate-genesis.sh
+        PRAGUE=${PRAGUE:-} bash /generate-genesis.sh
         /usr/local/bin/eth2-testnet-genesis deneb \
-        --config /config/network-config.yml \${L1_GENESIS_TIME:+--timestamp ${L1_GENESIS_TIME} \}
+        --config /config/network-config.yml \${L1_GENESIS_TIME:+--timestamp ${L1_GENESIS_TIME:-} \}
```
jonesho, line 83: To address the warning logs of missing env variables when running docker compose

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/conflation/GlobalBlobAwareConflationCalculator.kt
```diff
@@ -32,10 +35,101 @@ class GlobalBlobAwareConflationCalculator(
...
+    }
+  }
+
+  private fun recordBlobMetrics(blobInterval: BlockInterval, blobCompressedDataSize: Int) {
+    runCatching {
+      val gasUsedInBlob = blobBlockCounters
+        .filter { blobInterval.blocksRange.contains(it.blockNumber) }
+        .sumOf { it.gasUsed }
+      val uncompressedDataSizeInBlob = blobBlockCounters
+        .filter { blobInterval.blocksRange.contains(it.blockNumber) }
```
Filter94, line 116: We can do the filter once here and reuse the result

coordinator/core/build.gradle
Filter94, line 1: Side note. 

`implementation project(":jvm-libs:linea:metrics:micrometer")`

should be replaced by 
`testFixturesImplementation project(":jvm-libs:linea:metrics:micrometer")`

It would be nice to clean this up in this change. Our core implementation doesn't really depend on Micrometer

coordinator/core/src/test/kotlin/net/consensys/zkevm/ethereum/coordination/aggregation/ProofAggregationCoordinatorServiceTest.kt
```diff
@@ -74,7 +78,20 @@ class ProofAggregationCoordinatorServiceTest {
...
     val mockAggregationsRepository = mock<AggregationsRepository>()
     val mockProofAggregationClient = mock<ProofAggregationProverClientV2>()
-    val aggregationL2StateProvider = mock<AggregationL2StateProvider>()
+    val mockAggregationL2StateProvider = mock<AggregationL2StateProvider>()
+
+    // histogram metrics mocks
+    lateinit var mockAggregationSizeInBlocksHistogram: Histogram
+    lateinit var mockAggregationSizeInBatchesHistogram: Histogram
+    lateinit var mockAggregationSizeInBlobsHistogram: Histogram
+    val mockMetricsFacade: MetricsFacade = mock() {
```
Filter94: Isn't it easier to create a real instance instead? We have the micrometer dependency in the test fixtures anyway

coordinator/core/src/test/kotlin/net/consensys/zkevm/ethereum/coordination/conflation/GlobalBlobAwareConflationCalculatorTest.kt
```diff
@@ -43,6 +52,17 @@ class GlobalBlobAwareConflationCalculatorTest {
   private lateinit var blobs: MutableList<Blob>
   private val defaultBatchesLimit = 2U
   private val conflationTargetEndBlockNumbers: MutableSet<ULong> = mutableSetOf()
+  private lateinit var mockMetricsFacade: MetricsFacade
+
+  // histogram metrics mocks
+  private lateinit var mockGasUsedInBlobHistogram: Histogram
```
Filter94: I feel like this is an ideal case for a fake implementation backed by a list. Mocks instantiation is too repetitive IMO and the only advantage we're taking from the mocks is the `inOrder` verification. IMO a simple Histogram implementation will cover our needs much better with less repetitive and more readable code

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/conflation/GlobalBlobAwareConflationCalculator.kt
```diff
@@ -32,10 +35,101 @@ class GlobalBlobAwareConflationCalculator(
...
+  private fun recordBatchMetrics(conflation: ConflationCalculationResult) {
+    runCatching {
+      val gasUsedInBatch = blobBlockCounters
+        .filter { conflation.blocksRange.contains(it.blockNumber) }
+        .sumOf { it.gasUsed }
+      val uncompressedDataSizeInBatch = blobBlockCounters
+        .filter { conflation.blocksRange.contains(it.blockNumber) }
+        .sumOf { it.blockRLPEncoded.size }
+      val compressedDataSizeInBatch = blobCalculator.getCompressedDataSizeInCurrentBatch()
+      val numOfTransactionsInBatch = blobBlockCounters
```
Filter94: Can we reuse the result of the filtering?

Filter94 (approved):

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/conflation/GlobalBlobAwareConflationCalculator.kt
jonesho: Good catch, missed that one, thanks and updated

Filter94 (approved):

PR #660, opened 2025-02-06
feat: use linea_estimateGas instead of getFeeData for all L2 transact…

jonesho:
…ions

This PR implements issue(s) #

### Checklist


docker/compose-tracing-v1.yml
```diff
@@ -63,7 +63,7 @@ services:
       file: compose-spec-l2-services.yml
       service: coordinator
     environment:
-      config__override__l2-network-gas-pricing__json-rpc-pricing-propagation__disabled: true
+      config__override__l2-network-gas-pricing__json-rpc-pricing-propagation__disabled: false
```
jonesho: @jpnovais Why do we disable this? Do u think its okay to re-enable it? If okay, maybe we could just remove this line

VGau (approved):

PR #688, opened 2025-02-15
Prover: turns the plonk-in-wizard into a query

AlexandreBelling:
As part of the limitless prover effort. We need to turn the plonk-in-wizard into a query. Doing so means that instead of directly declaring the Plonk columns in the wizard we instead create a query that we can subsequently compile *after* the splitting has occurred.

### Checklist


prover/protocol/distributed/compiler/plonkinwizard/distribute.go
```diff
@@ -0,0 +1,69 @@
...
+
+		if !dataInModule {
+			continue
+		}
+
+		if dataInModule != selInModule {
+			utils.Panic("data and selector must be in the same module")
+		}
+
+		if hasMask && dataInModule != !maskInModule {
```
arijitdutta67: I think this should be `if hasMask && dataInModule != maskInModule`

prover/protocol/compiler/plonkinwizard/compile.go
```diff
@@ -0,0 +1,155 @@
...
+
+		ctx.CircuitMask = comp.InsertPrecomputed(
+			ifaces.ColIDf("%v_CIRCMASK", ctx.Q.ID),
+			smartvectors.NewRegular(circMaskVal),
+		)
+	}
+
+	commonconstraints.MustBeActivationColumns(comp, ctx.Q.Selector)
+
+	// This query ensures that sel[i] - sel[i+1] == 1 => mask[i] - mask[i+1] == -1
```
arijitdutta67: I would paraphrase this as sel[i+1] == sel[i] - 1 => mask[i+1] == mask[i] -1 i.e. mask decreases when sel decreases. Btw do we need to ensure that they are binary and once they are zero, they are never one again. Also I did not understand the whole comment. Maybe I am missing something but to me the query looks like "mask increases when sel decreases". Please check

arijitdutta67 (approved):
I had minor suggestions. But broadly LGTM.

prover/protocol/compiler/plonkinwizard/compile.go
```diff
@@ -0,0 +1,155 @@
...
+
+// checkActivators adds the constraints checking the activators are well-set w.r.t
+// to the circuit mask column. See [compilationCtx.Columns.Activators].
+func checkActivators(comp *wizard.CompiledIOP, ctx *context) {
+
+	var (
+		openings   = make([]query.LocalOpening, ctx.Q.GetMaxNbCircuitInstances())
+		mask       = ctx.Q.Selector
+		offset     = utils.NextPowerOfTwo(ctx.Q.GetNbPublicInputs())
+		activators = ctx.PlonkCtx.Columns.Activators
```
ivokub, line 97: I'm trying to follow here - according to `ctx.PlonkCtx.Columns.Activators` the verifier should check that this is actually an activation column, but I don't see it happening. We only check that the first value is correctly opened, but not that it is an activation column?

prover/protocol/compiler/plonkinwizard/compile.go
```diff
@@ -0,0 +1,155 @@
...
+	// is already constrained to never be "-1", this allows simplifying the constraint
+	// as follows. The constraint only works if the number of public inputs is not
+	// exactly a power of two but in this case, the implemented approach does not
+	// work.
+	comp.InsertGlobal(
+		round,
+		ifaces.QueryIDf("%v_MASK_DEC_WHEN_SEL_DEC", ctx.Q.ID),
+		sym.Mul(
+			sym.Sub(ctx.Q.Selector, column.Shift(ctx.Q.Selector, 1)),
+			sym.Sub(ctx.CircuitMask, column.Shift(ctx.CircuitMask, 1), -1),
```
ivokub: For me it seems the constant should be `1` here? Otherwise, this would only hold if we have 0->1 transition?

prover/protocol/compiler/plonkinwizard/compile.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+func compileQuery(comp *wizard.CompiledIOP, q *query.PlonkInWizard) {
+
+	var (
+		round          = max(q.Data.Round(), q.Selector.Round())
+		maxNbInstances = q.GetMaxNbCircuitInstances()
+		ctx            = &context{
+			Q:        q,
+			PlonkCtx: plonkinternal.PlonkCheck(comp, string(q.ID), round, q.Circuit, maxNbInstances),
```
ivokub: We're not passing plonk options?

prover/protocol/distributed/compiler/plonkinwizard/distribute.go
```diff
@@ -0,0 +1,69 @@
...
+
+// DistributePlonkInWizard scans initComp for non-ignored PlonkInWizard queries and
+// identifies the ones corresponding to moduleName. For each of those queries, the
+// function inserts a corresponding query into the module. The function will panic
+// if:
+//
+//   - the query is cross-module
+//
+// The new query is analog to the one in initComp but and shares the same name.
+func DistributePlonkInWizard(
```
ivokub, line 18: Method defined, but in practice not called yet. I don't know the details of limitless prover, but its not yet fully functional?

prover/protocol/distributed/compiler/plonkinwizard/distribute.go
ivokub: Agree, seems like so.

prover/protocol/distributed/compiler/plonkinwizard/distribute.go
```diff
@@ -0,0 +1,69 @@
...
+	moduleName distributed.ModuleName,
```
ivokub, line 19: Not directly about this PR, but seems that `distributed.ModuleName` is defined as an alias to `string`, this means that it is exacly the same as `string`. Wouldn't it be better to define it as a distinct type instead so we would have some type safety? A la `type ModuleName string` (note missing `=`)

ivokub (commented):
In general, looks good to me. However I think we have lost applications of `PlonkOption` during refactor and they are not applied anymore? See the comment `plonkinwizard/compile.go`. 

And I think it is also a bit inconvenient that now the `PlonkOption` fields are defined as `[]any` now. It seems due to trying to avoid import cycles as the options refer directly to the CompilationCtx and we cannot import it in `dedicated/plonk`. But I think we could define `Option` type in `dedicated/plonk` and then use implict interfaces to set the options.

So for now not accepting yet as I'm not sure we haven't lost functionality. Could you confirm?

prover/protocol/compiler/plonkinwizard/compile.go
AlexandreBelling: > Btw do we need to ensure that they are binary and once they are zero, they are never one again.
Yes, this is the role of the [commonconstraints.MustBeActivationColumns]

This ensures that the transition 1 -> 0 of sel can only happen at a position where the mask is going from 0 -> 1.

This looks like this;

```
sele:   1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 0 0 0 0 0 0 0 0 0 0 0
mask:   1 1 1 0 1 1 1 0 1 1 1 0 1 1 1 0 1 1 1 0 1 1 1 0 1 1 1 0
```

mask is like a "heartbeat" with a period occurring every 4 positions, and I want to enforce that sele cancel exactly at the end of 1 period.

prover/protocol/distributed/compiler/plonkinwizard/distribute.go
AlexandreBelling: I agree, I am actually refactoring that in my current branch

prover/protocol/distributed/compiler/plonkinwizard/distribute.go
AlexandreBelling: Also we are missing the big component to integrate everything, that's what I am working on. As far as it go, I cannot really confirm if this code will be integrated directly but I will use it as a draft.

prover/protocol/compiler/plonkinwizard/compile.go
AlexandreBelling: That's actually intended, 

0 -> 1, indicates the end of a mask period (e.g. a new plonk instance)
And I want the selector to cancel exactly just before the next instance. It's a bit over-constrained, I don't think it would be a big issue if we allowed sel: 1 -> 0 when mask: 0 -> * but this way there is only one valid assignment for sel which gives me more confidence.

prover/protocol/compiler/plonkinwizard/compile.go
AlexandreBelling: Yeah, the wording is not good (I just kept the current one). The [Activators] are columns of length 1 (thus single field elements) that are boolean and indicate of the corresponding Plonk-in-Wizard instance is actually used or not. It's a recent optimization (from September of last year iirc). If I have a circuit instance (say ECDSA) that I don't fully use, I assign all the columns of the circuit to zero and I cancel the related constraints (by multiplying them by the activator which is equal to zero).

AlexandreBelling: The checks that I am doing here is whether the circuit was indeed to be skipped or not according to the inputs of the circuit

ivokub: Aaah, I see. Yup in that case we don't need to check activation constraint.

prover/protocol/compiler/plonkinwizard/compile.go
ivokub: Got it 👍

AlexandreBelling:
Yeah we are not loosing functionality in this way. The tricky thing is that the Plonk options that we have today are options for the [PlonkCheck] function which was in `protocol/dedicated/plonk` but now this part of the code is part of a compiler (i.e. this is very internal: PlonkCheck should not be exposed to the user but the options should.

I'll make a "second" option type in the query package to have a cleaner separation

PR #694, opened 2025-02-17
feat: add support of eth_estimateGas for L1 transactions

jonesho:
This PR implements issue(s) #[3898](https://github.com/Consensys/zkevm-monorepo/issues/3898)

### Checklist


coordinator/clients/smart-contract-client/src/main/kotlin/net/consensys/linea/contract/Web3JContractAsyncHelper.kt
```diff
@@ -44,28 +47,117 @@ class Web3JContractAsyncHelper(
...
+        if (it.hasError()) {
+          log.info(
+            "eth_estimateGas failed for tx with blobCarrying={} error={} revertReason={}",
+            withBlobs,
+            it.error.message,
+            getRevertReason(it.error, smartContractErrors)
+          )
+          null
+        } else {
+          log.info(
```
Filter94: INFO level might be too much. I suggest lowering it to `DEBUG`

coordinator/clients/smart-contract-client/src/main/kotlin/net/consensys/linea/contract/l1/Web3JLineaRollupSmartContractClient.kt
```diff
@@ -112,32 +110,30 @@ class Web3JLineaRollupSmartContractClient internal constructor(
    */
   override fun submitBlobs(
     blobs: List<BlobRecord>,
-    gasPriceCaps: GasPriceCaps?
+    gasPriceCaps: GasPriceCaps?,
+    useEstimatedGas: Boolean
```
Filter94: I don't think we should change `LineaRollupSmartContractClient` interface. I suggest you creating another implementation of it instead, that would use `eth_estimateGas`

Filter94 (changes requested):

coordinator/clients/smart-contract-client/src/main/kotlin/net/consensys/linea/contract/l1/Web3JLineaRollupSmartContractClient.kt
jonesho: My initial thought is to treat the `useEstimatedGas` option similar as the `gasPriceCaps` option, but I agree we shouldn't have modified the `LineaRollupSmartContractClient`, I'd changed that to be handled by Web3JContractAsyncHelper to avoid passing the `useEstimatedGas` boolean flag around.

coordinator/clients/smart-contract-client/src/main/kotlin/net/consensys/linea/contract/Web3JContractAsyncHelper.kt
```diff
@@ -232,17 +330,32 @@ class Web3JContractAsyncHelper(
...
-    val gasLimit = gasProvider.getGasLimit(function.name)
-    val tx = Transaction.createFunctionCallTransaction(
-      transactionManager.fromAddress,
-      null,
-      null,
-      gasLimit,
-      contractAddress,
-      FunctionEncoder.encode(function)
-    )
+    return getGasLimit(function)
```
Filter94: This doesn't quite makes sense because `getGasLimit` already does what `executeEthCall` is supposed to do if `useEthEstimateGas` is enabled

jonesho: Yes, I was thinking the same but indeed it's a bit different, if `useEtheEstimateGas` is enabled, we would call `eth_estimateGas` to estimate the gas limit and then explicitly set it as the `gas` field in the subsequent `eth_call` to double check against if the "estimated gas limit" is sufficient enough, given that "estimated gas limit" is not a guarantee and could sometimes be insufficient when being used in `eth_call` or `eth_sendRawTransaction`

> In practice, this mismatch can happen, especially in complex transactions or when dealing with contracts that have unpredictable gas usage. To mitigate this:
> - Test with eth_call: Before relying on the estimate for a real transaction, simulate the call with eth_call using the estimated gas to see if it holds up.

wdyt?

Filter94:
I think we'll have an issue with blob submissions.

For blob submission the way things work now we'd send 1 `eth_estimateGas` for the first set of blobs, which would succeed. But if we're sending more than 6 blobs in a single `BlobSubmissionCoordinator` tick, I think `eth_estimateGas` would fail for the second sendBlobs transaction. Meaning that essentially `BlobSubmissionCoordinator` will be restricted to 1 blob submission per tick. That shouldn't disable anything functionally, but I think it may result in some errors in the logs

coordinator/clients/smart-contract-client/src/main/kotlin/net/consensys/linea/contract/Web3JContractAsyncHelper.kt
Filter94: It makes sense with a static limit, but when `useEtheEstimateGas` is set, I don't really see any point in `eth_estimateGas` -> `eth_call` -> `eth_estimateGas` -> `eth_sendRawTransaction` flow.

jonesho: To remove the extra call of `eth_estimateGas`, I'd changed the aggregation submission flow as:
`eth_estimateGas -> eth_call -> eth_sendRawTransaction`

jonesho:
> I think we'll have an issue with blob submissions.
> 
> For blob submission the way things work now we'd send 1 `eth_estimateGas` for the first set of blobs, which would succeed. But if we're sending more than 6 blobs in a single `BlobSubmissionCoordinator` tick, I think `eth_estimateGas` would fail for the second sendBlobs transaction. Meaning that essentially `BlobSubmissionCoordinator` will be restricted to 1 blob submission per tick. That shouldn't disable anything functionally, but I think it may result in some errors in the logs

Yes, that's true, the second sendBlobs transaction is most likely to fail by the `eth_estimateGas` (unless the first sendBlobs tx was quickly included in block), and seems like there's no good solution to that (if we want to send batch of "sequential logic" txs without waiting the receipt, we'd need static gas limit), so I'd suggest to not enable `eth_estimateGas` on blob submission (in coordinator config) given the point above and also provided that the cause of this PR (linea-incident-796) was aggregation submission which failed the `eth_call` with insufficient gas due to forwarded call to another contract with [63/64 remaining gas rule]([EIP-150](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-150.md)); whereas blob submission was relatively simple without any forward call with more stable gas usage.

Filter94:
> and seems like there's no good solution to that

We can try setting like a fixed margin (10% for example) to `eth_estimateGas`'s gas limit estimation. I think it can work well

Filter94 (approved):

Filter94 (approved):

PR #711, opened 2025-02-19
refactor: Remove prover support for v0 blobs

Tabaie:
* Maintain the v0 decompressor
* Remove the v0 decompression circuit
* Remove the v0 compressor
* Group blob decompression responses into an object


gbotrel (approved):
lgtm 👍 
I see no reasons to keep v0 prover around, but maybe @AlexandreBelling you do?

AlexandreBelling:
@jpnovais @Filter94 

Does any of you an issue if we remove this code. It implicitly means we won't be able to generate "V0" compression binaries if we ever need to. AFAICT, the version is still in-use on mainnet.

jpnovais:
Coordinator does not depend on this code, but rather on a compressor binary that was released before. So as long as those binaries are still avaible, coordinator will work just fine. 

In theory we should keep until it is prod version on mainnet, in case we find a bug and need to patch it. If you feel confidnet that not a issue I am ok deleting it.

PR #714, opened 2025-02-19
Remove the EIP-4844 Toggle From Decompression Circuit

Tabaie:
This can reduce the maximum blob size by 9.16 KB (from 793217 B to 802594 B) or open space for the execution data polynomial check


Tabaie:
Closing as the "calldata" (EIP-4844 = false) feature is still needed

PR #725, opened 2025-02-24
fix: bump gnark crypto to fix non std sis params init

gbotrel:
Fix failing test in compiler/vortex due to incorrect SIS initialization with test params.


prover/crypto/ringsis/ringsis_test.go
```diff
@@ -40,28 +40,13 @@ var testCasesKey = []struct {
...
-	// 	},
-	// },
-	// {
-	// 	Size: 256,
-	// 	Params: Params{
-	// 		LogTwoBound:  1,
-	// 		LogTwoDegree: 1,
-	// 	},
-	// },
+	{
```
AlexandreBelling, line 43: We can't just keep the "production" case. It has to work for all parameters even if it means that we use a fallback, inefficient implementation for those.

gbotrel: so, the only decision we made was limb size (bound) has to be a multiple of 8 (byte size). Rest works fine, this PR updates gnark-crypto because previously we incorrectly used the fast path in some cases we shouldn't have

srinathln7:
The latest status on `main` is that this [commit](https://github.com/Consensys/linea-monorepo/commit/43141fc13d2bd2d9420d2412962723051f1d50ae) was reverted due to the ring-SIS error. If this PR fixes the SIS error, we would ideally like to keep the rest of the changes introduced in the reverted commit (as I see a lot of changes were introduced) and undo the reverted commit.

I think the easiest way to achieve this is to cherry-pick the reverted commit and then apply the fix for the SIS error on top of it in the PR. What do you think?

@AlexandreBelling @gbotrel

0xy0k:
Since the SIS initialization was incorrectly using the fast path in some cases, does it need to extend test cases to cover edge cases explicitly?
Specifically, verifying that the fallback implementation works correctly in non-standard parameter cases could prevent future regressions.

prover/crypto/ringsis/ringsis_test.go
0xtinkle: Since the SIS initialization was incorrectly using the fast path in some cases, does it need to extend test cases to cover edge cases explicitly?
Specifically, verifying that the fallback implementation works correctly in non-standard parameter cases could prevent future regressions.

PR #726, opened 2025-02-24
Blob Size Maximizer - improvements

Tabaie:
The blob size maximizer is a little utility that compiles the decompression circuit under different maximum blob sizes to see how much the parameter can be pushed without exceeding 2^27 constraints.

This PR enables it to get more direction from the user (e.g. to binary search within a given bound.) It is useful when the process is run multiple times.


AlexandreBelling (approved):
LGTM

gbotrel (approved):

Issue #744, opened 2025-03-02
feat: enable compatibility check between `zkevm.bin` and `lt` trace file.


DavePearce:
## Description

Currently, both the `zkevm.bin` and the `lt` trace files contain embedded metadata about the underlying commit of `linea-tracer` used to generate them.  As a sanity check, we want to ensure that the same commit of `linea-constraints` is always used to generate both the `zkevm.bin` and the `lt` trace file.  The purpose of this issue is simply to introduce a check to ensure this.

PR #745, opened 2025-03-03
Prover/compatibility check between zkevmbin and lt trace file

DavePearce:
This PR implements issue(s) #744 

This adds a "strong compatibility check" between generated trace files and zkevm.bin constraint files.  Specifically, this check ensures that the `zkevm.bin` file used within the prover was generated from the same commit of `linea-constraints` as the generated `lt` trace file.  This is a strong check which panics if the check fails.


DavePearce:
@AlexandreBelling @gusiri This implements the check that was discussed last week.  I still have some concerns about the code:

* (**Error Handling**) In `arithmetization.go` in function `Assign` error messages arising from `RealLtTraces` are essentially ignored.  I have not changed this behaviour, but it seems iffy to me.  Should we not panic?

* (**Override**) I have added a flag to `arithmetization.Setting` called `RelaxedMode` which can be used to override the compatibility check.  Specifically, I added a CLI flag to `bin/checker` for this (I imagine it might be handy during debugging, etc).  Overall, I am a little concerned about the level of strictness here.

Also I updated the commit for `linea-constraints` as it is only from commit `73909cc` onwards that metadata is embedded within the `zkevm.bin` file, and likewise only from commit `2b6904` of `go-corset` that meta-data is available in the `lt` files.  And, for `linea-tracer` code for generating `lt` files with embedded metadata [will be merged shortly](https://github.com/Consensys/linea-tracer/pull/1862).

Finally, there are other ways we could implement this ... so happy to talk over any thoughts you have.

AlexandreBelling (commented):
Is this because we don't use this value anymore ?

DavePearce:
> Is this because we don't use this value anymore ?

@AlexandreBelling sorry for some reason I can't see what the comment is on.  Is that `constraint-versions.txt`?  Yes its because there is no tag on the commit for `constraints` I chose.

DavePearce (commented):
@gusiri Thanks for the updates

AlexandreBelling (approved):

AlexandreBelling:
Yeah, it seems I failed that comment somehow.

That was about the constraints-version.txt. I guess the version is directly inside of the zkevm.bin now

gusiri:
This PR simplifies the compatibility check. Now, we only need to verify whether the tracer and the prover share the same commit, regardless of version information or the regex pattern in the .lt file name.

In most cases, this will work smoothly, but I’d like to share a real-world scenario we encountered recently.

During the `beta-v1.2-rc3` deployment, we discovered a bug in the constraints. We fixed it, but we didn’t release a new tracer. These two are compatible. 

- Tracer version: `beta-v1.2-rc3` (constraints: `17c3732`)
- (Hotfix) Prover constraints: `beta-v1.2-rc3.1` (constraints: `e23c58a`)

With this PR, the prover in this scenario will now reject the trace.
In such cases, we would need to manually disable the compatibility check in the prover code rather than releasing and deploying a new tracer with the same constraints. Correct?

If so, should we also add a CLI flag for the full and partial prover? This would also require passing the flag in ArgoCD when necessary.

Please correct me if I am wrong. 
@DavePearce @AlexandreBelling What do you think?

prover/zkevm/checker.go
```diff
@@ -32,6 +32,7 @@ func CheckerZkEvm(tl *config.TracesLimits) *ZkEvm {
 			Arithmetization: arithmetization.Settings{
 				Limits:            tl,
 				OptimisationLevel: &mir.DEFAULT_OPTIMISATION_LEVEL,
+				RelaxedMode:       false,
```
gusiri: It would be great if we could rename `RelaxedMode` to something more intuitive, such as `SkipCompatibilityCheck`, `SkipConstraintVersionCheck`, `DisableCompatibilityCheck`, or `DisableConstraintVersionCheck`.

This would help others without prior context better understand the purpose of this configuration.
Especially if we decide to provide this as a CLI flag for the full or partial prover, a clearer name would make it easier for the Ops team and release managers to understand its function.

prover/zkevm/full.go
```diff
@@ -130,6 +130,7 @@ func fullZKEVMWithSuite(tl *config.TracesLimits, suite compilationSuite) *ZkEvm
 		Arithmetization: arithmetization.Settings{
 			Limits:            tl,
 			OptimisationLevel: &mir.DEFAULT_OPTIMISATION_LEVEL,
+			RelaxedMode:       false,
```
gusiri: Discussed this with Alex, and we think this mode in the full and partial prover need to be added to the config (the prover .toml config file) so that we can turn it on off when necessary.

prover/zkevm/checker.go
DavePearce: Yeah, fair enough. I have changed this now:  `RelaxedMode` => `IgnoreCompatibilityCheck`.

prover/zkevm/full.go
DavePearce: Ok, I will need to figure out how to do that.  Stay tuned :)

gusiri (dismissed):

prover/zkevm/arithmetization/arithmetization.go
```diff
@@ -41,19 +56,61 @@ func NewArithmetization(builder *wizard.Builder, settings Settings) *Arithmetiza
...
-
 	traceF := files.MustRead(traceFile)
-	trace, errT := ReadLtTraces(traceF, a.Schema)
+	// Parse trace file and extract raw column data.
+	rawColumns, metadata, errT := ReadLtTraces(traceF, a.Schema)
+
+	// Performs a compatibility check by comparing the constraints
+	// commit of zkevm.bin with the constraints commit of the trace file.
+	// Panics if an incompatibility is detected.
+	if *a.Settings.IgnoreCompatibilityCheck == false {
```
gusiri, line 77: @DavePearce I moved this to the top of Assign() so that constraints incompatibility fails earlier than other errors  (e.g., error loading the trace, fpath=%q, err=%v).

gusiri:
@DavePearce @AlexandreBelling I tested and confirmed that:

- For both the `checker` and `prover`, the compatibility check panics correctly when there's a constraints mismatch between zkevm.bin and the trace file.
- `ignore_compatibility_check` is set to `false` by default if not specified.
- When `ignore_compatibility_check = true`, the compatibility check is skipped.

gusiri (approved):

DavePearce:
Fantastic!!

AlexandreBelling (approved):

PR #771, opened 2025-03-12
feat: add send-bundle e2e test case with workaround to support traces…

jonesho:
…-v1 sequencer

This PR implements issue(s) #

### Checklist


e2e/src/common/utils.ts
```diff
@@ -57,6 +57,35 @@ export const encodeData = (types: string[], values: unknown[], packed?: boolean)
   return ethers.AbiCoder.defaultAbiCoder().encode(types, values);
 };
 
+export function generateUUIDv4(): string {
```
fluentcrafter: Why not use built-in [randomUUID](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) ?

```javascript
const crypto = require('crypto');
console.log(crypto.randomUUID().toString('hex'));
2e87767f-3d20-4ac6-b37c-7f01a10e4b52
```

e2e/src/common/utils.ts
```diff
@@ -121,6 +150,64 @@ export class LineaEstimateGasClient {
...
+        jsonrpc: "2.0",
+        method: "linea_sendBundle",
+        params: [
+          {
+            txs,
+            replacementUUID,
+            blockNumber,
+          },
+        ],
+        id: 1,
```
fluentcrafter: id should be dynamic and change on every request.

e2e/src/common/utils.ts
```diff
@@ -57,6 +57,35 @@ export const encodeData = (types: string[], values: unknown[], packed?: boolean)
...
 
+export function generateUUIDv4(): string {
+  return "xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx".replace(/[xy]/g, (c) => {
+    const r = (Math.random() * 16) | 0;
+    const v = c === "x" ? r : (r & 0x3) | 0x8;
+    return v.toString(16);
+  });
+}
+
+export async function pollForBlockNumber(
```
fluentcrafter: IMO this function can be improved:

 - Why passing a comparing function and expectedNumber? This is anti-pattern, either one or the other.
 - Using a comparingFunction it could be a generic polling function
 
 Something in the lines of
 ```javascript
async function awaitUntil<T>(
  callback: () => Promise<T>,
  stopRetry: (a: T) => boolean,
  pollingIntervalMs: number = 500,
  timeoutMs: number = 2 * 60 * 1000,
): Promise<T> {
 ```

e2e/src/send-bundle.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+          to: recipientAccount.address,
+          value: ethers.parseUnits("1000", "wei"),
+          maxPriorityFeePerGas: ethers.parseEther("0.000000001"), // 1 Gwei
+          maxFeePerGas: ethers.parseEther("0.00000001"), // 10 Gwei
+          nonce: senderNonce++,
+        };
+        txs.push(await getRawTransactionHex(txRequest, senderWallet));
+        txHashes.push(await getTransactionHash(txRequest, senderWallet));
+      }
+      const targetBlockNumber = (await config.getL2Provider().getBlockNumber()) + 5;
```
fluentcrafter, line 44: what is the semantics/contract of targetBlockNumber? Reason asking is that I see potential for this to become a flacky test because it was not included in the next 5 block....

e2e/src/send-bundle.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+      const replacementUUID = generateUUIDv4();
+
+      try {
+        await lineaSendBundleClient.lineaSendBundle(txs, replacementUUID, "0x" + targetBlockNumber.toString(16));
+      } catch (err) {
+        if (err instanceof Error) {
+          expect(err.message).toStrictEqual("Method not found");
```
fluentcrafter: Why are we suppressing the error of the the method is not found? Shouldn;t we fail the test?

e2e/src/send-bundle.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+      }
+
+      const hasReachedTargeBlockNumber = await pollForBlockNumber(config.getL2Provider(), targetBlockNumber);
+
+      expect(hasReachedTargeBlockNumber).toBeTruthy();
+      for (const tx of txHashes) {
+        const receipt = await config.getL2Provider().getTransactionReceipt(tx);
+        expect(receipt?.status).toStrictEqual(1);
+      }
+      const recipientBalance = await config.getL2Provider().getBalance(recipientAccount.address);
```
fluentcrafter: IMO this is unnecessary, we are testing the EVM with this assertion...

e2e/src/send-bundle.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+      expect(recipientBalance).toStrictEqual(ethers.parseUnits("3000", "wei"));
+    },
+    120_000,
+  );
+
+  it.concurrent(
+    "Should successfully call sendBundle to RPC node but the bundled txs should not get included",
```
fluentcrafter: can we improve the descriptions? it's it does not tell what scenario it is, that one get included the other not.

e2e/src/send-bundle.spec.ts
jonesho: The issue is that the besu nodes in tracing-v1 doesn't support bundle requests (not support in `linea-sequencer` plugin v0.1.4). I finally found a way to explicitly skip the tests by detecting the besu node's bundle support in runtime (its not straightforward, given that variables set in global setup are not directly accessible in test suites because Jest isolates test files to ensure test independence).

e2e/src/send-bundle.spec.ts
jonesho: This `targetBlockNumber` is to tell the sequencer to include the bundled txs in which block, current block number plus 5 blocks seems short but reasonable given that I haven't noticed any test failures because of this in local and pipeline e2e tests. I'd keep it for now to avoid adding run time to e2e tests.

e2e/src/send-bundle.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+  LineaBundleClient,
+  pollForBlockNumber,
+} from "./common/utils";
+import { ethers, TransactionRequest } from "ethers";
+
+describe("Send bundle test suite", () => {
+  const l2AccountManager = config.getL2AccountManager();
+  const lineaCancelBundleClient = new LineaBundleClient(config.getSequencerEndpoint()!);
+  const lineaSendBundleClient = new LineaBundleClient(config.getL2BesuNodeEndpoint()!);
+  const itif = shouldSkipBundleTests ? it.skip : it.concurrent;
```
fluentcrafter: Maybe this condition should be in the outer `describe`, instead of every individual `it`? WDYT?

fluentcrafter (dismissed):

e2e/src/send-bundle.spec.ts
jonesho: Yes, good point, changed to skip the entire describe block.

fluentcrafter (approved):

Issue #795, opened 2025-03-21
Bridge UI V2 E2E Tests


kyzooghost:
## Description
In rolling out the Bridge UI V2, we have disabled the suite of unit tests for the Bridge UI. We would like to add back a suite of unit and E2E tests.

## Motivation
Enhance code robustness

## Tasks
- [ ] Add baseline suite of unit tests for Bridge UI

## Acceptance criteria
- [ ] CI passing

## Risks
- [ ] Implementing brittle tests

## Remember to
 - [x] Add the `documentation` label in case there is an impact on the documentation
 - [x] Add `priority` and `team` labels
 - [x] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #814, opened 2025-03-25
[Feat] Bridge UI Test Workflow

kyzooghost:
This PR implements issue(s) #795 

### Checklist


.github/workflows/bridge-ui-e2e-tests.yml
```diff
@@ -24,38 +24,63 @@ jobs:
...
       - name: Setup nodejs environment
         uses: ./.github/actions/setup-nodejs
         with:
-          node-version: 18.17.0
+          node-version: 20.17.0
           pnpm-install-options: '--frozen-lockfile --prefer-offline'
 
       - name: Install Playwright
-        run: pnpm dlx playwright@1.45.3 install --with-deps
+        run: pnpm dlx playwright@1.51.1 install --with-deps chromium
```
VGau, line 47: Is there reason to add the chromium option ?

.github/workflows/bridge-ui-e2e-tests.yml
```diff
@@ -24,38 +24,63 @@ jobs:
...
+          # Can run unit tests in parallel
+          CI: "false"
+          NEXT_PUBLIC_INFURA_ID: ${{ secrets.PUBLIC_BRIDGE_UI_INFURA_ID }}
+          NEXT_PUBLIC_LIFI_API_KEY: "placeholder"
+          NEXT_PUBLIC_WALLET_CONNECT_ID: "placeholder"
+          NEXT_PUBLIC_DYNAMIC_ENVIRONMENT_ID: "placeholder"
+
+      # Prerequisite - Testing wallet must have >0 ETH and USDC on Sepolia
+      - name: Run E2E tests
         run: xvfb-run pnpm run -F bridge-ui test:e2e:headful
```
VGau, line 81: did you try to run the headless mode in the CI ?

bridge-ui/playwright.config.ts
```diff
@@ -1,14 +1,21 @@
 import { defineConfig, devices } from "@playwright/test";
+import "dotenv/config";
 
 export default defineConfig({
-  testDir: "./test/e2e",
-  timeout: 60_000,
+  testDir: ".",
```
VGau, line 5: Why did you remove the test dir ?

bridge-ui/src/components/ui/toggle-switch/index.tsx
```diff
@@ -8,7 +8,7 @@ type Props = {
   disabled?: boolean;
 };
 
-export default function ToggleSwitch({ checked = false, onChange, disabled }: Props) {
+export default function ToggleSwitch({ checked = false, onChange, disabled, ...rest }: Props) {
```
VGau, line 11: What is rest in this case ? The Props type only contain three params :)

bridge-ui/src/types/cctp.ts
```diff
@@ -3,7 +3,7 @@ export enum CctpAttestationMessageStatus {
   COMPLETE = "complete",
 }
 export type CctpAttestationMessage = {
-  attestation: `0x${string}`;
+  attestation: `0x${string}` | "PENDING";
```
VGau, line 6: What is the reason to add "PENDING" ?

bridge-ui/src/utils/cctp.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+  test("should return COMPLETE for a used nonce", async () => {
+    const resp = await getCctpTransactionStatus(toChainStub, cctpApiRespNoExpiry, usedCctpNonce);
+    expect(resp).toBe(TransactionStatus.COMPLETED);
+  });
+
+  test("should return PENDING for a truncated message", async () => {
+    // Immutable creation of new object
+    const corruptedCctpApiResp = { ...cctpApiRespNoExpiry };
+    // Chop off last character. Last 32-bytes of message should be expirationBlock as per https://developers.circle.com/stablecoins/message-format
+    corruptedCctpApiResp.message = corruptedCctpApiResp.message.slice(0, -1) as `0x${string}`;
```
VGau, line 114: General comment for slicing hex: you can use this utility function from viem: https://viem.sh/docs/utilities/slice

bridge-ui/src/utils/cctp.ts
```diff
@@ -25,12 +25,14 @@ const isCctpNonceUsed = async (
...
 
-const getCctpMessageExpiryBlock = (message: string): bigint | undefined => {
+export const getCctpMessageExpiryBlock = (message: string): bigint | undefined => {
   // See CCTPV2 message format at https://developers.circle.com/stablecoins/message-format
   const expiryInHex = message.substring(
     CCTP_V2_EXPIRATION_BLOCK_OFFSET,
     CCTP_V2_EXPIRATION_BLOCK_OFFSET + CCTP_V2_EXPIRATION_BLOCK_LENGTH,
   );
+  // Should be 32-bytes
+  if (expiryInHex.length !== 64) return undefined;
```
VGau, line 35: To compute bytes length you can use this function from viem https://viem.sh/docs/utilities/size

bridge-ui/test/constants.ts
```diff
@@ -15,6 +14,10 @@ export const LINEA_SEPOLIA_NETWORK = {
 };
 
 export const TEST_URL = "http://localhost:3000/";
-export const SEPOLIA_NETWORK_NAME = "Sepolia";
 export const WEI_AMOUNT = formatEther(BigInt(1)).toString();
-export const USDC_AMOUNT = formatUnits(BigInt(1), 6).toString();
+// Must be > minimum CCTP fee
+export const USDC_AMOUNT = formatUnits(BigInt(10), 6).toString();
```
VGau: Quick note here: You can use the following syntax to declare a bigint: `10n`

bridge-ui/src/types/cctp.ts
kyzooghost: It's a possible undocumented response from the CCTP API. If you query the CCTP API right after the `depositForBurn` function call, you will get the following API response

```
{
    attestation: "PENDING",
    message: "0x",
    eventNonce: "0xaf75c4d910592bc593c34bf6eb89937b1326ad43d9e3cf45581512efcf3e7da7",
    cctpVersion: 2,
    status: "pending_confirmations",
  };
```

.github/workflows/bridge-ui-e2e-tests.yml
kyzooghost: no I can try

.github/workflows/bridge-ui-e2e-tests.yml
kyzooghost: to only download the Chromium browser because this is the only browser we specify in `playwright.config.ts`

If you look at the latest `Install Playwright` run for main branch - https://github.com/Consensys/linea-monorepo/actions/runs/14087538514/job/39455459225 - it also downloads Firefox and Webkit which we do not use in our tests

kyzooghost: image

bridge-ui/playwright.config.ts
kyzooghost: This isn't removing the `test` dir, but is enabling Playwright to run tests in whole `/bridge-ui` repo

I added a unit test file `bridge-ui/src/utils/cctp.spec.ts` which Playwright could not find unless I changed the `testDir` config here. This can be run through `pnpm -F bridge-ui run test:unit` command

bridge-ui/src/components/ui/toggle-switch/index.tsx
kyzooghost: This is to enable passing down `data-testid` prop from a parent component - I do this in line 84 in `bridge-ui/src/components/setting/index.tsx`

bridge-ui/src/utils/cctp.spec.ts
kyzooghost: Had a look, I think the JS string slice function has a nicer API for this use case of cutting off the last X characters :)

bridge-ui/src/utils/cctp.ts
kyzooghost: I tried the viem 'size' function and it returns an unexpected result - `size(63_HEX_CHAR_STRING) = 32` so it doesn't detect the truncation of the last hex character in the unit test where I did that

.github/workflows/bridge-ui-e2e-tests.yml
kyzooghost: I tried, Metamask interactions don't work in headless mode

https://github.com/Consensys/linea-monorepo/actions/runs/14127092377/job/39578505269?pr=814

image

This is also documented here - https://docs.synpress.io/docs/known-issues#headless-mode-limitations-on-ci

bridge-ui/test/constants.ts
kyzooghost: done

bridge-ui/src/components/ui/toggle-switch/index.tsx
VGau: I am surprised that it is working but it is fine :)

bridge-ui/src/utils/cctp.ts
VGau: that's weird. let's keep it like this for now :)

VGau (approved):

Issue #817, opened 2025-03-27
Develop RLP Writer for building transactions


thedarkjester:
## Description
Develop an RLP Writer for building transactions

## Motivation
This will provide functionality that allows Linea transaction creation for users to submit Linea transactions on L1 Ethereum.

## Tasks
- [ ] Build RLP Writer for Type 2 transactions using EIP-2718 formatting ( `type[2] || payload` )
- [ ] Testing

## Acceptance criteria
- [ ] All permutations of Type 2 transactions work - calldata, access lists etc. 

see: 
- https://eips.ethereum.org/EIPS/eip-2718
- https://ethereum.org/en/developers/docs/data-structures-and-encoding/rlp/

PR #818, opened 2025-03-27
[Feat] 817 rlp tx writer

thedarkjester:
This PR implements issue(s) #817 

### Checklist


pnpm-lock.yaml
```diff
@@ -248,8 +248,6 @@ importers:
         specifier: 17.7.2
         version: 17.7.2
 
-  contracts/lib/forge-std: {}
```
kyzooghost: going to look into this - luckily it doesn't break the build whether it's there or not

contracts/src/libraries/RlpWriter.sol
```diff
@@ -0,0 +1,197 @@
...
+  /// @notice Encode integer in big endian binary form with no leading zeroes.
+  /// @param _uintValue The integer to encode.
+  /// @return binaryBytes RLP encoded bytes.
+  function _toBinary(uint256 _uintValue) private pure returns (bytes memory binaryBytes) {
+    assembly {
+      let ptr := mload(0x40) // Get free memory pointer
+
+      let i := 0
+      // Scan for first non-zero byte from MSB (big-endian)
+      for {
```
kyzooghost: if `iszero(_uintValue)` is true, then we can skip this loop as we know 'length' = 0

contracts/src/libraries/RlpWriter.sol
```diff
@@ -0,0 +1,197 @@
...
+
+      } lt(i, 32) {
```
kyzooghost: Also I was thinking - it may be possible to binary search (rather than linear search) for the MSB with a non-zero value. Could go something like this:

Let's represent our 32-byte number as - XXXX-XXYX-XXXX-XXXX-XXXX-XXXX-XXXX-XXXX
- Each X represents a byte which could be the most significant non-zero byte
- The most significant non-zero byte is the Y
- An O represents a byte we know is not the most significant non-zero byte

First-round 
- Check most-significant 16-bytes -> not zero
XXXX-XXYX-XXXX-XXXX-OOOO-OOOO-OOOO-OOOO

Second-round
- Check most significant 8-bytes -> not zero
XXXX-XXYX-OOOO-OOOO-OOOO-OOOO-OOOO-OOOO

Third-round
- Check most significant 4-bytes -> is zero
OOOO-XXYX-OOOO-OOOO-OOOO-OOOO-OOOO-OOOO

Fourth-round
- Check the most significant 2-bytes which could still hold the most significant non-zero byte -> is zero
OOOO-OOYX-OOOO-OOOO-OOOO-OOOO-OOOO-OOOO

Fifth-round
- Check the most significant 1-byte of the remaining possible byte - > not zero
OOOO-OOYO-OOOO-OOOO-OOOO-OOOO-OOOO-OOOO

We have found the answer in 5 rounds, a linear search would have taken 7 rounds

May have gotten some details wrong, was outlining a rough approach

contracts/src/libraries/RlpWriter.sol
```diff
@@ -0,0 +1,197 @@
...
+      mstore(binaryBytes, length)
+
+      // Write stripped bytes
+      for {
+        let j := 0
+      } lt(j, length) {
+        j := add(j, 1)
+      } {
+        let shift := mul(sub(length, add(j, 1)), 8)
+        let b := and(shr(shift, _uintValue), 0xff)
```
kyzooghost: Should we give a descriptive name for `b` such as `byteToAdd`?

contracts/src/libraries/RlpWriter.sol
```diff
@@ -0,0 +1,197 @@
...
+        }
+        break
+      }
+
+      let length := sub(32, i) // Number of non-zero bytes
+      binaryBytes := ptr
+      mstore(binaryBytes, length)
+
+      // Write stripped bytes
+      for {
```
kyzooghost: Thinking out loud - wondering if it's possible, instead of this loop, to use single mstore the entire truncated uint256 and update the free memory pointer accordingly.

Something like this:
- Let's take the number represented by 0x0000_0000_0000_0000_0000_0000_BBBB_BBBB, where B is a non-zero byte
- We find length = 8
- Use mstore on the truncated number such that memory layout from free memory pointer looks like:
[Length byte = 0x08][BBBB_BBBB_0000_0000_0000_0000_0000_0000]

- Update free memory pointer by adding 9 bytes
- End result is that there 24 zero-bytes in beyond the free memory pointer, that we can freely overwrite

[0x08][BBBB_BBBB_0000_0000_0000_0000_0000_0000]
__________________^

contracts/src/libraries/RlpWriter.sol
```diff
@@ -0,0 +1,197 @@
...
+        mstore(0x40, add(lengthBytes, 0x21)) // Advance free memory pointer
+      }
+      default {
+        // Case: long length
+        let temp := _itemLength
+        let lengthOfLength := 0
+
+        for {
+
+        } gt(temp, 0) {
```
kyzooghost: Similar to the linear byte-by-byte loop to find # of non-zero bytes for a uint256 in `_toBinary`, was thinking we could also use a binary search implementation here

kyzooghost: Seems to me that for `_toBinary` we do a linear search starting from the MSB, whereas here we do a do a linear search starting from the LSB.

Either way, we are doing the same query - for a uint256, find minimum # of bytes required to represent it - so this operation could be refactored into a separate function

contracts/src/libraries/RlpWriter.sol
thedarkjester: done locally - will be in the next commit

contracts/src/libraries/RlpWriter.sol
```diff
@@ -0,0 +1,197 @@
...
+  /// @return flattenedBytes The flattened byte string.
+  function _flatten(bytes[] memory _bytesList) private pure returns (bytes memory flattenedBytes) {
+    uint256 bytesListLength = _bytesList.length;
+    if (bytesListLength == 0) {
+      return new bytes(0);
+    }
+
+    uint256 flattenedBytesLength;
+    uint256 reusableCounter;
+    for (; reusableCounter < bytesListLength; reusableCounter++) {
```
VGau: Did you try to put the `reusableCounter++` in the unchecked section  ?

thedarkjester: Yeah, the compiler already takes care of that and the gas cost is the same

PR #820, opened 2025-03-28
feat: delete all aggregation records from last aggregated block plus o…

jonesho:
This PR implements issue(s) #824 

We recently had an incident on mainnet that after we increased the aggregation proof limit from 100 to 200, the finalization stuck. It was because, upon the coordinator's restart, the finalized block on L1 was `17410422` and it produced an aggregation proof of `17410423..17415478` but in the meantime the coordinator already submitted the old aggregation `17410423..17412896` on L1, and hence it would look for the next aggregation from `17412897` onwards which would never be found as the next aggregation in db was `17415479..17420546`

The cause was that the coordinator only removes aggregation records for "start block number >= last processed block + 1" and resumes aggregation from "last finalized block + 1", but indeed it should **remove aggregation records with "start block number >= highest consecutive aggregated block + 1" and resume aggregation from "highest consecutive aggregated block + 1" upon each coordinator's restart.**

### Checklist


Filter94:
I checked the logs and I see that at the time of Coordinator startup, last finalized block height was `17410422`. This means that with your solution we'd delete the sent aggregation and would create a new one on from `17410422`, yet `17410423..17412896` would be finalized anyway.

Since `17410423..17412896` was submitted , it means that this aggregation was already in DB. But on Coordinator startup we create `ProofAggregationCoordinatorService` which starts aggregation from `lastFinalizedBlock`. And I think this is the issue that has to be solved.

So the solution would be to continue from the last available proven aggregation from the DB rather than a finalized one

jonesho:
> I checked the logs and I see that at the time of Coordinator startup, last finalized block height was `17410422`. This means that with your solution we'd delete the sent aggregation and would create a new one on from `17410422`, yet `17410423..17412896` would be finalized anyway. It would lead to a corrupted DB because we wouldn't have the actual finalized proof in the DB.
> 
> Since `17410423..17412896` was submitted , it means that this aggregation was already in DB. But on Coordinator startup we create `ProofAggregationCoordinatorService` which starts aggregation from `lastFinalizedBlock`. And I think this is the issue that has to be solved.

@Filter94 I see your point, yes you're right, there would be a risk that we delete the aggregation records that were submitted but not yet finalized on L1. 

> So the solution would be to continue from the last available proven aggregation from the DB rather than a finalized one

A more complete solution would be finding the aggregation record with the highest consecutive batches and blobs, and remove all the subsequent records and resume aggregation from the record's "end block number + 1" upon restart, wdyt?

Filter94:
> finding the aggregation record with the highest consecutive batches and blobs

It's impossible to have a gap in aggregations and if we have an aggregation, then it doesn't have any gaps in batches and nonces, that's why I just said "last proven". And I think that's enough

jonesho:
> > finding the aggregation record with the highest consecutive batches and blobs
> 
> It's impossible to have a gap in aggregations and if we have an aggregation, then it doesn't have any gaps in batches and blobs, that's why I just said "last proven". And I think that's enough

I agree there should be no batch/blob gap in each aggregation (unless db corruption) but I think its still possible to have a gap in aggregations, given that only successfully generated proofs would be written into the db, and some proofs could be still in progress (or even failed) before the restart, and hence I think we should resume from the highest consecutive proven record

Filter94:
@jonesho I see. Theoretically it's possible yes, so I guess you're right from correctness PoV. Economically, we're potentially losing all the aggregation proofs starting from the gapped one. Could've been worse, but it's not a negligible impact either, so I'm not 100% certain on the solution here

jonesho:
> @jonesho I see. Theoretically it's possible yes, so I guess you're right from correctness PoV. Economically, we're potentially losing all the aggregation proofs starting from the gapped one. Could've been worse, but it's not a negligible impact either, so I'm not 100% certain on the solution here

Economically speaking, we're not losing all the prover resource spent on aggregation proofs starting from the gapped one, because the coordinator would check if the proof response was in the shared folder before we submit the request, the only thing we would spend again is couple RPC calls to get the L2 data, so I guess it should be fine. On the other hand, if we resume the aggregation after the highest proven record, the gaps would never be filled, which could be an issue.

coordinator/app/src/test/kotlin/net/consensys/zkevm/coordinator/app/L1DependentAppTest.kt
```diff
@@ -35,6 +35,7 @@ class L1DependentAppTest {
     val blobsRepository = mock<BlobsRepository>()
     val aggregationsRepository = mock<AggregationsRepository>()
     val lastProcessedBlock = 100uL
+    val lastConsecutiveAggregatedBlockNumber = 80uL
```
Filter94, line 38: IMO this test is useless because it's more complicated than the implementation

coordinator/persistence/aggregation/src/integrationTest/kotlin/net/consensys/zkevm/persistence/dao/aggregation/AggregationsPostgresDaoTest.kt
```diff
@@ -469,6 +469,107 @@ class AggregationsPostgresDaoTest : CleanDbTestSuiteParallel() {
...
+
+    val aggregationProof4 = createProofToFinalize(
+      firstBlockNumber = 40,
+      finalBlockNumber = 50,
+      parentAggregationLastBlockTimestamp = Instant.parse("2024-04-28T15:01:00Z"),
+      startBlockTime = Instant.fromEpochSeconds(0),
+      finalTimestamp = Instant.parse("2024-04-28T15:02:00Z")
+    )
+
+    val aggregation4 = Aggregation(
```
Filter94: From what I noticed, start and end block numbers are sometimes the same between `aggregationProof` and `Aggregation`. It would be easier to read the test if we had a method that is generating an aggregation, inserting it into a db and returning an aggregation proof object

coordinator/persistence/aggregation/src/integrationTest/kotlin/net/consensys/zkevm/persistence/dao/aggregation/AggregationsPostgresDaoTest.kt
```diff
@@ -469,6 +469,107 @@ class AggregationsPostgresDaoTest : CleanDbTestSuiteParallel() {
...
+      startBlockNumber = 40UL,
+      endBlockNumber = 50UL,
+      batchCount = 5UL,
+      aggregationProof = aggregationProof4
+    )
+    performInsertTest(aggregation4)
```
Filter94: We don't really need any verifications here. Since insertions aren't the focus of these tests we can only do the insert here

coordinator/persistence/aggregation/src/integrationTest/kotlin/net/consensys/zkevm/persistence/dao/aggregation/AggregationsPostgresDaoTest.kt
```diff
@@ -469,6 +469,107 @@ class AggregationsPostgresDaoTest : CleanDbTestSuiteParallel() {
...
+    val aggregationProof3 = createProofToFinalize(
+      firstBlockNumber = 21,
+      finalBlockNumber = 39,
+      parentAggregationLastBlockTimestamp = Instant.parse("2024-04-28T15:01:00Z"),
+      startBlockTime = Instant.fromEpochSeconds(0),
+      finalTimestamp = Instant.parse("2024-04-28T15:02:00Z")
+    )
+
+    val aggregation3 = Aggregation(
+      startBlockNumber = 31UL,
```
Filter94, line 486: Is this intentional that `startBlockNumber` is 31, while the proof has 21? Would be nice to be more explicit on it, maybe using variable names to hint the reader what is the scenario here

coordinator/persistence/aggregation/src/main/kotlin/net/consensys/zkevm/persistence/dao/aggregation/PostgresAggregationsDao.kt
```diff
@@ -296,6 +294,27 @@ class PostgresAggregationsDao(
...
+        Tuple.of(
+          fromBlockNumber,
+          aggregationStatusToDbValue(Aggregation.Status.Proven),
+          Int.MAX_VALUE
+        )
+      )
+      .toSafeFuture()
+      .thenApply { rowSet ->
+        if (rowSet.size() > 0) {
+          rowSet.last().getLong("end_block_number")
```
Filter94: Using `lastOrNull()` we can remove this if and simplify this significantly

coordinator/app/src/test/kotlin/net/consensys/zkevm/coordinator/app/L1DependentAppTest.kt
jonesho: I'd rather keep this as it reflects how the delete record functionss of the batch, blob, aggregation repositories are called.

coordinator/persistence/aggregation/src/integrationTest/kotlin/net/consensys/zkevm/persistence/dao/aggregation/AggregationsPostgresDaoTest.kt
jonesho: Agree, I'd even remove the creation of the aggregation proofs as they are not relevant to the test.

coordinator/persistence/aggregation/src/main/kotlin/net/consensys/zkevm/persistence/dao/aggregation/PostgresAggregationsDao.kt
jonesho: Also revised that in other usages.

coordinator/persistence/aggregation/src/integrationTest/kotlin/net/consensys/zkevm/persistence/dao/aggregation/AggregationsPostgresDaoTest.kt
```diff
@@ -469,6 +469,65 @@ class AggregationsPostgresDaoTest : CleanDbTestSuiteParallel() {
...
+      21L
+    ).get().also { highestEndBlockNumber ->
+      assertThat(highestEndBlockNumber).isNull()
+    }
+
+    // should return 50L as aggregations of blocks 31..39 and blocks 40..50 exist in db
+    aggregationsPostgresDaoImpl.findHighestConsecutiveEndBlockNumber(
+      31L
+    ).get().also { highestEndBlockNumber ->
+      assertThat(highestEndBlockNumber).isEqualTo(50L)
```
Filter94, line 520: Minor: instead of values like 50L, we can reference the `aggregation3.endBlockNumber`

coordinator/persistence/aggregation/src/integrationTest/kotlin/net/consensys/zkevm/persistence/dao/aggregation/AggregationsPostgresDaoTest.kt
```diff
@@ -469,6 +469,65 @@ class AggregationsPostgresDaoTest : CleanDbTestSuiteParallel() {
...
+  @Test
+  fun findHighestConsecutiveEndBlockNumber_returns_highest_consecutive_end_block_number_from_various_block_numbers() {
+    // insert aggregation of blocks 1..20 in db
+    val aggregation1 = Aggregation(
+      startBlockNumber = 1UL,
+      endBlockNumber = 20UL,
+      batchCount = 10UL,
+      aggregationProof = null // proof can be omitted here as not the focus of the test
+    )
+    aggregationsPostgresDaoImpl.saveNewAggregation(aggregation1).get()
```
Filter94, line 481: Minor: We're copypasting this creation and insertion for the same purpose multiple times. IMO it would be nice to reflect the similarity of their purpose by factoring them out into a function.

Filter94 (approved):

coordinator/persistence/aggregation/src/integrationTest/kotlin/net/consensys/zkevm/persistence/dao/aggregation/AggregationsPostgresDaoTest.kt
jonesho: This is intentional, imo its more readable this way for test cases.

PR #826, opened 2025-03-31
fix(state-manager): refactoring and removing debugging code.

bogdanbear:
This PR refactors the state-manager connector code between the HUB and Shomei traces, to improve readability.


AlexandreBelling (approved):
Approved. As a general comment, it's better to double-check what we commit to avoid adding debugging code :)

gusiri (approved):

PR #831, opened 2025-04-01
coordinator: update Shomei frontend at LATEST finalization

fluentcrafter:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/config/coordinator/coordinator-local-dev.config-traces-v2.overrides.toml
+++ b/config/coordinator/coordinator-local-dev.config-traces-v2.overrides.toml
@@ -23,3 +23,11 @@ endpoints=["http://127.0.0.1:8745"]
 
 [traces.conflation-v2]
 endpoints=["http://127.0.0.1:8745"]
+
+[type2-state-proof-provider]
+endpoints=[]
+
+[l2-network-gas-pricing.json-rpc-pricing-propagation]
+disabled=true
+geth-gas-price-update-recipients=[]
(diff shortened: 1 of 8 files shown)


config/coordinator/coordinator-local-dev.config-traces-v2.overrides.toml
```diff
@@ -23,3 +23,11 @@ endpoints=["http://127.0.0.1:8745"]
...
 [traces.conflation-v2]
 endpoints=["http://127.0.0.1:8745"]
+
+[type2-state-proof-provider]
+endpoints=[]
+
+[l2-network-gas-pricing.json-rpc-pricing-propagation]
+disabled=true
+geth-gas-price-update-recipients=[]
+besu-gas-price-update-recipients=["http://127.0.0.1:9045"]
```
jonesho, line 33: If `json-rpc-pricing-propagation` is disabled, why do we need to specify endpoints in `besu-gas-price-update-recipients`? shouldn't we just make it empty to avoid confusion?

alainncls (approved):

config/coordinator/coordinator-local-dev.config-traces-v2.overrides.toml
fluentcrafter: That's a fair point. Just left it there because tracing V2 uses Besu and in case someone want so enable it locally is handy.

Issue #804, opened 2025-03-24
Bridge UI V2 Refactor - remove ! anti-pattern


kyzooghost:
## Description
For Bridge UI V2, we would like to remove the `!` TypeScript antipattern for checking for null or undefined value. And replace with a more explicit check.

## Motivation
`!` checks for all falsy values in JS, not just undefined or null.

## Tasks
- [ ] Remove `!` antipattern for checking for null or undefined values

## Acceptance criteria
- [ ] CI passing

## Risks
- [ ] Break existing functionality

## Remember to
 - [x] Add the `documentation` label in case there is an impact on the documentation
 - [x] Add `priority` and `team` labels
 - [x] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

RasenGUY:
anyone working on this ?

PR #839, opened 2025-04-04
Fix/non specific !

kyzooghost:
This PR implements issue(s) #804 

### Checklist

### DIFF
--- a/bridge-ui/src/services/cctp.ts
+++ b/bridge-ui/src/services/cctp.ts
@@ -13,7 +13,7 @@ export async function fetchCctpAttestationByTxHash(
       },
     },
   );
-  if (response.ok === false) {
+  if (!response.ok) {
     throw new Error(
       `Error in fetchCctpAttestationByTxHash: isTestnet=${isTestnet} cctpDomain=${cctpDomain} transactionHash=${transactionHash}`,
     );
@@ -35,7 +35,7 @@ export async function fetchCctpAttestationByNonce(
       },
     },
   );
-  if (response.ok === false) {
+  if (!response.ok) {
     throw new Error(
       `Error in fetchCctpAttestationByNonce: isTestnet=${isTestnet} cctpDomain=${cctpDomain} nonce=${nonce}`,
     );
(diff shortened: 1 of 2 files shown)


bridge-ui/src/hooks/fees/useFees.ts
```diff
@@ -41,7 +42,7 @@ const useFees = () => {
     (fee: bigint) => {
       const zeroAddrLower = zeroAddress.toLowerCase();
       const price = tokenPrices?.[zeroAddrLower];
-      if (!price || price <= 0) return null;
+      if (isZero(price) || price <= 0) return null;
```
VGau: the price can be undefined as well :)

bridge-ui/src/hooks/transaction-args/cctp/useDepositForBurnTxArgs.ts
```diff
@@ -21,7 +21,14 @@ const useDepositForBurnTxArgs = ({ allowance }: UseDepositForBurnTxArgs) => {
   const fee = useCctpFee();
 
   return useMemo(() => {
-    if (!address || isNull(amount) || isUndefined(allowance) || allowance < amount || !recipient || !isCctp(token)) {
+    if (
+      isUndefinedOrEmptyString(address) ||
```
VGau: the address can be empty string if I am not wrong

kyzooghost: I had a closer look at this file and concluded that we don't need the `address` variable

I removed it and watched the E2E test in Github CI for USDC transfer - still works :)

bridge-ui/src/services/cctp.ts
```diff
@@ -13,7 +13,7 @@ export async function fetchCctpAttestationByTxHash(
       },
     },
   );
-  if (!response.ok) {
+  if (response.ok === false) {
```
VGau: There is not need to change the condition here :) `!response.ok` is fine
Same comment for similar conditions

kyzooghost: fixed

PR #852, opened 2025-04-08
Prover/rm closure rebase latest

srinathln7:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/prover/protocol/compiler/dummy/dummy.go
+++ b/prover/protocol/compiler/dummy/dummy.go
@@ -6,18 +6,13 @@ import (
 
 	"github.com/consensys/gnark/frontend"
 	"github.com/consensys/linea-monorepo/prover/protocol/column"
+	"github.com/consensys/linea-monorepo/prover/protocol/ifaces"
 	"github.com/consensys/linea-monorepo/prover/protocol/wizard"
 	"github.com/consensys/linea-monorepo/prover/utils"
 	"github.com/consensys/linea-monorepo/prover/utils/parallel"
 	"github.com/sirupsen/logrus"
 )
 
-/*
-Converts all the oracle commmitments into messages
-and ask the verifier to manually verify all queries.
-
-Primary use-case is testing.
-*/
 func Compile(comp *wizard.CompiledIOP) {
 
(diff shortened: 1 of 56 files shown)


prover/protocol/compiler/dummy/dummy.go
```diff
@@ -6,18 +6,13 @@ import (
...
 	"github.com/consensys/gnark/frontend"
 	"github.com/consensys/linea-monorepo/prover/protocol/column"
+	"github.com/consensys/linea-monorepo/prover/protocol/ifaces"
 	"github.com/consensys/linea-monorepo/prover/protocol/wizard"
 	"github.com/consensys/linea-monorepo/prover/utils"
 	"github.com/consensys/linea-monorepo/prover/utils/parallel"
 	"github.com/sirupsen/logrus"
 )
 
-/*
```
AlexandreBelling, line 15: I get you wanted to remove the doc because it's bad but replace it by something better if you do lol

prover/protocol/compiler/stitchsplit/splitter.go
```diff
@@ -44,116 +44,100 @@ func newSplitter(comp *wizard.CompiledIOP, size int) splitterContext {
...
 }
 
+type proveRoundProverAction struct {
+	ctx   *splitterContext
+	round int
+}
+
+func (a *proveRoundProverAction) Run(run *wizard.ProverRuntime) {
+	// Invoke the closure returned by Prove
+	proveFunc := a.ctx.Prove(a.round)
```
AlexandreBelling: I believe it would be simpler and cleaner to turn 'ctx.Prove' into the interface you are running here instead of calling it like that. The reason you had issues (I believe), is that we used to instantiate `ctx.Prove(run)` directly in the [ScanSplitCommit] (which is run at compilation time) and now it is being run in the [Run] function, done at prover runtime. This is difference is very likely what caused the issue.

Doing so, would  both simplify the code and solve the problem you tried to solve.

srinathln7: This is done.

AlexandreBelling (approved):

PR #853, opened 2025-04-08
coordinator: adds config v2 sample

fluentcrafter:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/config/coordinator/coordinator-config-v2-sample.toml
+++ b/config/coordinator/coordinator-config-v2-sample.toml
@@ -1,0 +1,302 @@
+testL1Disabled = false
+
+duplicated-logs-debounce-time = "PT15S"
+
+
+[protocol]
+[protocol.genesis]
+genesis-state-root-hash = "0x072ead6777750dc20232d1cee8dc9a395c2d350df4bbaa5096c6f59b214dcecd"
+# shnarf for contract V5
+# Keccak256(parentShnarf="0x00...00", snarkHash="0x00...00",
+#   parentStateRootHash="0x072ead6777750dc20232d1cee8dc9a395c2d350df4bbaa5096c6f59b214dcecd",
+#   evaludationClaim="0x00...00", evaludationPoint="0x00...00")
+genesis-shnarf-v5 = "0x47452a1b9ebadfe02bdd02f580fa1eba17680d57eec968a591644d05d78ee84f"
+[protocol.l1]
+contract-address = "0xCf7Ed3AccA5a467e9e704C703E8D87F634fB0Fc9"
+[protocol.l2]
+contract-address = "0xe537D669CA013d86EBeF1D64e40fC74CADC91987"
+# Number of children blocks to wait before considering a block "finalized"
+# and elegible for conflation and
+# this a workaround to mitigate Geth fork issues with Clique PoA
+# Coordinator will consider block as finalized after being included in the chain wtih children blocks-to-finalization
+# Recommended minimum of 2
+blocks-to-finalization = 2 # Used in MessageAnchoring and SafeBlockProvider
+
+
+[conflation]
+blocks-limit = 2
+conflation-calculator-version = "1.0.0"
+conflation-deadline = "PT6S" # =3*l2_block_time
+conflation-deadline-check-interval = "PT3S"
+conflation-deadline-last-block-confirmation-delay = "PT2S" # recommended: at least 2 * blockInterval
+
+# This is to prevent inflight trasactions that may change Smart contract state while coordinator is restarted.
+# Queries SMC for last finalised block, and keeps polling until this number of blocks observe the same state.
+# If state is updated meanwhile, it resets counter and restarts the polling.
+consistent-number-of-blocks-on-l1-to-wait = 1
+fetch-blocks-limit = 4000
+
+
+[blob-compression]
+blob-size-limit = 102400 # 100KB
+handler-polling-interval = "PT1S"
+# default batches limit is  aggregation-proofs-limit -1
+# batches-limit must be less than or equal to aggregation-proofs-limit-1
+batches-limit = 1
+
+
+[proof-aggregation]
+aggregation-proofs-limit = 3
+aggregation-deadline = "PT1M"
+aggregation-coordinator-polling-interval = "PT2S"
+deadline-check-interval = "PT8S"
+target-end-blocks = []
+
+
+[prover]
+version = "v2.0.0"
+fs-inprogess-request-writing-suffix = ".inprogress_coordinator_writing"
+fs-inprogess-proving-suffix-pattern = "\\.inprogress\\.prover.*"
+fs-polling-interval = "PT1S"
+fs-polling-timeout = "PT10M"
+[prover.execution]
+fs-requests-directory = "/data/prover-execution/requests"
+fs-responses-directory = "/data/prover-execution/responses"
+[prover.blob-compression]
+fs-requests-directory = "/data/prover-compression/requests"
+fs-responses-directory = "/data/prover-compression/responses"
+[prover.proof-aggregation]
+fs-requests-directory = "/data/prover-aggregation/requests"
+fs-responses-directory = "/data/prover-aggregation/responses"
+
+
+[zk-geth-traces]
+eth-api = "http://traces-node:8545"
+new-block-polling-interval = "PT1S"
+
+[traces]
+raw-execution-traces-version = "0.2.0"
+expected-traces-api-version = "0.2.0"
+[traces.counters]
+endpoints = ["http://traces-api:8080/"]
+request-limit-per-endpoint = 20
+[traces.counters.request-retries]
+max-attempts = 4
+backoff-delay = "PT1S"
+failures-warning-threshold = 2
+
+[traces.conflation]
+endpoints = ["http://traces-api:8080/"]
+request-limit-per-endpoint = 2
+[traces.conflation.request-retries]
+max-attempts = 4
+backoff-delay = "PT1S"
+failures-warning-threshold = 2
+
+[traces.file-manager]
+traces-file-extension = "json.gz"
+raw-traces-directory = "/data/traces/raw"
+non-canonical-raw-traces-directory = "/data/traces/raw-non-canonical"
+create-non-canonical-directory = true
+polling-interval = "PT1S"
+traces-file-creation-wait-timeout = "PT2M"
+
+[state-manager]
+version = "2.2.0"
+endpoints = ["http://shomei:8888/"]
+request-limit-per-endpoint = 3
+[state-manager.request-retries]
+max-attempts = 5
+backoff-delay = "PT2S"
+failures-warning-threshold = 2
+
+[type2-state-proof-provider]
+endpoints = ["http://shomei-frontend:8888/"]
+[type2-state-proof-provider.retries]
+max-attempts = 3
+backoff-delay = "PT1S"
+failures-warning-threshold = 2
+
+
+[l1-submission] # submission
+[l1-submission.dynamic-gas-price-cap]
+disabled = true
+[l1-submission.dynamic-gas-price-cap.gas-price-cap-calculation]
+adjustment-constant = 25
+blob-adjustment-constant = 25
+finalization-target-max-delay = "PT32H"
+base-fee-per-gas-percentile-window = "P7D"
+base-fee-per-gas-percentile-window-leeway = "PT10M"
+base-fee-per-gas-percentile = 10
+gas-price-caps-check-coefficient = 0.9
+[l1-submission.dynamic-gas-price-cap.fee-history-fetcher]
+fetch-interval = "PT1S"
+max-block-count = 1000
+reward-percentiles = [10, 20, 30, 40, 50, 60, 70, 80, 90, 100]
+[l1-submission.dynamic-gas-price-cap.fee-history-storage]
+storage-period = "P10D"
+
+[l1-submission.fallback-gas-price]
+fee-history-block-count = 10
+fee-history-reward-percentile = 15
+
+[l1-submission.blob]
+disabled = false
+rpc-endpoint = "http://l1-el-node:8545"
+submission-delay = "PT1S"
+submission-tick-interval = "PT1S"
+max-submissions-per-tick = 10
+db-max-blobs-to-return = 100
+[l1-submission.blob.gas]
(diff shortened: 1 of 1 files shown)


config/coordinator/coordinator-config-v2.toml
```diff
@@ -0,0 +1,322 @@
...
+[l1-submission.aggregation]
+disabled = false
+endpoint = "http://l1-el-node:8545"
+submission-delay = "PT1S"
+submission-tick-interval = "PT1S"
+max-submissions-per-tick = 10
+[l1-submission.aggregation.gas]
+gas-limit = 10_000_000
+max-fee-per-gas-cap = 100_000_000_000
+max-priority-fee-per-gas-cap = 10_000_000_001
```
jonesho: These two above, if follows the old config: `gas-price-cap-multiplier-for-finalization=2.0`, should be
```
max-fee-per-gas-cap = 200_000_000_000
max-priority-fee-per-gas-cap = 40_000_000_000
```

config/coordinator/coordinator-config-v2-override-local-dev.toml
Filter94, line 1: Can we remove the v2 from the name, since now it's supposed to be the only option?

Filter94: I can see that v2 is also in a package name. Is there an intention to remove it later or we just iterate with v3 and v4 later on?

config/coordinator/coordinator-config-v2.toml
```diff
@@ -0,0 +1,321 @@
+[defaults]
```
Filter94, line 1: I don't think introduction of defaults is a good idea. It introduces additional level of indirection and once it's ever defined, it will never go away, because with optional `l1Endpoint` and `l2Endpoint` inside feature specific configurations, it's impossible to tell from the configuration itself whether the default value is actually used anywhere or it being overriden everywhere. Explicit and non-optional specification of endpoints also helps users to understand what are the options to configure stuff and it can also help while investigating issues, by giving hints how and when given endpoint is used.

config/coordinator/coordinator-config-v2-override-local-dev.toml
fluentcrafter: Will do in a follow up clean up PR. It was just to make it obvious for now

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/DataBaseToml.kt
```diff
@@ -0,0 +1,36 @@
+package linea.coordinator.config.v2.toml
+
+import com.sksamuel.hoplite.Masked
+import linea.coordinator.config.v2.DatabaseConfig
+import kotlin.time.Duration.Companion.minutes
+import kotlin.time.Duration.Companion.seconds
+
+data class DataBaseToml(
```
Filter94: Let's stick to Database

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/L1SubmissionConfigToml.kt
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  data class BlobSubmissionConfigToml(
+    val disabled: Boolean = false,
+    val l1Endpoint: URL? = null,
+    val submissionDelay: Duration = 0.seconds,
+    val submissionTickInterval: Duration = 12.seconds,
+    val maxSubmissionTransactionsPerTick: UInt = 2u,
+    // eip-7691 on Prague fork allows up to 9 blobs per transaction.
+    // however, Geth nodes fail with "transaction too large" error. only 8 blobs are accepted
```
Filter94: Isn't it only 7?

coordinator/app/src/test/kotlin/linea/coordinator/config/v2/ApiConfigParsingTest.kt
```diff
@@ -0,0 +1,41 @@
...
+      observability-port = 9546
+    """.trimIndent()
+
+    val config = ApiConfigToml(
+      observabilityPort = 9546u,
+    )
+
+    val tomlMinimal = """
+
+    """.trimIndent()
```
Filter94:
```suggestion
    val tomlMinimal = ""
```

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/TracesToml.kt
```diff
@@ -0,0 +1,66 @@
...
+        endpoints = counters.endpoints,
+        requestLimitPerEndpoint = counters.requestLimitPerEndpoint,
+        requestRetries = counters.requestRetries.asDomain,
+      ),
+      conflation = TracesConfig.ClientApiConfig(
+        endpoints = conflation.endpoints,
+        requestLimitPerEndpoint = conflation.requestLimitPerEndpoint,
+        requestRetries = conflation.requestRetries.asDomain,
+      ),
+      /*
```
Filter94, line 44: I guess this will be removed in the follow-up PR?

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/SignerConfigToml.kt
```diff
@@ -0,0 +1,116 @@
...
+      result = 31 * result + endpoint.hashCode()
+      result = 31 * result + publicKey.contentHashCode()
+      return result
+    }
+  }
+
+  fun reified(): SignerConfig {
+    return SignerConfig(
+      type = type.reified(),
+      web3j = web3j?.let { SignerConfig.Web3jConfig(it.privateKey.value.decodeHex()) },
```
Filter94, line 105: Since we started to use decoders, we could use them to get rid of this structure of 2 optional values in favor of 1 non-optional

Filter94 (approved):

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/MessageAnchoringConfig.kt
```diff
@@ -0,0 +1,71 @@
...
+  ),
+  val anchoringTickInterval: Duration = 2.seconds,
+  val messageQueueCapacity: UInt = 10_000u,
+  val maxMessagesToAnchorPerL2Transaction: UInt = 100u,
+  val signer: SignerConfig,
+  val gas: GasConfig = GasConfig(),
+) : FeatureToggle {
+  init {
+    require(messageQueueCapacity >= 1u) {
+      "messageQueueCapacity=$messageQueueCapacity be equal or greater than 1"
```
Filter94: Copilot has a point
```suggestion
      "messageQueueCapacity=$messageQueueCapacity must be equal or greater than 1"
```

config/coordinator/coordinator-config-v2-override-local-dev.toml
fluentcrafter: >Is there an intention to remove it later or we just iterate with v3 and v4 later on?

That's the idea. This segregation makes the code reviews and reading easier IMO.

config/coordinator/coordinator-config-v2.toml
fluentcrafter: It's not clear to me: What are you concretely suggesting (please provide snippet) and why?

Filter94: I suggest to just remove defaults completely and specify l1 and l2 endpoints for every feature explicitly and mandatory

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/L1SubmissionConfigToml.kt
fluentcrafter: Yes, it is. Thanks for the call out.

config/coordinator/coordinator-config-v2.toml
fluentcrafter: That's overkilling. The purpose of this whole work was simplification. Having a single l1/l2 endpoint must work for the vast majority of cases. The others are there to tune them in **if necessary**. For this reason, I am keen in keeping it.

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/SignerConfigToml.kt
fluentcrafter: Maybe, but I don't think it's the high priority to invest time on it atm.

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/L1DependentApp.kt
```diff
@@ -167,101 +184,139 @@ class L1DependentApp(
...
+  }
+
+  private val l1FeeHistoriesRepository =
+    FeeHistoriesRepositoryImpl(
+      FeeHistoriesRepositoryImpl.Config(
+        rewardPercentiles = configs.l1Submission!!.dynamicGasPriceCap.feeHistoryFetcher
+          .rewardPercentiles.map { it.toDouble() },
+        // FIXME: there is not equivalent in the new config. Validate with Jones/Julien
+        minBaseFeePerBlobGasToCache = null,
+        fixedAverageRewardToCache = null,
```
fluentcrafter: @jonesho @julien-marchand this used to be instantiated in [PROD](https://github.com/Consensys/zk-apps-prod/blob/main/values/coordinator/mainnet/coordinator-config.yaml#L234-L235) with  
```
historic-base-fee-per-blob-gas-lower-bound=5000000000 # 5 Gwei
historic-avg-reward-constant=50000000 # 0.05 Gwei
```

@jonesho IIRC, when we were designing the new config structure, we dropped these configs; I don't recall the rationale... Can you please confirm they are not necessary?

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/TracesToml.kt
fluentcrafter: This was actually intended. Like the prover, I want to support new config. However, due to lack of time and other higher priorities I did not finish. So, left it here so when the time comes it can be reused and show the intent. It can be removed though If you think it's better.

Filter94: Up to you

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/L1DependentApp.kt
jonesho: @fluentcrafter I can confirm they are needed and should be placed under `l1-submission.dynamic-gas-price-cap.gas-price-cap-calculation`, they are used in part of dynamic gas pricing calculation

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/decoders/Decoders.kt
```diff
@@ -0,0 +1,83 @@
...
+      else -> { ConfigFailure.DecodeError(node, type).invalid() }
+    }
+  }
+
+  override fun supports(type: KType): Boolean {
+    return type.classifier == ByteArray::class
+  }
+}
+
+class TomlKotlinDurationDecoder : Decoder<Duration> {
```
Filter94, line 39: I don't see why would you need that. Hoplite supports things like `500m` out of the box. [Here](https://github.com/Consensys/maru/blob/main/config/src/test/kotlin/maru/config/HopliteFriendlinessTest.kt#L44) is an example in Maru

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/decoders/Decoders.kt
```diff
@@ -0,0 +1,83 @@
...
+import com.sksamuel.hoplite.StringNode
+import com.sksamuel.hoplite.decoder.Decoder
+import com.sksamuel.hoplite.fp.invalid
+import com.sksamuel.hoplite.fp.valid
+import linea.coordinator.config.v2.toml.SignerConfigToml
+import linea.kotlin.decodeHex
+import kotlin.reflect.KType
+import kotlin.time.Duration
+
+class TomlByteArrayHexDecoder : Decoder<ByteArray> {
```
Filter94, line 16: JFYI, I had an idea to write just the same in Maru, but probably it's not worthy extracting it for the reuse just for the sake of this single piece. But once there's more, we could do it in the future

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/decoders/Decoders.kt
fluentcrafter: What format/standard is that?

Current implementation relies on [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) duration format `PnDTnHnMn.nS` which is a well known/used format.

Filter94: https://github.com/sksamuel/hoplite/blob/master/README.md#duration-formats

I don't know if it's some kind of standard, but things like `500 milliseconds` make it very readable, which I prefer over ISO-8601

Filter94: Also, I don't know if things like "1 hour 30 minutes" are possible, but it can be worked around with `90 minutes`

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/L1DependentApp.kt
fluentcrafter: @jonesho configs added. Please take a look .

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/decoders/Decoders.kt
fluentcrafter: We are pegging ourselves to a specific framework/lib format, even though it may be more human-friendly, it brings a learning curve. People would need to read Holite documentation to know what and how is supported. 

So I will keep ISO format...

jonesho (approved):

PR #854, opened 2025-04-08
Message anchoring v2 part1

fluentcrafter:
Just a bit of TLC...


coordinator/ethereum/message-anchoring/build.gradle
```diff
@@ -3,19 +3,12 @@ plugins {
 }
 
 dependencies {
-  // TODO: review dependencies
-  implementation "io.vertx:vertx-core"
   implementation project(':jvm-libs:generic:vertx-helper')
```
jonesho, line 6: I think this `implementation project(':jvm-libs:generic:vertx-helper')` can be removed too, I was running [this](https://github.com/autonomousapps/dependency-analysis-gradle-plugin?tab=readme-ov-file) to check the unused dependencies, curious how did you get the list of removable dependencies?

fluentcrafter: I did it by "naked eye" and the run the build to check. 

The `implementation project(':jvm-libs:generic:vertx-helper')` is necessary, the build breaks without it.

Filter94 (approved):

Issue #843, opened 2025-04-04
Add modals for first time users to bridge.linea.build


nkkurt:
*Hey, the Growth Marketing Intake Typeform got a new response and a new ticket has been created:*
----

### Please enter your email address

<mailto:katie.wilkinson-ext@consensys.net|katie.wilkinson-ext@consensys.net>

### Due Date?

2025-04-11

### Which category best suits your request?

Linea

### What do you need help with?

Website

### What do you need help with?

Existing Webpage Update

### Existing Webpage URL

<https://bridge.linea.build/>

### Figma URL

<https://www.figma.com/design/5vnmyzpeOeO8ZjiHuqc3qk/Bridge-UI-Design?node-id=2001-697>

### Doc URL

n/a

### Describe the purpose/goal for updating the existing webpage

Add three modals to the Linea bridge that display only for first time users that explains how the bridge works. One displays when new users visit the "All Bridges" tab, another displays when new users visit the "Native Bridge" tab, and the last one when users visit the "Buy" tab

### Stakeholders?

Matthieu Bouchaud

### Is there anything more you'd like to share? 

You're welcome to provide additional details or ask any further questions.

n/a 

----

<@U085KLDF0FJ> *Here's the link to the Jira Ticket that has been created:*

<https://consensyssoftware.atlassian.net/browse/GMT-2851> 

----

[Slack Message](https://consensys.slack.com/archives/C04R631UMM0/p1743689594923909?thread_ts=1743689594.923909&cid=C04R631UMM0)

viphan007:
Hi @nkkurt, @MatthieuBouchaud,

I have created a PR for this ticket here: https://github.com/Consensys/linea-monorepo/pull/859

nkkurt:
Thanks @viphan007 . I see that the PR is now merged. Do you have any remaining work before we can close this ticket?

nkkurt:
The change is live and looks good - closing the ticket. Please feel free to reopen if needed.

PR #859, opened 2025-04-10
feat: add modal when user first visit

viphan007:
https://github.com/Consensys/linea-monorepo/issues/843


bridge-ui/src/components/modal/first-time-visit/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+  useEffect(() => {
+    if (!modalType) return;
+
+    const key = `hasVisited-${modalType}`;
+    if (localStorage.getItem(key) !== "true") {
+      setShowModal(true);
+      setShouldRenderModal(true);
+    }
+  }, [modalType]);
```
VGau: could you please use Zustand to store value instead of accessing the localStorage directly ? Add additional element to the ConfigState here: `bridge-ui/src/stores/configStore.ts`

VGau: to access the storage you can use the `useConfigStore` hook.
Here is a example to get the currency value:
`  const currency = useConfigStore((state) => state.currency);`

viphan007: I have updated, could you check it again?

viphan007: I have updated, could you check it again? @VGau

VGau (approved):

Issue #832, opened 2025-04-02
Expose metrics in the Postman


VGau:
## Description
We need to expose internal API metrics using the `prom-client` (Prometheus client for Node.js) in the Postman service. This will allow us to monitor performance, usage patterns, and identify bottlenecks or anomalies in real time.

The metrics should be exposed at a `/metrics` endpoint.

## Motivation
- Enables real-time observability of API usage and performance
- Allows integration with Grafana dashboards and alerting
- Helps identify slow endpoints, error rates, and traffic spikes

## Tasks
- [ ]  Install and configure prom-client in the Postman Node.js service
- [ ]  Add default metrics collection (e.g., memory usage, CPU, ....)
- [ ]  Expose a /metrics endpoint compatible with Prometheus
- [ ]  Add tests to validate that metrics are correctly emitted and formatted

## Acceptance criteria
- [ ] /metrics endpoint is live and returns Prometheus-formatted metrics
- [ ] Tests cover expected metric values and structure

PR #871, opened 2025-04-14
Feat: Add metrics to postman

VGau:
This PR implements issue(s) #832 

### Checklist


postman/src/services/processors/MessageClaimingProcessor.ts
```diff
@@ -225,15 +225,17 @@ export class MessageClaimingProcessor implements IMessageClaimingProcessor {
     maxFeePerGas: bigint,
   ): Promise<boolean> {
     if (isUnderPriced) {
```
kyzooghost, line 238: nit - `if (isUnderPriced && message.status !== MessageStatus.FEE_UNDERPRICED)` for less lines

postman/src/application/postman/api/metrics/MetricsService.ts
```diff
@@ -0,0 +1,147 @@
...
+    if (!counter) {
+      return undefined;
+    }
+
+    return (
+      (await counter.get()).values.find((value) => {
+        return Object.entries(labels).every(([key, val]) => value.labels[key] === val);
+      })?.value || 0
+    );
+  }
```
kyzooghost, line 62: Should we improve the readability here?

E.g. 

```ts
  private findMetricValueWithExactMatchingLabels(
    metricData: MetricObjectWithValues<MetricValue<string>>,
    labels: Record<string, string>,
  ): MetricValue<string> | undefined {
    return metricData.values.find((value) => Object.entries(labels).every(([key, val]) => value.labels[key] === val));
  }

  public async getCounterValue(name: string, labels: Record<string, string>): Promise<number | undefined> {
    const counter = this.counters.get(name);
    if (!counter) {
      return undefined;
    }

    const metricData = await counter.get();
    const metricValueWithMatchingLabels = this.findMetricValueWithExactMatchingLabels(metricData, labels);
    return metricValueWithMatchingLabels?.value || undefined;
  }
```

kyzooghost: Also wondering if we should return `undefined` as the fallback value rather than `0`

0 could be a valid metric value, but we wouldn't know if it represents the actual metric value or that no appropriate metric value was found

postman/src/application/postman/api/metrics/MetricsService.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  /**
+   * Creates histogram metric
+   */
+  public createHistogram(
+    name: string,
+    help: string,
+    labelNames: string[] = [],
+    buckets: number[] = [0.1, 0.5, 1, 2, 5, 10],
```
kyzooghost: Wondering if these default values for the histogram buckets serve a practical use for us, or are arbitrary values?

postman/src/application/postman/api/metrics/MessageMetricsService.ts
```diff
@@ -0,0 +1,60 @@
...
+  private async getMessagesCountByStatus(): Promise<{ status: MessageStatus; direction: Direction; count: number }[]> {
+    const totalNumberOfMessagesByStatusAndDirection = await this.entityManager
+      .createQueryBuilder(MessageEntity, "message")
+      .select("message.status", "status")
+      .addSelect("message.direction", "direction")
+      .addSelect("COUNT(message.id)", "count")
+      .groupBy("message.status")
+      .addGroupBy("message.direction")
+      .getRawMany();
+
```
kyzooghost, line 33:
```suggestion
// MessageStatus => MessageDirection => Count
```

To clarify the data structure :)

postman/src/application/postman/api/metrics/MessageMetricsService.ts
```diff
@@ -0,0 +1,60 @@
...
+    super();
+    this.createGauge("postman_messages_current", "Current number of messages by status", ["status", "direction"]);
+  }
+
+  public async initialize(): Promise<void> {
+    const fullResult = await this.getMessagesCountByStatus();
+    this.updateGauges(fullResult);
+  }
+
+  private async getMessagesCountByStatus(): Promise<{ status: MessageStatus; direction: Direction; count: number }[]> {
```
kyzooghost:
```suggestion
  private async getMessagesCountFromDatabase(): Promise<{ status: MessageStatus; direction: Direction; count: number }[]> {
```

postman/src/application/postman/api/__tests__/Api.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  });
+
+  it("should initialize the API", () => {
+    expect(api).toBeDefined();
+  });
+
+  it("should return metrics from the metric service", async () => {
+    await api.start();
+
+    const registry = api["metricsService"].getRegistry();
```
kyzooghost, line 32: Should we be be accessing private fields in the unit test? Could we use the public API in the unit test - GET request to the `/metrics` endpoint?

postman/src/application/postman/api/metrics/MessageMetricsService.ts
```diff
@@ -0,0 +1,60 @@
...
+          direction,
+          count: resultMap.get(status)?.get(direction) || 0,
+        });
+      }
+    }
+
+    return results;
+  }
+
+  private updateGauges(fullResult: { status: MessageStatus; direction: Direction; count: number }[]): void {
```
kyzooghost:
```suggestion
  private initializeGaugeValues(fullResult: { status: MessageStatus; direction: Direction; count: number }[]): void {
```

Since it seems that we update gauge values in the MessageStatusSubscriber class

postman/src/application/postman/persistence/subscribers/MessageStatusSubscriber.ts
```diff
@@ -0,0 +1,165 @@
...
+      }
+    }
+  }
+
+  private async updateMessageMetricsOnInsert(event: InsertEvent<MessageEntity>): Promise<void> {
+    try {
+      const messageStatus = event.entity.status;
+      const messageDirection = event.entity.direction;
+
+      const gauge = await this.metricsService.getGaugeValue("postman_messages_current", {
```
kyzooghost:
```suggestion
      const prevGaugeValue = await this.metricsService.getGaugeValue("postman_messages_current", {
```

postman/src/application/postman/persistence/subscribers/MessageStatusSubscriber.ts
```diff
@@ -0,0 +1,165 @@
...
+        status: messageStatus,
+        direction: messageDirection,
+      });
+
+      if (!gauge && gauge !== 0) {
```
kyzooghost:
```suggestion
      if (prevGaugeValue === undefined) {
```

kyzooghost: Also we would hit this case when the gauge with the matching labels don't exist, so it happen for `messageStatus` and `messageDirection` that we are not tracking?

postman/src/application/postman/persistence/subscribers/MessageStatusSubscriber.ts
```diff
@@ -0,0 +1,165 @@
...
+      this.logger.error("[MessageStatusSubscriber] Failed to update metrics:", error);
+    }
+  }
+
+  private async updateMessageMetricsOnRemove(event: RemoveEvent<MessageEntity>): Promise<void> {
+    try {
+      const messageStatus = event.databaseEntity.status;
+      const messageDirection = event.databaseEntity.direction;
+
+      const gauge = await this.metricsService.getGaugeValue("postman_messages_current", {
```
kyzooghost:
```suggestion
      const prevGaugeValue = await this.metricsService.getGaugeValue("postman_messages_current", {
```

postman/src/application/postman/persistence/subscribers/MessageStatusSubscriber.ts
```diff
@@ -0,0 +1,165 @@
...
+        status: messageStatus,
+        direction: messageDirection,
+      });
+
+      if (!gauge && gauge !== 0) {
```
kyzooghost:
```suggestion
      if (prevGaugeValue === undefined) {
```

postman/src/application/postman/persistence/subscribers/MessageStatusSubscriber.ts
```diff
@@ -0,0 +1,165 @@
...
+        }),
+      ]);
+
+      if (!previousGauge && previousGauge !== 0) {
+        return;
+      }
+
+      if (!gauge && gauge !== 0) {
+        return;
+      }
```
kyzooghost:
```suggestion
      if (previousGauge === undefined || gauge === undefined) {
        return;
      }
```

postman/src/application/postman/persistence/subscribers/MessageStatusSubscriber.ts
```diff
@@ -0,0 +1,165 @@
...
+    }
+  }
+
+  async afterRemove(event: RemoveEvent<MessageEntity>): Promise<void> {
+    if (event.entity) {
+      await this.updateMessageMetricsOnRemove(event);
+    }
+  }
+
+  async afterTransactionCommit(event: TransactionCommitEvent): Promise<void> {
```
kyzooghost, line 53: We use similar logic in this function as in `updateMessageMetricsOnUpdate`, should we use `updateMessageMetricsOnUpdate` in this function?

postman/scripts/runPostman.ts
```diff
@@ -145,8 +145,11 @@ async function main() {
         ? parseInt(process.env.DB_DAYS_BEFORE_NOW_TO_DELETE)
         : undefined,
     },
+    apiOptions: {
+      port: process.env.API_PORT ? parseInt(process.env.API_PORT) : undefined,
```
kyzooghost, line 157: Should we add `API_PORT` variable to `docker/config/postman/env` for the published Postman image?

postman/src/application/postman/persistence/repositories/TypeOrmMessageRepository.ts
```diff
@@ -175,15 +172,14 @@ export class TypeOrmMessageRepository<TransactionResponse extends ContractTransa
...
         )
         .andWhere(
-          "CAST(message.fee AS numeric) > :minimumMargin * ((:extraDataVariableCost * message.compressedTransactionSize) / message.claimTxGasLimit + :extraDataFixedCost) * message.claimTxGasLimit",
-          {
-            minimumMargin: feeEstimationOptions.minimumMargin,
-            extraDataVariableCost: feeEstimationOptions.extraDataVariableCost,
-            extraDataFixedCost: feeEstimationOptions.extraDataFixedCost,
-          },
+          new Brackets((qb) => {
+            qb.where("message.claimGasEstimationThreshold > :threshold", {
```
kyzooghost: Will this WHERE clause imposing a minimum threshold on `claimGasEstimationThreshold = messageFee / gasLimit` conflict with Postman message sponsoring?

postman/scripts/runPostman.ts
```diff
@@ -145,8 +145,11 @@ async function main() {
         ? parseInt(process.env.DB_DAYS_BEFORE_NOW_TO_DELETE)
         : undefined,
     },
+    apiOptions: {
```
jonesho, line 156: Can we name this as sth like `metricApiOptions`, as a reader this doesn't tell what the api options are for

postman/src/application/postman/api/metrics/MetricsService.ts
jonesho: I also think this would allow refactoring `getCounterValue` and `getGaugeValue` to share common codes.

postman/src/application/postman/api/metrics/MessageMetricsService.ts
```diff
@@ -0,0 +1,60 @@
+import { EntityManager } from "typeorm";
+import { Direction } from "@consensys/linea-sdk";
+import { MetricsService } from "./MetricsService";
+import { MessageEntity } from "../../persistence/entities/Message.entity";
+import { MessageStatus } from "../../../../core/enums";
+
+export class MessageMetricsService extends MetricsService {
+  constructor(private readonly entityManager: EntityManager) {
+    super();
+    this.createGauge("postman_messages_current", "Current number of messages by status", ["status", "direction"]);
```
jonesho: More intuitive to me `Current number of messages by status and direction`

postman/src/application/postman/persistence/subscribers/MessageStatusSubscriber.ts
```diff
@@ -0,0 +1,165 @@
...
+      if (gauge || gauge === 0) {
+        this.metricsService.incrementGauge("postman_messages_current", 1, {
+          status: updatedEntity.newStatus,
+          direction: updatedEntity.direction,
+        });
+      }
+    }
+  }
+
+  private async updateMessageMetricsOnInsert(event: InsertEvent<MessageEntity>): Promise<void> {
```
jonesho: `updateMessageMetricsOnInsert` and `updateMessageMetricsOnRemove` are almost the same, can we refactor to share common codes?

postman/src/application/postman/persistence/subscribers/MessageStatusSubscriber.ts
jonesho: I also think these two condition checks are confusing, just a check for `undefined` is more intuitive

postman/src/application/postman/persistence/subscribers/MessageStatusSubscriber.ts
jonesho: Also are we sure `afterTransactionCommit` and `afterUpdate` would only be called either one for an message db udpate? Otherwise we might have double inc/dec the gauges

postman/src/application/postman/api/metrics/MessageMetricsService.ts
jonesho: Also imo `postman_messages_current` has occurred too many times in the codebase, should we define it as a static variable? In addition, to align with coordinator metric naming, I would suggest to add prefix (`linea_`) to the metric name as sth like `linea_postman_messages` (we could even omit the `_current`), wdyt?

postman/src/services/persistence/LineaMessageDBService.ts
```diff
@@ -63,44 +61,20 @@ export class LineaMessageDBService extends MessageDBService implements IMessageD
    */
   public async getMessageToClaim(
     contractAddress: string,
-    _gasEstimationMargin: number,
+    gasEstimationMargin: number,
```
jonesho: comment: `_gasEstimationMargin` at line 57 need to be updated

postman/src/application/postman/persistence/repositories/TypeOrmMessageRepository.ts
```diff
@@ -175,15 +172,14 @@ export class TypeOrmMessageRepository<TransactionResponse extends ContractTransa
...
+              threshold: parseFloat(currentGasPrice.toString()) * gasEstimationMargin,
+            }).orWhere("message.claimGasEstimationThreshold IS NULL");
+          }),
         )
-        .orderBy("CAST(message.status as CHAR)", "ASC")
-        .addOrderBy("CAST(message.fee AS numeric)", "DESC")
+        .orderBy("CAST(message.status as CHAR)", "DESC")
```
jonesho: just curious, why change this to DESC?

postman/src/services/persistence/LineaMessageDBService.ts
```diff
@@ -63,44 +61,20 @@ export class LineaMessageDBService extends MessageDBService implements IMessageD
...
     maxRetry: number,
     retryDelay: number,
   ): Promise<Message | null> {
-    const feeEstimationOptions = await this.getClaimDBQueryFeeOptions();
+    const { maxFeePerGas } = await this.provider.getFees();
```
jonesho: iiuc, this `maxFeePerGas` is not from `linea_estimateGas` and hence its value is supposed to be greater than what returned by `linea_estimateGas`, which means it would be restricter to pick the message to claim in db than how we determine a transaction is "underpriced" (we use `getBlockExtraData` which should be the same as calling `linea_estimateGas`), what's the motivation of that? why not both use the same gas price from `getBlockExtraData`  like before?

postman/src/services/processors/MessageClaimingProcessor.ts
VGau: it is on purpose because I want to return true avec the tx is underpriced but I want to log and update the DB only when the `message.status !== MessageStatus.FEE_UNDERPRICED`

postman/src/application/postman/persistence/subscribers/MessageStatusSubscriber.ts
VGau: afterTransactionCommit is only call when a DB transaction commit happened.
afterUpdate is not called in this case :)

postman/src/application/postman/persistence/repositories/TypeOrmMessageRepository.ts
VGau: yes it will :) I will remove this condition on my side as well :)

postman/scripts/runPostman.ts
VGau: the API only contains metrics for the moment but it can contain health endpoints or other endpoints in the future. That's why I prefer to keep the naming generic

postman/src/application/postman/api/metrics/MessageMetricsService.ts
VGau: I can align the naming yes :) Will create a constant somewhere 👍

postman/src/application/postman/api/Api.ts
```diff
@@ -0,0 +1,65 @@
+import express, { Express, Request, Response } from "express";
+import { IMetricsService } from "../../../core/metrics/IMetricsService";
+import { ILogger } from "../../../core/utils/logging/ILogger";
+
+type ApiConfig = {
+  port: number;
+};
+
+export class Api {
```
jonesho, line 9: This class seems not generic and coupled with `IMetricService`, I think we should either make it generic or just call it `MetricApi`? wdyt?

postman/src/application/postman/persistence/repositories/TypeOrmMessageRepository.ts
VGau: It because I want to get messages with status `TRANSACTION_SIZE_COMPUTED` before messages with status `FEE_UNDERPRICED`

postman/src/application/postman/api/Api.ts
VGau: I think it is fine because this Api will possibly contain other endpoints like healthcheck and so on in the future.

postman/src/application/postman/api/__tests__/Api.test.ts
VGau: this is more like a e2e test or integration test :)

postman/src/application/postman/persistence/subscribers/MessageStatusSubscriber.ts
jonesho: Thanks Vic, can you remind me which cases would trigger `afterUpdate` and `afterTransactionCommit`? I'm kinda confused why we need both?

postman/src/application/postman/persistence/repositories/TypeOrmMessageRepository.ts
```diff
@@ -174,15 +169,7 @@ export class TypeOrmMessageRepository<TransactionResponse extends ContractTransa
...
-        .andWhere(
-          "CAST(message.fee AS numeric) > :minimumMargin * ((:extraDataVariableCost * message.compressedTransactionSize) / message.claimTxGasLimit + :extraDataFixedCost) * message.claimTxGasLimit",
-          {
-            minimumMargin: feeEstimationOptions.minimumMargin,
-            extraDataVariableCost: feeEstimationOptions.extraDataVariableCost,
-            extraDataFixedCost: feeEstimationOptions.extraDataFixedCost,
-          },
-        )
-        .orderBy("CAST(message.status as CHAR)", "ASC")
+        .orderBy("CAST(message.status as CHAR)", "DESC")
```
jonesho: @VGau I wanna clarify these changes, so does it mean we're good to try claiming messages that might be "underpriced" as long as they have the highest messagae fee in db? And we'd rely later on the other service (i.e.`LineaTransactionValidationService`) would actually check and update them if "underpriced"?

postman/src/application/postman/persistence/subscribers/MessageStatusSubscriber.ts
VGau:
```
await this.manager.transaction(async (entityManager) => {
  await entityManager.update(....);
  await entityManager.update(...);
})
```
The above example won't trigger the `afterUpdate`. It will only trigger the `afterTransactionCommit`.

postman/src/application/postman/persistence/repositories/TypeOrmMessageRepository.ts
VGau: We won't try to claim underpriced messages because there is a check in the `LineaTransactionValidationService` to check if the message is underpriced or not.

jonesho (dismissed):

kyzooghost (approved):

PR #885, opened 2025-04-17
Prover/Adding the Optional SIS Hashing Feature in the Vortex Protocol

arijitdutta67:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/prover/crypto/vortex/commitment.go
+++ b/prover/crypto/vortex/commitment.go
@@ -43,7 +43,7 @@ func (p *Params) CommitMerkleWithSIS(ps []smartvectors.SmartVector) (encodedMatr
 		// colHashes stores concatenation of SIS+MiMC hashes of the columns
 		// if isSISAppliedForRound is true, otherwise it stores the MiMC hashes
 		// of the columns.
-		colHashes = p.hashColumnsWithSIS(encodedMatrix)
+		colHashes = p.Key.TransversalHash(encodedMatrix)
 	})
 
 	timeTree := profiling.TimeIt(func() {
@@ -89,7 +89,7 @@ func (p *Params) CommitMerkleWithoutSIS(ps []smartvectors.SmartVector) (encodedM
 	timeTree := profiling.TimeIt(func() {
 		// colHashes stores the MiMC hashes
 		// of the columns.
-		colHashes = p.hashColumnsWithoutSIS(encodedMatrix)
+		colHashes = p.noSisTransversalHash(encodedMatrix)
 		leaves := make([]types.Bytes32, len(colHashes))
 		for i := range leaves {
 			leaves[i] = colHashes[i].Bytes()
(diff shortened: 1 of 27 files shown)


prover/crypto/vortex/commitment.go
```diff
@@ -31,7 +28,9 @@ type EncodedMatrix []smartvectors.SmartVector
 //
 // And can be safely converted to a field Element via
 // [field.Element.SetBytesCanonical]
-func (p *Params) CommitMerkle(ps []smartvectors.SmartVector) (encodedMatrix EncodedMatrix, tree *smt.Tree, colHashes []field.Element) {
+// We apply SIS+MiMC hashing on the columns to compute leaves only if
+// isSISAppliedForRound is true, otherwise we apply only the MiMC hashing.
+func (p *Params) CommitMerkle(ps []smartvectors.SmartVector, isSISAppliedForRound bool) (encodedMatrix EncodedMatrix, tree *smt.Tree, colHashes []field.Element) {
```
AlexandreBelling: Better to make two separate function and factor out the common parts, no?

AlexandreBelling: Ah, I see. You basically reused the code for replaceBySis.

prover/crypto/vortex/commitment.go
```diff
@@ -107,9 +108,9 @@ func (params *Params) encodeRows(ps []smartvectors.SmartVector) (encodedMatrix E
 // `encodedMatrix` sequentially.
 //
 // When SIS is used, `colHashes` stores the concatenation of the SIS hashes.
-func (params *Params) hashColumns(encodedMatrix EncodedMatrix) (colHashes []field.Element) {
+func (params *Params) hashColumns(encodedMatrix EncodedMatrix, isSISAppliedForRound bool) (colHashes []field.Element) {
 	// And obtain the hash of the columns
-	if !params.HasSisReplacement() {
+	if !params.HasSisReplacement() && isSISAppliedForRound {
```
AlexandreBelling: Wasn't the idea to remove "HasSisReplacement" ?

prover/crypto/vortex/commitment.go
```diff
@@ -40,15 +39,17 @@ func (p *Params) CommitMerkle(ps []smartvectors.SmartVector) (encodedMatrix Enco
...
+		// if isSISAppliedForRound is true, otherwise it stores the MiMC hashes
+		// of the columns.
+		colHashes = p.hashColumns(encodedMatrix, isSISAppliedForRound)
 	})
 
 	timeTree := profiling.TimeIt(func() {
 		var leaves []types.Bytes32
 
-		if !p.HasSisReplacement() {
+		if !p.HasSisReplacement() && isSISAppliedForRound {
```
AlexandreBelling: When you have a boolean flag that represent an option (which is a priori, not the default behaviour). Then, the practice is it make it false in the default case and true in the optional case.

Since the default behaviour is to use SIS, your flag should be "isSisReplacedByMiMC", so the contrary of that.

That way, all pieces of code that don't want to set this value won't have anything to change

prover/crypto/vortex/commitment.go
AlexandreBelling: But at this point, it is cleaner to have two function as this is a public function

prover/crypto/vortex/commitment.go
```diff
@@ -103,15 +141,18 @@ func (params *Params) encodeRows(ps []smartvectors.SmartVector) (encodedMatrix E
...
+func (params *Params) hashColumnsWithSIS(encodedMatrix EncodedMatrix) (colHashes []field.Element) {
+	// Obtain the hash of the columns
+	return params.Key.TransversalHash(encodedMatrix)
+
+}
+
+// hashColumnWithoutSIS returns a slice storing the hashes of the column of
+// `encodedMatrix` sequentially. `colHashes` stores the concatenation of the MiMC hashes.
+func (params *Params) hashColumnsWithoutSIS(encodedMatrix EncodedMatrix) (colHashes []field.Element) {
+	// Obtain the hash of the columns
```
AlexandreBelling: I think we can remove this flat-band function. e.g. a function that only returns the result of another one. Same for the above one.

AlexandreBelling (approved):
LGTM, just a minor comment. The update of the compiler looks great

PR #886, opened 2025-04-17
Message anchoring v2 part3

fluentcrafter:
This PR implements issue(s) #

### Checklist


jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClient.kt
```diff
@@ -0,0 +1,88 @@
...
+  override fun getBlockByNumber(blockParameter: BlockParameter): SafeFuture<Block?> {
+    return web3jClient
+      .ethGetBlockByNumber(blockParameter.toWeb3j(), true)
+      .sendAsync()
+      .thenCompose(::handleError)
+      .thenApply { block -> block?.toDomain() }
+      .toSafeFuture()
+  }
+
+  override fun getBlockByNumberWithoutTransactionsData(blockParameter: BlockParameter): SafeFuture<BlockWithTxHashes?> {
```
jonesho, line 50: imo, `getBlockByNumberWithoutTransactionsData` is a bit verbose, can we shorten it as `getBlockByNumberWithoutTxData`

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/EthGetBlockToLineaBlockMappers.kt
```diff
@@ -46,6 +47,38 @@ fun mapToDomain(web3jBlock: EthBlock.Block): Block {
...
+        throw IllegalArgumentException(
+          "Expected to be have EthBlock.TransactionHash. Got instance of ${web3jBlock.transactions[0]::class.java}"
+        )
+      }
+      web3jBlock.transactions.map { (it as EthBlock.TransactionHash).get().decodeHex() }
+    }
+  )
+  return block
+}
+
```
jonesho: Can we refactor `mapToDomain` and `mapToDomainWithTxHashes` by a helper function like:
```
fun <TxData> mapToDomainHelper(web3jBlock: EthBlock.Block, extractTxDataFunc: (EthBlock.Block) -> List<TxData>): BlockData<TxData> {
  return BlockData(
    number = web3jBlock.number.toULong(),
    hash = web3jBlock.hash.decodeHex(),
    parentHash = web3jBlock.parentHash.decodeHex(),
    ommersHash = web3jBlock.sha3Uncles.decodeHex(),
    miner = web3jBlock.miner.decodeHex(),
    nonce = web3jBlock.nonce.toULong(),
    stateRoot = web3jBlock.stateRoot.decodeHex(),
    transactionsRoot = web3jBlock.transactionsRoot.decodeHex(),
    receiptsRoot = web3jBlock.receiptsRoot.decodeHex(),
    logsBloom = web3jBlock.logsBloom.decodeHex(),
    difficulty = web3jBlock.difficulty.toULong(),
    gasLimit = web3jBlock.gasLimit.toULong(),
    gasUsed = web3jBlock.gasUsed.toULong(),
    timestamp = web3jBlock.timestamp.toULong(),
    extraData = web3jBlock.extraData.decodeHex(),
    mixHash = web3jBlock.mixHash.decodeHex(),
    baseFeePerGas = web3jBlock.baseFeePerGas?.toULong(), // Optional field for EIP-1559 blocks
    ommers = web3jBlock.uncles.map { it.decodeHex() }, // List of uncle block hashes
    transactions = run {
      if (web3jBlock.transactions.isNotEmpty() && web3jBlock.transactions[0] !is EthBlock.TransactionHash) {
        throw IllegalArgumentException(
          "Expected to be have EthBlock.TransactionHash. Got instance of ${web3jBlock.transactions[0]::class.java}"
        )
      }
      extractTxDataFunc(web3jBlock)
    }
  )
}
```

jonesho: E.g. `mapToDomain` can become:  
```
fun mapToDomain(web3jBlock: EthBlock.Block): Block {
  return mapToDomainHelper(web3jBlock) { block ->
    block.transactions.map { (it as EthBlock.TransactionObject).toDomain() }
  }
}
```

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClient.kt
```diff
@@ -0,0 +1,88 @@
...
+import linea.web3j.toDomain
+import net.consensys.linea.async.toSafeFuture
+import org.web3j.protocol.Web3j
+import org.web3j.protocol.core.methods.request.EthFilter
+import org.web3j.protocol.core.methods.response.Log
+import tech.pegasys.teku.infrastructure.async.SafeFuture
+
+/**
+ * Web3j Adapter of EthApiClient
+ * Request retries is reponsability of another class
```
jonesho: typo: `responsibility`

jvm-libs/linea/clients/eth-logs-searcher/src/main/kotlin/linea/ethapi/Web3JLogsSearcher.kt
```diff
@@ -70,15 +66,109 @@ class Web3JLogsSearcher(
       }
   }
 
-  private fun findLogLoop(
+  override fun getLogsRollingForward(
```
jonesho: Not see its usage yet, but I would assume this would be the function used by message anchoring service?

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClient.kt
fluentcrafter: `getBlockByNumberWithoutTxData` that was my 1st approach, then I found it confusing that it would not fetch any tx data (e.g basically the block header) and tx hashes are still some data. Then I changed to this.

jvm-libs/linea/clients/eth-logs-searcher/src/main/kotlin/linea/ethapi/Web3JLogsSearcher.kt
fluentcrafter: Yes, it's correct.

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/EthGetBlockToLineaBlockMappers.kt
fluentcrafter: I like your suggestion. Althogh it will need some adaptation because the `if (web3jBlock.transactions.isNotEmpty() && web3jBlock.transactions[0] !is EthBlock.TransactionHash) {` needs to be inside the extractTxDataFunc.

Thanks

jonesho (approved):

jvm-libs/linea/clients/eth-logs-searcher/src/main/kotlin/linea/ethapi/EthLogsSearcherImpl.kt
```diff
@@ -70,15 +66,109 @@ class Web3JLogsSearcher(
...
+    toBlock: ULong,
+    address: String,
+    topics: List<String?>,
+    chunkSize: UInt,
+    searchTimeout: Duration,
+    logsLimit: UInt?
+  ): SafeFuture<EthLogsSearcher.LogSearchResult> {
+    val cursor = ConsecutiveSearchCursor(fromBlock, toBlock, chunkSize.toInt(), SearchDirection.FORWARD)
+
+    val logsCollected: MutableList<EthLog> = CopyOnWriteArrayList()
```
Filter94, line 112: Is it being concurrently updated?

jvm-libs/linea/clients/interfaces/src/main/kotlin/linea/EthLogsSearcher.kt
```diff
@@ -24,10 +27,21 @@ interface EthLogsSearcher {
...
-  fun getLogs(
+  data class LogSearchResult(
+    val logs: List<EthLog>,
+    override val startBlockNumber: ULong,
+    override val endBlockNumber: ULong
+  ) : BlockInterval {
+    val isEmpty: Boolean = logs.isEmpty()
+  }
+
+  fun getLogsRollingForward(
```
Filter94, line 38: This method is more fit for a logs Collector, than a Searcher, because it doesn't search for a specific log event, but rather collects them. I suggest to split this interface in two

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClient.kt
```diff
@@ -0,0 +1,88 @@
...
+      /*address*/ address
+    ).apply {
+      topics.forEach { addSingleTopic(it) }
+    }
+
+    return web3jClient
+      .ethGetLogs(ethFilter)
+      .sendAsync()
+      .toSafeFuture()
+      .thenCompose(::handleError)
```
Filter94, line 77: Minor: This idiom can be factored out and reused, as this seems to be repeating

jvm-libs/linea/clients/eth-logs-searcher/src/main/kotlin/linea/ethapi/EthLogsSearcherImpl.kt
fluentcrafter: Yes, the timer and Future handler may run in different Threads.

Filter94: I don't get it. I found [this](https://github.com/Consensys/linea-monorepo/pull/886/files/274aef5657dff402136b48f587e257b6b72a8558#diff-b17bfc2505707e4971d05a6bf69d40bf089c733efbcd5fa4f8194056744a05bfR135) place where it's updated and nothing else. What is the other place I'm missing? What kind of timer?

jvm-libs/linea/clients/interfaces/src/main/kotlin/linea/EthLogsSearcher.kt
fluentcrafter: Thread safety is about writes as well as reads. That collection is being updated by a asynchronous Future's handler, which can be called in the context of different Threads. So it's not safe to use a regular collection

jvm-libs/linea/clients/eth-logs-searcher/src/main/kotlin/linea/ethapi/EthLogsSearcherImpl.kt
fluentcrafter: Thread safety is about **reads** and writes. The collection is updated inside an Async Futures handler that can be called by a different Thread Scope every call. So, it's not safe to use a regular collection.

Filter94: I'm still failing to see how does that happen concurrently. All I can see from the code is that `stopRetriesPredicate` is reading `logsCollected` after possible `addAll` inside the lambda, never concurrently. This makes the calls to logsCollected sequential, but from different threads. So I don't see why we'd need a thread safe implementation

fluentcrafter: Because without Threadsafety, even though they happen one after another, Thread may have cache outdated.

Filter94: So that means in that case a thread will read an outdated copy and won't fail because of that? In turn, this means that either `stopRetriesPredicate` may get fewer events and don't stop the logs collection at the right time.

I doubt if thread safety is a realistic concern here. I'll try to write a reproducer to get issues because of a thread cache and sequential operations

Filter94: 1. Futures are usually executed on the same thread unless they're composed with a future executed on a different thread
2. Here is the my attempt to reproduce the issue:
```

import java.util.concurrent.CompletableFuture
import java.util.concurrent.Executors
import org.apache.logging.log4j.LogManager

object App {
  @JvmStatic
  fun main(args: Array<String>) {
    val log = LogManager.getLogger(App.javaClass)

    val onePool = Executors.newFixedThreadPool(1)
    val otherPool = Executors.newFixedThreadPool(1)

    data class Some(val a: Int)

    repeat(10000000) { iteration ->
      println("Iteration $iteration")
      val logsCollected: MutableList<Some> = mutableListOf()
      CompletableFuture
        .runAsync(
          {
            repeat(1000) {
              logsCollected.add(Some(1))
            }
          },
          onePool,
        ).thenCompose {
          CompletableFuture.runAsync({
            if (logsCollected.size == 1000) {
              logsCollected.add(Some(6))
            } else {
              logsCollected.add(Some(5))
            }
          }, otherPool)
        }.get()

      if (logsCollected.last() == Some(5)) {
        throw IllegalStateException("Thread cache was outdated!")
      }
    }
  }
}
```
Here I'm enforcing execution of different stages of the future on separate threads and it didn't fail a single time so far

Filter94: With 1B iterations it didn't fail once on my machine too. It might be that I made a mistake in the reproducer somewhere, of course, but I'm prone to think that I don't know something that makes it work consistently, at least on my machine

PR #890, opened 2025-04-21
New compression dictionary and update guide

Tabaie:
This PR introduces a new compression dictionary trained on _all blocks_ from March 21 2025 to April 21 2025.

It also introduces a short guide on how to release new dictionaries.
@jpnovais I'm not too familiar with how the JNI libraries are used so that part is light on details. Please lmk if it can be improved.

Tests using the [blob simulator](https://github.com/Consensys/blobsim) on Linea blocks 18496135 to 18506189 shows a modest increase in the compression ratio:

**BEFORE** Using https://github.com/Consensys/linea-monorepo/blob/main/prover/lib/compressor/compressor_dict.bin
```
******************** AVERAGE STATS ********************
Note that there is a "theoretical" limit of 127 KB on blob size and hence 99.21% utilization
Average blob size: 127835 B, 124.84 KB
Average utilization 97.53%
Average compression ratio 6.33
Average payload size: 551213 B, 538.29 KB
TOTAL compression ratio 6.15
Minimum blob utilization 69.78
```
**AFTER** 
```
******************** AVERAGE STATS ********************
Note that there is a "theoretical" limit of 127 KB on blob size and hence 99.21% utilization
Average blob size: 127798 B, 124.80 KB
Average utilization 97.50%
Average compression ratio 6.76
Average payload size: 590585 B, 576.74 KB
TOTAL compression ratio 6.59
Minimum blob utilization 71.56
Minimum compression ratio 5.40
```


gbotrel (approved):

julien-marchand (approved):

AlexandreBelling:
Any reason to not merge this now? Seems like it won't affect the prod

Tabaie:
I agree that it doesn't affect prod. Still would like Pedro to take a look first right @julien-marchand?

Tabaie:
@fluentcrafter Exactly, except perhaps we should update State Reconstruction before the Coordinator (always downstream first)

PR #893, opened 2025-04-22
Message anchoring v2 part4

fluentcrafter:
This PR implements issue(s) #

### Checklist


coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/CapacityBoundedBlockingPriorityQueue.kt
```diff
@@ -0,0 +1,45 @@
...
+
+  override fun add(element: T): Boolean {
+    require(super.size < absoluteMaxCapacity.toInt()) {
+      "Queue is full size=$size, cannot add element: $element"
+    }
+    return super.add(element)
+  }
+
+  override fun addAll(elements: Collection<T>): Boolean {
+    require((remainingMaxCapacity() - super.size) >= elements.size) {
```
jonesho: Why do we need to substract `super.size` again? `remainingMaxCapacity()` should be enough, no?

coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/CapacityBoundedBlockingPriorityQueue.kt
```diff
@@ -0,0 +1,45 @@
...
+      "Queue absolute MaxremainingCapacity=${this.remainingMaxCapacity()} is less than elements size=${elements.size}"
```
jonesho: `MaxremainingCapacity` -> `remainingMaxCapacity`

coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/CapacityBoundedBlockingPriorityQueue.kt
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    return super.addAll(elements)
+  }
+
+  override fun remainingCapacity(): Int {
+    return (targetCapacity.toInt() - super.size)
```
jonesho: iiuc, this can potentially go negative, given that `super.size` is bounded by `absoluteMaxCapacity` (the check in line 24), and `super.size` is not necessary bounded by `targetCapacity` which was passed just as the `initialCapacity` of the underlying queue

coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/CapacityBoundedBlockingPriorityQueue.kt
```diff
@@ -0,0 +1,45 @@
+package linea.anchoring
+
+import java.util.concurrent.PriorityBlockingQueue
+
+/**
+ * A bounded priority queue that allows adding elements even when the queue is full.
+ */
+class CapacityBoundedBlockingPriorityQueue<T : Comparable<T>>(
```
jonesho: Just a comment, should we consider move this generic queue to `utils` package?

coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/MessageAnchoringApp.kt
```diff
@@ -0,0 +1,90 @@
...
+      l1ContractAddress = config.l1ContractAddress,
+      l1EthLogsClient = l1EthApiClient,
+      l2MessageService = l2MessageService,
+      eventsQueue = eventsQueue,
+      maxMessagesToAnchorPerL2Transaction = config.maxMessagesToAnchorPerL2Transaction,
+      l2HighestBlockTag = config.l2HighestBlockTag,
+      anchoringTickInterval = config.anchoringTickInterval
+    )
+
+  val queueMessages: List<MessageSentEvent>
```
jonesho: Seems not used for now, is there a need to expose the events in the queue to the caller?

coordinator/ethereum/message-anchoring-i2/build.gradle
```diff
@@ -0,0 +1,17 @@
+plugins {
```
jonesho, line 1: Why called folder name `message-anchoring-i2`? Are you intending to name as `message-anchoring-l2`?

coordinator/ethereum/message-anchoring-i2/src/test/kotlin/linea/anchoring/MessageAnchoringAppTest.kt
```diff
@@ -0,0 +1,243 @@
...
+    )
+  }
+
+  @Test
+  fun `should anchor messages from fresh deployment`() {
+    l1Client.setLogs(l1MessageSentLogs.flatten())
+
+    val anchoringApp = createApp(anchoringTickInterval = 100.milliseconds, l1EventSearchBlockChunk = 100u)
+    anchoringApp.start().get()
+    l1Client.setLatestBlockTag(l1MessageSentLogs.last().last().blockNumber + 10UL)
```
jonesho: imo, the latest block tag should be better set higher than the finalized block tag

jvm-libs/linea/clients/eth-logs-searcher/src/test/kotlin/linea/ethapi/LogsSearcherTest.kt
```diff
@@ -0,0 +1,159 @@
...
+  @Test
+  fun `should respect stopAfterTargetLogsCount`() {
+    val result = searcher.getLogsRollingForward(
+      fromBlock = BlockParameter.BlockNumber(0UL),
+      toBlock = BlockParameter.BlockNumber(3000UL),
+      address = testAddress,
+      topics = emptyList(),
+      chunkSize = 50U,
+      searchTimeout = 2.hours,
+      stopAfterTargetLogsCount = 2U
```
jonesho: can we set `stopAfterTargetLogsCount` as 1U? because the expected result is gonna be `initialLogs.take(2)` when `stopAfterTargetLogsCount = null`

coordinator/ethereum/message-anchoring-i2/src/test/kotlin/linea/anchoring/MessageAnchoringAppTest.kt
```diff
@@ -0,0 +1,243 @@
...
+    await()
+      .atMost(10.seconds.toJavaDuration())
+      .untilAsserted {
+        assertThat(l2MessageService.getLastAnchoredL1MessageNumber(block = BlockParameter.Tag.LATEST).get())
+          .isEqualTo(lastMessageSentEvent.event.messageNumber)
+      }
+
+    anchoringApp.stop().get()
+  }
+
```
jonesho, line 175: It would be great if we could add a similar test case as above but with many message events in a single L1 block which would require to collect the logs within same L1 block multiple times (i.e. the number of message events in a single L1 block is higher than the queue capacity atm)

coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/CapacityBoundedBlockingPriorityQueue.kt
fluentcrafter: Yes, thanks for the catch

coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/CapacityBoundedBlockingPriorityQueue.kt
fluentcrafter: Yes, I will fix it. thanks

coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/CapacityBoundedBlockingPriorityQueue.kt
fluentcrafter: I don't want to do this as I did not put effort into making it feature-proof for generic use. Perfer to carve-out after.

coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/MessageAnchoringApp.kt
fluentcrafter: It was for tests/metrics. But will remove it as it's not necessary.

coordinator/ethereum/message-anchoring-i2/build.gradle
fluentcrafter: The plan is to have the new code side-by-side for now. Once I replace old by new code will rename back to `message-anchoring`

This is a technique I use to avoid complex diffs, code reviews and merge issues.

coordinator/ethereum/message-anchoring-i2/src/test/kotlin/linea/anchoring/MessageAnchoringAppTest.kt
fluentcrafter: We only care about finalized tag atm actually, will remove the latest to avoid confusion

jvm-libs/linea/clients/eth-logs-searcher/src/test/kotlin/linea/ethapi/LogsSearcherTest.kt
fluentcrafter: Makes sense. thanks

coordinator/ethereum/message-anchoring-i2/src/test/kotlin/linea/anchoring/MessageAnchoringAppTest.kt
fluentcrafter: Very edge case, and good one will add it :)

coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/CapacityBoundedBlockingPriorityQueue.kt
```diff
@@ -0,0 +1,45 @@
...
+    require(targetCapacity.toInt() in validRange) { "targetCapacity=$targetCapacity must in range $validRange" }
+    require(absoluteMaxCapacity.toInt() in validRange) {
+      "absoluteMaxCapacity=$absoluteMaxCapacity must in range $validRange"
+    }
+    require(absoluteMaxCapacity >= targetCapacity) {
+      "absoluteMaxCapacity=$absoluteMaxCapacity must be greater than or equal to targetCapacity=$targetCapacity"
+    }
+  }
+
+  override fun add(element: T): Boolean {
```
Filter94: I couldn't find the usage of this method on this class and if we don't need it, I don't see why would we need 2 separate caps.

fluentcrafter: The method was overridden for consistency with addAll so it can be used interchangeably.

The 2 caps are because we collect L1 logs in chunks, if let's say remaining capacity is 100, we go to L1 and we get 125 it allows the remaining extra 25 in. Otherwise it would throw  already logged events away...

Filter94: Looking at the logic in `L1MessageSentEventsPoller`, which seems to be the only place where these 2 caps matter, we won't poll for more events if we reached the first cap and we won't want to lose events that we add on top of that if we happen to get a chunk with more events than the first cap.

What will happen if we exceed the second cap?

fluentcrafter: > What will happen if we exceed the second cap?

It would throw an error and retry the query in the next tick.

I did a small refactor to `L1MessageSentEventsPoller` that does not even need that queue.

coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/clients/L1MessageSentEventsFetcher.kt
```diff
@@ -0,0 +1,103 @@
...
+ * This class is responsible for fetching L1 MessageSent events from the Ethereum blockchain.
+ * It leverages the invariant of increasing message numbers to efficiently search for events on L1 by never going back
+ *
+ * This is relevant when MessageSent events are very sparse on the chain, it requires 10_000K l1 blocks
+ * to collect 100 messages.
+ */
+internal class L1MessageSentEventsFetcher(
+  private val l1SmartContractAddress: String,
+  private val l1EventsSearcher: EthLogsSearcher,
+  private val highestBlockNumber: BlockParameter
```
Filter94: `BlockNumber` part makes it confusing, because it's going to be a block tag more often. I suggest to rename it either to `blockParameter` or even `blockTag`, which is less precise, but IMO more pragmatic

coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/CapacityBoundedBlockingPriorityQueue.kt
fluentcrafter: https://github.com/Consensys/linea-monorepo/pull/893/commits/1cc3d66799d81b2c95cda0b1b8595b3d4b9a4c28

Filter94: That's where I was going with this. I see that it's not used anymore at all. Can we delete this class then? 🙂

coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/MessageAnchoringApp.kt
```diff
@@ -0,0 +1,96 @@
...
+      config = EthLogsSearcherImpl.Config(loopSuccessBackoffDelay = config.l1SuccessBackoffDelay)
+    )
+
+  // private val eventsQueue = CapacityBoundedBlockingPriorityQueue<MessageSentEvent>(
+  //   targetCapacity = config.messageQueueCapacity,
+  //   // allow at least one 1 L1 Block full of Linea Messages
+  //   // sendMessages takes 67,385 gas, 60Mgas/67k ~ 1000
+  //   absoluteMaxCapacity = (config.messageQueueCapacity * 2u).coerceAtLeast(1000u)
+  // )
+  private val eventsQueue = PriorityBlockingQueue<MessageSentEvent>(config.messageQueueCapacity.toInt())
```
Filter94, line 48: Don't forget to delete the comment above

coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/MessageAnchoringService.kt
```diff
@@ -0,0 +1,93 @@
...
+        eventsQueue.removeIf { it.messageNumber <= lastAnchoredL1MessageNumber }
+
+        eventsQueue
+          .toArray(emptyArray<MessageSentEvent>())
+          .filter { it.messageNumber > lastAnchoredL1MessageNumber }
+          .sortedBy { it.messageNumber }
+          .take(maxMessagesToAnchorPerL2Transaction.toInt())
+      }.thenCompose { eventsToAnchor ->
+        if (eventsToAnchor.isEmpty()) {
+          // log.debug("No new events to anchor")
```
Filter94: I think it's a useful log. Would be useful to log in `else` as well

coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/clients/L1MessageSentEventsFetcher.kt
```diff
@@ -0,0 +1,103 @@
...
+) {
+  private val lastSearch: AtomicPair = AtomicPair()
+
+  private class AtomicPair(
```
Filter94: minor: Consider an atomic reference to a data class instead

coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/clients/L1MessageSentEventsFetcher.kt
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * This is relevant when MessageSent events are very sparse on the chain, it requires 10_000K l1 blocks
+ * to collect 100 messages.
+ */
+internal class L1MessageSentEventsFetcher(
+  private val l1SmartContractAddress: String,
+  private val l1EventsSearcher: EthLogsSearcher,
+  private val highestBlockNumber: BlockParameter
+) {
+  private val lastSearch: AtomicPair = AtomicPair()
```
Filter94: It would be nice to log its state sometimes on DEBUG level

coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/clients/L1MessageSentEventsPoller.kt
```diff
@@ -0,0 +1,74 @@
...
+    val queueLastMessage = eventsQueue.lastOrNull()
+    if (queueLastMessage != null) {
+      return SafeFuture.completedFuture(queueLastMessage.messageNumber.inc())
+    } else {
+      return l2MessageService.getLastAnchoredL1MessageNumber(block = highestBlockNumber)
+        .thenApply { it.inc() }
+    }
+  }
+
+  fun queueRemainingCapacity(): Int {
```
Filter94: Can be private

coordinator/ethereum/message-anchoring-i2/src/test/kotlin/linea/anchoring/MessageAnchoringAppTest.kt
```diff
@@ -0,0 +1,268 @@
...
+  private fun createL1MessageSentV1Logs(
+    l1BlocksWithMessages: List<ULong>,
+    numberOfMessagesPerBlock: Int,
+    startingMessageNumber: ULong = 1UL
+  ): List<L1MessageSentV1EthLogs> {
+    val ethLogs = mutableListOf<L1MessageSentV1EthLogs>()
+    var messageNumber = startingMessageNumber
+    l1BlocksWithMessages.forEach { blockNumber ->
+      for (i in 0 until numberOfMessagesPerBlock) {
```
Filter94: Minor: `i` isn't used. I think this can be replaced by `repeat(numberOfMessagesPerBlock)`

coordinator/ethereum/message-anchoring-i2/src/test/kotlin/linea/anchoring/events/L1RollingHashUpdatedEventTest.kt
```diff
@@ -0,0 +1,48 @@
...
+    )
+
+    // When
+    val result = L1RollingHashUpdatedEvent.fromEthLog(ethLog)
+
+    // Then
+    val expectedEvent = L1RollingHashUpdatedEvent(
+      messageNumber = 0xB415A.toULong(),
+      rollingHash = "0x7abd5eea8cbb46bba0aa83369dcc0d9b18931a825b73f45d98da586070eafa8b".decodeHex(),
+      messageHash = "0x24dca2d33621322ef7c85d7cea38b673c06cbb86a7f15c8aa5f658485f932fd0".decodeHex()
```
Filter94, line 39: Minor: we could move these into constants so it's easier to track what goes from where

coordinator/ethereum/message-anchoring-i2/src/test/kotlin/linea/anchoring/fakes/FakeL2MessageServiceTest.kt
```diff
@@ -0,0 +1,104 @@
...
+    ).get()
+      .also { txHash ->
+        val lastMessageNumber = fakeL2MessageService.getLastAnchoredL1MessageNumber(BlockParameter.Tag.LATEST).get()
+        val lastRollingHash = fakeL2MessageService.getLastAnchoredRollingHash()
+        val anchoredMessages = fakeL2MessageService.getAnchoredMessageHashes()
+
+        assertThat(txHash).isNotEmpty()
+        assertThat(lastMessageNumber).isEqualTo(3UL)
+        assertThat(lastRollingHash)
+          .isEqualTo("0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa".decodeHex())
```
Filter94, line 54: Minor: Here as well, it would be slightly easier to follow if we reuse `"0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa".decodeHex()` and `3UL`

coordinator/ethereum/message-anchoring-i2/src/test/kotlin/linea/anchoring/MessageAnchoringAppTest.kt
```diff
@@ -0,0 +1,268 @@
...
+
+        assertThat(l2MessageService.getAnchoredMessageHashes().map { it.encodeHex() })
+          .isEqualTo(ethLogs.drop(1).map { it.messageSent.event.messageHash.encodeHex() })
+      }
+
+    anchoringApp.stop().get()
+  }
+
+  @Test
+  fun `should anchor messages in chunks when too many messages are waiting on L1`() {
```
Filter94, line 149: Nit: `in chunks` part doesn't seem to be important for this test. From the test it seems that all the messages are being anchored, but we don't verify if it's in chunks or not

jvm-libs/linea/clients/interfaces/src/test/kotlin/linea/ethapi/FakeEthApiClientTest.kt
```diff
@@ -0,0 +1,189 @@
...
+  @Test
+  fun `should return logs within block range`() {
+    val logs = fakeEthApiClient.getLogs(
+      fromBlock = 0UL.toBlockParameter(),
+      toBlock = 300UL.toBlockParameter(),
+      address = testAddress,
+      emptyList()
+    ).get()
+
+    assertThat(logs).isEqualTo(initialLogs.take(2))
```
Filter94: This test is wrong. The last log is being filtered out because of the address, not because of the block range. Consider changing the range to 100 .. 100 and expect the first log, or better add another log for block 300 for the testAddress

jvm-libs/linea/clients/interfaces/src/test/kotlin/linea/ethapi/FakeEthApiClientTest.kt
```diff
@@ -0,0 +1,189 @@
...
+  @Test
+  fun `should filter logs by topics`() {
+    fakeEthApiClient.getLogs(
+      fromBlock = 0UL.toBlockParameter(),
+      toBlock = 300UL.toBlockParameter(),
+      address = testAddress,
+      listOf(testTopic1)
+    ).get()
+      .also { logs ->
+        assertThat(logs).isEqualTo(initialLogs.take(2))
```
Filter94, line 85: Here as well. Last log event doesn't have a chance to be included because of the address filter. Consider adding another log event with another topic, but the same testAddress as the first 2. Or just leave the second case with `testTopic2`

jvm-libs/linea/clients/eth-logs-searcher/src/test/kotlin/linea/ethapi/LogsSearcherTest.kt
```diff
@@ -0,0 +1,159 @@
...
+      chunkSize = 100U,
+      searchTimeout = 5.seconds,
+      stopAfterTargetLogsCount = null
+    ).get()
+
+    assertThat(result.logs).isEqualTo(listOf(initialLogs[2]))
+  }
+
+  @Test
+  fun `should filter logs by topics`() {
```
Filter94, line 100: Consider adding more logs with `testTopic2` and `testAddress`, because this test is indistinguishable from the previous one, that is testing filtering by address

coordinator/ethereum/message-anchoring-i2/src/test/kotlin/linea/anchoring/MessageAnchoringAppTest.kt
fluentcrafter: I forgot to add the chunks bit actually.

coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/clients/L1MessageSentEventsFetcher.kt
fluentcrafter: Added some logging, please check

jvm-libs/linea/clients/interfaces/src/test/kotlin/linea/ethapi/FakeEthApiClientTest.kt
fluentcrafter: Done, please check

jonesho (approved):

jvm-libs/linea/clients/interfaces/src/test/kotlin/linea/ethapi/FakeEthApiClientTest.kt
fluentcrafter: It was fixed, please double-check. Thanks for the feedback

Filter94 (approved):

Issue #878, opened 2025-04-16
run e2e tests when building linea-besu-package


fluentcrafter:
## Description
We shall run linea e2e stack when building a new linea-besu-package to ensure compatibility between the besu version and plugins as well as full integration between packages. 

The linea-besu-package shall be published to Docker Hub after the e2e test suite passes against the newly built linea-besu-package;

**Note**: Given linea will decommission Geth tracing soon, we don't need to adapt current ci flow for this use case. It's good enough to just fovous on linea v1.4 onwards

## Motivation
- Reliability on the `linea-besu-package` distributions
- Avoid deploying images in production that may break at runtime
- Save costs on publishing unnecessary images to docker hub;

## Tasks
- [ ] Adapt linea-besu-package CI to run end2end tests before publishing to docker hub

## Acceptance criteria
- [ ] linea-besu-package is published only if the e2e pass

Issue #877, opened 2025-04-16
move build of linea-besu-package inside Linea Monorepo


fluentcrafter:
## Description
Move the build of `linea-besu-package` inside Linea Monorepo. 

Note, this issue shall be implemented with #878 in mind

## Motivation
Leverage Linea e2e tests to test the integration between Linea Besu and Linea Besu Package. 

## Tasks
- [ ] Create CI to build Linea Besu Package in monorepo
- [ ] We shall fix the current issue that the update of one environment, build all environments unnecessarily
- [ ] allow to tag (not mandatory) the images with a semantic name tag: 
   - [ ] linea-besu-package:v1.4-mainnet-8cbfc75

## Acceptance criteria
- [ ] linea-besu-package is built and published automatically to dockerhub when a version in (dev,sepolia,mainnet) is updated
   - [ ] only updated environments are build/plublished
- [ ] commit tag shall match the author's original commit with corect diff (no merge commit as we have atm, is confusing)

PR #900, opened 2025-04-23
Feat/integrate linea besu package

jonesho:
This PR implements issue(s) #877 and #878

In this PR:
- Move linea-besu-package repo into linea-monorepo (under `./linea-besu-package` folder)
- There would only be one single manual release workflow (i.e. linea-besu-package-release) which accepts an input for the release tag prefix such as the arithmetization version (e.g. 2.2.2) and the resultant docker image tag would be (e.g. `2.2.2-20250429141855-5bdf21f`) as suggested in this [doc](https://docs.google.com/document/d/1tGS93zyG6Dt3-k3q1n_9XRdlmt81qhguOvzsmPrrJSA/edit?tab=t.0#heading=h.c2bh57tnzwny), and the release tag would be (e.g. `linea-besu-package-2.2.2-20250429141855-5bdf21f`), the docker image and release would only be published if the besu and e2e tests are passed
- Any commit changes related to linea-besu-package would trigger the workflow automatically, e.g. changes in the `versions.env` file would triggered the e2e test and the image would be pushed to dockerhub with prefix automatically set as the `LINEA_TRACER_PLUGIN_VERSION` from `versions.env` (e.g. image tag would be `beta-v2.1-rc14-20250429141855-5bdf21f`), changes in profile or config files would triggered the besu test and the image would be pushed to dockerhub, etc..
- The docker image tag will always be `[releasePrefix]-[YYYYMMDDHHMMSS]-[shortenCommitHash]`
- The release info for the pushed docker images (including those from PRs) can be found at https://github.com/Consensys/linea-monorepo/releases

### Checklist


fluentcrafter:
> Each env will have its own manual release workflow (i.e. linea-besu-package-{devnet, sepolia, mainnet}) and each of them needs to be given a release tag prefix such as the arithmetization version (e.g. 2.2.2) and the resultant docker image tag would be (e.g. 2.2.2-20250429141855-5bdf21f) as suggested in this [doc](https://docs.google.com/document/d/1tGS93zyG6Dt3-k3q1n_9XRdlmt81qhguOvzsmPrrJSA/edit?tab=t.0#heading=h.c2bh57tnzwny), and the release tag would be (e.g. linea-besu-package-2.2.2-20250429141855-5bdf21f), the docker image and release would only be published if the besu and e2e tests are passed

One of the goals was to make the image build environment agnostic. So the same package shall work on all environments is this something we can achieve?

@jonesho @julien-marchand WDYT?

fab-10:
adding @Gabriel-Trintinalia @joshuafernandes @jflo to the loop

fab-10 (approved):

fab-10 (approved):

linea-besu-package/.gitignore
fluentcrafter, line 1: Why do we need this here?

linea-besu-package/.gitattributes
fluentcrafter, line 1: Should this be merged with the file in the root folder?

linea-besu-package/LICENSE
fluentcrafter, line 1: This will make 2 licences for the repo. Do we want that?

maybe we should remove this one in favour of the one in repo [root ](https://github.com/Consensys/linea-monorepo/blob/main/LICENSE)

@julien-marchand WDYT?

linea-besu-package/.gitignore
jonesho: Agree, this can be removed

linea-besu-package/.gitattributes
jonesho: Yes, I'll just simply delete it

linea-besu-package/LICENSE
jonesho: Agree, it should be deleted, I dun see a point for keeping 2 licenses for monorepo, will delete for now

fluentcrafter (approved):

PR #910, opened 2025-04-26
Message anchoring v2 part6

fluentcrafter:
This PR implements issue(s) #

### Checklist

* [ ] I wrote new tests for my new core changes.
* [ ] I have successfully ran tests, style checker and build against my new changes locally.
* [ ] I have informed the team of any breaking changes if there are any.

PR #913, opened 2025-04-29
Message anchoring v2 part7 - clean up of legacy clients

fluentcrafter:
This PR implements issue(s) #

### Checklist


besu-plugins/state-recovery/appcore/logic/src/main/kotlin/linea/staterecovery/BlockImporter.kt
```diff
@@ -42,17 +42,19 @@ class BlockImporterAndStateVerifierV1(
   private val stateManagerImportTimeoutPerBlock: Duration
 ) : BlockImporterAndStateVerifier {
   override fun importBlocks(blocks: List<BlockFromL1RecoveredData>): SafeFuture<ImportResult> {
+    val sortedBlocks = blocks.sortedBy { it.header.blockNumber }
```
Filter94, line 45: This doesn't seem to be related to the anchoring. Why do we need it?

jvm-libs/linea/clients/interfaces/src/test/kotlin/linea/ethapi/FakeEthApiClientTest.kt
```diff
@@ -132,14 +132,14 @@ class FakeEthApiClientTest {
       )
     ).also { client ->
       client.setFinalizedBlockTag(500UL)
-      assertThat(client.getBlockByNumber(BlockParameter.Tag.LATEST).get()?.number).isEqualTo(500UL)
-      assertThat(client.getBlockByNumber(BlockParameter.Tag.SAFE).get()?.number).isEqualTo(500UL)
-      assertThat(client.getBlockByNumber(BlockParameter.Tag.FINALIZED).get()?.number).isEqualTo(500UL)
+      assertThat(client.findBlockByNumber(BlockParameter.Tag.LATEST).get()?.number).isEqualTo(500UL)
```
Filter94: Isn't it better to use `getBlockByNumber` here? In case a test fails with NullPointerException, we'll see exactly why in the stacktrace anyway

jvm-libs/linea/clients/linea-l1-contract-client/src/main/kotlin/linea/contract/l2/Web3JL2MessageServiceSmartContractClient.kt
```diff
@@ -116,6 +119,26 @@ class Web3JL2MessageServiceSmartContractClient(
 
   override fun getAddress(): String = contractAddress
   override fun getVersion(): SafeFuture<L2MessageServiceSmartContractVersion> = getSmartContractVersion()
+  override fun getDeploymentBlock(): SafeFuture<ULong> {
```
Filter94, line 122: I would remove this method and instead I'd compute the deployment block once at startup and would store it in the only place it's used, in `AggregationL2StateProviderImpl`. Memory-wise it would be the same, since we wouldn't have to maintain the cache here, but would have to store the constant value in `AggregationL2StateProviderImpl` instead. But it would make the code more straightforward, with less branching and futures

jonesho:
@fluentcrafter Just to check, are these comments from the previous PR gonna be addressed in this PR?
- https://github.com/Consensys/linea-monorepo/pull/910#discussion_r2065557040
- https://github.com/Consensys/linea-monorepo/pull/910#discussion_r2065575560
- https://github.com/Consensys/linea-monorepo/pull/910#discussion_r2065680172

besu-plugins/state-recovery/appcore/logic/src/main/kotlin/linea/staterecovery/BlockImporter.kt
fluentcrafter: It was because staterecovery tests were failing in CI and by the logs somehow it seemed that they were replayed out of order. Hence I added this. Although, I don't have confidence on that. But since it does not hurt I would leave it

jvm-libs/linea/clients/interfaces/src/test/kotlin/linea/ethapi/FakeEthApiClientTest.kt
fluentcrafter: getBlockByNumber did not exist when I when wrote the test. Will adapt it.

jvm-libs/linea/clients/linea-l1-contract-client/src/main/kotlin/linea/contract/l2/Web3JL2MessageServiceSmartContractClient.kt
fluentcrafter: That was my first thought & approach. But the code was getting less clear (at the instantiation level), and IMO the logic of what/how to look for shall be encapsulated inside the smart contract client and not leak to the callers. 

Also, this approach has advantages that allow for instantiating the client without having the contract deployed right away, which can be convenient. 
For now I will leave it as is...

fluentcrafter:
@jonesho @Filter94 thanks for your review. Addressed your feedback, thanks.

Filter94 (dismissed):

Filter94 (approved):

PR #915, opened 2025-04-29
feat: BLS12-381 precompiles glue

ivokub:
Adds glue for integrating BLS precompiles from gnark with the data provided in the arithmetization.

WIP

### Checklist

### DIFF
--- a/prover/zkevm/prover/bls/element.go
+++ b/prover/zkevm/prover/bls/element.go
@@ -1,0 +1,44 @@
+package bls
+
+import (
+	"github.com/consensys/gnark/frontend"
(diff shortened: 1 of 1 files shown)


ivokub:
BLS glue itself is ready, but there are a few blockers:
* to update to the latest zkevm_prague we need need go-corset dependency.
* there is potentially memory regression with GKR, see https://github.com/Consensys/gnark/pull/1616. I have to check how it works and possibly backport it on top of gnark master

I also have some temp branch locally where I have tested implementation against actual traces, but it updates go-corset and I don't want to force it right now in this PR as different versions seem to be incompatible with each other.

AlexandreBelling (approved):

PR #917, opened 2025-04-30
minor(race): solves race condition for the FFT

AlexandreBelling:
This PR implements issue(s) #

### Checklist


AlexandreBelling:
I don't think this is the solution to the prod problem but I think we should merge it

gusiri (approved):

PR #924, opened 2025-04-30
Tried improving the queue usage for the anchoring

Filter94:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/MessageAnchoringService.kt
+++ b/coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/MessageAnchoringService.kt
@@ -16,7 +16,7 @@ import java.util.Queue
 import kotlin.time.Duration
 
 class MessageAnchoringService(
-  private val vertx: Vertx,
+  vertx: Vertx,
   private val l1ContractAddress: String,
   private val l1EthLogsClient: EthLogsClient,
   private val l2MessageService: L2MessageServiceSmartContractClient,
@@ -40,7 +40,6 @@ class MessageAnchoringService(
 
         eventsQueue
           .toTypedArray()
-          .filter { it.messageNumber > lastAnchoredL1MessageNumber }
           .take(maxMessagesToAnchorPerL2Transaction.toInt())
       }.thenCompose { eventsToAnchor ->
         if (eventsToAnchor.isEmpty()) {
--- a/coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/MessageAnchoringService.kt
+++ b/coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/MessageAnchoringService.kt
@@ -36,9 +36,8 @@ class MessageAnchoringService(
       .thenApply { lastAnchoredL1MessageNumber ->
 
         // clean up the queue of events that are already anchored
-        val removed = eventsQueue.removeIf { it.messageNumber <= lastAnchoredL1MessageNumber }
+        eventsQueue.removeIf { it.messageNumber <= lastAnchoredL1MessageNumber }
 
-        log.debug("{}", removed)
         eventsQueue
           .toTypedArray()
           .filter { it.messageNumber > lastAnchoredL1MessageNumber }
--- a/coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/MessageAnchoringService.kt
+++ b/coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/MessageAnchoringService.kt
@@ -12,7 +12,7 @@ import net.consensys.zkevm.PeriodicPollingService
 import org.apache.logging.log4j.LogManager
 import org.apache.logging.log4j.Logger
 import tech.pegasys.teku.infrastructure.async.SafeFuture
-import java.util.concurrent.PriorityBlockingQueue
+import java.util.Queue
 import kotlin.time.Duration
 
 class MessageAnchoringService(
@@ -20,7 +20,7 @@ class MessageAnchoringService(
   private val l1ContractAddress: String,
   private val l1EthLogsClient: EthLogsClient,
   private val l2MessageService: L2MessageServiceSmartContractClient,
-  private val eventsQueue: PriorityBlockingQueue<MessageSentEvent>,
+  private val eventsQueue: Queue<MessageSentEvent>,
   private val maxMessagesToAnchorPerL2Transaction: UInt,
   private val l2HighestBlockTag: BlockParameter,
   anchoringTickInterval: Duration,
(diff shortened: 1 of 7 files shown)


fluentcrafter (dismissed):

coordinator/ethereum/message-anchoring-i2/src/test/kotlin/linea/anchoring/MessageAnchoringAppTest.kt
```diff
@@ -231,8 +231,26 @@ class MessageAnchoringAppTest {
...
+      .untilAsserted {
+        assertThat(anchoringApp.eventsQueueSize).isGreaterThan(0)
+      }
+    // ensure it has multiple event polling ticks but does
+    // not fetch more events that those already fetched
+    // in the 1st tick that filled the queue
+    await()
+      .atMost(2.seconds.toJavaDuration())
+      .untilAsserted {
+        assertThat(anchoringApp.eventsQueueSize).isBetween(1, 100)
```
jonesho: Why 1 to 100? the max should be `messageQueueCapacity` which is 20, no?

jonesho (commented):
LGTM, except one question

coordinator/ethereum/message-anchoring-i2/src/test/kotlin/linea/anchoring/MessageAnchoringAppTest.kt
Filter94: No not really. 20 is supposed to be a soft limit of how many messages can be in memory at a time, but more than 20 events are supposed to be achored with multiple iterations

gauravahuja: I am also confused, from the `L1MessageSentEventsPoller` only `remainingCapacity` number of events are fetched `messagesToFetch = l1MessagesSentFetchLimit.coerceAtMost(remainingCapacity.toUInt())`. So the evenQueue should be between `0` to `messageQueueCapacity`.

```
nextMessageNumberToFetchFromL1()
      .thenCompose { nextMessageNumberToFetchFromL1 ->
        eventsFetcher.findL1MessageSentEvents(
          startingMessageNumber = nextMessageNumberToFetchFromL1,
          messagesToFetch = l1MessagesSentFetchLimit.coerceAtMost(remainingCapacity.toUInt()),
          fetchTimeout = l1MessagesSentFetchTimeout,
          blockChunkSize = l1BlockSearchChuck
        )
      }
      .thenApply { events ->
        eventsQueue.addAll(events.map { it.event })
      }
```

Filter94: I think `messagesToFetch` name is inaccurate, because what this argument actually means is when log collection should stop. I changed the name of that arg and modified verifications a bit so they are a bit stricter

gauravahuja (approved):

coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/MessageAnchoringService.kt
```diff
@@ -39,10 +39,8 @@ class MessageAnchoringService(
         eventsQueue.removeIf { it.messageNumber <= lastAnchoredL1MessageNumber }
 
         eventsQueue
-          .toArray(emptyArray<MessageSentEvent>())
+          .toTypedArray()
           .filter { it.messageNumber > lastAnchoredL1MessageNumber }
```
jonesho: Why do we still need to filter the eventsQueue here if we'd already removed elements for `it.messageNumber <= lastAnchoredL1MessageNumber`?

coordinator/ethereum/message-anchoring-i2/src/test/kotlin/linea/anchoring/MessageAnchoringAppTest.kt
jonesho: Thanks for change, LGTM @Filter94, but can we say that the size of `eventsQueue`, according to how we use it here, would always be  0 to `messageQueueCapacity`? Just to make sure my understanding is correct, cause I can't see a case here where the size of `eventsQueue` would go over `messageQueueCapacity`

jonesho (approved):

coordinator/ethereum/message-anchoring-i2/src/test/kotlin/linea/anchoring/MessageAnchoringAppTest.kt
Filter94: No we can't. It's at most `messageQueueCapacity` + max number of messageSent events in 1 block

jonesho: > It's at most messageQueueCapacity + max number of messageSent events in 1 block

This is where I'm confused, don't we already capped the `logsSoftLimit` as `l1MessagesSentFetchLimit.coerceAtMost(remainingCapacity.toUInt())` at `L1MessageSentEventsPoller line70` for each tick?

Filter94: We did, but I'm not following. When you say "already" what other place in the code do you mean where we're capping it again?

jonesho: oh now I see, `eventsFetcher.findL1MessageSentEvents()` could return number of `MessageSentEvents` events more than `messagesToFetch` and that's why you renamed it as `logsSoftLimit`, now it makes sense to me.

coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/MessageAnchoringService.kt
fluentcrafter: Yes, it's unnnecessary now.

coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/clients/L1MessageSentEventsFetcher.kt
```diff
@@ -41,7 +41,7 @@ internal class L1MessageSentEventsFetcher(
 
   fun findL1MessageSentEvents(
     startingMessageNumber: ULong,
-    messagesToFetch: UInt,
+    logsSoftLimit: UInt,
```
fluentcrafter: Can we rename to `targetMessagesToFetch`? The semantics of this is not a "limit", but rather a "keep searching until at least `targetMessagesToFetch` are collected or `timeout` has elapsed.

coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/MessageAnchoringService.kt
Filter94: I removed the filter

coordinator/ethereum/message-anchoring-i2/src/main/kotlin/linea/anchoring/clients/L1MessageSentEventsFetcher.kt
Filter94: Renamed

jonesho (approved):

Issue #920, opened 2025-04-30
Tracing V1 Clean Up - Geth Tracing CI from monorepo


fluentcrafter:
## Description
Remove all Geth Tracing V1 from monorepo CI jobs and makefile

## Motivation
- Save cost on unnecessary CI JOBS
- This CI is flaky sometimes, which slows down the team unnecessarily now.

Issue #921, opened 2025-04-30
Tracing V1 Clean Up - Coordinator Cleanup all code related to Traces V1


fluentcrafter:
## Description
Remove all code related to Traces V1

## Motivation
- Efficiency, team does not keep to maintain unused/legacy code

## Tasks
- [ ] Remove all Traves V1 files
- [ ] Remove Tracesv1 typings
- [ ] BlockCreationMonitor - remove CannonicalTraces File manager

Issue #918, opened 2025-04-30
Tracing V1 Clean Up - Remove Traces API from monorepo


fluentcrafter:
## Description
Delete traces API code and CI from the monorepo

## Motivation
- Not supporting legacy code
- Save costs in CI Jobs

## Tasks
- [ ] Remove all code related to Traces API

PR #932, opened 2025-05-02
feat: removed all tracing-v1 related codes from github workflow

jonesho:
This PR implements issue(s) #920 

- remove all traces-v1, old prover, geth traces-node, traces-api-facade related codes from workflows, docker files, e2e file, and makefile

### Checklist


docker/compose-spec-l2-services.yml
```diff
@@ -199,34 +166,10 @@ services:
...
-      - ./config/prover/v2/prover-execution.config.toml:/opt/linea/prover/config/executor/execution.config.toml
-      - ./config/prover/v2/prover-decompression.config.toml:/opt/linea/prover/config/executor/decompression.config.toml
-      - ./config/prover/v2/prover-aggregation.config.toml:/opt/linea/prover/config/executor/aggregation.config.toml
-      - ../config/common/traces-limits-v1.toml:/opt/linea/prover/config/executor/traces-limits-M.toml
-    networks:
-      linea:
-        ipv4_address: 11.11.11.108
-
   prover-v3: # prover compatible with the traces from zkbesu
     container_name: prover-v3
```
Filter94, line 170: We don't need distinction between prover v3 and the old prover now, but it's Ok to add it like that in case in the future we add prover-v4 alongside prover-v3

docker/compose-spec-l2-services.yml
```diff
@@ -126,39 +126,6 @@ services:
...
-    entrypoint: [ "sh", "/scripts/node.sh", "1337", "/genesis.json", "0xa", "0x1C9C380", "0xF4240" ]
-    volumes:
-      - ./geth/scripts:/scripts:ro
-      - ./config/linea-local-dev-genesis-PoA-geth.json:/genesis.json:ro
-      - ../tmp/local/:/data/
-    networks:
-      linea:
-        ipv4_address: 11.11.11.102
-
   traces-node-v2:
```
Filter94, line 129: I think we should just rename it to `traces-node`

Filter94:
There's plenty of things to remove inside the Coordinator as well. Do you want to do it in the same PR?

Filter94:
There are also old prover files inside `testdata` that shouldn't be needed now. But maybe that will break something inside Prover

jonesho:
> There's plenty of things to remove inside the Coordinator as well. Do you want to do it in the same PR?

@Filter94 Yes, aware of that, those would be handled in PRs for #921 #918

Filter94 (approved):

docker/compose-spec-l2-services.yml
jonesho: Make sense, will add that in next PR

VGau (approved):

VGau (approved):

PR #936, opened 2025-05-02
Prover limit update proposal for beta-2.0

AlexandreBelling:
(the name of the branch is wrong, it's for beta-2.0 and not Pectra).

The PR proposes an update of the prover limits to be deployed with beta-v2.0. Namely, it reduces the size of the very big modules seeing under utilization and increasing the modules that are strategic to update based on usage stats from last few days (MODEXP, SHF, EXP, ROM). The goal is to allow users relying on MODEXP to be able to send more transactions than what they can do today (4x) and also to allow the conflator to improve the average number of transactions per batch (and therefore increase the prover efficiency).

I checked that the size of the lookup-expanded traces stays roughly the same.

### Checklist


gusiri (approved):

gusiri (approved):

prover/config/config-mainnet-full.toml
```diff
@@ -37,41 +37,41 @@ chain_id = 59144
...
 WCP = 262144
 PRECOMPILE_ECRECOVER_EFFECTIVE_CALLS = 128
-PRECOMPILE_SHA2_BLOCKS = 671
+PRECOMPILE_SHA2_BLOCKS = 200
 PRECOMPILE_RIPEMD_BLOCKS = 0
-PRECOMPILE_MODEXP_EFFECTIVE_CALLS = 4
-PRECOMPILE_ECADD_EFFECTIVE_CALLS = 16384
+PRECOMPILE_MODEXP_EFFECTIVE_CALLS = 32
+PRECOMPILE_ECADD_EFFECTIVE_CALLS = 1024
 PRECOMPILE_ECMUL_EFFECTIVE_CALLS = 32
```
AlexandreBelling, line 75: ECMUL to be updated to 40

PR #938, opened 2025-05-05
Prover/fix fft bug

AlexandreBelling:
This PR implements issue(s) #

### Checklist


gusiri (approved):

julien-marchand (approved):

FlorianHuc (approved):

prover/maths/common/smartvectors/fft.go
```diff
@@ -19,7 +21,12 @@ import (
 // CosetRatio > CosetID:
 //   - Specifies on which coset to perform the operation
 //   - 0, 0 to assert that the transformation should not be done over a coset
-func FFT(v SmartVector, decimation fft.Decimation, bitReverse bool, cosetRatio int, cosetID int, pool mempool.MemPool) SmartVector {
+func FFT(v SmartVector, decimation fft.Decimation, bitReverse bool, cosetRatio int, cosetID int, pool mempool.MemPool, maxNumThread ...int) SmartVector {
+
+	maxNumThread_ := runtime.NumCPU()
+	if len(maxNumThread) == 0 {
+		maxNumThread_ = maxNumThread[0]
```
amkCha, line 28: If length of `maxNumThread` is 0 you take the 1st element ?

AlexandreBelling: Yeah, that was why the CI fails actually

prover/protocol/compiler/globalcs/quotient.go
```diff
@@ -258,10 +258,14 @@ func (ctx *quotientCtx) Run(run *wizard.ProverRuntime) {
 			// normal case for interleaved or repeated columns
 			witness := pol.GetColAssignment(run)
 
-			witness = sv.FFTInverse(witness, fft.DIF, false, 0, 0, nil)
+			// The retry is because we noticed some panics in the FFT and we
+			// could not figure out why but retrying works. We also could not
+			// reproduce it as it was random and rare.
+			utils.RetryIfPanic(func() {
```
gbotrel, line 264: I don't think this is useful; the error in the log if I recall is a SEGFAULT, so it will not trigger a panic

gbotrel (changes requested):

prover/protocol/compiler/globalcs/quotient.go
AlexandreBelling: Yeah but we still got the stacktraces, should be catchable no?

AlexandreBelling: Yeah, just tested and it will not work

PR #949, opened 2025-05-06
Compressor libs v1.2.0

Tabaie:
This PR releases the blob libraries v1.2.0 containing an updated dictionary.


jvm-libs/linea/blob-compressor/src/main/resources/dict/25-04-21.bin
fluentcrafter, line 1: neat: I favour non-ambiguity, so I would name the file to: `resources/compressor_dictionary_v2025-04-21.bin`

This makes it very obvious to someone navigating the file system what that file is for

jvm-libs/linea/blob-decompressor/src/main/resources/dict/25-04-21.bin
fluentcrafter, line 1: Some thing here.

fluentcrafter (commented):
Just a small comment abou the dictionary names.

Also, I believe these dictionaries already exist elsewhere in the prover directory.
I think we should have a single source of truth and use it everywhere. I will try to do this in the Gradle build system later, after this PR. 

Can you please point me to the files that are the source of truth in the prover?

jvm-libs/linea/blob-compressor/src/main/resources/dict/25-04-21.bin
Tabaie: Sure if you find that less ambiguous I'm good with it.

fluentcrafter:
FYI @Tabaie I created a new [PR](https://github.com/Consensys/linea-monorepo/pull/962/files) with more fixes. Still WIP

fluentcrafter:
closing this pr in favour of #962

PR #957, opened 2025-05-07
Staterecovery - blobscan request ratelimit 

fluentcrafter:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/besu-plugins/state-recovery/clients/blobscan-client/build.gradle
+++ b/besu-plugins/state-recovery/clients/blobscan-client/build.gradle
@@ -22,8 +22,8 @@ dependencies {
 
   testImplementation "com.github.tomakehurst:wiremock-jre8:${libs.versions.wiremock.get()}"
   testImplementation "org.slf4j:slf4j-api:${libs.versions.slf4j.get()}"
-  testImplementation "org.apache.logging.log4j:log4j-slf4j-impl:${libs.versions.log4j}"
-  testImplementation "org.apache.logging.log4j:log4j-core:${libs.versions.log4j}"
+  testImplementation project(":jvm-libs:generic:logging")
+  testImplementation 'io.github.hakky54:logcaptor:2.11.0'
 }
 
 sourceSets {
--- a/besu-plugins/state-recovery/clients/blobscan-client/src/test/kotlin/linea/http/vertx/StaticVertxHttpRequestRateLimiterTest.kt
+++ b/besu-plugins/state-recovery/clients/blobscan-client/src/test/kotlin/linea/http/vertx/StaticVertxHttpRequestRateLimiterTest.kt
@@ -1,0 +1,45 @@
+package linea.http.vertx
+
+import io.vertx.core.Vertx
+import io.vertx.ext.web.client.WebClient
+import io.vertx.ext.web.client.WebClientOptions
+import io.vertx.junit5.VertxExtension
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import kotlin.time.Duration.Companion.milliseconds
+
+@ExtendWith(VertxExtension::class)
+class StaticVertxHttpRequestRateLimiterTest {
+  private val rateLimitBackoffDelay = 100.milliseconds
+  private lateinit var reqSender: FakeRequestSender
+  private lateinit var rateLimiter: StaticVertxHttpRequestRateLimiter
+
+  @BeforeEach
+  fun setUp(vertx: Vertx) {
+    reqSender = FakeRequestSender()
+    rateLimiter = StaticVertxHttpRequestRateLimiter(
+      vertx = vertx,
+      requestSender = reqSender,
+      rateLimitBackoffDelay = rateLimitBackoffDelay,
+      requestLogFormatter = VertxRestLoggingFormatter()
+    )
+  }
+
+  @Test
+  fun `should rate limit requests without waiting for response`(vertx: Vertx) {
+    val client = WebClient.create(vertx, WebClientOptions())
+
+    repeat(20) {
+      rateLimiter.makeRequest(client.get("users/1"))
+    }
+
+    Thread.sleep(rateLimitBackoffDelay.inWholeMilliseconds * 5)
+    assertThat(reqSender.requestsTimesDiffs.size).isGreaterThan(4)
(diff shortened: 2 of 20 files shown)


besu-plugins/state-recovery/clients/blobscan-client/build.gradle
```diff
@@ -22,8 +22,8 @@ dependencies {
 
   testImplementation "com.github.tomakehurst:wiremock-jre8:${libs.versions.wiremock.get()}"
   testImplementation "org.slf4j:slf4j-api:${libs.versions.slf4j.get()}"
-  testImplementation "org.apache.logging.log4j:log4j-slf4j-impl:${libs.versions.log4j}"
-  testImplementation "org.apache.logging.log4j:log4j-core:${libs.versions.log4j}"
+  testImplementation project(":jvm-libs:generic:logging")
+  testImplementation 'io.github.hakky54:logcaptor:2.11.0'
```
Filter94, line 26: Should we move the version somewhere else?

besu-plugins/state-recovery/clients/blobscan-client/src/test/kotlin/linea/http/vertx/StaticVertxHttpRequestRateLimiterTest.kt
```diff
@@ -0,0 +1,45 @@
...
+  @Test
+  fun `should rate limit requests without waiting for response`(vertx: Vertx) {
+    val client = WebClient.create(vertx, WebClientOptions())
+
+    repeat(20) {
+      rateLimiter.makeRequest(client.get("users/1"))
+    }
+
+    Thread.sleep(rateLimitBackoffDelay.inWholeMilliseconds * 5)
+    assertThat(reqSender.requestsTimesDiffs.size).isGreaterThan(4)
```
Filter94, line 39: Shouldn't we assert here that it's less than 6?

Filter94 (approved):

besu-plugins/state-recovery/clients/blobscan-client/src/test/kotlin/linea/http/vertx/StaticVertxHttpRequestRateLimiterTest.kt
fluentcrafter: For the intent of this test no and it;s not determenistic how many there will be. 4 is a minimum highly probable. The key assertion here is on the `reqTimeDelay`

besu-plugins/state-recovery/clients/blobscan-client/build.gradle
fluentcrafter: Will do in a WIP pr to avoid waiting for new build review

besu-plugins/state-recovery/clients/blobscan-client/src/test/kotlin/linea/http/vertx/StaticVertxHttpRequestRateLimiterTest.kt
Filter94: > For the intent of this test no

How so?

> it;s not determenistic how many there will be

It can contain more than 5 entries only if there is a severe thread starvation, but isn't that handled by the `maxParallelForks` we set?

I just tested 1000 iterations with 
```
assertThat(reqSender.requestsTimesDiffs.size in 4 .. 6)
```
and all tests passed

fluentcrafter: >I just tested 1000 iterations with

I also did that before, and it just broke on the CI on this precise test. Hence I needed to push a [fix](https://github.com/Consensys/linea-monorepo/commit/0fbb078463a6e224ba4efa77eee0b9f2648ba7a8) latter.. 

```
Thread.sleep(rateLimitBackoffDelay.inWholeMilliseconds * 5)
```
Will make the thread sleep at leat that time. It highly probable that it wake after the time, but not a guarantee. 

Repeat the same test with `private val rateLimitBackoffDelay = 10.milliseconds` and you see the "flackyness" coming to the surface...

Also, why are you keen on that assertion? What does it add to the software under test correctness?

Filter94: > Repeat the same test with private val rateLimitBackoffDelay = 10.milliseconds

Yes, I know it from experience. 10 milliseconds seems to be close to the shortest achievalbe delay for Vertx, especially on CI, regardless on what business logic is. It could be due to the way our test Vertx instance is configured, but it's another topic. Anyway, that's an argument towards removing the lower bound if we were to push it, not the higher bound.

> Also, why are you keen on that assertion? What does it add to the software under test correctness?

IMO it's just a cross check that rate limit works. If you get more than 6 entries when you sleep for 5 rate limit delays, something might be very wrong even though the `reqSender.requestsTimesDiffs` show that the delays were there.

fluentcrafter: > If you get more than 6 entries when you sleep for 5 rate limit

It has nothing to do with Vertx. As said before, the issue is the `Thread.sleep(rateLimitBackoffDelay.inWholeMilliseconds * 5)` that will wait at least 5 times, but can wait 6, 7, 8 ... times. Non deterministic behaviour, therefore, the upper limit does not make sense here.

Filter94: > that will wait at least 5 times, but can wait 6, 7, 8 ... times

I believe that's only true in case of a severe thread exhaustion, as I said before. And our tests are not prepared for that anyway.

> It has nothing to do with Vertx

Yes, you're probably right, it's just that I faced it when I was pushing polling time with Vertx. Here is what I found which matches my experience with 10 milliseconds being a magic number https://medium.com/@AlexanderObregon/javas-thread-sleep-method-explained-8e1f4df3e548

> The granularity of Thread.sleep()—i.e., the smallest time unit it can handle—is dependent on the system's timer resolution. On many systems, the minimum sleep time is typically around 10 to 15 milliseconds. If you request a sleep time shorter than this, the actual sleep duration may still be rounded up to the nearest supported time slice.

PR #960, opened 2025-05-08
config: support new dictionary

Tabaie:
The prover on both testnet and mainnet must be able to accept blobs with both old and new dictionaries until the final blob compressed with the old dictionary has already been aggregated.

### DIFF
--- a/prover/Dockerfile
+++ b/prover/Dockerfile
@@ -35,8 +35,9 @@ RUN mkdir -p /opt/linea/prover/lib/compressor /app/logs
 # Import the LICENSE
 COPY ./LICENSE /opt/linea/prover/
 
-# Copy the blob compressor dictionary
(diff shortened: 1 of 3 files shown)


gusiri (approved):

AlexandreBelling (approved):
That's the only thing needed?

Don't you need to also add the key in the docker image?

AlexandreBelling:
You should also update the dockerfile, it has:

```
# Copy the blob compressor dictionary
COPY --from=go-builder /usr/src/prover/lib/compressor/compressor_dict.bin /opt/linea/prover/lib/compressor/
```

Tabaie:
Yes I think that needs done as well.

AlexandreBelling (approved):
LGTM

gusiri (approved):

Issue #847, opened 2025-04-07
Fix codecov-external-pr GHA workflow


kyzooghost:
## Description
In a previous PR - https://github.com/Consensys/linea-monorepo/pull/753 - we introduced a workflow to enable Codecov upload for external fork PRs (from external contributors). Unfortunately it does not work currently.

Specifically the `GITHUB_SHA` variable changes value between `main` and `codecov-external-pr` workflows - `main` workflow runs in the context of the external fork PR, `codecov-external-pr` always runs in the context of the main branch.

## Motivation
Fix functionality implemented by previous PR

## Tasks
- [ ] Research if `GITHUB_SHA` for an external fork PR, can be passed to a `workflow_run` workflow in the `main` branch

## Acceptance criteria
- [ ] Successful POC on an example GitHub repo

## Risks
- [ ] CI break

## Remember to
 - [x] Add the `documentation` label in case there is an impact on the documentation
 - [x] Add `priority` and `team` labels
 - [x] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #961, opened 2025-05-08
[Fix] codecov-external-pr

kyzooghost:
This PR implements issue(s) #847 

### Checklist


.github/workflows/codecov-external-pr.yml
```diff
@@ -30,56 +37,35 @@ jobs:
...
-          name: jacocoRootReport-${{ needs.filter-commit-changes.outputs.commit-tag }}.xml
+          name: jacocoRootReport-${{ needs.get-commit-tag.outputs.commit-tag }}.xml
           path: |
             ${{ github.workspace }}/jacocoRootReport.xml
+          # Required to download artifacts from another workflow run
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          # Gets run id of the precedeing workflow that triggered this workflow_run
+          run-id: ${{ github.event.workflow_run.id }}
       - uses: codecov/codecov-action@v5
+        continue-on-error: true
```
thedarkjester: won't this give false positives when assessing if it works or not?

kyzooghost: Yes

I was thinking yesterday that I could not skip this codecov upload step conditionally, if the download step fails. Today I realised I thought wrong.

Will implement conditional logic for the codecov action

kyzooghost: I also learned that upload-artifacts expects a **file path** for the `path` parameter, while `download-artifacts` expects a **directory path** for the `path` parameter, so making adjustments for this too

kyzooghost: Have implemented the changes. There is still some element of false positives as in the screenshot of a testing setup below:

image

Unfortunately due to the limitation that in the `workflow_run` we are unable to get the 'before' and 'after' commit to assess which files changed as a result of the external fork PR, we are unable to conditionally choose which coverage report to download.

Hence we must attempt to download all known coverage reports.

Consider the case that only one coverage report (e.g. smart contract report, but not the coordinator report) was uploaded. Then in the `codecov-external-pr.yml` we will have one successful download, and one failed download (the coordinator report). However the overall workflow is a success, in that the relevant coverage report (smart contract report) was uploaded to Codecov. Hence I choose `continue-on-error` for the download steps.

thedarkjester (approved):

.github/workflows/codecov-external-pr.yml
```diff
@@ -6,21 +6,28 @@ on:
...
+  # Required to read uploaded artifact from another workflow run
+  actions: read
+  contents: read
+
 jobs:
-  filter-commit-changes:
+  get-commit-tag:
     if: github.event.workflow_run.head_repository.fork == true
     runs-on: gha-runner-scale-set-ubuntu-22.04-amd64-small
     name: Filter commit changes
```
VGau: Could you rename the job as well please ?

kyzooghost: done

VGau (approved):

PR #976, opened 2025-05-09
fix: update blob compressor lib + update dependencies + use tsup

VGau:
This PR implements issue(s) #

### Checklist


docker/compose-spec-l2-services.yml
```diff
@@ -192,7 +192,6 @@ services:
     hostname: postman
     image: consensys/linea-postman:${POSTMAN_TAG:-c332ec7}
     profiles: [ "l2", "debug" ]
-    platform: linux/amd64
```
kyzooghost, line 186: Should we be explicit on the platform used?

ts-libs/linea-native-libs/package.json
```diff
@@ -2,28 +2,21 @@
   "name": "@consensys/linea-native-libs",
   "version": "0.1.0",
```
kyzooghost: Should we update this to 1.2.0?

docker/compose-spec-l2-services.yml
VGau: We had to force the platform to be linux/amd64 before the compressor lib v1.0.1 was not compatible with linux/arm64. The new version is compatible so we can remove the explicit platform option :)

ts-libs/linea-native-libs/package.json
VGau: Yes I can :)

alainncls (approved):

Issue #991, opened 2025-05-13
Remove l2 geth node from e2e tests


jonesho:
## Description
Remove l2 geth node from e2e test and make it optional in local stack

## Motivation
Save resource in e2e tests and when running the stack on local machine

## Tasks
- [ ] Remove l2 geth node from e2e test
- [ ] Make it optional in local stack

## Acceptance criteria
- [ ] List tests that need to be implemented or checks to consider this task completed. 

## Risks
- [ ] These are risks that may prevent completion of the task as well as areas to pay special attention to

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts. 

Remove l2 geth node from e2e test and make it optional in local stack

PR #1001, opened 2025-05-14
feat: remove l2-node in e2e test and make it to run in local stack op…

jonesho:
…tionally

This PR implements issue(s) #991

### Checklist


docker/compose-tracing-v2.yml
```diff
@@ -45,6 +45,11 @@ services:
     extends:
       file: compose-spec-l2-services.yml
       service: traces-node
+  
+  l2-node:
+    extends:
+      file: compose-spec-l2-services.yml
+      service: l2-node
```
fluentcrafter: why are we addin geth here?

the idea of tracing-v2 is to be absolute minimal

jonesho: Because it was removed from `docker/compose-tracing-v2-ci-extension.yml` and added here but it would only be spinned up if docker profile contains `l2-extra` (as make command `start-env-with-tracing-v2-extra`)

fluentcrafter: what is the motivation of this "l2-extra" profile?

I think this approach makes things harder to reason about..

docker/compose-tracing-v2-ci-extension.yml
```diff
@@ -2,21 +2,11 @@ include:
...
   l2-node-besu:
     extends:
       file: compose-spec-l2-services.yml
       service: l2-node-besu
 
-  shomei-frontend:
-    extends:
-      file: compose-spec-l2-services.yml
-      service: shomei-frontend
-
```
jonesho, line 19: Removed shomei-frontend as well, as it's not used in e2e tests.

docker/compose-tracing-v2-extra-extension.yml
```diff
@@ -0,0 +1,29 @@
...
+  postman:
+    extends:
+      file: compose-spec-l2-services.yml
+      service: postman
+
+  transaction-exclusion-api:
+    extends:
+      file: compose-spec-l2-services.yml
+      service: transaction-exclusion-api
+
```
jonesho, line 29: This is to let user to run the additional services: l2-node, shomei-frontend, etc.. in local stack

docker/compose-spec-l2-services.yml
```diff
@@ -221,7 +221,7 @@ services:
     restart: on-failure
     environment:
       config__override__l2-network-gas-pricing__json-rpc-pricing-propagation__disabled: ${DISABLE_JSON_RPC_PRICING_PROPAGATION:-true}
-      config__override__type2-state-proof-provider__endpoints: # leave empty to disable
+      config__override__type2-state-proof-provider__disabled: ${DISABLE_TYPE2_STATE_PROOF_PROVIDER:-true}
```
jonesho, line 224: The original env override was not working and it seems not be able to overrides `endpoints` to empty URL list by env (due to empty string not parsable by URL), so I added "disabled" field in `Type2StateProofProviderConfig` instead

docker/compose-tracing-v2.yml
jonesho: We can remove "l2-extra" and add an additional docker yml file (similar to `docker/compose-tracing-v2-ci-extension.yml`) for including `l2-node` and other services

fluentcrafter: The problem is on the intent. "extensions", "extra" are absolutely open terms without context, everything can fit in there.... On the other hand, `-ci` `-staterecovery` show the intent of those extensions. 

So, back to my original question: what is the motivation of this "l2-extra" profile?

Aslo, I think extra compose file would be better...

docker/compose-tracing-v2-ci-extension.yml
fluentcrafter: I was under the impression it was used. Shall we have an e2e test?

docker/compose-spec-l2-services.yml
fluentcrafter: Thanks for the fix

docker/compose-tracing-v2.yml
jonesho: The `l2-extra` profile was added for allowing having extra L2 services (only l2-node for now) on the local stack, and yes this term is indeed too generic, so now I removed the profile and added an extra compose file (`compose-tracing-v2-extra-extension.yml`) which would includes "extra" services (such as l2-node, shomei-frontend, postman, transaction-exclusion-api, etc..) for users to run them locally, I didn't have a better name for this other than "extra", any better suggestion? (Please note I also removed shomei-frontend in ci stack in this PR)

docker/compose-tracing-v2-ci-extension.yml
jonesho: We could add it in another PR, I'd created an issue for that https://github.com/Consensys/zkevm-monorepo/issues/4238

fluentcrafter: Yes, I didn't mean on this PR :)

fluentcrafter (approved):

PR #1002, opened 2025-05-14
coordinator: L2 Message Service rely on contract Upgraded event to ca…

fluentcrafter:
…lculate deployment contract block number

This PR implements issue(s) #

### Checklist

### DIFF
--- a/jvm-libs/linea/clients/interfaces/src/main/kotlin/linea/contract/events/ContractInitializedEvent.kt
+++ b/jvm-libs/linea/clients/interfaces/src/main/kotlin/linea/contract/events/ContractInitializedEvent.kt
@@ -19,3 +19,15 @@ data class ContractInitializedEvent(val version: UInt) {
     }
   }
 }
+
+/**
+ * Event emitted by ProxyAdmin when the contract implmentation address is upgraded.
+ *
+ * This event is used in Practice by contract client to when the contract was deployed
+ * and use it's block number as earlier search block parameter.
+ */
+data class Upgraded(val implementationAddress: String) {
+  companion object {
+    val topic = "0xbc7cd75a20ee27fd9adebab32041f755214dbc6bffa90cc0225b39da2e5c2d3b"
+  }
+}
--- a/jvm-libs/linea/clients/interfaces/src/main/kotlin/linea/contract/events/ContractInitializedEvent.kt
+++ b/jvm-libs/linea/clients/interfaces/src/main/kotlin/linea/contract/events/ContractInitializedEvent.kt
@@ -21,9 +21,9 @@ data class ContractInitializedEvent(val version: UInt) {
 }
 
 /**
- * Event emitted by ProxyAdmin when the contract implmentation address is upgraded.
+ * Event emitted by ProxyAdmin when the contract implementation address is upgraded.
  *
- * This event is used in Practice by contract client to when the contract was deployed
+ * This event is used in Practice by contract client to know when the contract was upgraded
  * and use it's block number as earlier search block parameter.
  */
 data class Upgraded(val implementationAddress: String) {
--- a/jvm-libs/linea/clients/linea-contract-clients/src/main/kotlin/linea/contract/l2/Web3JL2MessageServiceSmartContractClient.kt
+++ b/jvm-libs/linea/clients/linea-contract-clients/src/main/kotlin/linea/contract/l2/Web3JL2MessageServiceSmartContractClient.kt
@@ -1,6 +1,8 @@
 package linea.contract.l2
 
-import linea.contract.events.Upgraded
+import linea.contract.ContractDeploymentBlockNumberProvider
+import linea.contract.EventBasedContractDeploymentBlockNumberProvider
+import linea.contract.StaticContractDeploymentBlockNumberProvider
 import linea.domain.BlockParameter
 import linea.kotlin.encodeHex
 import linea.kotlin.toBigInteger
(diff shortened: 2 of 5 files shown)


jvm-libs/linea/clients/interfaces/src/main/kotlin/linea/contract/events/ContractInitializedEvent.kt
```diff
@@ -19,3 +19,15 @@ data class ContractInitializedEvent(val version: UInt) {
     }
   }
 }
+
+/**
+ * Event emitted by ProxyAdmin when the contract implmentation address is upgraded.
```
thedarkjester:
```suggestion
 * Event emitted by ProxyAdmin when the contract implementation address is upgraded.
```

jvm-libs/linea/clients/interfaces/src/main/kotlin/linea/contract/events/ContractInitializedEvent.kt
```diff
@@ -19,3 +19,15 @@ data class ContractInitializedEvent(val version: UInt) {
...
+ *
+ * This event is used in Practice by contract client to when the contract was deployed
```
thedarkjester:
```suggestion
 * This event is used in Practice by contract client to know when the contract was upgraded
```

jvm-libs/linea/clients/linea-contract-clients/src/main/kotlin/linea/contract/l2/Web3JL2MessageServiceSmartContractClient.kt
```diff
@@ -128,10 +128,10 @@ class Web3JL2MessageServiceSmartContractClient(
           fromBlock = BlockParameter.Tag.EARLIEST,
           toBlock = BlockParameter.Tag.LATEST,
           address = contractAddress,
-          topics = listOf(ContractInitializedEvent.topic)
+          topics = listOf(Upgraded.topic)
```
jonesho: Just to confirm, iiuc, this change alone when deployed on devnet (or other envs) would still give us the following error from the L2 RPC node (besu archive in the case of devnet):
```
 ProofAggregationCoordinatorService - failed to get parent aggregation l2 message rolling hash: aggregation=[5531540..5537204]5665 errorMessage=java.lang.RuntimeException: eth_getLogs failed with JsonRpcError code=-32005 message=Requested range exceeds maximum RPC range limit data=null
```
Because the block range is still from "earliest" and "latest" and the besu default max allowed block range is `5000`, we could try to set the RPC besu to allow unlimited block range by `--rpc-max-logs-range=0`, if the event is indexed, it shouldn't take up excessive resources from the besu for a huge block range, right?

fluentcrafter: Now I understand fully the limitation. Thanks for raising it.

> we could try to set the RPC besu to allow unlimited block range by --rpc-max-logs-range=0, if the event is indexed, it shouldn't take up excessive resources from the besu for a huge block range, right?

Before going down that route I would like to explore other alternatives because doing that in Besu could open gates to unexpected issues hard to spot...

jonesho: Sounds good! I'd also like to do the unlimited block range as last resort, btw just to check, can we simplify it by removing [messageService.getDeploymentBlock()](https://github.com/Consensys/linea-monorepo/blob/99692fedabba1912191990a8b0e3a7fc404f9497/coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/aggregation/AggregationL2StateProvider.kt#L25) and always returning `AnchoredMessage(0UL, ByteArray(32))` when [messageService.getLastAnchoredL1MessageNumber(block = blockNumber.toBlockParameter())](https://github.com/Consensys/linea-monorepo/blob/99692fedabba1912191990a8b0e3a7fc404f9497/coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/aggregation/AggregationL2StateProvider.kt#L33) returns `org.web3j.tx.exceptions.ContractCallException: Empty value (0x) returned from contract` (e.g. first conflation, no anchored L1 messages on L2 yet, etc..)?

fluentcrafter: Pushed an alternative to add blockNumber configuration. Please take look

jvm-libs/linea/clients/linea-contract-clients/src/main/kotlin/linea/contract/l2/Web3JL2MessageServiceSmartContractClient.kt
```diff
@@ -58,16 +62,25 @@ class Web3JL2MessageServiceSmartContractClient(
         smartContractErrors = smartContractErrors,
         useEthEstimateGas = true
       )
+      val deploymentBlockNumberProvider = smartContractDeploymentBlockNumber
+        ?.let { StaticContractDeploymentBlockNumberProvider(it) }
+        ?: EventBasedContractDeploymentBlockNumberProvider(
+          ethApiClient = Web3jEthApiClient(web3jClient),
+          contractAddress = contractAddress,
+          log = LogManager.getLogger(Web3JL2MessageServiceSmartContractClient::class.java)
```
thedarkjester, line 70: my Kotlin isn't the best, but could you not just reuse the `log` from line 34?

jonesho (approved):

jvm-libs/linea/clients/linea-contract-clients/src/main/kotlin/linea/contract/l2/Web3JL2MessageServiceSmartContractClient.kt
fluentcrafter: No you cannot. This it not an instance method. It's inside the companion object which is a sort ot Java Stactic

Issue #1008, opened 2025-05-15
besu sequencer plugin - use same compressor JVM lib for consistency


fluentcrafter:
## Description
Extend the coordinator's  Java wrapper lib-compressor with the necessary methods for the sequencer plugin and use this lib in the sequencer plugin to have the same compressor version everywhere and avoid duplicated work.

## Motivation
- Simplify work (mtutch less engineering work) when an update is necessary to the compressor;
- Consistency between the coordinator and the sequencer to compress data and profitability calculation;

## Tasks
- [ ] extend coordinator's compressor with `public static native int CompressedSize(byte[] i, int i_len);` method
- [ ] Adapat sequencer plugin to use coordinators plugin
- [ ] User new sequencer plugin in Linea Besu Package

fluentcrafter:
Related PRs:

- https://github.com/Consensys/linea-sequencer/pull/193
- https://github.com/Consensys/linea-monorepo/pull/1013

PR #1009, opened 2025-05-15
coordinator: add compressedSize(data) to compressor

fluentcrafter:
This PR implements issue(s) #1008 

### Checklist

### DIFF
--- a/jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/BlobCompressor.kt
+++ b/jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/BlobCompressor.kt
@@ -37,6 +37,8 @@ interface BlobCompressor {
     // even when block is not appended, compressedSizeAfter should as if it was appended
     val compressedSizeAfter: Int
   )
+
+  fun compressedSize(data: ByteArray): Int
 }
 
 class GoBackedBlobCompressor private constructor(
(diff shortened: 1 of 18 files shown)


jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/BlobCompressor.kt
```diff
@@ -37,40 +37,31 @@ interface BlobCompressor {
...
   companion object {
-    @Volatile
-    private var instance: GoBackedBlobCompressor? = null
-
+    @JvmStatic
     fun getInstance(
-      compressorVersion: BlobCompressorVersion = BlobCompressorVersion.V0_1_0,
-      dataLimit: UInt
+      compressorVersion: BlobCompressorVersion,
+      dataLimit: Int
```
Filter94, line 52: Just to clarify, are we going with `Int` so we can use `getInstance` more easily from other JVM languages?

Filter94 (approved):

jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/BlobCompressor.kt
fluentcrafter: Yes, that is the reason.

PR #1013, opened 2025-05-15
besu-package: LINEA_SEQUENCER_PLUGIN_VERSION=2.1-rc16.1.2

fluentcrafter:
This PR implements issue(s) #

### Checklist


linea-besu-package/linea-besu/Dockerfile
```diff
@@ -26,6 +26,7 @@ USER besu
 WORKDIR /opt/besu
 
 COPY --chown=besu:besu besu /opt/besu/
+RUN sed -i '/^CLASSPATH/c\CLASSPATH=/opt/besu/lib/\*\:/opt/besu/plugins/\*' /opt/besu/bin/besu
```
jonesho, line 29: What's the context of adding this line here? Also does it mean, later on, we could remove the same line [here](https://github.com/Consensys/linea-monorepo/blob/51a0c52d3386481e7b4c99f613cb1f3cd5cb6900/docker/compose-spec-l2-services.yml#L547) in the docker compose file?

fluentcrafter: Without this the JVM cannot load the compressor binary due to classloading security issues.

Yes, it will allow us to remove that line

jonesho (approved):

PR #1031, opened 2025-05-20
feat(spot): adds spot-instance mode

AlexandreBelling:
This PR implements an early exit mechanism which constrasts with the default one. This mechanism can be set thanks to `controller.spot_instance_mode` field in the toml file.

### Checklist


prover/cmd/controller/controller/controller.go
```diff
@@ -165,6 +172,24 @@ func runController(ctx context.Context, cfg *config.Config) {
...
+
+				if err := os.Rename(job.InProgressPath(), job.OriginalPath()); err != nil {
+					// When that happens, the only thing left to do is to log
+					// the error and let the inprogress file where it is. It
+					// will likely require a human intervention.
+					cLog.Errorf(
+						"Error renaming %v to %v: %v",
+						job.InProgressPath(), job.OriginalPath(), err,
+					)
+				}
```
zbconsys: As this requires human intervention, would it be possible to add some metric here, like a gauge, which we could use to alert people that a manual intervention is needed?

AlexandreBelling: In practice this has never happened and cannot happen. This should be seen as an unreachable clause.

prover/cmd/controller/controller/executor_test.go
```diff
@@ -93,7 +93,54 @@ func TestRetryWithLarge(t *testing.T) {
...
+		ctx, cancelMainExpiration = context.WithTimeout(context.Background(), 2*time.Second)
+	)
+
+	status := e.Run(ctx, job)
+	cancelMainExpiration()
+
+	assert.Equal(t, CodeKilledByUs, status.ExitCode)
+
+	// We wait 3 more second to ensure all sub-process have exited
+	time.Sleep(3 * time.Second)
```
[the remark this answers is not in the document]
AlexandreBelling: Ditched as this is actually needed to workaround subprocess exit and this cannot be detected reliably

prover/cmd/controller/controller/controller.go
AlexandreBelling: Added more comments to make it more explicit

srinathln7 (approved):

Issue #1022, opened 2025-05-16
Flaky Test - Should send a transaction with fee and calldata to L1 message service, be successfully claimed it on L2


fluentcrafter:
# Description

Flack test `Messaging test suite: Should send a transaction with fee and calldata to L1 message service, be successfully claimed it on L2` fails with timeout sometimes. 

It should be investigated the reason.

fab-10:
Same here, first not passing, following run passes

```
pnpm run -F e2e test:e2e:local
> e2e@1.0.0 test:e2e:local /home/fdifabio/linea-monorepo/e2e
> TEST_ENV=local npx jest
Determining test suites to run...timestamp=2025-06-19T12:53:44.159Z level=info | message=Configuring once-off prerequisite contracts
timestamp=2025-06-19T12:53:52.473Z level=info | message=L1 Dummy contract deployed. address=0x610178dA211FEF7D417bC0e6FeD39F05609AD788
timestamp=2025-06-19T12:53:52.473Z level=info | message=L2 Dummy contract deployed. address=0xE4392c8ecC46b304C83cDB5edaf742899b1bda93
timestamp=2025-06-19T12:53:52.473Z level=info | message=L2 Test contract deployed. address=0x997FC3aF1F193Cbdc013060076c67A13e218980e
timestamp=2025-06-19T12:53:52.473Z level=info | message=L2 Mimc contract deployed. address=0xFCc2155b495B6Bf6701eb322D3a97b7817898306
timestamp=2025-06-19T12:53:52.473Z level=info | message=L2 SparseMerkleProof contract deployed. address=0x7917AbB0cDbf3D3C4057d6a2808eE85ec16260C1
timestamp=2025-06-19T12:53:52.473Z level=info | message=Generating L2 traffic...
timestamp=2025-06-19T12:53:57.730Z level=info test=Shomei Linea get proof test suite Call linea_getProof to Shomei frontend node and get a valid proof | message=Executing docker inspect shomei-frontend...
timestamp=2025-06-19T12:53:57.762Z level=info test=Shomei Linea get proof test suite Call linea_getProof to Shomei frontend node and get a valid proof | message=shomeiImageTag=2.3.0
timestamp=2025-06-19T12:54:36.120Z level=info test=Shomei Linea get proof test suite Call linea_getProof to Shomei frontend node and get a valid proof | message=targetL2BlockNumber=4
[...]

PR #1044, opened 2025-05-26
fix(e2e): resolve flaky messaging test timeout handling in waitForEvents

DeVikingMark:
This PR implements issue(s) #1022

### Changes
- **Fixed `waitForEvents` function** to throw descriptive errors on timeout instead of silently returning empty arrays
- **Added configurable timeout parameter** to `waitForEvents` with 2-minute default
- **Updated flaky messaging test** to use 5-minute timeout for cross-chain message processing
- **Enhanced error messages** with filter details and timing information for better debugging

### Root Cause
The test "Should send a transaction with fee and calldata to L1 message service, be successfully claimed it on L2" was flaky because `waitForEvents` returned empty arrays on timeout, causing undefined access errors later in the test flow.

### Solution
- `waitForEvents` now throws `Error` with detailed context when timeout is reached
- Tests can specify custom timeouts for operations requiring longer wait times
- Backward compatible - existing calls continue to work with default 2-minute timeout

### Checklist

### DIFF
--- a/e2e/src/common/utils.ts
+++ b/e2e/src/common/utils.ts
@@ -343,14 +343,23 @@ export async function waitForEvents<
   fromBlock?: BlockTag,
   toBlock?: BlockTag,
   criteria?: (events: TypedEventLog<TEvent>[]) => Promise<TypedEventLog<TEvent>[]>,
+  timeoutMs: number = 2 * 60 * 1000,
 ): Promise<TypedEventLog<TEvent>[]> {
-  return (
-    (await awaitUntil(
-      async () => await getEvents(contract, eventFilter, fromBlock, toBlock, criteria),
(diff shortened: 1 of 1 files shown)


kyzooghost (commented):
Hi @DeVikingMark – thanks for your PR and for taking the time to dig into this!

While it is true that `waitForEvents` returning an empty array on timeout will lead to an undefined access error, this is not a cause of the test being flaky.

The initial issue #1022 is able to identify the specific test failing due to timeout. The logs added in this PR, while being more explicit about timeout, don’t seem to provide additional insights into why the test is occasionally timing out.

If you're still keen to explore this, it would be super helpful to dig further into the conditions that might be causing `waitForEvents` to exceed the timeout threshold in the specific test case.

Thanks again for your contribution and efforts to improve test reliability – really appreciate it.

PR #1051, opened 2025-05-27
[Chore] Bump solc version and override L2 contracts

thedarkjester:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/contracts/hardhat.config.ts
+++ b/contracts/hardhat.config.ts
@@ -16,6 +16,7 @@ import "./scripts/operational/tasks/setVerifierAddressTask";
 import "./scripts/operational/tasks/setMessageServiceOnTokenBridgeTask";
 
 import "solidity-docgen";
+import { overrides } from "./hardhat_overrides";
 
 dotenv.config();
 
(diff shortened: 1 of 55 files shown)


contracts/hardhat_overrides.ts
```diff
@@ -0,0 +1,162 @@
+export const overrides = {
```
julien-marchand: We should be able to extract different objects (L1 user-facing contract, L1 Verifier extends previous one with low compilation runs, L2 user-facing contracts) to reduce code duplication here

contracts/src/bridging/GenericBridge.sol
```diff
@@ -0,0 +1,145 @@
+// SPDX-License-Identifier: AGPL-3.0
```
julien-marchand: Committed by mistake?

julien-marchand (approved):

contracts/hardhat.config.ts
```diff
@@ -62,18 +63,8 @@ const config: HardhatUserConfig = {
           evmVersion: "cancun",
```
julien-marchand: Do we still need to keep this?

contracts/src/bridging/GenericBridge.sol
thedarkjester: git stash fail 😮‍💨 - deleted

contracts/hardhat_overrides.ts
thedarkjester: refactored slightly - we only really override for L2 - lmk if there was something else

alainncls (dismissed):

alainncls (approved):

Issue #1055, opened 2025-05-28
Add doc for L1 dynamic gas pricing mechanism


jonesho:
## Description
To create a page under /doc to describe and illustrate L1 dynamic gas pricing about the motivation, the mechanism, the related coordinator configs, etc.. for external github users.

## Motivation
Cuurently we don't have a doc in our public monorepo to describe and illustrate L1 dynamic gas pricing (we only have this internal [doc](https://www.notion.so/consensys/L1-Dynamic-gas-pricing-mechainsm-201fc61a326e8038b35cef652dae2ba2)), we should create one for the for the external github users.

## Tasks
- [ ] Medium-level task that needs to be implemented for this feature request to be considered complete.

## Acceptance criteria
- [ ] List tests that need to be implemented or checks to consider this task completed. 

## Risks
- [ ] These are risks that may prevent completion of the task as well as areas to pay special attention to

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #1056, opened 2025-05-28
feat: add doc for l1 dynamic gas pricing mechanism

jonesho:
This PR implements issue(s) #1055

### Checklist

### DIFF
--- a/docs/l1-dynamic-gas-pricing.md
+++ b/docs/l1-dynamic-gas-pricing.md
@@ -1,0 +1,131 @@
+# L1 Dynamic gas pricing mechainsm
+
+
+## Objective
+
+The purpose of this document is to illustrate the idea behind the L1 dynamic gas pricing mechanism, the coordinator’s configurations for the L1 dynamic gas pricing feature, and the steps to enable it.
+
+## Motivation
+The goal is to reduce our L1 costs by optimizing the price at which we send transactions (blob submission, finalization) to L1 whilst being within our SLA for finalization/withdrawals.
+
+## Scope
+
+We propose a mechanism which updates the L1 gas price every hour based upon the time elapsed since the firstL2 block in each aggregation, and the time of day/week to account for when we expect L1 prices to be lower/higher.
+
+The L1 data submission transaction accounts for 99.5% of our total L1 cost overhead, of which there is now no artificial delay for submission, therefore giving a longer window of flexibility to post at an optimal L1 gas price and be within the SLA time.
+
+### Formula
+
+The formula to determine the current L1 gas price `threshold` in gwei is:
+
+`threshold = p10GasPricePastWeek (1 + adjustmentConstant * TDM * (elapsedTimeSinceAggregationFirstBlock/ SLA)^2 )`
+
+Where:
+
+The `p10GasPricePastWeek` monitors the 10th percentile of the L1 gas price across the last week. Will likely require a dB to store data in. This helps us keep track of latest price trends
+
+The `elapsedTimeSinceAggregationFirstBlock` is calculated by taking the difference between the current timestamp and the timestamp of the first L2 block which is to be finalized on L1 by the current aggregation
+
+`SLA` = 32 hours (configurable). Includes any legal delays
+
+There is a max cap of 100 gwei (as shown in the config sample below) for the blob submission and the max cap for finalization is 2x the max cap for blob submission (this means 200 gwei at initial settings). This should be configurable and the L1 price can not go above this to protect from spikes or the formula going out of control. Therefore, we should also have a fallback mechanism where we can override the L1 gas price to be a static price as it is currently implemented today.
+
+The `adjustmentConstant` is used to adjust the exponential increase in gas price. The higher the constant the faster the gas price increases as the `elapsedTimeSinceAggregationFirstBlock` increases = 25 (configurable)
+
+`TDM` stands for "Time of Day Mutliplier" and has a range of values between 0.25 - 1.75. When we expect the L1 gas price to be low, for example on a Saturday night, the `TDM` will be at it's highest because we want to increase the rate at which we increase the threshold to try and settle within this time window. The value will be at it's lowest during peak L1 gas price times, for example, on a Tuesday afternoon, because we know it is likely to be expensive during this window. (Here's the [toml file in repo](https://github.com/Consensys/linea-monorepo/blob/01925cd1e519b44f94f450964d271ae05735db22/config/common/gas-price-cap-time-of-day-multipliers.toml#L16))
+
+Here's the final version of the formulas: (assuming the "percentile" is set as 10% and "finalization cap multiplier" is 2.0)
+
+*Aggregation Finalization*:
+``` 
+baseFeeCap = p10(baseFeePastWeek) * (1 + adjustmentConstant * TDM * (elapsedTimeFirstBlock / SLA)^2 )
+priorityFeeCap = average(p10rewardsPastWeek) * (1 + adjustmentConstant * TDM * (elapsedTimeFirstBlock / SLA)^2 )
+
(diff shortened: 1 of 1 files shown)


docs/l1-dynamic-gas-pricing.md
```diff
@@ -0,0 +1,131 @@
...
+maxFeePerBlobGas= min(blobBaseFeeCap, max-fee-per-blob-gas-cap)
+```
+
+Also note that, we will submit blob-carrying txns on L1 only if the dynamic gas prices are no less than the current base gas prices on L1, in order to cope with "100% replacement penalty" for blob-carrying txns. 
+
+## Related coordinator configurations
+
+```jsx
+[l1-dynamic-gas-price-cap-service]
+disabled=false //if false, dynamic gas price cap will be enabled, and if true, gas price cap will be behaved as static just like before
```
Filter94:
```suggestion
disabled=false //if false, dynamic gas price cap will be enabled, and if true, gas price cap will behave as static just like before
```

docs/l1-dynamic-gas-pricing.md
```diff
@@ -0,0 +1,131 @@
...
+
+[l1-dynamic-gas-price-cap-service.gas-price-cap-calculation]
+adjustment-constant=25 //used in the dynamic gas price cap equation
+blob-adjustment-constant=25 //used in the dynamic gas price cap equation
+finalization-target-max-delay="PT32H" //used in the dynamic gas price cap equation which is derived by our service level agreement that a L2 block should be finalized on L1 within 32 hours
+gas-fee-percentile-window="P7D" //used by fee history repository to calculate the p10(baseFeePastWeek)
+gas-fee-percentile-window-leeway="PT10M" //used by gas price cap provider to determine whether the currently cached fee history data is enough for gas price cap calculation, e.g. this means 50400 - 50 = 50350 blocks of fee history data from the past 7 days would be sufficient to proceed the calculation
```
Filter94: Unrelated to this PR. Why do we need it in the first place? Why can't we set 7 days - 10 minutes right away, without a special configuration?

docs/l1-dynamic-gas-pricing.md
```diff
@@ -0,0 +1,131 @@
...
+- `max-priority-fee-per-gas-cap` if we want to change the max cap of `maxPriorityFeePerGas`
+- `gas-price-cap-multiplier-for-finalization` to set the multiplier of the max caps of `maxFeePerGas` and `maxPriorityFeePerGas` with respect to that for data submission, e.g. if the multiplier is set as `2` and `max-fee-per-gas-cap` is set as `52 Gwei` which means the max cap of `maxFeePerGas` for finalization txn would be `52 * 2 = 104 Gwei` , same for max cap of `maxPriorityFeePerGas`
+
+## Steps to enable
+
+1. We can enable the L1 dynamic gas pricing mechanism by setting the following config, given that the other parameters as mentioned above have been properly set:
+    - `[l1-dynamic-gas-price-cap-service] disabled=false`
+3. Start/Restart the coordinator
+
+## Steps to verify
```
Filter94, line 111: Maybe we can mention the metrics caveat somewhere, that some metrics turn to 0 when there are no blob submission happening

Filter94 (approved):

docs/l1-dynamic-gas-pricing.md
```diff
@@ -0,0 +1,131 @@
...
+## Objective
+
+The purpose of this document is to illustrate the idea behind the L1 dynamic gas pricing mechanism, the coordinator’s configurations for the L1 dynamic gas pricing feature, and the steps to enable it.
+
+## Motivation
+The goal is to reduce our L1 costs by optimizing the price at which we send transactions (blob submission, finalization) to L1 whilst being within our SLA for finalization/withdrawals.
+
+## Scope
+
+We propose a mechanism which updates the L1 gas price every hour based upon the time elapsed since the firstL2 block in each aggregation, and the time of day/week to account for when we expect L1 prices to be lower/higher.
```
julien-marchand:
```suggestion
We propose a mechanism that updates the L1 gas price every hour based on the time elapsed since the first L2 block in each aggregation and the time of day/week to account for when we expect L1 prices to be lower/higher.
```

docs/l1-dynamic-gas-pricing.md
```diff
@@ -0,0 +1,131 @@
...
+
+The L1 data submission transaction accounts for 99.5% of our total L1 cost overhead, of which there is now no artificial delay for submission, therefore giving a longer window of flexibility to post at an optimal L1 gas price and be within the SLA time.
```
julien-marchand:
```suggestion
The L1 data submission transaction accounts for the major part of our total L1 cost overhead. There is now no artificial delay for submission, giving a longer window of flexibility to post at an optimal L1 gas price and be within the SLA time.
```
I'm tuning the wording a bit, as this figure isn't true anymore now that blob fees have been drastically reduced.

docs/l1-dynamic-gas-pricing.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Formula
+
+The formula to determine the current L1 gas price `threshold` in gwei is:
+
+`threshold = p10GasPricePastWeek (1 + adjustmentConstant * TDM * (elapsedTimeSinceAggregationFirstBlock/ SLA)^2 )`
+
+Where:
+
+The `p10GasPricePastWeek` monitors the 10th percentile of the L1 gas price across the last week. Will likely require a dB to store data in. This helps us keep track of latest price trends
```
julien-marchand: `Will likely` should be changed

docs/l1-dynamic-gas-pricing.md
```diff
@@ -0,0 +1,131 @@
...
+
+The `elapsedTimeSinceAggregationFirstBlock` is calculated by taking the difference between the current timestamp and the timestamp of the first L2 block which is to be finalized on L1 by the current aggregation
+
+`SLA` = 32 hours (configurable). Includes any legal delays
```
julien-marchand: Is 32h the default? Otherwise, we should use 16h (current value)

julien-marchand (dismissed):
Some of the wording is conditional - probably based on the original issue. Could you review, confirm what's the effective implementation, and change the tense accordingly?

docs/l1-dynamic-gas-pricing.md
jonesho: The `gas-fee-percentile-window` is used for defining the definite block range to get the fee history data from (e.g. past 7 days worth of history data), but the `gas-fee-percentile-window-leeway` is used for defining the number of missing blocks that we can tolerate from that definite block range, so let say if the fee history data stored in db from that 7 days block range (expected to have 50400 data entries) is less than 50350 (leeway set as 50), then coordinator would disable dynamic gas pricing and fall back to the static gas pricing mechanism. If we want a less stricted fallback criteria, one could set the leeway with higher value.

Filter94: It makes sense now. Thanks!

Filter94 (approved):

docs/l1-dynamic-gas-pricing.md
```diff
@@ -0,0 +1,149 @@
...
+
+Also note that, we will submit blob-carrying txns on L1 only if the dynamic gas prices are no less than the current base gas prices on L1, in order to cope with "100% replacement penalty" for blob-carrying txns.
+
+## Related coordinator configurations
+
+```jsx
+[l1-dynamic-gas-price-cap-service]
+disabled=false //if false, dynamic gas price cap will be enabled, and if true, gas price cap will behave as static just like before
+
+[l1-dynamic-gas-price-cap-service.gas-price-cap-calculation]
```
fluentcrafter, line 73: @jonesho shall we update this to match the new configuration structure?

PR #1088, opened 2025-06-04
Make MetricsCategory an interface and add default tags

gauravahuja:
This PR implements issue(s) https://github.com/Consensys/maru/issues/97

In this PR:
- Make metrics category an interface instead of an enum. This way services that import metrics package can define their own metrics category.
- Move LineaMetricsCategory enum out of the metrics package
- Add default tags to the Micrometer Metrics Facade. This is useful to set the service version or node id that is common for all metrics being published by the service.

### Checklist

### DIFF
--- a/jvm-libs/linea/core/metrics/src/main/kotlin/net/consensys/linea/metrics/MetricsFacade.kt
+++ b/jvm-libs/linea/core/metrics/src/main/kotlin/net/consensys/linea/metrics/MetricsFacade.kt
@@ -7,18 +7,8 @@ import java.util.function.Supplier
 
 data class Tag(val key: String, val value: String)
 
-enum class LineaMetricsCategory {
-  AGGREGATION,
-  BATCH,
-  BLOB,
-  CONFLATION,
-  GAS_PRICE_CAP,
-  TX_EXCLUSION_API,
-  ;
-
-  override fun toString(): String {
-    return this.name.replace('_', '.').lowercase()
-  }
+interface MetricsCategory {
+  val name: String
 }
 
 interface Counter {
@@ -38,7 +28,7 @@ interface TimerCapture<T> {
 
 interface MetricsFacade {
   fun createGauge(
-    category: LineaMetricsCategory? = null,
+    category: MetricsCategory? = null,
     name: String,
     description: String,
     measurementSupplier: Supplier<Number>,
@@ -46,14 +36,14 @@ interface MetricsFacade {
   )
 
   fun createCounter(
-    category: LineaMetricsCategory? = null,
+    category: MetricsCategory? = null,
     name: String,
     description: String,
     tags: List<Tag> = emptyList(),
   ): Counter
 
   fun createHistogram(
-    category: LineaMetricsCategory? = null,
+    category: MetricsCategory? = null,
     name: String,
     description: String,
     tags: List<Tag> = emptyList(),
@@ -62,14 +52,14 @@ interface MetricsFacade {
   ): Histogram
 
   fun <T> createSimpleTimer(
-    category: LineaMetricsCategory? = null,
+    category: MetricsCategory? = null,
     name: String,
     description: String,
     tags: List<Tag> = emptyList(),
   ): TimerCapture<T>
 
   fun <T> createDynamicTagTimer(
-    category: LineaMetricsCategory? = null,
+    category: MetricsCategory? = null,
     name: String,
     description: String,
     tagKey: String,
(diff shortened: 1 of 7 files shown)


jvm-libs/linea/core/metrics/src/main/kotlin/net/consensys/linea/metrics/MetricsFacade.kt
```diff
@@ -38,22 +28,22 @@ interface TimerCapture<T> {
 
 interface MetricsFacade {
   fun createGauge(
-    category: LineaMetricsCategory? = null,
+    category: MetricsCategory? = null,
```
Filter94: Why do we want to allow `null-s`? I think if we only stick to non-null-s, it's great that `MetricsFacade` enforces a convention to categorize metrics

jvm-libs/linea/metrics/micrometer/src/main/kotlin/net/consensys/linea/metrics/micrometer/MicrometerMetricsFacade.kt
```diff
@@ -17,6 +17,7 @@ import io.micrometer.core.instrument.Timer as MicrometerTimer
 class MicrometerMetricsFacade(
   private val registry: MeterRegistry,
   private val metricsPrefix: String? = null,
+  private val defaultTags: List<Tag> = emptyList(),
```
Filter94: These aren't default, since they're applied to every measurement. These are common tags

jvm-libs/linea/metrics/micrometer/src/main/kotlin/net/consensys/linea/metrics/micrometer/MicrometerMetricsFacade.kt
```diff
@@ -37,26 +38,28 @@ class MicrometerMetricsFacade(
...
     measurementSupplier: Supplier<Number>,
     tags: List<Tag>,
   ) {
-    if (category != null) requireValidMicrometerName(category.toString())
+    if (category != null) requireValidMicrometerName(category.toValidMicrometerName())
     requireValidMicrometerName(name)
     val builder = Gauge.builder(metricHandle(category, name), measurementSupplier)
-    if (tags.isNotEmpty()) {
-      val flatTags = tags.flatMap {
+    if (tags.isNotEmpty() || defaultTags.isNotEmpty()) {
```
Filter94: I think we should reuse this piece of logic for other kinds of metrics

gauravahuja: Done

jvm-libs/linea/metrics/micrometer/src/main/kotlin/net/consensys/linea/metrics/micrometer/MicrometerMetricsFacade.kt
gauravahuja: renamed to commonTags

jvm-libs/linea/core/metrics/src/main/kotlin/net/consensys/linea/metrics/MetricsFacade.kt
gauravahuja: Made it non null, had to provide the category in some places where it wasn't provided.

jvm-libs/linea/metrics/micrometer/src/main/kotlin/net/consensys/linea/metrics/micrometer/MicrometerMetricsFacade.kt
```diff
@@ -37,74 +38,64 @@ class MicrometerMetricsFacade(
...
   init {
     if (metricsPrefix != null) requireValidMicrometerName(metricsPrefix)
+    commonTags.forEach { requireValidMicrometerName(it.key) }
   }
 
-  private fun metricHandle(category: LineaMetricsCategory?, metricName: String): String {
+  private fun metricHandle(category: MetricsCategory, metricName: String): String {
     val prefixName = if (metricsPrefix == null) "" else "$metricsPrefix."
-    val categoryName = if (category == null) "" else "$category."
+    val categoryName = "${category.toValidMicrometerName()}."
```
Filter94, line 46: Nit: I think it's more logical to move the dot to the pattern string below. `categoryName` defines the name, the string defines the pattern.

jvm-libs/linea/metrics/micrometer/src/main/kotlin/net/consensys/linea/metrics/micrometer/MicrometerMetricsFacade.kt
Filter94: I meant that we should reuse this piece of code. Would something like this be worth it?

```
private inline fun <T> applyTagsIfNotEmpty(
    builder: T,
    tags: List<Tag>,
    crossinline tagsApplier: T.(Array<out String>) -> T
  ): T = flattenTags(tags).takeIf { it.isNotEmpty() }?.let { builder.tagsApplier(it.toTypedArray()) } ?: builder
```

And you call it like 
```
applyTagsIfNotEmpty(builder, tags) { tags(*it) }
```

Filter94 (dismissed):

julien-marchand (approved):

PR #1095, opened 2025-06-04
feat(bridge-ui): add base path config

VGau:
This PR implements issue(s) #

### Checklist


bridge-ui/environment.d.ts
```diff
@@ -5,6 +5,7 @@ declare global {
       E2E_TEST_PRIVATE_KEY: string;
       E2E_TEST_WALLET_PASSWORD: string;
       NEXT_PUBLIC_INFURA_ID: string;
+      NEXT_PUBLIC_BASE_PATH: string;
```
alainncls, line 8: We should add it to the `.env.template` file for local development/testing

VGau: This variable is automatically populated in the next config :) I will add the NEXT_PUBLIC_ENVIRONMENT variable

PR #1097, opened 2025-06-05
Add new timer and counter methods to metrics facade

gauravahuja:
This PR implements issue(s) https://github.com/Consensys/maru/issues/97

In this PR
- Add counter and timer Factory that can create meters with common tags. This is useful in cases where the timers or counters only differ by tags. eg: Timers in an Engine API client implementation. There will be one timer provider which will create timers for each method. See example in unit test.
- Add Timer Interface that is agnostic of the return type of the method/future being timed.
- Remove SimpleTimerCapture and replace it with Timer.
- Update the future extensions.

### Checklist


jvm-libs/linea/core/metrics/build.gradle
```diff
@@ -3,3 +3,7 @@ plugins {
 }
 
 description="Linea metrics facade"
+
+dependencies {
+  api "io.vertx:vertx-core"
```
fluentcrafter: The idea of this Module was to be Kotlin/Java vanilla. Can we please keep that and remove dependencies on vertx?

I see it's here because of `fun <T> captureTime(f: Future<T>): Future<T>` but that can be added as extension function where necessary.

jvm-libs/linea/metrics/micrometer/src/test/kotlin/net/consensys/linea/metrics/micrometer/MicrometerTimerAdapterTest.kt
```diff
@@ -0,0 +1,150 @@
...
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+import tech.pegasys.teku.infrastructure.async.SafeFuture
+import java.util.concurrent.ExecutionException
+import java.util.concurrent.TimeUnit
+
+class MicrometerTimerAdapterTest {
+
+  @Test
+  fun `timer with no extra tags`() {
```
fluentcrafter: it's good practice to write test cases with `should ...` or `when .... then/should` so wit states the scenario under test

jvm-libs/linea/core/metrics/build.gradle
gauravahuja: What about SafeFuture?

jvm-libs/linea/metrics/micrometer/build.gradle
```diff
@@ -7,4 +7,5 @@ description="Micrometer metrics implementation for Linea"
 dependencies {
   api "io.vertx:vertx-micrometer-metrics"
```
Filter94: Can you please get rid of it in favor of
```
api "io.micrometer:micrometer-registry-prometheus:${libs.versions.micrometer.get()}"
```
 like I did in my [change](https://github.com/Consensys/linea-monorepo/pull/1107/files#diff-f6a4e0528e5f7de546e6ad297253a5b2bbb23b812827f06cd211aecd9976428eR8)?

jvm-libs/linea/metrics/micrometer/src/main/kotlin/net/consensys/linea/metrics/micrometer/AbstractTimerCapture.kt
```diff
@@ -3,18 +3,15 @@ package net.consensys.linea.metrics.micrometer
...
+import net.consensys.linea.metrics.TimerCapture
 
 /**
  * An abstract class which contains everything is needed to make pretty much any Timer capture.
  * Mimicks the Builder pattern and allows captureTime implementation to perform different kinds of
  * captures TODO: In order to improve performance, TimerCapture instances can be cached into a
  * thread safe Map just like it's made in MeterRegistry class
  */
-abstract class AbstractTimerCapture<T> {
+abstract class AbstractTimerCapture<T> : TimerCapture<T> {
```
Filter94: I don't think this class is useful for anything anymore

jvm-libs/linea/metrics/micrometer/src/main/kotlin/net/consensys/linea/metrics/micrometer/DynamicTagTimerCapture.kt
```diff
@@ -15,7 +14,7 @@ import java.util.function.Function
  * nanosecond-level measurements. Related issue:
  * https://github.com/micrometer-metrics/micrometer/issues/535
  */
-class DynamicTagTimerCapture<T> : AbstractTimerCapture<T>, TimerCapture<T> {
+class DynamicTagTimerCapture<T> : AbstractTimerCapture<T> {
```
Filter94: You can improve this class by onboarding [my change ](https://github.com/Consensys/linea-monorepo/pull/1107/files#diff-4b230feff290192bebef11dee694282e81d195b1e9abbb3dafab9cb953638a9dR11)to split it in to a Builder and the `TimerCapture`. But probably this class needs a redesign anyway. Up to you

jvm-libs/linea/core/metrics/src/main/kotlin/net/consensys/linea/metrics/MetricsFacade.kt
```diff
@@ -21,6 +21,19 @@ interface Histogram {
...
+interface TimerFactory {
+  fun create(tags: List<Tag> = emptyList()): Timer
+}
+
+interface Timer {
+  fun <T> captureTime(f: CompletableFuture<T>): CompletableFuture<T>
+  fun <T> captureTime(action: Callable<T>): T
+}
+
 interface TimerCapture<T> {
```
Filter94: Having both `Timer` and `TimerCapture` still feels wrong. I tried a few things, but it seems like there's no easy way to fix this. Let's at least slightly improve it by renaming this interface to `DynamicTagTimerCapture` and name the implementation `DynamicTagTimerCaptureImpl`, accordingly to emphasize that it's there only for the `DynamicTagTimerCapture`. I feel like `DynamicTagTimerCapture` needs a redesign, but we need to move forward on this

Filter94 (approved):

jvm-libs/linea/metrics/micrometer/src/main/kotlin/net/consensys/linea/metrics/micrometer/AbstractTimerCapture.kt
gauravahuja: Removed the AbstractTimerCapture class

jvm-libs/linea/core/metrics/src/main/kotlin/net/consensys/linea/metrics/MetricsFacade.kt
gauravahuja: I redesigned the `DynamicTagTimerCapture` and renamed it to `DynamicTagTimerImpl` for consistency between SimpleTimer and DynamicTimer.

jvm-libs/linea/metrics/micrometer/src/main/kotlin/net/consensys/linea/metrics/micrometer/MicrometerTimerAdapter.kt
```diff
@@ -0,0 +1,56 @@
...
+  private val timerProvider = MicrometerTimer.builder(name)
+    .description(description)
+    .tags(*commonTags.flatMap { listOf(it.key, it.value) }.toTypedArray())
+    .withRegistry(meterRegistry)
+
+  init {
+    commonTags.forEach { it.requireValidMicrometerName() }
+  }
+
+  fun getTimer(tags: List<Tag>): io.micrometer.core.instrument.Timer {
```
Filter94: Nit. We could use `MicrometerTimer` here I believe

jvm-libs/linea/metrics/micrometer/src/main/kotlin/net/consensys/linea/metrics/micrometer/MicrometerTimerAdapter.kt
```diff
@@ -0,0 +1,56 @@
...
+      return action.call()
+    } finally {
+      timerSample.stop(adaptee)
+    }
+  }
+}
+
+class TimerFactoryImpl(
+  private val meterRegistry: MeterRegistry,
+  private val name: String,
```
Filter94: Nit: Not all the val-s need to be val-s here

Filter94 (approved):

PR #1101, opened 2025-06-05
Feat/integrate ZkCounter into the Sequencer

fab-10:
This PR implements issue(s) #

### Checklist


besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/selectors/TraceLineLimitTransactionSelector.java
letypequividelespoubelles, line 1: the `ZkCounter`uses the Besu hook `tracePrecompileCall(MessageFrame frame, long gasRequirement, Bytes output)`, so I think it should be called here too ?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/selectors/TraceLineLimitTransactionSelector.java
```diff
@@ -239,5 +259,120 @@ public void traceEndBlock(final BlockHeader blockHeader, final BlockBody blockBo
...
+    }
+
+    @Override
+    public void tracePostExecution(
+        final MessageFrame frame, final Operation.OperationResult operationResult) {
+      delegate.tracePostExecution(frame, operationResult);
+    }
+
+    @Override
+    public void tracePrecompileCall(
```
fab-10, line 316: @letypequividelespoubelles isn't this the call you say it is missing?

letypequividelespoubelles: oh yes it is, sorry my bad I didn't see it

besu-plugins/linea-sequencer/acceptance-tests/src/test/java/linea/plugin/acc/test/TransactionTraceLimitOverflowLimitlessTest.java
```diff
@@ -0,0 +1,87 @@
...
+    expectedConfirmedTxs.addAll(
+        minerNode.execute(
+            accountTransactions.createIncrementalTransfers(fewLinesSender, recipient, 4)));
+
+    expectedConfirmedTxs.stream()
+        .map(Hash::toHexString)
+        .forEach(hash -> minerNode.verify(eth.expectSuccessfulTransactionReceipt(hash)));
+
+    // assert that tx over line count limit is not confirmed and is removed from the pool
+    minerNode.verify(
```
letypequividelespoubelles, line 83: shouldn't we call popTransaction to check that the popping mechanism is working as intended ?

besu-plugins/linea-sequencer/acceptance-tests/src/test/resources/strictModuleLimitsLimitless.toml
```diff
@@ -0,0 +1,17 @@
...
+
+[traces-limits]
+MODEXP  = 1
+RIP     = 1
+BLAKE   = 1
+
+#
+# Block-specific limits
+#
+BLOCK_L1_SIZE      = 2000
```
letypequividelespoubelles, line 16: what's the diff betewwen the strict and non strict toml files ?

letypequividelespoubelles (approved):
LGTM. 
Two more questions:
- what about for L1_BLOCK_SIZE & L2_L1_LOGS ? 
- what about testing the commit / pop mechanism ?

besu-plugins/linea-sequencer/acceptance-tests/src/test/resources/strictModuleLimitsLimitless.toml
fab-10: The strict file has a block size limit of 2000 and is used to check that limit in `TransactionTraceLimitOverflowLimitlessTest`

besu-plugins/linea-sequencer/acceptance-tests/src/test/java/linea/plugin/acc/test/TransactionTraceLimitOverflowLimitlessTest.java
fab-10: At the acceptance test level, we do not have direct visibility of the `popTransaction` method, but we know it is called, since we sent 1(big)+4(small) txs, but only 4(small) are added to a block, so the first was popped

fab-10:
> LGTM. Two more questions:
> 
> * what about for L1_BLOCK_SIZE & L2_L1_LOGS ?
> * what about testing the commit / pop mechanism ?

 `L1_BLOCK_SIZE` is tested in `TransactionTraceLimitOverflowLimitlessTest`
 for `L2_L1_LOGS` idk how to make such calls, but it will be nice to have, I will create a ticket for that and ask to @julien-marchand @FlorianHuc to prioritize
 
 About the commit/pop mechanism at the acceptance test level, we do not have direct visibility of these methods, but we know they are called every time a tx is evaluated and discarded, an example [is here](https://github.com/fab-10/linea-monorepo/blob/33bbef47220a18bd4ddb8990827851ba90a7ff10/besu-plugins/linea-sequencer/acceptance-tests/src/test/java/linea/plugin/acc/test/ExcludedPrecompilesLimitlessTest.java#L202) where we check the log to be sure the tx is popped due to the line count limit overflow.

PR #1114, opened 2025-06-06
coordinator: reduce unnecessary JSON-RPC requests

fluentcrafter:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/coordinator/ethereum/message-anchoring/src/main/kotlin/linea/anchoring/clients/L1MessageSentEventsPoller.kt
+++ b/coordinator/ethereum/message-anchoring/src/main/kotlin/linea/anchoring/clients/L1MessageSentEventsPoller.kt
@@ -10,6 +10,7 @@ import org.apache.logging.log4j.LogManager
 import org.apache.logging.log4j.Logger
 import tech.pegasys.teku.infrastructure.async.SafeFuture
 import java.util.Deque
+import java.util.concurrent.atomic.AtomicLong
 import kotlin.time.Duration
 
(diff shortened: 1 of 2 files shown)


coordinator/ethereum/message-anchoring/src/main/kotlin/linea/anchoring/clients/L1MessageSentEventsPoller.kt
```diff
@@ -37,14 +38,18 @@ class L1MessageSentEventsPoller(
     l1HighestBlock = l1HighestBlock,
     log = log,
   )
+  private val lastFetchedMessageNumber: AtomicLong = AtomicLong(0L)
```
Filter94, line 41: Maybe it would make more sense to introduce this cache on the client level? Also, maybe it would be good to force this cache only for `FINALIZED` L1 blocks. I know that `message-anchoring.l1-highest-block-tag` can't really be anything other than finalized, but for safety, maybe it's worth to enforce on the client level for this case specifically.

Filter94 (approved):

coordinator/ethereum/message-anchoring/src/main/kotlin/linea/anchoring/clients/L1MessageSentEventsPoller.kt
fluentcrafter: Maybe. Will merge as it is working and simple enough.

PR #1115, opened 2025-06-09
Use distribution artifact instead of uber jar for Tracer plugin

fab-10:
This PR implements issue(s) #

Since https://github.com/Consensys/linea-tracer/pull/1996 the Tracer is not publishing any more the uber jar, and the plugin is only available as a ZIP archive that needs to be unpacked in the `plugins` dir, this PR adjust the `linea-besu-package` to use the new distribution artifact.

### Checklist


jonesho:
> Since https://github.com/Consensys/linea-tracer/pull/1996 the Tracer is not publishing any more the uber jar, and the plugin is only available as a ZIP archive that needs to be unpacked in the plugins dir, this PR adjust the linea-besu-package to use the new distribution artifact.
@fab-10 if that's the case then why linea-tracer still publish the jar file? [linea-tracer-beta-v3.1-rc2.jar](https://github.com/Consensys/linea-tracer/releases/download/beta-v3.1-rc2/linea-tracer-beta-v3.1-rc2.jar)

PR #1151, opened 2025-06-12
Prover/koalabear migration rebased bis 2

ThomasPiellard:
# Description

WIP 

refactoring and migrating to koalabear on the prover.
* prover/maths package now uses koalabear
* The fft package has been removed, we use directly the FFT in gnark-crypto
* prover/crypto/vortex package now uses koalabear, and usess by default SIS for hashing the columns, and poseidon2 for the Merkle tree.

## Type of change

- [x] New feature (non-breaking change which adds functionality)


arijitdutta67:
Hi @ThomasPiellard , thanks for the work. I have the following concerns:

- I see you use Posseidon hash on top of SIS hash to compute the leaf. With the optional sis hashing feature we sometime use direct hashing on the columns of the Vortex matrices avoiding SIS. For reference, it is already implemented on the `prover/limitless-top-level-rebased` branch. Will suggest you to have a look at `crypto/vortex` and `compiler/vortex` folders.
- The major concern is that, in self recursion, we prove that the MiMC hashing is done correctly to compute the leaf using a wizard gadget. For Posseidon hash, we need a similar gadget, otherwise self-recursion will not work. See [here](https://github.com/Consensys/linea-monorepo/blob/a7f59c3d34727c061e43625d2cb944a026124314/prover/protocol/compiler/selfrecursion/column_opening.go#L263) for reference.
CC: @AlexandreBelling

arijitdutta67:
> yes by default in this version it is SIS for the columns + poseidon2 for the Merkle tree. But it is possible to change them, by providing either `ColumnHasher` or `MerkleHasher` to `Params` (so we can use mimc for the columns if needed). In the previous version, there was something called "LeafHashFunc" and "MerkleHashFunc", and the functions to hash the columns were duplicated according the LeafHashFunc being SIS or something else, and the naming was a bit confusing.
> 
> * My main question to know if what we did makes sense is : is it correct that there are only 2 hash function at play (I am not counting FS, I am just talking about vortex, assuming that the random coins are already computed), one to hash the columns, and one for the Merkle tree ? For instance you cannot have a configuration where you hash the columns with SIS, then hash the SIS hashed columns using LeafHashFunc, then build the Merkle tree using MerkleHashFunc where LeafHashFunc != MerkleHashFunc ?

I like the modification you did on the `ColumnHasher()` and `MerkleHasher()` part. For us, the other hash function for both column and merkle hashing was MiMC. But I am not aware of any problem if it is the case: LeafHashFunc != MerkleHashFunc .

PR #1161, opened 2025-06-17
coordinator: adds FakeLineaRollupSmartContractClient

fluentcrafter:
This is necessary for tests in Maru, but it can be useful for coordinator or besu plugins tests.

### DIFF
--- a/jvm-libs/linea/clients/linea-contract-clients/src/testFixtures/kotlin/linea/contract/l1/FakeLineaRollupSmartContractClient.kt
+++ b/jvm-libs/linea/clients/linea-contract-clients/src/testFixtures/kotlin/linea/contract/l1/FakeLineaRollupSmartContractClient.kt
@@ -43,10 +43,8 @@ class FakeLineaRollupSmartContractClient(
   _finalizedBlocks: List<FinalizedBlock> = listOf(FinalizedBlock(0uL, Clock.System.now(), Random.nextBytes(32))),
   _messageRollingHashes: Map<ULong, ByteArray> = emptyMap(),
 ) : LineaRollupSmartContractClientReadOnly {
+  @get:Synchronized @set:Synchronized
   var contractVersion: LineaContractVersion = LineaContractVersion.V6
-    @Synchronized get
-
-    @Synchronized set
 
   val messageRollingHashes: MutableMap<ULong, ByteArray> = ConcurrentHashMap(_messageRollingHashes)
   val finalizedBlocks: MutableMap<ULong, FinalizedBlock> = ConcurrentHashMap()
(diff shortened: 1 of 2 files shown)


jvm-libs/linea/clients/linea-contract-clients/src/testFixtures/kotlin/linea/contract/l1/FakeLineaRollupSmartContractClient.kt
```diff
@@ -0,0 +1,106 @@
...
+  }
+}
+
+class FakeLineaRollupSmartContractClient(
+  val contractAddress: String = Random.nextBytes(20).encodeHex(),
+  _finalizedBlocks: List<FinalizedBlock> = listOf(FinalizedBlock(0uL, Clock.System.now(), Random.nextBytes(32))),
+  _messageRollingHashes: Map<ULong, ByteArray> = emptyMap(),
+) : LineaRollupSmartContractClientReadOnly {
+  @get:Synchronized @set:Synchronized
+  var contractVersion: LineaContractVersion = LineaContractVersion.V6
```
gauravahuja: Why does contractVersion need to be mutable? Can't it be set in the constructor?

gauravahuja (approved):
Just the one comment otherwise lgtm

jvm-libs/linea/clients/linea-contract-clients/src/testFixtures/kotlin/linea/contract/l1/FakeLineaRollupSmartContractClient.kt
fluentcrafter: To allow for the future to simulate scenarios of a smart contract upgrade. We already had these cases in prod.

fluentcrafter: However, the initial version can be in the constructor. Will add that.

gauravahuja: > To allow for the future to simulate scenarios of a smart contract upgrade. We already had these cases in prod.

I guess then the base interface should be modified to allow for this.

gauravahuja (approved):

jvm-libs/linea/clients/linea-contract-clients/src/testFixtures/kotlin/linea/contract/l1/FakeLineaRollupSmartContractClient.kt
fluentcrafter: > I guess then the base interface should be modified to allow for this.

No, because that is read from the smart contract. Is an async read-only operation. However, in the tests we can force whatever scerario.

PR #1166, opened 2025-06-17
fix: added test fixtures dependencies for linea-contract-clients

jonesho:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/jvm-libs/linea/clients/linea-contract-clients/build.gradle
+++ b/jvm-libs/linea/clients/linea-contract-clients/build.gradle
@@ -5,6 +5,15 @@ plugins {
 
 description = "Linea L1 smart contract client"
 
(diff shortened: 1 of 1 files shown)


jvm-libs/linea/clients/linea-contract-clients/build.gradle
```diff
@@ -5,6 +5,15 @@ plugins {
 
 description = "Linea L1 smart contract client"
 
+configurations {
+  testFixturesDeps {
+    extendsFrom implementation // Inherit plugin's dependencies
+    canBeConsumed = false
+    canBeResolved = false
```
fluentcrafter, line 12: what do these mean, why do we need them?

PR #1173, opened 2025-06-17
Feat: Update Linea Besu package to latest Besu and plugin versions

fab-10:
Related to https://github.com/Consensys/linea-monorepo/pull/1115 

Updated Besu and plugins to latest version, with support for the limitless prover.

Disable parallel tx processing in Besu until it is fully supported by Linea, since it is creating concurrency issues when used with the Shomei plugin and the Tracer.

Removed also the publication of the jar, since only the zip is needed now.

NOTE: if you wonder why in some places `Xbonsai-parallel-tx-processing-enabled=false` is used, while in other `bonsai-parallel-tx-processing-enabled=false` is used, it is because the former is compatible also with the previous version of Besu so it must be used for tests that will be updated only when the new Linea Besu Package is released after this PR is merged, at that point we can aligned the options to only used the new version.

### Checklist


garyschulte (approved):
🚢

docker/config/l2-node-besu/l2-node-besu-config.toml
```diff
@@ -63,3 +63,5 @@ bonsai-historical-block-limit=1024
 
 Xin-process-rpc-enabled=true
 Xin-process-rpc-apis=["ETH", "MINER"]
+
+Xbonsai-parallel-tx-processing-enabled=false
```
jonesho, line 67: @fab-10 so this mean we'd also need to disable "parallel-tx-processing" when we deploy this `linea-besu-package` image on our production environments, until we have a new release with the fix, right?

e2e/src/send-bundle.spec.ts
```diff
@@ -69,7 +71,7 @@ describe("Send bundle test suite", () => {
       for (let i = 0; i < 3; i++) {
         const txRequest: TransactionRequest = {
           to: recipientAccount.address,
-          value: ethers.parseUnits("1000", "wei"),
+          value: ethers.parseEther("5"),
```
jonesho, line 74: Just curious, why changed it from 1000 wei to 5 ETH? just for better readability?

docker/config/l2-node-besu/l2-node-besu-config.toml
fab-10: yes, asked to devops to apply the configuration change

e2e/src/send-bundle.spec.ts
fab-10: The change is needed because the new version of Besu has a better gas estimation algorithm that also takes in account the sender balance, so the previous version failed due to `Upfront cost exceeds account balance`

```
2025-07-10 08:28:43.808+00:00 | vert.x-worker-thread-14 | TRACE | JsonRpcExecutor | JSON-RPC request -> eth_estimateGas [{"maxFeePerGas":"0x2540be400","maxPriorityFeePerGas":"0x3b9aca00","nonce":"0x0","value":"0x3e8","from":"0xf2a5b3e0f9b3460f0f3469b2ab10bd7acb59e936","to":"0x54ae6f52ce019989d87e60392205f411ae68f3ac"}]
2025-07-10 08:28:43.809+00:00 | vert.x-worker-thread-14 | TRACE | LondonFeeMarket | block #11 parentBaseFee: 0x0000000000000000000000000000000000000000000000000000000000000007 parentGasUsed: 21000 parentGasTarget: 1000000000 baseFee: 0x0000000000000000000000000000000000000000000000000000000000000007
2025-07-10 08:28:43.809+00:00 | vert.x-worker-thread-14 | TRACE | TransactionSimulator | Simulated block header: ProcessableBlockHeader{number=11, parentHash=0x538eadf4518b2523fd311d13cb4df0f60619c1231c49a4501e566cd0b88cd755, coinbase=0x0ad2f22bd2dfdc523269028a3e4346230eff96cd, difficulty=0x0000000000000000000000000000000000000000000000000000000000000001, gasLimit=2000000000, timestamp=1752136123, baseFee=0x0000000000000000000000000000000000000000000000000000000000000007, mixHashOrPrevRandao=null, }
2025-07-10 08:28:43.812+00:00 | vert.x-worker-thread-14 | TRACE | AbstractJsonRpcExecutor | {"jsonrpc":"2.0","id":2,"error":{"code":-32004,"message":"Upfront cost exceeds account balance"}}
```

jonesho: ic, just to clarify, so u mean, in the prevoius test implementation, the sender would only leave with 500 wei (1500 - 1000 wei) which was not sufficient to pay for the gas fee (now being checked by new version of Besu) and hence would lead to "Upfront cost exceeds account balance" error, that's why we changed the initial sender balance from 1000 wei to 5 eth?

fab-10: correct

PR #1197, opened 2025-06-19
Feat: Improve the logging of Besu during e2e tests

fab-10:
* Log RPC requests and responses
* Collect Besu L1 node logs
* Remove extra GC options, just go with the default

### Checklist

### DIFF
--- a/docker/compose-spec-l1-services.yml
+++ b/docker/compose-spec-l1-services.yml
@@ -14,7 +14,7 @@ services:
       retries: 120
     environment:
       LOG4J_CONFIGURATION_FILE: /var/lib/besu/log4j.xml
-      JAVA_OPTS: -Xmx1536m -XX:+UnlockExperimentalVMOptions -XX:-UseG1GC -XX:+UseZGC
+      JAVA_OPTS: -Xmx1536m
     entrypoint:
       - /bin/bash
       - -c
--- a/docker/config/l1-node/el/log4j.xml
+++ b/docker/config/l1-node/el/log4j.xml
@@ -27,6 +27,9 @@
     <Logger name="io.opentelemetry" level="WARN" additivity="false">
       <AppenderRef ref="Console"/>
     </Logger>
+    <Logger name="org.hyperledger.besu.ethereum.api.handlers.AbstractJsonRpcExecutor" level="TRACE">
+      <AppenderRef ref="Console"/>
+    </Logger>
     <Root level="INFO">
       <AppenderRef ref="Console"/>
     </Root>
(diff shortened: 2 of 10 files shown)


docker/compose-spec-l1-services.yml
```diff
@@ -14,7 +14,7 @@ services:
       retries: 120
     environment:
       LOG4J_CONFIGURATION_FILE: /var/lib/besu/log4j.xml
-      JAVA_OPTS: -Xmx1536m -XX:+UnlockExperimentalVMOptions -XX:-UseG1GC -XX:+UseZGC
+      JAVA_OPTS: -Xmx1536m
```
jonesho, line 17: Why removed the other options?

docker/config/l1-node/el/log4j.xml
```diff
@@ -27,6 +27,9 @@
     <Logger name="io.opentelemetry" level="WARN" additivity="false">
       <AppenderRef ref="Console"/>
     </Logger>
+    <Logger name="org.hyperledger.besu.ethereum.api.handlers.AbstractJsonRpcExecutor" level="TRACE">
```
jonesho, line 30: If enabling TRACE level here, would that add a lot noise in the `l1-node` logs?

docker/compose-spec-l1-services.yml
fab-10: Because the optimal GC for Besu is G1GC

docker/config/l2-node-besu/log4j.xml
```diff
@@ -37,6 +37,9 @@
     <Logger name="org.hyperledger.besu.ethereum.eth.transactions.TransactionPool" level="TRACE" additivity="false">
       <AppenderRef ref="Console"/>
     </Logger>
+    <Logger name="org.hyperledger.besu.ethereum.api.handlers.AbstractJsonRpcExecutor" level="TRACE">
```
jonesho, line 40: Same, if enabling TRACE level here, would that add a lot noise in the `l2-node-besu` logs?

jonesho (approved):
Approved, but just left a few questions

docker/config/l1-node/el/log4j.xml
fab-10: This is need to log the RPC request/response without it we are kinda blind on what is happening, and I think these logs are used only for debugging purposes, and we need some more details to make them useful

docker/config/l2-node-besu/log4j.xml
fab-10: s.a.

jonesho (approved):

PR #1201, opened 2025-06-20
Feat: Add a transfer bot to keep alive block production on e2e network

fab-10:
When debugging e2e tests about bundles, I found that if you try to run only that test in isolation 
```
pnpm run -F e2e test:e2e:local -- e2e/src/send-bundle.spec.ts
```
it will never work, since bundles are specific for inclusion in a future block, but since there is no traffic, due to no other e2e running in parallel, bundle tests just wait in vain for the right block to be built.

So I added a simple bot that just sends a transfer every 2 seconds to keep alive block production.

### Checklist


fab-10:
Superseded by https://github.com/Consensys/linea-monorepo/pull/1203

PR #1212, opened 2025-06-24
Update Besu to 25.5.0-linea2.1 in package

fab-10:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/.github/workflows/linea-besu-package-release.yml
+++ b/.github/workflows/linea-besu-package-release.yml
@@ -39,6 +39,7 @@ permissions:
   actions: read
   id-token: write
   contents: write
+  packages: write
 
 jobs:
   filter-commit-changes:


.github/workflows/linea-besu-package-release.yml
```diff
@@ -39,6 +39,7 @@ permissions:
   actions: read
   id-token: write
   contents: write
+  packages: write
```
FlorianHuc, line 42: What does this feature means?

alainncls: We've recently tightened permissions across all pipelines, requiring them to explicitly declare what they need.
Here, the pipeline needs to publish packages, so we added `packages: write`.
Since the pipeline hadn't run yet, we didn't catch the issue before, now it's fixed!

amkCha (approved):

PR #1235, opened 2025-07-02
Chore: cleanup and simplify log4j conf for Besus running in e2e tests

fab-10:
This PR implements issue(s) #

### Checklist


docker/config/l2-node-besu/log4j.xml
```diff
@@ -12,34 +12,16 @@
...
-    </Logger>
+    <Logger name="org.hyperledger.besu.ethereum.eth.sync.fullsync.FullSyncTargetManager" level="INFO" />
+    <Logger name="org.hyperledger.besu.ethereum.blockcreation" level="INFO" />
+    <Logger name="org.hyperledger.besu.consensus.merge.blockcreation" level="INFO" />
+    <Logger name="org.hyperledger.besu.ethereum.api.jsonrpc" level="TRACE" />
+    <Logger name="io.opentelemetry" level="WARN" />
+    <Logger name="net.consensys.linea.sequencer.txselection.selectors" />
+    <Logger name="org.hyperledger.besu.ethereum.eth.transactions.TransactionPool" level="TRACE" />
+    <Logger name="org.hyperledger.besu.ethereum.api.handlers.AbstractJsonRpcExecutor" level="TRACE" />
     <Root level="${sys:root.log.level}">
```
fluentcrafter: Since we are hands on it: shall we set the lever here directly?

I never understood -- any code base not just Linea -- the purpose of creating a property just fo the root log level 🤷🏻‍♂️ it just feels like #overengineering

fluentcrafter (approved):
Thanks for the clean up.

I would just go a step further and clean up all

```
<Properties>
    <Property name="root.log.level">WARN</Property>
  </Properties>
```

I set it directly
```
<Root level="WARN">
```

docker/config/l2-node-besu/log4j.xml
fab-10: Fine for me, I think it was done like that to be able to change the log level externally w/o changing the file, but never used that feature

fab-10: done

PR #1242, opened 2025-07-04
Feat: linea_sendBundle add validation of transactions

fab-10:
This PR implements issue(s) #

This PR adds validation of transactions contained in a bundle, according to Linea rules: allowed address, gas limit per tx, calldata per tx, excluding for the moment the profitability and the simulation check, that can be added later.

### Checklist

### DIFF
--- a/besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/rpc/services/LineaBundleEndpointsPlugin.java
+++ b/besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/rpc/services/LineaBundleEndpointsPlugin.java
@@ -10,11 +10,17 @@
 package net.consensys.linea.rpc.services;
 
 import com.google.auto.service.AutoService;
+import java.util.Arrays;
+import java.util.Optional;
 import net.consensys.linea.AbstractLineaRequiredPlugin;
 import net.consensys.linea.rpc.methods.LineaCancelBundle;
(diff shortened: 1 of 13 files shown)


besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/rpc/services/LineaBundleEndpointsPlugin.java
```diff
@@ -43,6 +49,22 @@ public void doRegister(final ServiceManager serviceManager) {
         lineaCancelBundleMethod::execute);
   }
 
+  public PluginTransactionPoolValidator createTransactionValidator() {
+    final var validators =
+        new PluginTransactionPoolValidator[] {
+          new AllowedAddressValidator(transactionPoolValidatorConfiguration().deniedAddresses()),
```
macfarla, line 55: @fab-10 in reality how often does this get called (loading the denied addresses) - is it only on startup?

macfarla (approved):

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/rpc/services/LineaBundleEndpointsPlugin.java
fab-10: yes it is only at startup

PR #1261, opened 2025-07-24
Prover: State Manager Fix on Deletion Segments

bogdanbear:
This PR implements issue(s) [1262](https://github.com/Consensys/linea-monorepo/issues/1262) by adding a filter on Shomei storage key in the case of deletion filter. 

### Checklist

* [x] New filter for storage keys on Shomei's side. 
* [x] Tested on mainnet traces without issues.


AlexandreBelling:
Merged in prover/execution-limitless-backend-integration

PR #1264, opened 2025-07-25
Prover/invalidity proofs

Soleimani193:
This PR implements the flow for the invalidity proofs. It supports only two cases of invalidity: bad--transaction-nonce and bad-account-balance. 

### Checklist


prover/circuits/aggregation/circuit_test.go
```diff
@@ -46,6 +46,10 @@ func TestPublicInput(t *testing.T) {
 			L1RollingHashMessageNumber:              549263,
 			L2MsgRootHashes:                         []string{"0xfb7ce9c89be905d39bfa2f6ecdf312f127f8984cf313cbea91bca882fca340cd"},
 			L2MsgMerkleTreeDepth:                    5,
+			LastFinalizedRollingHashNumberTx:        3,
```
AlexandreBelling: Can we have a more specific name to be sure this is never confusion with L1MsgRollingHash?

PR #1273, opened 2025-07-29
Fix: handle nil trace from go-corset expansion in arithmetization

### LINKED ISSUE #198: Handling Trace Expansion Errors from `go-corset`
The prover is not properly handling error cases returned when `go-corset` fails in trace expansion.

## Impact

I don't believe any users are affected at this time.

## Describe the bug

In the file `prover/zkevm/arithmetization/files.go`, we have this sequence:

```go
expTraces, errs := schema.NewTraceBuilder(sch).Build(rawTraces)
if len(errs) > 0 {
	logrus.Warnf("corset expansion gave the following errors: %v", errors.Join(errs...).Error())
}

return expTraces, nil
```

The problem is that, when `len(errs) > 0` we have (or can have) `expTraces == nil`.  This arises when `go-corset` believes it is unable to continue with trace expansion (typically because there is some column missing from the trace).

### Steps to Reproduce

To reproduce this bug you need a constraint file which is missing one or more columns required by the constraints.

### Actual result

The prover will `panic`

### Expected result

The prover should fail gracefully.

dinitheth:
### Summary

This PR fixes an issue where the prover panics if go-corset returns a nil expanded trace.

When `schema.NewTraceBuilder(...).Build()` fails due to malformed input (e.g. missing columns in `.lt` file), it returns a nil trace and errors. Previously, the code logged a warning but continued, causing a nil pointer panic.

**This patch:**
- Aborts early using `logrus.Panic` if the trace is nil or any errors occurred.
- Joins all error messages via `errors.Join` for clarity.
- Prevents passing nil traces to `AssignFromLtTraces`.

### ✅ Fixes

Closes #198

### Files Changed
- `prover/zkevm/arithmetization/arithmetization.go`

### Why it matters

This change makes trace expansion failure safe and debuggable, avoiding silent panics in CI and developer environments.

### DIFF
--- a/prover/zkevm/arithmetization/arithmetization.go
+++ b/prover/zkevm/arithmetization/arithmetization.go
@@ -18,41 +18,24 @@ import (
 // Settings specifies the parameters for the arithmetization part of the zkEVM.
 type Settings struct {
 	Limits *config.TracesLimits
-	// IgnoreCompatibilityCheck disables the strong compatibility check.
-	// Specifically, it does not require the constraints and the trace file to
-	// have both originated from the same commit.  By default, the compability
-	// check should be enabled.
 	IgnoreCompatibilityCheck *bool
-	// OptimisationLevel determines the optimisation level which go-corset will
-	// apply when compiling the zkevm.bin file to AIR constraints.  If in doubt,
-	// use mir.DEFAULT_OPTIMISATION_LEVEL.
-	OptimisationLevel *mir.OptimisationConfig
+	OptimisationLevel        *mir.OptimisationConfig
 }
 
 // Arithmetization exposes all the methods relevant for the user to interact
-// with the arithmetization of the zkEVM. It is a sub-component of the whole
-// ZkEvm object as it does not includes the precompiles, the keccaks and the
-// signature verification.
+// with the arithmetization of the zkEVM.
 type Arithmetization struct {
 	Settings *Settings
-	// Schema defines the columns, constraints and computations used to expand a
-	// given trace, and to subsequently to check satisfiability.
-	Schema *air.Schema
-	// Metadata embedded in the zkevm.bin file, as needed to check
-	// compatibility.  Guaranteed non-nil.
+	Schema   *air.Schema
 	Metadata typed.Map
 }
 
-// NewArithmetization is the function that declares all the columns and the constraints of
-// the zkEVM in the input builder object.
 func NewArithmetization(builder *wizard.Builder, settings Settings) *Arithmetization {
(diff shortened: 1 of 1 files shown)


prover/zkevm/arithmetization/arithmetization.go
```diff
@@ -18,86 +18,57 @@ import (
 // Settings specifies the parameters for the arithmetization part of the zkEVM.
 type Settings struct {
 	Limits *config.TracesLimits
-	// IgnoreCompatibilityCheck disables the strong compatibility check.
```
AlexandreBelling, line 21: Why removing all that doc?

prover/zkevm/arithmetization/arithmetization.go
```diff
@@ -18,86 +18,57 @@ import (
...
 	traceF := files.MustRead(traceFile)
-	// Parse trace file and extract raw column data.
 	rawColumns, metadata, errT := ReadLtTraces(traceF, a.Schema)
 
-	// Performs a compatibility check by comparing the constraints
-	// commit of zkevm.bin with the constraints commit of the trace file.
-	// Panics if an incompatibility is detected.
 	if !*a.Settings.IgnoreCompatibilityCheck {
-		var errors []string
+		var compatibilityErrors []string
```
AlexandreBelling, line 51: Thinking about it, but we should be using an `errors.Join` pattern I think here.

prover/zkevm/arithmetization/arithmetization.go
```diff
@@ -106,11 +77,22 @@ func (a *Arithmetization) Assign(run *wizard.ProverRuntime, traceFile string) {
 	if errT != nil {
 		fmt.Printf("error loading the trace fpath=%q err=%v", traceFile, errT.Error())
 	}
-	// Perform trace expansion
+
+	// 🛠️ Patch: safer handling of expansion failure
```
AlexandreBelling, line 81: That comment should:
1) Not have emojis in it
2) It should explain why the handling of `expandedTrace` is needed here. More generally, a comment should be explaining "why the code is like that" instead of explaining what a diff is.

Second, the control flow of the handling can be vastly simplified:

```
...

traceExpansionErr := errors.Join(errs...) // will be nil if no errors

// explain why this is needed
if expandedTrace == nil {
    traceExpansionErr = errors.Join(tracesExpansionErr, errors.New("corset expansion returned nil trace")
}

if traceExpansionErr != nil {
     utils.Panic("corset expansion failed: %v", traceExpansionErr.Error())
}

...
```

Issue #1246, opened 2025-07-14
Besu Plugin - FinalizedTagUpdater - Stop After Paris/Maru is enabled


fluentcrafter:
## Description
When Linea updates to use Maru (QBFT consensus), Maru will have the responsibility to update the FINALIZED block through `engine_forkChoiceUpdate` call, hence `LineaFinalizationTagUpdater` should stop otherwise, it will override/conflict with Maru information.

There are 2 possible ways to achieve this

## A - Adapt LineaFinalizationTagUpdater aware of Fork
We adapt LineaFinalizationTagUpdater to be aware of which fork Besu is running on and stops updating the finalised block after the Paris fork because it''s Maru's responsibility

## B - no Development - Manual intervention

After the Paris block is reached, we update the Besu nodes to disable `LineaFinalizationTagUpdater`. 
However, there may be a small overlapping period where the plugin and Maru may be "stepping on each other's toes". This scenario shall be tested and to evaluate if Option B is valid. 

This approach has a big UX disadvantage for external node runners that may need to be aware these intricacies.

PR #1275, opened 2025-07-30
feat: change stop point from PRAGUE to PARIS

jonesho:
This PR implements issue(s) #1246

### Checklist


Filter94 (approved):
I assume the manual test is passing. Also, it would be nice to test both Prague and Paris forks in the test

PR #1276, opened 2025-07-30
blob compressor v2 - limitless prover

fluentcrafter:
This PR implements issue(s) #

### Checklist


coordinator/app/src/test/resources/configs/coordinator-traces-v2-override.config.toml
```diff
@@ -10,7 +10,7 @@ fs-requests-directory = "/data/prover/v3/aggregation/requests"
 fs-responses-directory = "/data/prover/v3/aggregation/responses"
 
 [traces]
-blob-compressor-version="V1_2"
+blob-compressor-version="V2_0"
```
jonesho: This I believe can be removed, we moved `blob-compressor-version` to `[conflation.blob-compression]`, also there's another one in coordinator/app/src/test/resources/configs/coordinator-traces-v2-override.config.toml, can we remove as well?

fluentcrafter: This file is not even used at all. So I will removed it..

jonesho (approved):

PR #1288, opened 2025-08-04
feat: added support of tls connection and chore for code readibility

jonesho:
This PR implements issue(s) #1285

### Checklist


docker/config/linea-besu-sequencer/tls-files/sequencer_client_keystore.p12
kyzooghost, line 1: Is there a security vulnerability from committing .p12 file to public Github repo? Could it be an K8s ExternalSecret with the actual secret stored somewhere else?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/liveness/LineaLivenessService.java
kyzooghost, line 1: I find `shouldBuildLivenessBundle()` challenging to follow due to the cognitive overhead of mentally tracking ~10 variables with subtle differences throughout the function. Would it make sense to create a tech debt ticket to revisit this function for improved readability?

Some refactoring ideas:
1. Reduce variable scope – Variables like isNotFirstBlockTimestampToCheck seem to span ~40-50 lines. Could we perhaps use early returns to limit their lifespan to just a few lines where they're actually needed?
2. Favor early exits – perhaps we could focus cases where shouldBuild = false and return early
3. Extract helper methods – Some of the variable-heavy logic could potentially be extracted into smaller, focused methods.

jonesho: Yes, if it's still hard to following logically, we could follow this up in a tech debt ticket

docker/config/linea-besu-sequencer/tls-files/sequencer_client_keystore.p12
jonesho: It should be fine as these keystore and truststore files are just for local stack run, and we already did the same for coordinator to facilitate tls connection to web3signer locally. When deploying to our environments, we would use standard k8s practice to store the secret, and we'd already done that for coordinator and web3signer tls connection across all environments (e.g. sepolia, mainnet)

kyzooghost (approved):

Issue #1270, opened 2025-07-29
Update linea-monorepo compose file with more recent images and config files


jonesho:
🔐 🔐 `Should this be a security advisory instead?` 🔐 🔐

## Description
We should update the local stack by updating the linea-monorepo compose and config files for the followings:
- update `LINEA_BESU_TAR_GZ=25.7.0-linea2`,  in `versions.env`
- update `LINEA_SEQUENCER_PLUGIN_VERSION` (#281) and `LINEA_FINALIZED_TAG_UPDATER_PLUGIN_VERSION` (#1246) in `versions.env`
- update `expected-traces-api-version = "beta-v3.1-rc6"` in `coordinator-config-v2.toml`
- update `linea-besu-package` version in `compose-spec-l2-services.yml`
- update `coordinator` version in `compose-spec-l2-services.yml` for blob compressor version update to 2.0.0

## Motivation
This section should indicate what value we are receiving from it.

## Tasks
- [ ] Medium-level task that needs to be implemented for this feature request to be considered complete.

## Acceptance criteria
- [ ] List tests that need to be implemented or checks to consider this task completed. 

## Risks
- [ ] These are risks that may prevent completion of the task as well as areas to pay special attention to

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #1298, opened 2025-08-06
feat: update versions of linea-besu, linea-sequencer, finalized-tag-u…

jonesho:
…pdater

This PR implements issue(s) #1270

### Checklist


.github/workflows/reuse-run-e2e-tests.yml
```diff
@@ -133,7 +133,7 @@ jobs:
           docker container ls -a
       - name: Run e2e tests
         id: run_e2e_tests
-        timeout-minutes: 4
+        timeout-minutes: 6
```
jonesho, line 136: The liveness e2e needs to be started after the usual e2e tests ended because the sequencer will be restarted in liveness e2e which might disrupt the result of usual e2e tests if they are both running in parallel. Usually the liveness test alone only takes 30 sec but added timeout to 6 mins just to avoid flaky liveness tests as we'd seen sequencer took a bit longer time to return `eth_getLogs` result after restart

VGau (approved):

jonesho:
@VGau could you pls help review again as I'd switched to use fetch JSON-RPC call to retrieve liveness eth logs to avoid flaky issues

Issue #1266, opened 2025-07-28
Prevent Type 4 (EIP-7702) transaction for inclusion


kyzooghost:
## Description
Prevent mining EIP-7702 (Type 4) transactions on L2. We need to prevent type 4 transactions for inclusion:
- During block import (so that if a misbehaving sequencer includes it, it gets rejected by other validators)
- In the transaction pool as well (to avoid saturating the transaction pools)
- During block building (to avoid mining them)
  - Might be unnecessary with the above
 
Implement this as a feature back for future rollback

## Motivation
Linea does not plans to support Type 4 EIP-7702 transactions at a later date

## Tasks
- [ ] Extend the [https://github.com/Consensys/linea-sequencer](https://github.com/Consensys/linea-monorepo/tree/main/besu-plugins/linea-sequencer) to:
  - [ ] Drop type 4 transaction during block building
  - [ ] Consider a block invalid if it contains type 3 transaction during block import
  - [ ] Drop type 4 transactions, in the transaction pool
- [ ] Implement as a feature flag for future rollback

fluentcrafter:
@kyzooghost can you cofirm if this is already delived and close if so?

PR #1305, opened 2025-08-12
[Feat] Sequencer plugin to block Type4 Tx

kyzooghost:
This PR implements issue(s) #1266

### Checklist

### DIFF
--- a/besu-plugins/linea-sequencer/acceptance-tests/src/test/java/linea/plugin/acc/test/EIP7702TransactionDenialTest.java
+++ b/besu-plugins/linea-sequencer/acceptance-tests/src/test/java/linea/plugin/acc/test/EIP7702TransactionDenialTest.java
@@ -1,0 +1,271 @@
+/*
+ * Copyright Consensys Software Inc.
+ *
+ * This file is dual-licensed under either the MIT license or Apache License 2.0.
+ * See the LICENSE-MIT and LICENSE-APACHE files in the repository root for details.
+ *
+ * SPDX-License-Identifier: MIT OR Apache-2.0
+ */
+package linea.plugin.acc.test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import okhttp3.Response;
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.bytes.Bytes32;
+import org.hyperledger.besu.datatypes.Address;
+import org.hyperledger.besu.datatypes.BlobGas;
+import org.hyperledger.besu.datatypes.Hash;
+import org.hyperledger.besu.datatypes.RequestType;
+import org.hyperledger.besu.ethereum.api.jsonrpc.internal.parameters.EnginePayloadParameter;
+import org.hyperledger.besu.ethereum.core.BlockHeader;
+import org.hyperledger.besu.ethereum.core.Difficulty;
+import org.hyperledger.besu.ethereum.core.Request;
+import org.hyperledger.besu.ethereum.mainnet.BodyValidation;
+import org.hyperledger.besu.ethereum.mainnet.MainnetBlockHeaderFunctions;
+import org.hyperledger.besu.tests.acceptance.dsl.account.Accounts;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+import org.web3j.crypto.Credentials;
+import org.web3j.protocol.Web3j;
+
+/**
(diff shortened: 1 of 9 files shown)


besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/config/LineaTransactionValidatorCliOptions.java
```diff
@@ -27,6 +30,14 @@ public class LineaTransactionValidatorCliOptions implements LineaCliOptions {
       description = "Enable blob transactions? (default: ${DEFAULT-VALUE})")
   private boolean blobTxEnabled = DEFAULT_BLOB_TX_ENABLED;
 
+  @CommandLine.Option(
+      names = {DELEGATE_CODE_TX_ENABLED},
+      arity = "0..1",
+      hidden = true,
+      paramLabel = "<BOOLEAN>",
+      description = "Enable delegate code transactions? (default: ${DEFAULT-VALUE})")
```
fluentcrafter: shall we add the correspondent EIP to the description?

besu-plugins/linea-sequencer/acceptance-tests/src/test/java/linea/plugin/acc/test/EIP7702TransactionDenialTest.java
fluentcrafter, line 1: IIUC, this test ensures that `engine_newPayload` does not import blocks with EIP7702 txs. However, is there a way to make sure that when we call `engine_forkChoiceUpdated(..., newPayloadAttributes)` that triggers block building and then `engine_getPayload`, the created payload does not contain that tx?  

This would ensure that no matter what, the sequencer does not include the tx....

Otherwise, I see a gap: TX makes to TX pool of the sequencer through P2P (e.g Geth nodes, partners running Besu without plugin enabled) and may get included...

kyzooghost: Hmm the test `eip7702TransactionIsRejectedFromTransactionPool` in this file is explicitly testing for rejection from tx pool. To be precise it's testing for txRequests sent directly from a user to the node, rather than via P2P.

kyzooghost: I can look into testing the `engine_forkChoiceUpdated(..., newPayloadAttributes)` route to be sure

kyzooghost: This is the raw `engine_forkChoiceUpdated` JSON-RPC call I got from the Besu logs

```
{
  "jsonrpc": "2.0",
  "method": "engine_forkchoiceUpdatedV3",
  "params": [
    {
      "headBlockHash": "0xe4f2c57368f85809e04fbe90310052da8880b553a4a30fa2aa341f6a5a1bb36a",
      "safeBlockHash": "0xe4f2c57368f85809e04fbe90310052da8880b553a4a30fa2aa341f6a5a1bb36a",
      "finalizedBlockHash": "0xe4f2c57368f85809e04fbe90310052da8880b553a4a30fa2aa341f6a5a1bb36a"
    },
    {
      "timestamp": 5,
      "prevRandao": "0x0000000000000000000000000000000000000000000000000000000000000000",
      "suggestedFeeRecipient": "0x0000000000000000000000000000000000000000",
      "withdrawals": [],
      "parentBeaconBlockRoot": "0x0000000000000000000000000000000000000000000000000000000000000000"
    }
  ],
  "id": 67
}
```

There's no fields to put a transaction in there. So the eip7702 tx must already be in the node txPool prior to this call.

However `eip7702TransactionIsRejectedFromTransactionPool()` tests that we cannot insert the eip7702 tx into the node txPool.

So this prevents us from reaching the scenario you've described of `engine_forkChoiceUpdated(..., newPayloadAttributes)` triggering block building including eip7702 tx.

kyzooghost: This is my mental model of the situation to clarify
image

fluentcrafter: > However the eip7702TransactionIsRejectedFromTransactionPool tests that we cannot put the eip7702 tx into the node txPool.

That prevent's it only if sequencer has the plugin validation in JSON-RPC and P2P broadcasting. I am not sure if the plugin will be called when tx is received through P2P. Maybe @fab-10 can clarify?

fab-10: The validation plugin is always called regardless the source of the tx, so if the tx type is disabled it will never be added to the pool

fluentcrafter (approved):

fab-10 (approved):
LGTM, just one suggestion: could be nicer to have a single option, es. `disable-tx-types`, with a list of type, instead of multiple independent options, that will also support future tx types without new development

kyzooghost:
> LGTM, just one suggestion: could be nicer to have a single option, es. `disable-tx-types`, with a list of type, instead of multiple independent options, that will also support future tx types without new development

Will keep this in mind for the next tx-type block or refactor ticket :)

PR #1307, opened 2025-08-12
perf(mpts): use the batch evaluate lagrange function

AlexandreBelling:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/prover/protocol/compiler/mpts/prover.go
+++ b/prover/protocol/compiler/mpts/prover.go
@@ -222,10 +222,7 @@ func (re RandomPointEvaluation) Run(run *wizard.ProverRuntime) {
 
 	}
 
-	ys := make([]fext.Element, len(polyVals))
-	for i := range ys {
-		ys[i] = smartvectors.EvaluateLagrangeFullFext(polyVals[i], r) //TODO@yao : check here EvaluateLagrangeMixed or EvaluateLagrangeFullFext, check witness and poly type,
-	}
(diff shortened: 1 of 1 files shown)


prover/protocol/compiler/mpts/prover.go
```diff
@@ -222,10 +222,7 @@ func (re RandomPointEvaluation) Run(run *wizard.ProverRuntime) {
 
 	}
 
-	ys := make([]fext.Element, len(polyVals))
-	for i := range ys {
-		ys[i] = smartvectors.EvaluateLagrangeFullFext(polyVals[i], r) //TODO@yao : check here EvaluateLagrangeMixed or EvaluateLagrangeFullFext, check witness and poly type,
-	}
+	ys := smartvectors.BatchEvaluateLagrangeExt(polyVals, r)
```
YaoJGalteland, line 225: go test -timeout 30s -run ^TestCompiler$ github.com/consensys/linea-monorepo/prover/protocol/compiler/mpts 

this test fails with a panic: 
--- FAIL: TestCompiler (0.00s)
--- FAIL: TestCompiler/two-poly-one-point-different-sizes-simple-values (0.00s)
panic: Had a panic: vector.InnerProduct: vectors don't have the same length

AlexandreBelling: Alright, surely we should panic when all vectors don't have the same length

PR #1325, opened 2025-08-20
Added a special mode to the TracesGeneratorJsonRpcClientV2 to ignore …

Filter94:
…Traces Generator's failures

This PR implements issue(s) #

### Checklist


coordinator/clients/traces-generator-api-client/src/main/kotlin/net/consensys/zkevm/coordinator/clients/TracesGeneratorJsonRpcClientV2.kt
```diff
@@ -65,10 +87,22 @@ class TracesGeneratorJsonRpcClientV2(
...
         )
+        if (config.ignoreTracesGeneratorErrors && !result.isSuccess()) {
+          Ok(createFallbackTracesCountersResponse())
+        } else {
+          result
+        }
+      }
+      .exceptionally { throwable ->
+        if (config.ignoreTracesGeneratorErrors) {
+          Ok(createFallbackTracesCountersResponse())
```
jonesho: Is this exceptionally branch really needed here? `if (config.ignoreTracesGeneratorErrors && !result.isSuccess())` should catch the errors from traces counter response, no?

Filter94: If the Traces Generator is unavailable, it's a network error and there is no RPC response, it's an exception

jonesho: So what I'm confused is why would we want to return `createFallbackTracesCountersResponse` for network errors, I thought we only want to do so when we got bad responses from Tracer?

Filter94: It's only useful when `traces.conflation.request-retries.max-retries` are set. And if they are set and the Traces Generator goes down, conflation may just stop, because the max number of retries was reached and traces count or conflation process thrown an exception and it won't be retried

jonesho: So let say if there were network errors or tracers went down which reached max retries, are we okay to return fallback counter or conflated trace responses, assuming that the requests are legit and valid responses should have been returned if the tracer were up, I'm concern in this case, would those fallback counter or conflated trace responses lead to invalid finalization proof later on?

Filter94: It's only applicable to the cases when Dev Prover is used, I.e. we don't care about the finalization

jonesho (approved):

coordinator/clients/traces-generator-api-client/src/main/kotlin/net/consensys/zkevm/coordinator/clients/TracesGeneratorJsonRpcClientV2.kt
```diff
@@ -41,65 +45,111 @@ class TracesGeneratorJsonRpcClientV2(
...
+      tracesCounters = TracesCountersV2.EMPTY_TRACES_COUNT,
+      tracesEngineVersion = config.expectedTracesApiVersion,
+    )
+  }
+
+  private fun createFallbackConflatedTracesResponse(
+    startBlockNumber: ULong,
+    endBlockNumber: ULong,
+  ): GenerateTracesResponse {
+    val defaultFileName = "$startBlockNumber-$endBlockNumber.conflated.${config.expectedTracesApiVersion}.lt"
```
fluentcrafter: we should add a label to the name indicating they are fake/empty
```suggestion
    val defaultFileName = "$startBlockNumber-$endBlockNumber.fake-empty.conflated.${config.expectedTracesApiVersion}.lt"
```

fluentcrafter (dismissed):
LGTM, just left small suggestion on the file name

jonesho (approved):

Issue #1328, opened 2025-08-20
Controller might skip performing MessageSent event log search for certain block range in face of exceptions 


jonesho:
🔐 🔐 `Should this be a security advisory instead?` 🔐 🔐

Controller would possibly skip performing a MessageSent event log search for certain block range if there were exceptions raised during the search loop.

## Impact
__environment [Mainnet]__

__an user is impacted [Y], if yes type of user: [External]__

We'd an incident on mainnet where the coordinator was failing to anchore any message for a long period time, by checking the coordinator log, we found out that shortly after the last anchoring tx, there was an socket timeout exception raised in `L1MessageSentEventsPoller`, but since then the coordinator was no longer anchoring any messages, even though we could see there were many new messages submitted on L1 during that period of time. The timeout exception might not be relevant to the issue (as we can see it happened few time before and after the incident and coordinator was able to resume the anchoring from it), but it led us to check against the error handling inside `L1MessageSentEventsPoller`

By checking the code, we noticed that if there were exception raised inside the retry action in `getLogsLoopingForward`, `cursor.next()` would still be called which would cause the loop to permanently skip the event log search for the particular block range denoted by the previous cursor.

In order to fix this, we should advance the cursor (i.e. to call `cursor.next()`) only when the block range denoted by the current cursor had been searched.

### Steps to Reproduce
_A bug that we can't reproduce is hard to resolve, to ensure the bug is fixed in a timely manner, it is crucial to indicate how to reproduce it. If the bug can't be reproduced, give specific details on how it happened._

Steps to reproduce the bug:

### Actual result
Evidence of the bug including log extracts, screen shot, wallet addresses, Tx hash, ...

### Expected result

_Remember that an issue without proper labels and without assignee is unlikely to get prioritized -> contact the team or the product owner to get it included in the plan._

PR #1330, opened 2025-08-20
1328 controller skip log search for exceptions

jonesho:
This PR implements issue(s) #1328

### Checklist

### DIFF
--- a/jvm-libs/linea/clients/eth-logs-searcher/src/test/kotlin/linea/ethapi/LogsSearcherTest.kt
+++ b/jvm-libs/linea/clients/eth-logs-searcher/src/test/kotlin/linea/ethapi/LogsSearcherTest.kt
@@ -87,7 +87,7 @@ class LogsSearcherTest {
     fakeElClient.setFinalizedBlockTag(450UL)
     // The search should retry on errors when find logs (all will throw 5 times)
     // and not to skip retrieving any of the event logs
-    fakeElClient.setGetLogsThrowErrorAtAndTimes(
+    fakeElClient.setGetLogsBlocksForcedErrorsCounts(
       mutableMapOf(
         1UL to 5U, // will throw error 5 times when find logs in range 1..100
         101UL to 5U, // will throw error 5 times when find logs in range 101..200
@@ -113,7 +113,7 @@ class LogsSearcherTest {
   @Test
   fun `should throw exception when the search timeout reached in face of consistent errors`() {
     fakeElClient.setFinalizedBlockTag(450UL)
-    fakeElClient.setGetLogsThrowErrorAtAndTimes(
+    fakeElClient.setGetLogsBlocksForcedErrorsCounts(
       mutableMapOf(
         401UL to null, // will always throw error when find logs in range 301..450
       ),
--- a/jvm-libs/linea/clients/interfaces/src/testFixtures/kotlin/linea/ethapi/FakeEthApiClient.kt
+++ b/jvm-libs/linea/clients/interfaces/src/testFixtures/kotlin/linea/ethapi/FakeEthApiClient.kt
@@ -36,7 +36,7 @@ class FakeEthApiClient(
 
   // key = which block number to throw error
   // value = how many times left to throw error, null means always throw
-  private var getLogsThrowErrorAtAndTimes: MutableMap<ULong, ULong?> = mutableMapOf()
+  private var getLogsBlocksForcedErrorsCounts: MutableMap<ULong, ULong?> = mutableMapOf()
 
   init {
     require(initialTagsBlocks.keys.size == BlockParameter.Tag.entries.size) {
@@ -76,8 +76,8 @@ class FakeEthApiClient(
   }
 
   @Synchronized
-  fun setGetLogsThrowErrorAtAndTimes(getLogsThrowErrorAtAndTimes: MutableMap<ULong, ULong?>) {
-    this.getLogsThrowErrorAtAndTimes = getLogsThrowErrorAtAndTimes
+  fun setGetLogsBlocksForcedErrorsCounts(getLogsBlocksForcedErrorsCounts: MutableMap<ULong, ULong?>) {
+    this.getLogsBlocksForcedErrorsCounts = getLogsBlocksForcedErrorsCounts
   }
 
   @Synchronized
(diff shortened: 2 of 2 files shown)


jvm-libs/linea/clients/eth-logs-searcher/src/main/kotlin/linea/ethapi/EthLogsSearcherImpl.kt
```diff
@@ -123,8 +124,12 @@ class EthLogsSearcherImpl(
 
         enoughLogsCollected || collectionTimeoutElapsed || noMoreChunksToCollect
       },
+      timeout = searchTimeout,
```
jonesho, line 127: I think we should add a timeout here, otherwise, it would keep retrying indefinitely on errors

jvm-libs/linea/clients/eth-logs-searcher/src/main/kotlin/linea/ethapi/EthLogsSearcherImpl.kt
```diff
@@ -123,8 +124,12 @@ class EthLogsSearcherImpl(
...
     ) {
-      val chunk = cursor.next()
+      if (chunk == null || lastSearchedChunk.get() == chunk) {
+        chunk = cursor.next()
+      }
```
jonesho, line 133: We only advance the cursor if the `lastSearchedChunk` is equal to the current chunk to avoid cursor being advanced when errors happened e.g. `getLogs`

jvm-libs/linea/clients/eth-logs-searcher/src/test/kotlin/linea/ethapi/LogsSearcherTest.kt
```diff
@@ -81,6 +82,57 @@ class LogsSearcherTest {
     assertThat(result.endBlockNumber).isEqualTo(450UL)
   }
 
+  @Test
+  fun `should return logs within block range in face of temporary errors`() {
```
jonesho, line 86: This test would fail (not able to retrieve any event logs) if the newly added condition (https://github.com/Consensys/linea-monorepo/pull/1330/files#diff-b17bfc2505707e4971d05a6bf69d40bf089c733efbcd5fa4f8194056744a05bfR129:~:text=if%20(chunk%20%3D%3D%20null%20%7C%7C%20lastSearchedChunk.get()%20%3D%3D%20chunk) was absent

jvm-libs/linea/clients/interfaces/src/testFixtures/kotlin/linea/ethapi/FakeEthApiClient.kt
```diff
@@ -206,6 +214,20 @@ class FakeEthApiClient(
     val fromBlockNumber: ULong = blockParameterToBlockNumber(fromBlock)
     val toBlockNumber: ULong = blockParameterToBlockNumber(toBlock)
 
+    var shouldThrow = false
+    errorBlockNumbersAndThrowTimes.forEach { errorBlockNumber, throwTimes ->
+      if ((fromBlockNumber..toBlockNumber).contains(errorBlockNumber)) {
+        if (throwTimes != null && throwTimes > 0UL) {
+          errorBlockNumbersAndThrowTimes[errorBlockNumber] = errorBlockNumbersAndThrowTimes[errorBlockNumber]!! - 1UL
+        }
+        shouldThrow = shouldThrow || throwTimes == null || errorBlockNumbersAndThrowTimes[errorBlockNumber]!! > 0UL
```
fluentcrafter: This logic must be in another function: `isInRange()` has a simple and clear responsibility and returns a boolean, and it's used as a filter. Hijacking it's logic this way is an anti-pattern and code-smell. Therefore, the throw error logic shall be somewhere else IMO.

Also, the name and method above, `private var errorBlockNumbersAndThrowTimes: MutableMap<ULong, ULong?> = mutableMapOf()`, suggest that `eth_getBlockByNumber` would fail as well...

We should make it clear that it applies to the eth_getLogs response.

jvm-libs/linea/clients/eth-logs-searcher/src/main/kotlin/linea/ethapi/EthLogsSearcherImpl.kt
fluentcrafter: - We should add a comment here why this `if`, it's not obvious.

fluentcrafter (approved):

jvm-libs/linea/clients/interfaces/src/testFixtures/kotlin/linea/ethapi/FakeEthApiClient.kt
```diff
@@ -34,6 +34,10 @@ class FakeEthApiClient(
   private val logsDb: MutableList<EthLog> = mutableListOf()
   private val blocksDb: MutableMap<ULong, Block> = mutableMapOf()
 
+  // key = which block number to throw error
+  // value = how many times left to throw error, null means always throw
+  private var getLogsThrowErrorAtAndTimes: MutableMap<ULong, ULong?> = mutableMapOf()
```
fluentcrafter:
```suggestion
  private var getLogsBlocksForcedErrorsCounts: MutableMap<ULong, ULong?> = mutableMapOf()
```

jvm-libs/linea/clients/interfaces/src/testFixtures/kotlin/linea/ethapi/FakeEthApiClient.kt
```diff
@@ -198,6 +215,19 @@ class FakeEthApiClient(
     return logsDb.filter { isInRange(it.blockNumber, fromBlock, toBlock) }
   }
 
+  private fun isAtBlockNumberToThrow(fromBlockNumber: ULong, toBlockNumber: ULong): Boolean {
+    var shouldThrow = false
+    getLogsThrowErrorAtAndTimes.forEach { errorBlockNumber, throwTimes ->
+      if ((fromBlockNumber..toBlockNumber).contains(errorBlockNumber)) {
```
fluentcrafter: nit: Depending on how smart the compiler is, this may be very inefficient because it's creating a range every iteration of the loop. However, `(fromBlockNumber..toBlockNumber)` can be cached outside the loop.

Issue #1379, opened 2025-09-03
Intergrate Maru with the E2E local stack


Filter94:
## Description
Since we're soon switching from Clique -> Paris -> Prague and it needs Maru CL client.

## Motivation
We need Maru in our E2E stack to test if all the components are working well with Maru and with Paris and Prague EVM versions

## Tasks
- [ ] Add Maru client to the local stack to drive the chain
- [ ] Adjust E2E tests to generate Prague timestamp

## Acceptance criteria
- [ ] E2E tests work with Maru, Prague transactions are executable in the E2E stack

## Risks
- [ ] ???

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #1381, opened 2025-09-03
feat: update besu and plugin versions

jonesho:
This PR implements issue(s) #1379

### Checklist

### DIFF
--- a/config/common/traces-limits-v2.toml
+++ b/config/common/traces-limits-v2.toml
@@ -85,6 +85,6 @@ POINT_EVAL          = 0
 
 # Introduced from beta-v4.0-rc6-CANCUN and present on Cancun
 BLS_DATA                = 1073741824
-RLP_UTILS               = 131072 # Note: need to confirm if this limit could have the same limit as RLP_TXN
+RLP_UTILS               = 131072
 POWER_REFERENCE_TABLE   = 2147483647
 BLS_REFERENCE_TABLE     = 2147483647


coordinator/clients/traces-generator-api-client/src/main/kotlin/net/consensys/zkevm/coordinator/clients/TracesClientResponsesParser.kt
```diff
@@ -55,17 +53,25 @@ object TracesClientResponsesParser {
...
+          .onFailure {
+            log.error(
+              "Failed to parse Evm module ${traceModule.name}='$counterValue' to UInt. errorMessage={}",
+              it.message,
+              it,
+            )
+          }
+          .getOrThrow()
+      } else {
+        0U
```
Filter94, line 73: Why was this fallback added? Does Trace Generator return different set of modules depending on the fork?

docker/config/l2-genesis-initialization/genesis-maru.json.template
```diff
@@ -0,0 +1,20 @@
...
+    "0": {
+      "type": "difficultyAwareQbft",
+      "blockTimeSeconds": 1,
+      "postTtdConfig": {
+        "validatorSet": ["0x083e11f14b68d5ebeb4be3e3e0237ebbc5675ab5"],
+        "elFork": "Paris"
+      },
+      "terminalTotalDifficulty": 8
+    },
+    "1789631893": {
```
Filter94, line 13: This is necessary, but the timestamp is set far ahead, so we never hit it

fluentcrafter (approved):

PR #1397, opened 2025-09-05
coordinator: optimize conflation to skip trace conflation and shomei …

fluentcrafter:
# Description 
optimise conflation flow by to skipping 1) trace-api conflation and shomei when proof response already exists

Biggest inpact upon coordinator restart and conflation catch up.

### Checklist

### DIFF
--- a/coordinator/core/src/main/kotlin/net/consensys/zkevm/coordinator/clients/ProverClient.kt
+++ b/coordinator/core/src/main/kotlin/net/consensys/zkevm/coordinator/clients/ProverClient.kt
@@ -1,10 +1,18 @@
 package net.consensys.zkevm.coordinator.clients
 
+import net.consensys.zkevm.domain.ProofIndex
 import net.consensys.zkevm.domain.ProofToFinalize
 import net.consensys.zkevm.domain.ProofsToAggregate
 import tech.pegasys.teku.infrastructure.async.SafeFuture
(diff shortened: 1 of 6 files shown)


coordinator/core/src/main/kotlin/net/consensys/zkevm/coordinator/clients/ProverClient.kt
```diff
@@ -1,10 +1,18 @@
 package net.consensys.zkevm.coordinator.clients
 
+import net.consensys.zkevm.domain.ProofIndex
 import net.consensys.zkevm.domain.ProofToFinalize
 import net.consensys.zkevm.domain.ProofsToAggregate
 import tech.pegasys.teku.infrastructure.async.SafeFuture
 
-interface ProverClient<ProofRequest, ProofResponse> {
+interface ProverProofResponseChecker<ProofResponse> {
```
jonesho, line 8: nit: can we rename this to `ProofResponseChecker`, imo `ProverProofResponseChecker` is a bit verbose

fluentcrafter: Verbose Yes, also it helps with consistency of having Prover prefix which is easier to reason about and search through the code, that was my motivation

PR #1404, opened 2025-09-09
perf: revamp `symbolic.EvalBoard`

gbotrel:
Most of this PR is in `evaluation.go` . 
It gets rid of the multiple `Evaluate` `EvaluateMixed` and `EvaluatedExt` similar functions to keep only one.

The `boardAssignment` is replaced by the `evaluation` struct that allocate the needed memory for each go routine before looping on the chunks it needs to solve. The new evaluation logic also bypasses the `arithmetic_ext / _ mixed etc`.go files. The inputs are smart vectors, they are written in the pre-allocated arrays in the `evaluation` struct, which account for their type (constant, rotated etc). 

Since memory is known in advance for each chunk (i.e. set of rows we process in a go routine), there is no need for the memory pool either.

Perf against `dev-small-fields` for realistic segment:

```
benchmark                                  old ns/op        new ns/op        delta
BenchmarkCompiler/realistic-segment-96     184955073455     156563834619     -15.35%

benchmark                                  old allocs      new allocs      delta
BenchmarkCompiler/realistic-segment-96     22627949643     22340680827     -1.27%

benchmark                                  old bytes         new bytes        delta
BenchmarkCompiler/realistic-segment-96     1071664302168     800542197232     -25.30%
```

(this is without @YaoJGalteland other branches on poseidon2 bytes consumtions, fft domains, avx512 method usage etc).


prover/circuits/prove.go
```diff
@@ -86,7 +86,6 @@ func ProveCheck(setup *Setup, assignment frontend.Circuit, opts ...any) (plonk.P
 			assignment,
 			assignment,
 			setup.Circuit.Field(),
-			test.WithBackendProverOptions(proverOpts...),
```
Tabaie, line 89: @ivokub argues that removing this is okay because it is only ever used for hint substitution, which would not be necessary for the test engine.

prover/symbolic/evaluation.go
```diff
@@ -2,195 +2,460 @@ package symbolic
...
 
-			var (
-				chunkStart = chunkID * MaxChunkSize
-				chunkStop  = (chunkID + 1) * MaxChunkSize
-			)
+	})
+	return sv.NewRegularExt(res)
+}
+
+func newEvaluation[T chunkValue](b *ExpressionBoard) evaluation[T] {
```
YaoJGalteland, line 152: The current evaluation system only supports the same type for all operations and does not support mixed type operations. For example, `PolyEval` only computes when both the `input` and `x` are either base elements or ext elements. 
- The PolyEval in evalNodeBase, where both input and x are base elements, doesn’t make sense and is never called. x is ext value by default
-  A potential case I haven’t found is PolyEval with input = base and x = ext, but this is probably covered in evalNodeExt by lifting input to ext at some point. A more fine-grained mixed case could be done in another PR if we plan to optimize it, but it’s probably not so straightforward.

gbotrel: yes I benchmarked and check the realistic segment, doing the mixed arithemtic complicates things for virtually no gains.

PR #1414, opened 2025-09-11
Fix the sticky failing client in LoadBalancingJsonRpcClient

gauravahuja:
[This PR fixes the issue in mentioned the slack message](https://consensys.slack.com/archives/C05J55YUDUZ/p1757403729927089?thread_ts=1757349472.306389&cid=C05J55YUDUZ)
### Checklist


fluentcrafter (commented):
Thanks for the fix. Can you please add a short comment in lines 115 why we selecte by min totalReqeusts and what problem it solves? 

The test helps on that, but not require some effrot to understand.

Issue #1416, opened 2025-09-11
Add proven aggregation block tracking metric for coordinator troubleshooting


jasonyic:
## Description
Add a new metric to track the highest consecutive proven aggregation block number in the coordinator dashboard to improve troubleshooting capabilities.

## Motivation
Currently, we lack visibility into the highest consecutive proven aggregation block number, which is crucial for troubleshooting coordinator issues. This metric would help identify gaps in the aggregation proving process and potential bottlenecks in the pipeline.

## Tasks
Add a new row/panel to the coordinator dashboard displaying "Highest Consecutive Proven Aggregation Block Number" with the following implementation options:

Prometheus Metric (Recommended)
Add a new Prometheus metric coordinator_highest_consecutive_proven_block to the coordinator service
Implement periodic calculation and export of this metric
Add corresponding Grafana panel using PromQL query

The metric should be calculated using the following logic:

```
WITH previous_ends AS (
  SELECT *,
    LAG(end_block_number, 1) OVER (ORDER BY end_block_number ASC) AS previous_end_block_number
  FROM $aggregationsTable
  WHERE start_block_number >= $1
  ORDER BY start_block_number ASC
),
first_gapped_aggregation AS (
  SELECT start_block_number
  FROM previous_ends
  WHERE start_block_number > $1 
    AND start_block_number - 1 != previous_end_block_number
  LIMIT 1
)
SELECT * FROM previous_ends
WHERE EXISTS (SELECT 1 FROM previous_ends WHERE start_block_number = $1)
  AND (previous_ends.previous_end_block_number = previous_ends.start_block_number - 1 
       OR previous_ends.start_block_number = $1)
  AND ((SELECT COUNT(1) FROM first_gapped_aggregation) = 0 
       OR previous_ends.start_block_number < (SELECT * FROM first_gapped_aggregation))
LIMIT $2
```
Where:

$1 = lastFinalizedBlock + 1
$2 = reasonable limit for query performance
$aggregationsTable = aggregations table name

## Acceptance criteria
- [ ] List tests that need to be implemented or checks to consider this task completed. 

## Risks
- [ ] These are risks that may prevent completion of the task as well as areas to pay special attention to

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

jasonyic:
[...]

Issue #1415, opened 2025-09-11
Coordinator should add retry mechanism for retrieving L2 data when creating aggregation proof requests


jonesho:
🔐 🔐 `Should this be a security advisory instead?` 🔐 🔐

## Description
Coordinator should add retry mechanism for L2 data retrieval when creating aggregation proof requests. On mainnet, we had an incident of coordinator not creating certain aggregation proof requests due to some L2 RPC errors when creating aggregation proof requests. The lack of retry would require coordinator to maually restart to recover which is not ideal, we should add retry mechanism on aggregation proof creation just like what we did for exexcution proof creation (i.e. retry on exceptions including conflated-traces or StateMerkleProof request failures)

E.g. Coordinator Logs (these three aggregation proof requests were missing due to the RPC errors):
```
time=2025-09-10T16:56:51,429Z level=ERROR message=failed to get parent aggregation l2 message rolling hash: aggregation=[23160065..23160315]251 errorMessage=java.net.SocketTimeoutException: timeout | logger=ProofAggregationCoordinatorService thread=pool-2-thread-272 | java.util.concurrent.CompletionException: java.net.SocketTimeoutException: timeout\n	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(Unknown Source)\n...
	
time=2025-09-10T16:56:40,293Z level=ERROR message=failed to get parent aggregation l2 message rolling hash: aggregation=[23160316..23160474]159 errorMessage=java.net.SocketTimeoutException: timeout | logger=ProofAggregationCoordinatorService thread=pool-2-thread-304 | java.util.concurrent.CompletionException: java.net.SocketTimeoutException: timeout\n...

time=2025-09-10T17:03:40,173Z level=ERROR message=failed to get parent aggregation l2 message rolling hash: aggregation=[23160647..23160820]174 errorMessage=java.net.ConnectException: Failed to connect to linea-geth-archive-v2/172.20.12.1:8545 | logger=ProofAggregationCoordinatorService thread=pool-2-thread-145 | java.util.concurrent.CompletionException: java.net.ConnectException: Failed to connect to linea-geth-archive-v2/172.20.12.1:8545\n...
```
## Motivation
This section should indicate what value we are receiving from it.

## Tasks
- [x] Implement retry mechanism for aggregation proof request creation
- [x] Add metric to show "highest consecutive proven aggregation block number" (#1416)

## Acceptance criteria
[...]

PR #1433, opened 2025-09-15
feat: add retry for creating aggregation proof request and add metric…

jonesho:
… for consecutive aggregation proofs

This PR implements issue(s) #1415 and #1416

### Checklist

### DIFF
--- a/besu-plugins/linea-sequencer/acceptance-tests/src/test/resources/moduleLimits.toml
+++ b/besu-plugins/linea-sequencer/acceptance-tests/src/test/resources/moduleLimits.toml
@@ -94,5 +94,4 @@ BLOCK_TRANSACTIONS = 200 # max number of tx in an L2 block
 # Fixed size, static tables
 #
 BIN_REFERENCE_TABLE = 262144 # contains 3 * 256^2 + 256 data rows + 1 padding row
-SHF_REFERENCE_TABLE = 4096   # contains 9 * 256         data rows + 1 padding row
 INSTRUCTION_DECODER = 512    # contains 256             data rows + 1 padding row
--- a/besu-plugins/linea-sequencer/acceptance-tests/src/test/resources/moduleLimits_sendRawTx.toml
+++ b/besu-plugins/linea-sequencer/acceptance-tests/src/test/resources/moduleLimits_sendRawTx.toml
@@ -57,7 +57,6 @@ BLOCK_TRANSACTIONS       = 200 # max number of tx in an L2 block
 # Fixed size, static tables
 #
 BIN_REFERENCE_TABLE = 262144 # contains 3 * 256^2 + 256 data rows + 1 padding row
-SHF_REFERENCE_TABLE = 4096   # contains 9 * 256         data rows + 1 padding row
 INSTRUCTION_DECODER = 512    # contains 256             data rows + 1 padding row
 
 #
--- a/besu-plugins/linea-sequencer/acceptance-tests/src/test/resources/noModuleLimits.toml
+++ b/besu-plugins/linea-sequencer/acceptance-tests/src/test/resources/noModuleLimits.toml
@@ -49,7 +49,6 @@ WCP                 = 262144
 # Fixed size, static tables
 #
 BIN_REFERENCE_TABLE = 262144 # contains 3 * 256^2 + 256 data rows + 1 padding row
-SHF_REFERENCE_TABLE = 4096   # contains 9 * 256         data rows + 1 padding row
 INSTRUCTION_DECODER = 512    # contains 256             data rows + 1 padding row
 
 #
--- a/besu-plugins/linea-sequencer/acceptance-tests/src/test/resources/strictModuleLimits.toml
+++ b/besu-plugins/linea-sequencer/acceptance-tests/src/test/resources/strictModuleLimits.toml
@@ -58,7 +58,6 @@ BLOCK_TRANSACTIONS       = 200 # max number of tx in an L2 block
 # Fixed size, static tables
 #
 BIN_REFERENCE_TABLE = 262144 # contains 3 * 256^2 + 256 data rows + 1 padding row
-SHF_REFERENCE_TABLE = 4096   # contains 9 * 256         data rows + 1 padding row
 INSTRUCTION_DECODER = 512    # contains 256             data rows + 1 padding row
 
 #
--- a/besu-plugins/linea-sequencer/acceptance-tests/src/test/resources/txOverflowModuleLimits.toml
+++ b/besu-plugins/linea-sequencer/acceptance-tests/src/test/resources/txOverflowModuleLimits.toml
@@ -57,7 +57,6 @@ BLOCK_TRANSACTIONS       = 200 # max number of tx in an L2 block
 # Fixed size, static tables
 #
 BIN_REFERENCE_TABLE = 262144 # contains 3 * 256^2 + 256 data rows + 1 padding row
-SHF_REFERENCE_TABLE = 4096   # contains 9 * 256         data rows + 1 padding row
 INSTRUCTION_DECODER = 512    # contains 256             data rows + 1 padding row
 
 #
--- a/besu-plugins/linea-sequencer/sequencer/src/test/resources/sequencer/line-limits.toml
+++ b/besu-plugins/linea-sequencer/sequencer/src/test/resources/sequencer/line-limits.toml
@@ -50,7 +50,6 @@ WCP                 = 262144
 # Fixed size, static tables
 #
 BIN_REFERENCE_TABLE = 262144 # contains 3 * 256^2 + 256 data rows + 1 padding row
-SHF_REFERENCE_TABLE = 4096   # contains 9 * 256         data rows + 1 padding row
 INSTRUCTION_DECODER = 512    # contains 256             data rows + 1 padding row
 
 #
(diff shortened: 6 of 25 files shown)


coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/aggregation/ProofAggregationCoordinatorService.kt
```diff
@@ -215,6 +207,57 @@ class ProofAggregationCoordinatorService(
...
+              }
+              .whenException {
+                log.error(
+                  "Error getting consecutive aggregation end block number from DB: aggregation={} errorMessage={}",
+                  blobsToAggregate.intervalString(),
+                  it.message,
+                  it,
+                )
+              }
+          }
```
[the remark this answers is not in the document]
jonesho: False postive, that's the intended behavior

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflation/ConflationApp.kt
```diff
@@ -264,6 +272,8 @@ class ConflationApp(
         targetEndBlockNumbers = configs.conflation.proofAggregation.targetEndBlocks ?: emptyList(),
         metricsFacade = metricsFacade,
         provenAggregationEndBlockNumberConsumer = { aggEndBlockNumber -> highestAggregationTracker(aggEndBlockNumber) },
```
gauravahuja, line 290: Is `provenAggregationEndBlockNumberConsumer` one of any use now?

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/aggregation/ProofAggregationCoordinatorService.kt
```diff
@@ -167,33 +172,23 @@ class ProofAggregationCoordinatorService(
     }
     val blockIntervals = BlockIntervals(startingBlockNumber, upperBoundaries)
 
-    aggregationL2StateProvider
-      .getAggregationL2State(blockNumber = blobsToAggregate.startBlockNumber.toLong() - 1)
-      .whenException {
-        log.error(
-          "failed to get parent aggregation l2 message rolling hash: aggregation={} errorMessage={}",
-          blobsToAggregate.intervalString(),
+    AsyncRetryer.retry(
```
gauravahuja, line 178: Why no make the flow inside aggregationL2StateProvider retryable instead of adding a retry here?

gauravahuja (approved):
LGTM, just a few clarifying questions

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/aggregation/ProofAggregationCoordinatorService.kt
```diff
@@ -215,6 +210,57 @@ class ProofAggregationCoordinatorService(
...
+          .thenPeek {
+            aggregationsRepository.findHighestConsecutiveEndBlockNumber()
+              .thenApply { it ->
+                if (it != null) {
+                  provenConsecutiveAggregationEndBlockNumberConsumer.accept(it.toULong())
+                }
+              }
+              .whenException {
+                log.error(
+                  "Error getting consecutive aggregation end block number from DB: aggregation={} errorMessage={}",
```
fluentcrafter: maybe this shall be a warning: 
- 1) It will be retried later. 
- 2) it does not affect the correctness or core functionality of the system. If DB fails will have real errors on other critical areas e.g saving to the DB

fluentcrafter (dismissed):

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflation/ConflationApp.kt
jonesho: Yes, it's been used in the dashboard here: https://grafana.o11y.web3factory.consensys.net/d/admexryn432m8d/linea-coordinator?var-rate_interval=1h&var-network=mainnet&var-cluster=linea-prod-eks&from=2025-09-16T02:43:27.176Z&to=2025-09-16T05:43:27.176Z&timezone=browser&orgId=17&refresh=5s&viewPanel=panel-27

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/aggregation/ProofAggregationCoordinatorService.kt
jonesho: Yes, agree

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/aggregation/ProofAggregationCoordinatorService.kt
jonesho: It's because I want to `aggregationL2StateProvider` to be solely responsible for getting the l2 state data, and the retry would be up to the caller of the functions from `aggregationL2StateProvider`

coordinator/core/src/test/kotlin/net/consensys/zkevm/ethereum/coordination/aggregation/ProofAggregationCoordinatorServiceTest.kt
Filter94, line 1: I think we should add a test to test the resilience against failures

Filter94 (approved):

coordinator/persistence/aggregation/src/main/kotlin/net/consensys/zkevm/persistence/dao/aggregation/PostgresAggregationsDao.kt
```diff
@@ -295,20 +304,30 @@ class PostgresAggregationsDao(
   }
 
   override fun findHighestConsecutiveEndBlockNumber(
-    fromBlockNumber: Long,
+    fromBlockNumber: Long?,
```
fluentcrafter, line 337: Why are we changing this? IIRC, this was meant to be called with `lastFinalizedBlockNumber + 1` to guarantee aggregation continuity. 

With this change, if we have lastFinalizedBlockNumber = 99, aggregations: [100..109, 110..119, 120..129]
and  110..119, 120..129 are proven, it will return 129 when it should retrun null. This can be problematic because will give the illusion the system is working fine, but there is a bug in the metrics...

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/aggregation/ProofAggregationCoordinatorService.kt
```diff
@@ -215,6 +210,57 @@ class ProofAggregationCoordinatorService(
               it,
             )
           }
+          .thenPeek {
+            aggregationsRepository.findHighestConsecutiveEndBlockNumber()
```
fluentcrafter: This can have a bug. Please see my comment on the DAO class

We shall fix this.

fluentcrafter (changes requested):

coordinator/persistence/aggregation/src/main/kotlin/net/consensys/zkevm/persistence/dao/aggregation/PostgresAggregationsDao.kt
jonesho: @fluentcrafter The current logic here is that if `fromBlockNumber` is not given, it will set the `startBlockNumber` from the first row as `fromBlockNumber`, and given that the first row representing the "last finalized aggregation" should always be there in db (by coordinator current implementation), then in your case, it will return 99 instead (i.e. find the highest consecutive end block number from first row if `fromBlockNumber` is not given)

The only corner case for the metric is that, when the coordinator and db are started from scratch, before the first aggregation created [1..10], there could be a slight possibility that the second aggregation [11, 20] created before the first one, and hence the metric "consecutive highest proven aggregation end block number" would jump from 0 to 20, instead of 0 to 10 to 20.

fluentcrafter: > and given that the first row representing the "last finalized aggregation" should always be there in db (by coordinator's current implementation),

That works. However, it's not a good design principle because the Dao has some logic that makes assumptions about other logic outside of the Dao component.

IMO, a better alternative would be to `ProofAggregationCoordinatorService` to have a `lastFinilizedBlockNumberSupplier: () -> ULong` and call `findHighestConsecutiveEndBlockNumber(lastFinilizedBlockNumberSupplier())`

WDYT?

jonesho: @fluentcrafter thanks for the suggestion, but imo, the `PostgresAggregationsDao` doesn't make any assumption about the logic outside of the Dao component, all it does is to try fetch the first row from the aggregation table as `fromBlockNumber` if it's not explicitly passed to the `findHighestConsecutiveEndBlockNumber` call, it aligns with the function name which is to `findHighestConsecutiveEndBlockNumber` from the aggregation table

Anyway, I'd still like to go for your suggestion, as to address the corner cases like the one mentioned above

fluentcrafter (approved):

PR #1434, opened 2025-09-16
perf: better transversal hash and improve split extension

gbotrel:
Improve parallelization and cache locality of `TransversalHash` and some tweaks to split.

```
benchmark                                  old ns/op       new ns/op       delta
BenchmarkCompiler/realistic-segment-96     41610556430     12732767142     -69.40%

benchmark                                  old allocs     new allocs     delta
BenchmarkCompiler/realistic-segment-96     115030145      112697284      -2.03%

benchmark                                  old bytes        new bytes        delta
BenchmarkCompiler/realistic-segment-96     185684402344     172041072160     -7.35%
```

### DIFF
--- a/prover/crypto/ringsis/ringsis.go
+++ b/prover/crypto/ringsis/ringsis.go
@@ -252,7 +252,8 @@ func (s *Key) TransversalHash(v []smartvectors.SmartVector, res []field.Element)
 		for col := start; col < end; col += windowSize {
 			for i := 0; i < nbRows; i++ {
 				for j := range transposed {
-					transposed[j][i] = v[i].Get(col)
+					// TODO need careful review; tests passes with v[i].Get(col) and v[i].Get(col+j)
+					transposed[j][i] = v[i].Get(col + j)
 				}
 			}
 			for j := range transposed {


prover/protocol/compiler/splitextension/split.go
```diff
@@ -264,27 +266,39 @@ func (vctx *VerifierCtx) RunGnark(api frontend.API, run wizard.GnarkRuntime) {
...
-	for i := 0; i < 4; i++ {
-		buf[i] = make([]field.Element, size)
-	}
-	parallel.Execute(size, func(start, end int) {
-		for i := start; i < end; i++ {
+
+	switch sv.(type) {
+	case *smartvectors.Regular, *smartvectors.Constant, *smartvectors.PaddedCircularWindow:
+		return [4]smartvectors.SmartVector{
+			sv, // TODO @gbotrel check if we need to copy here.
```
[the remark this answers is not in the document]
gbotrel: cc @AlexandreBelling / @ThomasPiellard  is it ok here if I don't do a deep copy of the input ?

prover/crypto/ringsis/ringsis.go
```diff
@@ -225,63 +211,56 @@ func (s *Key) FlattenedKey() []field.Element {
...
+			windowSize /= 2
+		}
+		transposed := make([][]field.Element, windowSize)
+		for i := range transposed {
+			transposed[i] = make([]field.Element, nbRows)
+		}
+		for col := start; col < end; col += windowSize {
+			for i := 0; i < nbRows; i++ {
+				for j := range transposed {
+					// TODO need careful review; tests passes with v[i].Get(col) and v[i].Get(col+j)
```
gbotrel, line 255: this is odd -- not sure the vortex package tests correctly / extensively. 
can we add some extra tests ? cc @YaoJGalteland  / @ThomasPiellard

YaoJGalteland: The `ringsis` package does not test `TransversalHash`, so regardless of if you use `v[i].Get(col)` or `v[i].Get(col+j)`, the tests will pass. We have plenty of test cases in `TestProver` in crypto/vortex/commitment_test.go, which calls `TransversalHash`, and in that case `v[i].Get(col)` will fail, only `v[i].Get(col+j)` can pass.

PR #1437, opened 2025-09-16
Implemented confaltion calculators cutting off the hard forks

Filter94:
This PR implements issue(s) #

### Checklist


coordinator/app/src/main/kotlin/linea/coordinator/config/v2/ConflationConfig.kt
```diff
@@ -42,5 +43,11 @@ data class ConflationConfig(
     val coordinatorPollingInterval: Duration = 3.seconds,
     val targetEndBlocks: List<ULong>? = null,
     val aggregationSizeMultipleOf: UInt = 1u,
+    val hardForks: HardForks = HardForks(),
+  )
+
+  data class HardForks(
+    val timestampBasedForks: List<Instant> = emptyList(),
+    val totalTerminalDifficulty: ULong = ULong.MAX_VALUE,
```
fluentcrafter: I suggest making it nullable/optional and null by default.

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/ConflationConfig.kt
```diff
@@ -42,5 +43,11 @@ data class ConflationConfig(
     val coordinatorPollingInterval: Duration = 3.seconds,
     val targetEndBlocks: List<ULong>? = null,
     val aggregationSizeMultipleOf: UInt = 1u,
+    val hardForks: HardForks = HardForks(),
```
fluentcrafter: Suggestions:
- 1. put `hardforks` at parent conflation level because it applies to batches/blobs on cutoff (blobs need to be sealed there even if it almost empty)
- 2. Make it optional, default to null. Otherwise, after we reach the hardforks, we always have at least the TTD one set, which is noisy IMO and there is no value on keeping it after they get finalized on L1.

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/ConflationToml.kt
```diff
@@ -47,6 +49,7 @@ data class ConflationToml(
     val coordinatorPollingInterval: Duration = 3.seconds,
     val targetEndBlocks: List<ULong>? = null,
     val aggregationSizeMultipleOf: UInt = 1u,
+    val hardForks: HardForksToml = HardForksToml(),
```
fluentcrafter: same notes as above in CoordinatorConfig class.

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflation/ConflationApp.kt
```diff
@@ -465,12 +471,46 @@ class ConflationApp(
...
+        TimestampHardForkConflationCalculator(
+          hardForkTimestamps = configs.conflation.proofAggregation.hardForks.timestampBasedForks,
+          initialTimestamp = initialTimestamp,
+        ),
+      )
+      log.info(
+        "Added timestamp-based hard fork calculator with {} timestamps, initialized at {}",
+        configs.conflation.proofAggregation.hardForks.timestampBasedForks.size,
+        initialTimestamp,
+      )
```
fluentcrafter: IMO its usefull to log the full list of forks.

```suggestion
      log.info(
        "Added timestamp-based hard fork calculator: initialized at {} with timestamps={}, ",
        initialTimestamp,
        configs.conflation.proofAggregation.hardForks.timestampBasedForks
      )
```

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflation/ConflationApp.kt
```diff
@@ -465,12 +471,46 @@ class ConflationApp(
...
+  }
+
+  private fun addTtdHardForkCalculator(calculators: MutableList<ConflationCalculator>, block: Block) {
+    calculators.add(
+      TtdHardForkConflationCalculator(
+        totalTerminalDifficulty = configs.conflation.proofAggregation.hardForks.totalTerminalDifficulty,
+        initialTotalDifficulty = block.totalDifficulty,
+      ),
+    )
+    log.info(
```
fluentcrafter: We should really make hardforks and TTD optional, other this line will always be logged and can be confusing after we reach TTD finalized on L1, becasue we should be able to just remove this...

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflation/ConflationApp.kt
```diff
@@ -373,11 +378,12 @@ class ConflationApp(
     lastProvenConsecutiveBatchBlockNumberProvider
   }
 
+  private val extendedWeb3JImpl = ExtendedWeb3JImpl(l2Web3jClient)
```
fluentcrafter:
```suggestion
  private val extendedWeb3J = ExtendedWeb3JImpl(l2Web3jClient)
```

coordinator/core/src/main/kotlin/net/consensys/zkevm/domain/Conflation.kt
```diff
@@ -48,11 +48,11 @@ enum class ConflationTrigger(val triggerPriority: Int) {
   // Business logic needs priority to pick the trigger in case multiple calculators trigger conflation.
   // TARGET_BLOCK_NUMBER needs to be the highest priority as it is used as conflation, blob and aggregation boundary.
   TARGET_BLOCK_NUMBER(1),
-  DATA_LIMIT(2),
-  TRACES_LIMIT(3),
-  TIME_LIMIT(4),
-  BLOCKS_LIMIT(5),
-  SWITCH_CUTOFF(6),
```
fluentcrafter, line 55: why supressing the `SWITCH_CUTOFF` ? 
This was used before in scenarios without EVM upgrades or hardforks, e.g: Linea Protocol upgrade requiring a prover/verifier upgrade. 
So IMO we should keep the support for it.

coordinator/core/src/main/kotlin/net/consensys/zkevm/domain/Conflation.kt
```diff
@@ -75,14 +75,14 @@ data class BlockCounters(
...
 ) {
-
   override fun toString(): String {
     return "BlockCounters(blockNumber=$blockNumber, " +
       "blockTimestamp=$blockTimestamp, " +
       "tracesCounters=$tracesCounters, " +
-      "blockRLPEncoded=${blockRLPEncoded.size}bytes), " +
-      "numOfTransactions=$numOfTransactions"
+      "blockRLPEncoded=${blockRLPEncoded.size}bytes, " +
+      "totalDifficulty=$totalDifficulty,)"
```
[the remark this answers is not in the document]
fluentcrafter: cursor is rigth :)

coordinator/app/src/test/kotlin/linea/coordinator/config/v2/ConflationParsingTest.kt
```diff
@@ -34,6 +35,10 @@ class ConflationParsingTest {
       coordinator-polling-interval = "PT2S"
       deadline-check-interval = "PT8S"
       target-end-blocks = [10, 20, 30_000]
+
+      [conflation.proof-aggregation.hard-forks]
```
fluentcrafter: inline with above suggestion, i think it should be `conflation.hard-forks`

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/conflation/GlobalBlockConflationCalculator.kt
```diff
@@ -75,7 +75,7 @@ class GlobalBlockConflationCalculator(
     )
     val triggers = calculators.mapNotNull {
       val overflowTrigger = it.checkOverflow(blockCounters)
-      log.trace("CHECK: calculator=${it.id}, blockNumber=${blockCounters.blockNumber}, trigger=$overflowTrigger")
+      log.trace("CHECK: calculator={}, blockNumber={}, trigger={}", it.id, blockCounters.blockNumber, overflowTrigger)
```
fluentcrafter, line 78: nice touch 👍

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/conflation/TimestampHardForkConflationCalculator.kt
```diff
@@ -0,0 +1,73 @@
...
+  override val id: String = "TIMESTAMP_HARD_FORK"
+
+  private var lastProcessedTimestamp: Instant = initialTimestamp
+
+  init {
+    require(hardForkTimestamps.isNotEmpty()) {
+      "Instantiation of TimestampHardForkConflationCalculator is pointless with empty timestamp list"
+    }
+    require(hardForkTimestamps.toSet().size == hardForkTimestamps.size) {
+      "Timestamps list contains duplicates! Probably a misconfiguration!"
```
fluentcrafter, line 29: I suggest to add this check to the HardForksToml class instantiation as well. Easeir to understand when it fails...

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/conflation/TimestampHardForkConflationCalculator.kt
```diff
@@ -0,0 +1,73 @@
...
+
+    // Check if this block crosses any hard fork timestamp boundary
+    val applicableTimestamp = hardForkTimestamps.find { timestamp ->
+      // If the last processed block was before the timestamp and this block is at or after the timestamp
+      lastProcessedTimestamp < timestamp && blockTimestamp >= timestamp
+    }
+
+    if (applicableTimestamp != null) {
+      log.info(
+        "Hard fork detected at block {} with timestamp {}. Fork timestamp: {}",
```
fluentcrafter:
```suggestion
        "Hard fork detected at block={} with timestamp={} forkTimestamp={}",
```

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/ConflationConfig.kt
Filter94: The list is near the `targetEndBlocks`. If we move the list of hard forks, `targetEndBlocks` should move too. Since first and foremost we want to split aggregations, to me it makes more sense to keep both in `ProofAggregation` config

Filter94: >Make it optional, default to null. Otherwise, after we reach the hardforks, we always have at least the TTD one set, which is noisy IMO and there is no value on keeping it after they get finalized on L1

TTD is not a thing. I don't see a downside in empty list for timestamps as a default

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflation/ConflationApp.kt
Filter94: It's behind an if now

coordinator/core/src/main/kotlin/net/consensys/zkevm/domain/Conflation.kt
Filter94: I didn't find its usages. If it's not used anywhere, what would support entail?

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/conflation/TimestampHardForkConflationCalculator.kt
Filter94: I added it to `ProofAggregation`

coordinator/core/src/main/kotlin/net/consensys/zkevm/domain/Conflation.kt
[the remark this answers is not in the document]
Filter94: I did it intentionally for the same reason our linter leaves commas, but I can remove it if you think it doesn't make sense

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/ConflationConfig.kt
fluentcrafter: > TTD is not a thing. 
What do you mean by that?

>The list is near the targetEndBlocks.
Yes, because the goal was to have `targetEndBlocks` for blobs as well for the cases when we had issues with blobs and avoid resending. But, I am fine as it is as well

Filter94: > What do you mean by that?

TTD calculator doesn't exist anymore

> Yes, because the goal was to have targetEndBlocks for blobs as well for the cases when we had issues with blobs and avoid resending. But, I am fine as it is as well

That's exactly the same case with timestamp hard forks. We want to cut off conflations, blobs and aggregations

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/conflation/TimestampHardForkConflationCalculator.kt
fluentcrafter: IMO still relevant in HardForksToml because it's better UX in case of failure...

Filter94: I can add it there as well, but I don't understand how it's better. Can you please explain?

coordinator/core/src/test/kotlin/net/consensys/zkevm/ethereum/coordination/conflation/ConflationCalculatorByTimeDeadlineTest.kt
fluentcrafter, line 1: Why remove this whole test class? Does it require a heavy refactor?

Filter94: It was by mistake

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/ConflationConfig.kt
fluentcrafter: > TTD calculator doesn't exist anymore

Why did do that? I think it was useful because block time is not 100% deterministic in production (e.g sequencer upgrade) so it would be more reliable than using timestamp fork for Paris, 
But I am ok

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/conflation/TimestampHardForkConflationCalculator.kt
fluentcrafter: It will fail earlier with a toml parsing error and the error will say precisely which property on the config file is failing. As it is now it will fial on `reify()` at best without pointing to the toml field.

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/ConflationConfig.kt
Filter94: > Why did do that? 

1. I faced an issue with aggregation calculator. In order to use the totalDifficulty in `BlobCounters`, we'd have to persist it in the DB. And I thought that doing the DB migration and add a field that will be useful only once per lifetime is very wasteful
2. I realized that since we already have `target-end-blocks` and TTD determines the block number, we don't really need the TTD in the first place

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/conflation/TimestampHardForkConflationCalculator.kt
Filter94: Ok, I moved it there

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/ConflationConfig.kt
Filter94: > so it would be more reliable than using timestamp fork for Paris

To be clear, we won't rely on timestamps for Paris. We'll rely on the block number computed from the TTD and passed as `target-end-blocks`

coordinator/core/src/test/kotlin/net/consensys/zkevm/ethereum/coordination/conflation/GlobalBlockConflationCalculatorIntTest.kt
```diff
@@ -214,6 +224,63 @@ class GlobalBlockConflationCalculatorIntTest {
...
+    )
+    assertThat(conflations[1]).isEqualTo(
+      ConflationCalculationResult(
+        startBlockNumber = 2uL,
+        endBlockNumber = 3uL,
+        conflationTrigger = ConflationTrigger.HARD_FORK,
+        tracesCounters = block3Counters.tracesCounters.add(block4Counters.tracesCounters),
+      ),
+    )
+  }
```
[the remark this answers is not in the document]
jonesho: Yes, it was a mistake, will fix it

config/common/traces-limits-v2.toml
```diff
@@ -84,7 +84,7 @@ BLAKE               = 0
 POINT_EVAL          = 0
 
 # Introduced from beta-v4.0-rc6-CANCUN and present on Cancun
-BLS_DATA                = 2147483647
-RLP_UTILS               = 2147483647 # Note: need to confirm if this limit could have the same limit as RLP_TXN
+BLS_DATA                = 8192
```
jonesho: @Filter94 why change these two from max int to 8192, same as `TXN_DATA`? `BLS_DATA` would be always 15 from Cancun

Filter94: I'll set it to `1073741824`. It needs to be a power of 2

jonesho (approved):

PR #1441, opened 2025-09-17
Feat/e2e switch from london to cancun

jonesho:
This PR implements issue(s) #

### Checklist


docker/config/l2-genesis-initialization/init.sh
```diff
@@ -1,28 +1,29 @@
...
 sed -i "s/%PRAGUE_TIME%/$prague_timestamp/g" genesis-besu.json
-sed -i "s/%PRAGUE_TIME%/$prague_timestamp/g" genesis-geth.json
 
 CREATE_EMPTY_BLOCKS="${CREATE_EMPTY_BLOCKS:-false}"
 sed -i "s/%CREATE_EMPTY_BLOCKS%/$CREATE_EMPTY_BLOCKS/g" genesis-besu.json
+
+shanghai_timestamp_ms=$((shanghai_timestamp * 1000))
+cancun_timestamp_ms=$((cancun_timestamp * 1000))
+prague_timestamp_ms=$((prague_timestamp * 1000))
+sed -i'' "s/^\(timestamp-based-hard-forks[ ]*=[ ]*\).*/\1[${shanghai_timestamp_ms}, ${cancun_timestamp_ms}]/" coordinator/coordinator-config-v2.toml
```
Filter94: We shouldn't modify the files checked into git. Please create a copy and reference it in the docker setup

docker/config/maru/follower.config.toml
Filter94, line 1: Why do we need Maru Follower?

docker/config/l2-genesis-initialization/genesis-besu.json.template
```diff
@@ -13,28 +13,60 @@
...
-    "1b9abeec3215d8ade8a33607f2cf0f4f60e5f0d0": {
+    "0x1b9abeec3215d8ade8a33607f2cf0f4f60e5f0d0": {
+      "balance": "90000000000000000000000",
       "@WARNING": "FOR LOCAL DEV ONLY - DO NOT REUSE THESE KEYS ELSEWHERE",
-      "privateKey": "0x1dd171cec7e2995408b5513004e8207fe88d6820aeff0d82463b3e41df251aae",
-      "comment": "Contract deployer account",
+      "@privateKey": "0x1dd171cec7e2995408b5513004e8207fe88d6820aeff0d82463b3e41df251aae",
+      "@comment": "Contract deployer account"
+    },
+    "0xa05b21E5186Ce93d2a226722b85D6e550Ac7D6E3": {
```
Filter94, line 43: Is there a reason to add this new address?

jonesho: No, it was just a copy and paste from Maru `docker/initialization/genesis-besu.json.template`, shall we remove it?

docker/config/maru/follower.config.toml
jonesho: Just to test the maru follower setup (with static peers) in our e2e test, if you think that's out-of-scope for the e2e test, I'm also keen to remove it, as one of the cons is that it would consume more resources (around 320MB mem) in our e2e test

docker/config/l2-genesis-initialization/genesis-besu.json.template
Filter94: I'd rather remove it

docker/config/maru/follower.config.toml
Filter94: That's out of scope, because peering is a feature of Maru that is maintained and tested in a separate repo. For These E2E tests we need the simplest setup possible

docker/config/l2-genesis-initialization/genesis-besu.json.template
jonesho: Tried to remove, but then e2e test failed because the genesis state root hash would be changed, and you'd also need to regenerate the integration test data, let me know if its worth the effort

Filter94: I guess not at this point

Filter94 (approved):

Issue #1488, opened 2025-09-25
Calldata based pricing needs to reference previous block extra data on chain and supports for eth_gasPrice


jonesho:
🔐 🔐 `Should this be a security advisory instead?` 🔐 🔐

## Description
In sepolia, we noticed that when L2 block creation halts and if the calldata size average of the last 5 blocks is higher than half of the max calldata size, the variable cost would keep increasing each time the coordinator update the variable cost to sequencer.

To fix this, we should use the extra data from the latest L2 block to extract the "previous_block_variable_cost" (mentioned in the equation from the calldata-based pricing [issue](https://github.com/Consensys/zkevm-monorepo/issues/4249)) instead of using the in-memory last calculated variable cost, this would help avoid variable cost increase during block stallment and also void calldata-based pricing reset after coordinator restart 

Also we need to support that, when calldata-based pricing is enabled, eth_gasPrice should also increase accordingly which is not the case atm, given that `legacyGasPricingCalculator` is using the `boundedVariableCostCalculator` (the one that is used when calldata-based pricing is disabled). We should change it to use `L2CalldataBasedVariableFeesCalculator` when calldata-based pricing is enabled.

## Motivation
- To avoid unnecessary variable cost increases when L2 block halts 
- To ensure the value returned by eth_gasPrice would align with calldata-based pricing 
- To avoid the state of calldata-based pricing being reset after coordinator restart  

## Tasks
- [ ] Medium-level task that needs to be implemented for this feature request to be considered complete.

## Acceptance criteria
- [ ] List tests that need to be implemented or checks to consider this task completed. 

## Risks
- [ ] These are risks that may prevent completion of the task as well as areas to pay special attention to

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #1494, opened 2025-09-28
feat: calldata-based pricing to extract last variable cost from lates…

jonesho:
…t block extra data and support calldata-based pricing in eth_gasPrice

This PR implements issue(s) #1488 

- `L2CalldataBasedVariableFeesCalculator` would extract the variableCost from the latest block extraData to use as previous variableCost instead of the last in-memory variableCost
- `legacyGasPricingCalculator` which is an instance of `TransactionCostCalculator` would use `L2CalldataBasedVariableFeesCalculator` as the fee calculator instead of just the plain `varaibleFeesCalculator` to allow eth_gasPrice to align with `L2CalldataBasedVariableFeesCalculator` gas pricing
- Added 2 prom metrics for block extra data: `linea_l2_pricing_ethgasprice` and `linea_l2_pricing_variablecost` for better observability

### Checklist

---

> [!NOTE]
> Reworks L2 gas pricing to derive variable fees from the latest block’s extraData with calldata-based adjustments, adds L2 pricing metrics, and refactors configs/APIs accordingly.
> 
> - **L2 Gas Pricing Service**:
>   - **Config refactor**: Replace separate `l2CalldataSizeAccumulatorConfig`/`l2CalldataBasedVariableFeesCalculatorConfig` with `L2CalldataPricingConfig`; enable only when `calldataSumSizeBlockCount > 0`.
>   - **Calculator composition**: Build `variableCostCalculator` that conditionally wraps `VariableFeesCalculator` with `L2CalldataBasedVariableFeesCalculator` using `L2CalldataSizeAccumulatorImpl` and new `HistoricVariableCostProviderImpl`.
>   - **Legacy calculator**: Apply bounds conditionally when calldata-based pricing is disabled.
>   - **Wiring**: Pass shared `variableCostCalculator` to `MinerExtraDataV1CalculatorImpl`; inject `metricsFacade` into `L2NetworkGasPricingService` and `ExtraDataV1PricerService`.
> - **Static-cap module**:
>   - **New** `HistoricVariableCostProviderImpl`: fetches variable cost (wei) from latest L2 block `extraData` with caching and error handling.
>   - **`L2CalldataBasedVariableFeesCalculator`**: now uses latest block number, `L2CalldataSizeAccumulator.getSumOfL2CalldataSize(blockNumber)`, and `HistoricVariableCostProvider`; removes internal state; enforces `calldataSizeBlockCount > 0`.
>   - **`L2CalldataSizeAccumulatorImpl`**: API now requires a `blockNumber`; adds caching of last `(blockNumber, sum)` and improved logging.
>   - **`ExtraDataV1PricerService`**: records gauges `linea_l2_pricing_variablecost` and `linea_l2_pricing_ethgasprice`; track last extra data via `AtomicReference`.
> - **Core/Metrics**:
>   - Add `LineaMetricsCategory.L2_PRICING`.
> - **Interfaces**:
>   - Update `L2CalldataSizeAccumulator.getSumOfL2CalldataSize(blockNumber)`; add `HistoricVariableCostProvider`.
> - **Tests/Build**:
>   - Add tests for new provider and updated calculators/accumulator; update existing tests to new signatures and metrics; add test fixtures dependency.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bcd97051e0ac118cccdb54aaf26725da27e5b0e1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/coordinator/ethereum/gas-pricing/static-cap/src/main/kotlin/net/consensys/linea/ethereum/gaspricing/staticcap/HistoricVariableCostProviderImpl.kt
+++ b/coordinator/ethereum/gas-pricing/static-cap/src/main/kotlin/net/consensys/linea/ethereum/gaspricing/staticcap/HistoricVariableCostProviderImpl.kt
@@ -27,7 +27,7 @@ class HistoricVariableCostProviderImpl(
             "Block {} not found, the requesting node could be out-of-sync",
             blockParameter.getNumber(),
           )
-          throw IllegalStateException("Block ${blockParameter.getNumber()} not found")
+          throw IllegalStateException("Block $blockParameter not found")
         } else {
           try {
             MinerExtraDataV1.decodeV1(block!!.extraData.encodeHex())


coordinator/ethereum/gas-pricing/static-cap/src/main/kotlin/net/consensys/linea/ethereum/gaspricing/staticcap/L2CalldataSizeAccumulatorImpl.kt
```diff
@@ -29,24 +33,34 @@ class L2CalldataSizeAccumulatorImpl(
...
-              .map { blockNumber ->
-                web3jClient.ethGetBlockSizeByNumber(blockNumber.toLong())
-              }
+        if (lastCalldataSizeSum.get().first == currentBlockNumber) {
+          log.debug(
+            "Use cached lastCalldataSizeSum={} currentBlockNumber={}",
+            lastCalldataSizeSum.get().second,
+            currentBlockNumber,
+          )
+          SafeFuture.completedFuture(lastCalldataSizeSum.get().second)
```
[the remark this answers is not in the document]
fluentcrafter: @jonesho this is a relevant comment.

coordinator/ethereum/gas-pricing/static-cap/src/main/kotlin/net/consensys/linea/ethereum/gaspricing/staticcap/L2CalldataSizeAccumulatorImpl.kt
```diff
@@ -29,24 +33,34 @@ class L2CalldataSizeAccumulatorImpl(
...
+        } else if (config.calldataSizeBlockCount > 0u && currentBlockNumber.toUInt() >= config.calldataSizeBlockCount) {
```
fluentcrafter: For cosystency and safty, let's use ULong for block numbers
```suggestion
        } else if (config.calldataSizeBlockCount > 0u && currentBlockNumber >= config.calldataSizeBlockCount.toULong()) {
```

coordinator/ethereum/gas-pricing/static-cap/src/main/kotlin/net/consensys/linea/ethereum/gaspricing/staticcap/L2CalldataSizeAccumulatorImpl.kt
```diff
@@ -29,24 +33,34 @@ class L2CalldataSizeAccumulatorImpl(
...
+          val blockRange =
+            (currentBlockNumber.toUInt() - config.calldataSizeBlockCount + 1U)..currentBlockNumber.toUInt()
```
fluentcrafter: same here, favour ULong

fluentcrafter (commented):
IIUC, both the HistoricVariableCostProvider.getLatestVariableCost and L2CalldataSizeAccumulatorImpl.getRecentL2CalldataSize should return data that is pegged ti the same block number.

However, with corrent design that may not happen, of it does is an obscour correlation.

I thick we should redesigh these classes.

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/L2NetworkGasPricingService.kt
```diff
@@ -62,28 +64,59 @@ class L2NetworkGasPricingService(
...
-    val variableCostCalculator = VariableFeesCalculator(
-      config.variableFeesCalculatorConfig,
-    )
-    BoundableFeeCalculator(
-      config = config.variableFeesCalculatorBounds,
-      feesCalculator = variableCostCalculator,
-    )
+  private fun isL2CalldataBasedVariableFeesEnabled(config: Config): Boolean {
+    return config.l2CalldataBasedVariableFeesCalculatorConfig != null &&
+      config.l2CalldataSizeAccumulatorConfig != null &&
```
Filter94: Why `l2CalldataSizeAccumulatorConfig` isn't under `l2CalldataBasedVariableFeesCalculatorConfig`?

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/L2NetworkGasPricingService.kt
```diff
@@ -62,28 +64,59 @@ class L2NetworkGasPricingService(
...
+      config.l2CalldataBasedVariableFeesCalculatorConfig.calldataSizeBlockCount > 0u
```
Filter94: If `l2CalldataBasedVariableFeesCalculatorConfig` is defined, why `calldataSizeBlockCount` can be 0 at all? Shouldn't it be a validation on the config level?

Filter94:
> L2CalldataBasedVariableFeesCalculator would extract the variableCost from the latest block extraData to use as previous variableCost instead of the last in-memory variableCost

Is this to curb the exponential growth when the block production is stuck?

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/L2NetworkGasPricingService.kt
```diff
@@ -62,28 +64,58 @@ class L2NetworkGasPricingService(
...
-      baseCalculator,
-    )
+    if (isL2CalldataBasedVariableFeesEnabled(config)) {
+      baseCalculator
+    } else {
+      BoundableFeeCalculator(
+        config.legacy.legacyGasPricingCalculatorBounds,
+        baseCalculator,
+      )
+    }
```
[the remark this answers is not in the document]
Filter94: I think this is relevant. Even with `L2CalldataBasedVariableFeesCalculator`, the bounds have to be applied

coordinator/ethereum/gas-pricing/static-cap/src/main/kotlin/net/consensys/linea/ethereum/gaspricing/staticcap/HistoricVariableCostProviderImpl.kt
```diff
@@ -0,0 +1,70 @@
...
+
+  private fun getHistoricVariableCostInWei(blockParameter: BlockParameter): SafeFuture<Double> {
+    return web3jClient.ethGetBlock(blockParameter)
+      .thenApply { block ->
+        try {
+          MinerExtraDataV1.decodeV1(block!!.extraData.encodeHex())
+            .variableCostInKWei.toDouble() * OneKWei
+        } catch (th: Throwable) {
+          if (block != null) {
+            log.debug(
```
Filter94, line 36: Should it be a WARN?

coordinator/ethereum/gas-pricing/static-cap/src/main/kotlin/net/consensys/linea/ethereum/gaspricing/staticcap/HistoricVariableCostProviderImpl.kt
```diff
@@ -0,0 +1,70 @@
...
+              "Will return historic variable cost as zero due to failure in decoding extra data: {}",
+              th.message,
+            )
+            0.0
```
Filter94, line 40: I think if we failed to parse the last block's extraData we should instead fail here too. Hopefully on the next tick we'll hit a block that has a parseable extraData. Otherwise we're risking to set a nonsensical gas price for the next block

coordinator/ethereum/gas-pricing/static-cap/src/test/kotlin/net/consensys/linea/ethereum/gaspricing/staticcap/L2CalldataBasedVariableFeesCalculatorTest.kt
```diff
@@ -172,52 +223,158 @@ class L2CalldataBasedVariableFeesCalculatorTest {
...
+      historicVariableCostProvider = fakeHistoricVariableCostProvider,
+    )
+
+    // call calculateFees first to instantiate the lastVariableCost
+    feesCalculator.calculateFees(feeHistory).let {
+      fakeHistoricVariableCostProvider.setLatestVariableCost(it)
+    }
+
+    assertThat(feesCalculator.calculateFees(feeHistory))
+      .isEqualTo(15000.0)
```
Filter94: Where does the value of 15000.0 comes from? It's unclear for me from the test that it's the original

jonesho: Made it more clear about `15000` is the original variable fee

coordinator/ethereum/gas-pricing/static-cap/src/main/kotlin/net/consensys/linea/ethereum/gaspricing/staticcap/HistoricVariableCostProviderImpl.kt
jonesho: I'd keep it as debug as it does happen during the first few blocks in our local stack (before we set the first miner extra data for variable cost)

coordinator/ethereum/gas-pricing/static-cap/src/main/kotlin/net/consensys/linea/ethereum/gaspricing/staticcap/HistoricVariableCostProviderImpl.kt
jonesho: When running the e2e in our local stack, extraData of the first few blocks were not parsable (they are the miner signature), so if we just fail on unparsable extra data, we would never be able to set block extra data with variable cost, given that for now, if calldata-based pricing is enabled, we would always parse the latest block extra data to get the previous variable cost. I think the risk is quite minimal, given that we're still having single sequencer now.

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/L2NetworkGasPricingService.kt
jonesho: Setting `calldataSizeBlockCount` to zero means disabling the calldata-based pricing (from the initial PR): in the coordinator config, when we disable calldata-based pricing, we want to keep the calldata-based configuration in config file by simply setting the `calldataSizeBlockCount` to zero

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/L2NetworkGasPricingService.kt
jonesho: In the Config of the `L2NetworkGasPricingService`, we separate `config.l2CalldataSizeAccumulatorConfig` from `config.l2CalldataBasedVariableFeesCalculatorConfig` for passing them directly to the constructor of `l2CalldataSizeAccumulator` and `L2CalldataBasedVariableFeesCalculator` respectively

coordinator/core/src/main/kotlin/net/consensys/linea/metrics/LineaMetricsCategory.kt
```diff
@@ -6,4 +6,5 @@ enum class LineaMetricsCategory : MetricsCategory {
   BLOB,
   CONFLATION,
   GAS_PRICE_CAP,
+  MINER_EXTRA_DATA,
```
fluentcrafter: I recommend calling this `L2_PRICING` which is what this is about. 

Minor extra data refers to the implementation details of how we propagate that information.

coordinator/ethereum/gas-pricing/src/main/kotlin/net/consensys/linea/ethereum/gaspricing/DynamicGasPrice.kt
```diff
@@ -12,7 +12,11 @@ interface FeesFetcher {
 }
 
 interface L2CalldataSizeAccumulator {
-  fun getSumOfL2CalldataSize(): SafeFuture<BigInteger>
+  fun getSumOfL2CalldataSize(latestBlockNumber: ULong): SafeFuture<BigInteger>
+}
+
+interface HistoricVariableCostProvider {
+  fun getVariableCost(blockNumber: ULong): SafeFuture<Double>
```
fluentcrafter, line 19: If the interfaces are supposed to be called with the same block number, they should have consistent naming: `blockNumber` in both is my recommendation.

coordinator/ethereum/gas-pricing/static-cap/src/test/kotlin/net/consensys/linea/ethereum/gaspricing/staticcap/HistoricVariableCostProviderImplTest.kt
```diff
@@ -0,0 +1,118 @@
...
+import org.junit.jupiter.api.Test
+import org.mockito.kotlin.doReturn
+import org.mockito.kotlin.eq
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.times
+import org.mockito.kotlin.verify
+import tech.pegasys.teku.infrastructure.async.SafeFuture
+
+class HistoricVariableCostProviderImplTest {
+  val mockEthBlock = mock<Block> {
```
fluentcrafter: In cases like this, we should favour the real data class instead. We have testing fixtures for that, `linea.domain.BlockFactory`

coordinator/ethereum/gas-pricing/static-cap/src/test/kotlin/net/consensys/linea/ethereum/gaspricing/staticcap/HistoricVariableCostProviderImplTest.kt
```diff
@@ -0,0 +1,118 @@
...
+      ).isEqualTo(expectedVariableCost)
+    }
+
+    // verified it only called ethGetBlock once
+    verify(mockWeb3jClient, times(1))
+      .ethGetBlock(eq(targetBlockNumber.toBlockParameter()))
+  }
+
+  @Test
+  fun test_getLatestVariableCost_throws_error_when_ethGetBlock_returns_null() {
```
fluentcrafter: What's the value of this test, asserting it throws an NPE? That's not Kotlin idiomatic. One of Kotlin's first and biggest value propositions is no NPE. If we have, then we did something wrong... 

After reading the code again, yes, I think we have a design flaw and this test brings it to the surface... Will leave the comment on the interface

coordinator/ethereum/gas-pricing/static-cap/src/main/kotlin/net/consensys/linea/ethereum/gaspricing/staticcap/HistoricVariableCostProviderImpl.kt
```diff
@@ -0,0 +1,70 @@
...
+import java.util.concurrent.atomic.AtomicReference
+
+class HistoricVariableCostProviderImpl(
+  private val web3jClient: ExtendedWeb3J,
+) : HistoricVariableCostProvider {
+  private val log: Logger = LogManager.getLogger(this::class.java)
+  private var lastVariableCost: AtomicReference<Pair<ULong, Double>> =
+    AtomicReference(0UL to 0.0)
+
+  private fun getHistoricVariableCostInWei(blockParameter: BlockParameter): SafeFuture<Double> {
```
fluentcrafter, line 22: After reading the code and the tests (see my comment [here](https://github.com/Consensys/linea-monorepo/pull/1494/files#r2391167559)), I think this has a design flaw:

If I read and understand correctly, we have 3 cases here:
- 1. block for the given blockNumber does not exist, `block!!.extraData.` will throw NPE
- 2. block exists
   - 2.1 - it does not have extra data prices
   - 2.2 - has extra data - happy path - is parsed and return

So it seems to me that when:
1. happens, it was an invalid state, because we call this function with the LATEST blockNumber, which means the node was out of synch. Promise should be rejected with `IlegalStateException("block $blockNumber nout found")`
2.1 happens, we are at the beggining, legit case - we should retrun no data, so the interface should have a nullable and the caller defaults when no value is pressent. We shall update the signature to `getHistoricVariableCostInWei(blockParameter: BlockParameter): SafeFuture<Double?>`
2.2 happy path :) 

@Filter94 WDYT?

coordinator/ethereum/gas-pricing/static-cap/src/main/kotlin/net/consensys/linea/ethereum/gaspricing/staticcap/L2CalldataBasedVariableFeesCalculator.kt
```diff
@@ -51,22 +54,22 @@ class L2CalldataBasedVariableFeesCalculator(
...
     val delta = (
       l2CalldataSizeAccumulator
-        .getSumOfL2CalldataSize().get().toDouble()
+        .getSumOfL2CalldataSize(latestBlockNumber.toULong()).get().toDouble()
         .minus(callDataTargetSize)
       )
       .div(callDataTargetSize)
       .coerceAtLeast(-1.0)
       .coerceAtMost(1.0)
+    val latestVariableCost = historicVariableCostProvider.getVariableCost(latestBlockNumber.toULong()).get()
```
fluentcrafter: please make the 2 calls: 
`l2CalldataSizeAccumulator.getSumOfL2CalldataSize(latestBlockNumber.toULong())` and `historicVariableCostProvider.getVariableCost(latestBlockNumber.toULong())` in parallel

coordinator/ethereum/gas-pricing/static-cap/src/main/kotlin/net/consensys/linea/ethereum/gaspricing/staticcap/L2CalldataBasedVariableFeesCalculator.kt
```diff
@@ -51,22 +54,22 @@ class L2CalldataBasedVariableFeesCalculator(
       .times(config.calldataSizeBlockCount)
       .toDouble().div(2.0)
 
+    val latestBlockNumber = web3jClient.ethBlockNumber().get()
```
fluentcrafter:
```suggestion
    val latestBlockNumber = web3jClient.ethBlockNumber().get().toULong()
```

this makes code simpler below

coordinator/ethereum/gas-pricing/static-cap/src/main/kotlin/net/consensys/linea/ethereum/gaspricing/staticcap/HistoricVariableCostProviderImpl.kt
Filter94: I don't see how 2.1 should be handled on the caller side. From Jones' explanation [here](https://github.com/Consensys/linea-monorepo/pull/1494#discussion_r2391014960), it should be just 0 instead

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/L2NetworkGasPricingService.kt
Filter94: I think we should either make `l2CalldataBasedVariableFeesCalculatorConfig` nullable and make `l2CalldataBasedVariableFeesCalculatorConfig == 0` an invalid value OR make `l2CalldataBasedVariableFeesCalculatorConfig` == 0 an indicator that `CalldataBasedVariableFeesCalculator` shouldn't be used. The first option is cleaner IMO

The current encoding of the information whether `CalldataBasedVariableFeesCalculator` should be used is redundant.

coordinator/ethereum/gas-pricing/static-cap/src/main/kotlin/net/consensys/linea/ethereum/gaspricing/staticcap/HistoricVariableCostProviderImpl.kt
fluentcrafter: On 2.1 am fine with returning 0

Filter94: So if we're returning 0 for 2.1, why don't we return it for 1?

fluentcrafter: Because, as I explained above, there are different cases: 1 can happen at getHistoricVariableCostInWei(block= 1_000_000) if that node is at block 999_999, there will definitely be extraData and returning 0 is not the truth. So, we should fail and retry on the next tick.

coordinator/ethereum/gas-pricing/static-cap/src/main/kotlin/net/consensys/linea/ethereum/gaspricing/staticcap/L2CalldataBasedVariableFeesCalculator.kt
jonesho: It was done on previous [commit](https://github.com/Consensys/linea-monorepo/pull/1494/commits#:~:text=feat%3A%20add%20target%20block%20number%20for%20getSumOfL2CalldataSize%20and%20getVariableCost), but was reverted due to a cursor comment regarding thread blocking, which after rethinking, running these two in parallel shouldn't be worse than running them one by one synchronously, I'd put them back

coordinator/ethereum/gas-pricing/static-cap/src/main/kotlin/net/consensys/linea/ethereum/gaspricing/staticcap/L2CalldataBasedVariableFeesCalculator.kt
```diff
@@ -34,39 +38,35 @@ class L2CalldataBasedVariableFeesCalculator(
...
+        l2CalldataSizeAccumulator.getSumOfL2CalldataSize(latestBlockNumber.toULong())
+          .thenCombine(
+            historicVariableCostProvider.getVariableCost(latestBlockNumber.toULong()),
+          ) { sumOfL2CalldataSize, latestVariableCost ->
+            sumOfL2CalldataSize to latestVariableCost
+          }
+      }.get()
+
+    val delta = sumOfL2CalldataSize.toDouble()
+      .minus(callDataTargetSize)
```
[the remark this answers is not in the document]
jonesho: init check callDataTargetSize cannot be zero

coordinator/ethereum/gas-pricing/static-cap/src/main/kotlin/net/consensys/linea/ethereum/gaspricing/staticcap/L2CalldataBasedVariableFeesCalculator.kt
fluentcrafter: I don't know what cursor said, I am confident we can - and should - do them in parallel in a safe manner to save time and make it more efficient (time wise).

jonesho: Yes, that's why I reverted the removal

coordinator/ethereum/gas-pricing/static-cap/src/main/kotlin/net/consensys/linea/ethereum/gaspricing/staticcap/HistoricVariableCostProviderImpl.kt
```diff
@@ -0,0 +1,74 @@
...
+    return web3jClient.ethGetBlock(blockParameter)
+      .thenApply { block ->
+        if (block == null) {
+          log.warn(
+            "Block {} not found, the requesting node could be out-of-sync",
+            blockParameter.getNumber(),
+          )
+          throw IllegalStateException("Block ${blockParameter.getNumber()} not found")
```
fluentcrafter: nit: although it's supposed to be called with a number, `blockParameter.getNumber()` may throw if it's a TAG.   So just delegate to the underlying to String and it will solve it for us :) 
```suggestion
          throw IllegalStateException("Block $blockParameter not found")
```

fluentcrafter (approved):

fluentcrafter (approved):

PR #1553, opened 2025-10-14
Implement configuration reload for LineaTransactionPoolValidatorPlugin

Matilda-Clerke:
This PR implements issue(s) #
https://github.com/Consensys/protocol-misc/issues/956

### Usage

1. Update the denylist file on disk
2. call https://besu.hyperledger.org/public-networks/reference/api#plugins_reloadpluginconfig with the argument `net.consensys.linea.sequencer.txpoolvalidation.LineaTransactionPoolValidatorPlugin` in order to trigger a reload of the LineaTransactionPoolValidatorPlugin's configuration. (see TransactionPoolDenyListReloadTest)

### Checklist

---

> [!NOTE]
> Implements runtime reload of transaction pool and bundle deny lists via plugins API, refactoring validators to use AtomicReference and adding acceptance/unit tests.
> 
> - **Tx Pool Validation**:
>   - `LineaTransactionPoolValidatorPlugin`: implement `reloadConfiguration()` to refresh denied addresses; keep a reference to `LineaTransactionPoolValidatorFactory`.
>   - `LineaTransactionPoolValidatorFactory`: switch denied list to `AtomicReference` and expose `setDeniedAddresses`.
>   - `AllowedAddressValidator`: now uses `AtomicReference<Set<Address>>` for dynamic updates.
> - **Bundle Endpoints**:
>   - `LineaBundleEndpointsPlugin`: maintain `bundleDeniedAddresses` via `AtomicReference`, initialize on start, and implement `reloadConfiguration()` to re-parse override deny list.
> - **CLI/Config**:
>   - `LineaTransactionPoolValidatorCliOptions`: make `parseDeniedAddresses` public for reuse during reload.
> - **Tests**:
>   - Acceptance: add `TransactionPoolDenyListReloadTest` to verify deny list reload behavior; enable `PLUGINS` RPC API in `LineaPluginTestBase`.
>   - Unit: update `AllowedAddressValidatorTest` to the new `AtomicReference` API.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e069ddd521d6eaafe829989097113af5e68a1700. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


besu-plugins/linea-sequencer/acceptance-tests/src/test/java/linea/plugin/acc/test/TransactionPoolDenyListReloadTest.java
```diff
@@ -0,0 +1,140 @@
...
+  private static final BigInteger GAS_PRICE = Convert.toWei("20", Convert.Unit.GWEI).toBigInteger();
+  private static final BigInteger GAS_LIMIT = BigInteger.valueOf(210000);
+  private static final BigInteger VALUE = BigInteger.ONE; // 1 wei
+
+  @TempDir static Path tempDir;
+  static Path tempDenyList;
+
+  @Override
+  public List<String> getTestCliOptions() {
+    tempDenyList = tempDir.resolve("denyList.txt");
```
macfarla, line 48: think if the tempDir is only used here you could simplify and just use `@TempDir` for the tempDenyList itself?

Matilda-Clerke: Tbh, I haven't used `@TempDir` before. I just copied this test class from your linea-sequencer PR. I'll have a look and see if there's a neater way to do it.

Matilda-Clerke: Looks like this is the generally accepted way to create a temporary file with JUnit 5

besu-plugins/linea-sequencer/acceptance-tests/src/test/java/linea/plugin/acc/test/TransactionPoolDenyListReloadTest.java
```diff
@@ -0,0 +1,140 @@
...
+    Credentials allowedCredentials = Credentials.create(Accounts.GENESIS_ACCOUNT_TWO_PRIVATE_KEY);
+    final Web3j miner = minerNode.nodeRequests().eth();
+
+    RawTransactionManager transactionManager =
+        new RawTransactionManager(miner, allowedCredentials, CHAIN_ID);
+    assertAddressAllowed(transactionManager, allowedCredentials.getAddress());
+  }
+
+  @Test
+  public void emptyDenyList_thenDenySender_cannotAddTxToPool() throws Exception {
```
fab-10: Nice test! 
Please add also the opposite, start denied and the removed from the list

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/rpc/services/LineaBundleEndpointsPlugin.java
```diff
@@ -27,6 +34,11 @@ public class LineaBundleEndpointsPlugin extends AbstractLineaRequiredPlugin {
   private LineaSendBundle lineaSendBundleMethod;
   private LineaCancelBundle lineaCancelBundleMethod;
 
+  private Optional<LineaTransactionPoolValidatorConfiguration>
+      lineaTransactionPoolValidatorConfiguration = Optional.empty();
```
fab-10: Why optional? that configuration must be always present at this point, and if not present it is an internal error and better to fail fast, rather than silently skip the reload

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txpoolvalidation/LineaTransactionPoolValidatorPlugin.java
```diff
@@ -39,6 +44,10 @@ public class LineaTransactionPoolValidatorPlugin extends AbstractLineaRequiredPl
   private TransactionPoolValidatorService transactionPoolValidatorService;
   private TransactionSimulationService transactionSimulationService;
   private Optional<JsonRpcManager> rejectedTxJsonRpcManager = Optional.empty();
+  private Optional<LineaTransactionPoolValidatorFactory> lineaTransactionPoolValidatorFactory =
+      Optional.empty();
+  private Optional<LineaTransactionPoolValidatorConfiguration>
+      lineaTransactionPoolValidatorConfiguration = Optional.empty();
```
fab-10: s.a.

besu-plugins/linea-sequencer/acceptance-tests/src/test/java/linea/plugin/acc/test/TransactionPoolDenyListReloadTest.java
Matilda-Clerke: Good idea, will do.

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/rpc/services/LineaBundleEndpointsPlugin.java
Matilda-Clerke: I used an optional because it follows an existing pattern for fields to be set in `doStart` or `doRegister`. Looks like it can just be replaced by the actual object and initialized during construction, so that's probably best,

Matilda-Clerke: Actually, it looks like the BesuContext isn't in place yet so we can't construct the Configuration class until doStart is called. I'll make it throw an exception if not present when used rather than silently skip the reload

Matilda-Clerke:
RE: implementing reload in `AbstractLineaSharedPrivateOptionsPlugin` 
I think that would be a pretty significant break from the existing convention of each plugin class managing its own config reload. I'm not totally opposed to a higher level config reload system which automatically updates the config the plugin class uses, but I feel like that should be a separate PR.

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/rpc/services/LineaBundleEndpointsPlugin.java
```diff
@@ -76,6 +87,19 @@ public void doStart() {
     // set the pool
     lineaSendBundleMethod.init(bundlePoolService, createTransactionValidator());
     lineaCancelBundleMethod.init(bundlePoolService);
+
+    lineaTransactionPoolValidatorConfiguration =
+        Optional.of(transactionPoolValidatorConfiguration());
+    bundleDeniedAddresses.set(
+        lineaTransactionPoolValidatorConfiguration.get().bundleDeniedAddresses());
+  }
```
fab-10: As you can see from the previous code in the `doRegister` that is called before `doStart`
```
new AllowedAddressValidator(
              transactionPoolValidatorConfiguration().bundleDeniedAddresses()),
```
here it is guaranteed that transactionPoolValidatorConfiguration() always returns not null, so you can get rid of the optionality, that because if the configuration was not available then we would have already failed.

In a future PR we can add a not null annotation to the return value of the method to make this clear.

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txpoolvalidation/LineaTransactionPoolValidatorPlugin.java
```diff
@@ -82,19 +90,23 @@ public void doStart() {
                               besuConfiguration.getDataPath(),
                               lineaRejectedTxReportingConfiguration)
                           .start());
-
+      lineaTransactionPoolValidatorConfiguration =
+          Optional.of(transactionPoolValidatorConfiguration());
```
fab-10: same here, `transactionPoolValidatorConfiguration()` must not return null at this point

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txpoolvalidation/LineaTransactionPoolValidatorPlugin.java
```diff
@@ -141,6 +153,17 @@ public void doStart() {
...
+  @Override
+  public CompletableFuture<Void> reloadConfiguration() {
+    var conf = this.lineaTransactionPoolValidatorConfiguration.orElseThrow(() -> new RuntimeException("LineaTransactionPoolValidatorConfiguration is not available, but reloadConfiguration called"));
+    if (lineaTransactionPoolValidatorFactory.isPresent()) {
+      Set<Address> newDeniedAddresses =
+          LineaTransactionPoolValidatorCliOptions.create().parseDeniedAddresses(conf.denyListPath());
+      lineaTransactionPoolValidatorFactory.get().setDeniedAddresses(newDeniedAddresses);
+    }
+    return CompletableFuture.completedFuture(null);
+  }
```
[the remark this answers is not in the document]
fab-10: do you think this applies?

fab-10 (commented):
There is one comment for Cursor that seems relevant, and check my comment about the config not allowed to be null

fab-10:
> [CLA assistant check](https://cla-assistant.io/Consensys/linea-monorepo?pullRequest=1553) Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/Consensys/linea-monorepo?pullRequest=1553) before we can accept your contribution.You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/Consensys/linea-monorepo?pullRequest=1553) it.

@Matilda-Clerke have you already signed the CLA?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txpoolvalidation/LineaTransactionPoolValidatorPlugin.java
```diff
@@ -141,6 +153,16 @@ public void doStart() {
...
   }
 
+  @Override
+  public CompletableFuture<Void> reloadConfiguration() {
+    var conf = this.lineaTransactionPoolValidatorConfiguration.orElseThrow(() -> new RuntimeException("LineaTransactionPoolValidatorConfiguration is not available, but reloadConfiguration called"));
+    var lineaTransactionPoolValidatorFactory = this.lineaTransactionPoolValidatorFactory.orElseThrow(() -> new RuntimeException("LineaTransactionPoolValidatorFactory is not available, but reloadConfiguration called"));
+      Set<Address> newDeniedAddresses =
+          LineaTransactionPoolValidatorCliOptions.create().parseDeniedAddresses(conf.denyListPath());
+    lineaTransactionPoolValidatorFactory.setDeniedAddresses(newDeniedAddresses);
+    return CompletableFuture.completedFuture(null);
```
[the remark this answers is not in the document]
Matilda-Clerke: I'm going to resolve this as it will be fixed in a future PR when we move configuration lifecycle to `AbstractLineaSharedPrivateOptionsPlugin`

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/rpc/services/LineaBundleEndpointsPlugin.java
Matilda-Clerke: Oh, I see how it works now. Will update

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txpoolvalidation/LineaTransactionPoolValidatorPlugin.java
```diff
@@ -141,6 +148,20 @@ public void doStart() {
     }
   }
 
+  @Override
+  public CompletableFuture<Void> reloadConfiguration() {
+    var lineaTransactionPoolValidatorFactory =
+        this.lineaTransactionPoolValidatorFactory.orElseThrow(
+            () ->
+                new RuntimeException(
```
fab-10: I think Cursor bot is suggestion that instead of throwing this should return `CompletableFuture::completeExceptionally`

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txpoolvalidation/LineaTransactionPoolValidatorPlugin.java
```diff
@@ -141,6 +148,20 @@ public void doStart() {
...
+                    "LineaTransactionPoolValidatorFactory is not available, but reloadConfiguration called"));
+    Set<Address> newDeniedAddresses =
+        LineaTransactionPoolValidatorCliOptions.create()
+            .parseDeniedAddresses(transactionPoolValidatorConfiguration().denyListPath());
```
fab-10: Also here the exception should be catch and returned via the completable future

fab-10 (approved):

PR #1573, opened 2025-10-16
perf: rework quotient computation

gbotrel:
This PR improve end to end time for some proofs by 30%, and reduce memory usage by a similar metric.

---

> [!NOTE]
> Major performance refactor: chunked arena-based symbolic evaluation, reworked quotient computation, parallel/hasher-reuse SMT build, arena-backed RS encoding, and broad vector/math optimizations; plus API shifts to BoardListVariableMetadata and minor runtime fixes.
> 
> - **Performance/Core**:
>   - **Quotient**: New refine-and-evaluate flow with constant propagation, precomputed annulator inverses, per-level coset re-eval, arena-backed buffers, and reduced GC; replaces prior board/roots plumbing.
>   - **Symbolic Evaluation**: Replaced old evaluator with chunked, arena-based engine; dynamic chunk sizing via memory limit; removes `evaluation_board.go`; adds `CountNodesFilterConstants`.
>   - **Smart Vectors/Field**:
>     - Vector ops now use `field.Vector`; new `ExpVec`/`ExpInt64`; faster lincomb/product paths; batch interpolate streamlined; FFT/window/rotated helpers (e.g., `Interval()`, `WriteSubVectorInSlice`).
>   - **SMT**: Reuse hasher (`hashLR`), parallel bottom-up build with worker pool; added benchmark; update proof/update paths to pass hasher.
>   - **Vortex/RS**: `rsEncode` supports arena to avoid allocs; encoding uses per-thread arenas; column hashing unchanged but parallelized.
> - **API/Usage Changes**:
>   - Expressions: introduce `expr.BoardListVariableMetadata()`; migrate callers across protocol (globalcs, distributed, wizard, queries, etc.).
>   - ZkEVM: `RunStatRecords` now `(cfg, witness)`; bootstrapper rescaling used in stats path.
>   - Arithmetization assignment: per-module column assignment parallelized; ensure left-padding where required.
>   - Logs PI: assign `FilterArith` as left-padded (via `TryReduceSizeLeft`) to curb over-segmentation.
> - **Misc**:
>   - Remove CSV/evaluation board utilities; minor tiling/parallelism in stitching interleave; add SMT benchmark; small cleanups/TODOs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit feeb6664b960f4d7c9b4bae11388bbb02fdb9227. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/config/coordinator/coordinator-config-v2.toml
+++ b/config/coordinator/coordinator-config-v2.toml
@@ -78,7 +78,7 @@ fs-responses-directory = "/data/prover/v3/aggregation/responses"
 #fs-responses-directory = "/data/prover/v3/aggregation/responses"
 
 [traces]
-expected-traces-api-version = "beta-v4.0-rc17"
+expected-traces-api-version = "beta-v4.0-rc18"
 [traces.counters]
 endpoints = ["http://traces-node:8545/"]
 request-limit-per-endpoint = 1
--- a/docker/compose-spec-l2-services.yml
+++ b/docker/compose-spec-l2-services.yml
@@ -39,7 +39,7 @@ services:
   sequencer:
     hostname: sequencer
     container_name: sequencer
-    image: consensys/linea-besu-package:${BESU_PACKAGE_TAG:-beta-v4.0-rc17-20251024131506-d32162b}
+    image: consensys/linea-besu-package:${BESU_PACKAGE_TAG:-beta-v4.0-rc18-20251030110723-de8c2f3}
     profiles: [ "l2", "l2-bc", "debug", "external-to-monorepo" ]
     depends_on:
       l2-genesis-initialization:
@@ -114,7 +114,7 @@ services:
   l2-node-besu:
     hostname: l2-node-besu
     container_name: l2-node-besu
-    image: consensys/linea-besu-package:${BESU_PACKAGE_TAG:-beta-v4.0-rc17-20251024131506-d32162b}
+    image: consensys/linea-besu-package:${BESU_PACKAGE_TAG:-beta-v4.0-rc18-20251030110723-de8c2f3}
     profiles: [ "l2", "l2-bc", "debug", "external-to-monorepo" ]
     depends_on:
       sequencer:
@@ -155,7 +155,7 @@ services:
   l2-node-besu-follower:
     hostname: l2-node-besu-follower
     container_name: l2-node-besu-follower
-    image: consensys/linea-besu-package:${BESU_PACKAGE_TAG:-beta-v4.0-rc17-20251024131506-d32162b}
+    image: consensys/linea-besu-package:${BESU_PACKAGE_TAG:-beta-v4.0-rc18-20251030110723-de8c2f3}
     profiles: [ "l2", "l2-bc", "debug", "external-to-monorepo" ]
     depends_on:
       sequencer:
@@ -196,7 +196,7 @@ services:
   traces-node:
     hostname: traces-node
     container_name: traces-node
-    image: consensys/linea-besu-package:${BESU_PACKAGE_TAG:-beta-v4.0-rc17-20251024131506-d32162b}
+    image: consensys/linea-besu-package:${BESU_PACKAGE_TAG:-beta-v4.0-rc18-20251030110723-de8c2f3}
     profiles: [ "l2", "l2-bc", "debug", "external-to-monorepo" ]
     depends_on:
       sequencer:
@@ -374,7 +374,7 @@ services:
       - l1network
 
   zkbesu-shomei:
-    image: consensys/linea-besu-package:${BESU_PACKAGE_TAG:-beta-v4.0-rc17-20251024131506-d32162b}
+    image: consensys/linea-besu-package:${BESU_PACKAGE_TAG:-beta-v4.0-rc18-20251030110723-de8c2f3}
     hostname: zkbesu-shomei
     container_name: zkbesu-shomei
     profiles: [ "l2", "l2-bc", "external-to-monorepo" ]
@@ -591,7 +591,7 @@ services:
         ipv4_address: 10.10.10.205
 
   zkbesu-shomei-sr:
-    image: consensys/linea-besu-package:${BESU_PACKAGE_TAG:-beta-v4.0-rc17-20251024131506-d32162b}
+    image: consensys/linea-besu-package:${BESU_PACKAGE_TAG:-beta-v4.0-rc18-20251030110723-de8c2f3}
     hostname: zkbesu-shomei-sr
     container_name: zkbesu-shomei-sr
     profiles: [ "staterecovery", "external-to-monorepo" ]


prover/maths/field/bls12_377.go
```diff
@@ -157,3 +157,67 @@ func FromBool(b bool) Element {
...
+		// ensure that vector and a are not the same slice; else we need to copy a into base
+		v0 := &vector[0] // #nosec G602 we check that N > 0 above
+		a0 := &a[0]      // #nosec G602 we check that N > 0 above
+		if v0 == a0 {
+			base = make(Vector, N)
+			copy(base, a)
+		}
+	}
+
+	copy(vector, base)
```
ivokub, line 190: I think this comes from bavard, but imo we could have a version when `a==vector` where we don't do two copies, but instead directly copy `copy(vector, a)`.

prover/maths/field/bls12_377.go
```diff
@@ -157,3 +157,67 @@ func FromBool(b bool) Element {
...
+		}
+		return
+	}
+	base := a
+	exp := k
+	if k < 0 {
+		// call batch inverse
+		base = BatchInvert(a)
+		exp = -k // if k == math.MinInt64, -k overflows, but uint64(-k) is correct
+	} else if N > 0 {
```
ivokub, line 180: These two conditions seem to be independent? Shouldn't we have:
```go
if k < 0 {
    base = BatchInvert(a)
    exp = -k
}
if N > 0 {
    ....
}
```

prover/maths/field/bls12_377.go
```diff
@@ -157,3 +157,67 @@ func FromBool(b bool) Element {
...
+		// ensure that vector and a are not the same slice; else we need to copy a into base
+		v0 := &vector[0] // #nosec G602 we check that N > 0 above
+		a0 := &a[0]      // #nosec G602 we check that N > 0 above
+		if v0 == a0 {
+			base = make(Vector, N)
```
ivokub, line 185: And dunno how much this is called, maybe it would be worth it to reuse it?

prover/protocol/compiler/stitchsplit/stitcher.go
```diff
@@ -106,13 +107,28 @@ func (a *StitchColumnsProverAction) Run(run *wizard.ProverRuntime) {
...
-				(*assignement)[i+j*maxSizeGroup] = witnesses[i].Get(j)
+
+		const tileSize = 128
+		rows := witnesses[0].Len()
+		cols := len(subColumns)
+
+		parallel.Execute(rows, func(start, end int) {
+			for tileStart := start; tileStart < end; tileStart += tileSize {
+				tileEnd := tileStart + tileSize
+				if tileEnd > end {
```
ivokub:
```go
tileEnd = min(tileStart+tileSize, end)
```

prover/utils/arena/vector_arena.go
```diff
@@ -50,9 +50,21 @@ func (a *VectorArena) Offset() int64 {
 	return atomic.LoadInt64(&a.offset)
 }
 
+// Remaining returns the number of elements of type T that can still be allocated
+// from the arena before it is exhausted.
+func Remaining[T any](a *VectorArena) int {
```
ivokub, line 55: I guess you already tried, but generically there is also Go experiment for arenas https://go.dev/src/arena/arena.go

I think implementation-wise it is similar, but I understand GC knows arenas better so for example if you would do experiments Arena.Free then memory can be reused without running GC.

We could have two implementations as is and hide the experiment arena behind the `goexperiment.arena` build tag.

prover/utils/collection/mapping.go
```diff
@@ -19,6 +19,12 @@ func NewMapping[K comparable, V any]() Mapping[K, V] {
 	}
 }
 
+func NewMappingWithCapacity[K comparable, V any](capacity int) Mapping[K, V] {
```
ivokub, line 22: Dunno if it is worth it - we work a lot with maps with string keys, but maybe we could have a separate implementation where we map the string keys to ints with fnv for example.

I think this could be in general done when we change `Mapping` to interface instead of struct type and then have separate implementation which we initialize in `NewMapping` if the first type parameter is `string` (or string-like `~string`).

prover/utils/parallel/pool/pool.go
```diff
@@ -18,7 +18,9 @@ func ExecutePoolChunky(nbIterations int, work func(k int)) {
 	for i := 0; i < nbIterations; i++ {
```
ivokub, line 18: I understand this method allows to perform parallel work fixed to OS thread. Keep in mind that with new Go 1.25 `runtime.GOMAXPROCS` is dynamic. I think it never happens in our context, but we should keep in mind as we have the `available` chan with buffer depending on it.

prover/utils/parallel/parallel.go
```diff
@@ -48,6 +48,8 @@ func Execute(nbIterations int, work func(int, int), maxCpus ...int) {
 		}
 
 		go func() {
+			runtime.LockOSThread()
```
ivokub: Here and elsewhere - why lock work to particular thread? Does it allow to use some caches more efficiently? In that case, I think a small comment would be useful.

prover/zkevm/arithmetization/assignment.go
```diff
@@ -65,25 +66,28 @@ func AssignFromLtTraces(run *wizard.ProverRuntime, schema *air.Schema, expTraces
...
+			}
+
+			plain := make([]field.Element, data.Len())
+			for i := range plain {
+				plain[i] = data.Get(uint(i))
+			}
 
-		for i := range plain {
-			plain[i] = data.Get(uint(i))
+			run.AssignColumn(ifaces.ColID(name), smartvectors.LeftPadded(plain, padding, wCol.Size()))
```
ivokub: Looking at run.AssignColumn implementation it uses lock in any case. Maybe we could also remove it there.

prover/maths/common/smartvectors/rotated.go
```diff
@@ -242,3 +242,50 @@ func SoftRotate(v SmartVector, offset int) SmartVector {
...
+		for i := 0; i < stop-start; i++ {
+			s[i] = r.v.Regular[start+r.offset+i]
+		}
+		return
+	}
+	size := r.Len()
+	spanSize := stop - start
+
+	// compact boundary checks
+	if stop <= start || start < 0 || stop > size {
```
ivokub: We already do the checks in the preambule

prover/maths/common/smartvectors/rotated.go
```diff
@@ -242,3 +242,50 @@ func SoftRotate(v SmartVector, offset int) SmartVector {
 	panic("unreachable")
 
 }
+
+// WriteSubVectorInSlice writes the subvector [start:stop) into the provided
+// slice s. The slice s must be of length stop-start.
+// This is equivalent to doing r.SubVector(start, stop).WriteInSlice(s) but
+// avoids an allocation.
+func (r *Rotated) WriteSubVectorInSlice(start, stop int, s []field.Element) {
```
ivokub, line 213: With this method, I would refactor also the `SubVector` method to be:
```go
func (r *Rotated) SubVector(start, stop int) SmartVector {
    if stop+r.offset < len(r.v.Regular) && start+r.offset > 0 {
		res := Regular(r.v.Regular[start+r.offset : stop+r.offset])
		return &res
	}
    res := make([]field.Element, stop-start)
    r.WriteSubVectorInSlice(start, stop, res)
    return Regular(res)
}
```

prover/maths/common/smartvectors/rotated.go
```diff
@@ -242,3 +242,50 @@ func SoftRotate(v SmartVector, offset int) SmartVector {
...
+	// sanity checks on start / stop / len(s)
+	if start < 0 || stop > r.Len() || stop <= start || len(s) != stop-start {
+		utils.Panic("invalid start/stop/len(s): start=%v, stop=%v, len(s)=%v, vector len=%v", start, stop, len(s), r.Len())
+	}
+	if stop+r.offset < len(r.v.Regular) && start+r.offset > 0 {
+		for i := 0; i < stop-start; i++ {
+			s[i] = r.v.Regular[start+r.offset+i]
+		}
+		return
+	}
```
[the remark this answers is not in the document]
ivokub: Indeed I think this comment makes sense.

ivokub (commented):
Looks good to me, added a few inline comments.

Maybe we could also look into the Go experimental memory arenas? I understand it is more compatible with GC so that GC never touches arena and when its freed then frees it directly without GC. It requres running Go experiment but we can conditionally build it with `goexperiment.arenas` tag. If it is worth it I could try to have a look?

Dunno if it affects performance a lot, but we have a lot of maps with strings as keys. I did a short benchmark and it seems when we use int64 keys then map accesses could be faster (20ns vs 10ns). But I don't have  intuition how many accesses we do. If its 100_000 accesses with 10ns improvement, then we save million ns which is worthless :)

And I guess we could also extend on the Vector API - right now we have vector-vector ops, but we also use quite a lot of vector-element ops which are manually unrolled.

There is also a comment about if-elseif statement in bls12_377.go - for me it seems a bug, but perhaps its not? Can you confirm and I'll approve.

prover/maths/common/smartvectors/rotated.go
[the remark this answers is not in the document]
gbotrel: yes just copy pasted the existing code -- didn't fix but will do :)

prover/maths/field/bls12_377.go
gbotrel: no because `BatchInvert` already allocates a new slice, so no-need to recopy it in that case. But worth a comment

prover/maths/field/bls12_377.go
gbotrel: don't think so because in the loop below we overwrite vector, and we need to keep base intact

prover/utils/arena/vector_arena.go
gbotrel: "> The proposal to add arenas to the standard library is on indefinite hold due to concerns about API pollution." 

I did this `vector_arena` thing because of that, the experiment flag could dissapear unexpectedly

prover/utils/collection/mapping.go
gbotrel: yesss so many maps :( . So many strings :(  . 
Thought about that too, but sometime we do iterate on the keys of the map too, so ... but will try for one or two places I have in mind, don't suspect it's noticeable on the over all trace though.

prover/zkevm/arithmetization/assignment.go
gbotrel: ?? remove what ?

ivokub: Currently this part is run within `parallel.Execute`, but when we call `run.AssignColumn` then it will be synchronized as it locks a `run.lock`.

My idea was either to remove that lock there (but it may have some consequences), or perhaps move `run.AssignmentColumn` here into a goroutine so that we could run next loop iteration here.

prover/utils/arena/vector_arena.go
ivokub: Agree. I see from the discussions that they want to implement less-polluting regions package. In that case, indeed lets avoid implementing using Go experimental arenas.

prover/utils/parallel/parallel.go
gbotrel: it actually doesn't help much on the perf side, had some cleaner traces (as in go tool trace) and thought at some point it helped, but will remove, it can (with lots of go routine) introduce some subtelties.

prover/zkevm/arithmetization/assignment.go
gbotrel: ha I see -- yes you are right, but I think that would be part of a larger refactor aiming to reduce all the maps.
Here the goal of parallelizing was mostly for the operation above, i.e. accessing the data and building the plain vector.

AlexandreBelling (approved):

Issue #1577, opened 2025-10-17
EthApi - Generic Interface


fluentcrafter:
## Description
Extend the current EthApiClient interface with the necessary eth_API methods and Kotlin native methods/classes, decoupled from Web3J.

## Motivation
Make Coordinator and Maru code and business logic simpler and agnostic to Web3J (or any other framework). This will avoid big refactors when Web3J brings breaking changes.

## Tasks
- [ ] Design remaing eth_API methods for the interface
- [ ] Implement an adapter that can delegate to Web3J
- [ ] Migrate coordinator code to use the new interface instead of Web3J

PR #1584, opened 2025-10-17
EthApi Interface - new eth methods

fluentcrafter:
This PR implements issue(s) #1577

### Checklist

---

> [!NOTE]
> Refactors EthApi into modular interfaces with many new eth_* methods, implements them in Web3j clients (with retries), adds domain models/mappers, and updates coordinator and log searcher to new APIs.
> 
> - **API Surface (major refactor)**:
>   - Split `EthApiClient` into modular interfaces: `EthApiChainIdClient`, `EthApiFeeClient`, `EthApiSimulationClient`, `EthApiBlockClient`, `EthLogsClient`, `EthApiAccountClient`, `EthApiTransactionClient`, `EthApiExecutionClientInfo`.
>   - Add methods: `ethChainId`, `ethBlockNumber`, block getters (`ethFind/GetBlockByNumber{FullTxs,TxHashes}`), account (`ethGetBalance`, `ethGetTransactionCount`), fees (`ethGasPrice`, `ethMaxPriorityFeePerGas`, `ethFeeHistory`), txs (`ethGetTransactionByHash`, `ethGetTransactionReceipt`, `ethSendRawTransaction`), node info (`ethProtocolVersion`, `ethCoinbase`, `ethMining`), simulation (`ethCall`, `ethEstimateGas`).
> - **Domain Models & Mappers**:
>   - New `TransactionForEthCall` and `TransactionReceipt` models; add mappers (`TransactionForEthCallMapper`, `TransactionReceiptMapper`).
> - **Client Implementations**:
>   - `Web3jEthApiClient` implements new methods; `Web3jEthApiClientWithRetries` wraps them with tailored retry predicates (incl. execution-reverted stop).
>   - `EthLogsClient` adds `FilterOptions` + `ethGetLogs` helper.
>   - `FakeEthApiClient` updated to new interface/methods.
> - **Call Sites Updated**:
>   - Replace `getChainId()` with `ethChainId()` and `getBlockByNumberWithoutTransactionsData()` with `ethGetBlockByNumberTxHashes()` in coordinator components and log searcher.
> - **Tests**:
>   - Update tests to use new method names (`ethGetBlockByNumber{FullTxs,TxHashes}`) and behaviors.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit edcdd265e46760ccf7ee14fdc3e04d1bb42cc73d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClientWithRetries.kt
+++ b/jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClientWithRetries.kt
@@ -153,9 +153,22 @@ class Web3jEthApiClientWithRetries(
     blockParameter: BlockParameter,
     stateOverride: StateOverride?,
   ): SafeFuture<ByteArray> {
-    // FIXME: stop retry when revert reason is returned as JSON RPC error
-    return retry(stopRetriesPredicateForTag(blockParameter)) {
+    val predicate = stopRetriesPredicateForTag(blockParameter).or(stopOnExecutionRevertedPredicate)
+    return retry(predicate) {
       ethApiClient.ethCall(transaction, blockParameter, stateOverride)
+    }
+  }
+
+  val stopOnExecutionRevertedPredicate: Predicate<Throwable> = Predicate { th ->
+
+    if (th.cause is linea.error.JsonRpcErrorResponseException) {
+      val rpcError = th.cause as linea.error.JsonRpcErrorResponseException
+      // -32000 = "execution reverted", no point in retrying
+      rpcError.rpcErrorCode == -32000 ||
+        rpcError.rpcErrorMessage.contains("reverted", ignoreCase = true) ||
+        rpcError.rpcErrorData.toString().contains("reverted", ignoreCase = true)
+    } else {
+      false
     }
   }
 


jvm-libs/linea/clients/interfaces/src/main/kotlin/linea/ethapi/EthApiClient.kt
```diff
@@ -3,37 +3,119 @@ package linea.ethapi
...
   }
+  // ==============================================================================
+  // ==============================================================================
+
+  // Ethereum JSON-RPC API methods with eth prefix
+  // Protocol version and chain information
+  fun ethChainId(): SafeFuture<ULong>
+  fun ethProtocolVersion(): SafeFuture<Int>
+
+  fun ethCoinbase(): SafeFuture<ByteArray>
```
Filter94: 1. I think that if we implement the entirety of the Eth API, we unnecessarily take the burden of maintaining it in the future
2. Having many methods implemented in the same interface forces mocking / a whole fake Eth client implementation. I think if we use functional interfaces / lambdas for the interaction with the chain we'll get a better testability + we still have an option of using a comprehensive Eth Client, serving as a fake chain. I think a fake Eth client is only useful in scenarios that require cohesive state updates on the fake Eth client side. It would just implement all the individual functional interfaces

fluentcrafter: 1. I agree, hence I didn't create all the methods. The idea is to implemente the ones we need (or very likely to need) 
2. I think it's bit overkill. Mocking or faking all the methods it's not entirely true was in pratice only need to mock/fake those that are used in practice. Also, the idea of fakes is to "write ince use many". Also, the approach of using segregated methods/lambas has 2 small disavantages IMO: 1) unnecessary long constructors when a class uses more than one, 2) when there are refactor and class changes the used methods make the refactor even bigger.

Also, this interface does not prevent/stops the develper from using those lambads in it's class, because it case just point to a function of the interface.

Filter94: >when there are refactor and class changes the used methods make the refactor even bigger

You mean because any individual "usage" (method) is a 1 argument in the constructor?

Filter94: >I agree, hence I didn't create all the methods

But are we using state overrides, getBalance, getCoinbase? I'd rather remove them if we don't

Filter94: > it's not entirely true was in pratice only need to mock/fake those that are used in practice

I don't get that. 
1. If we're going with a fake implementation we either use an approach similar to mocking, i.e. set state for a part of the functionality that is interconnected, and then have calls only for that functionality, otherwise the test double can fail or return default data. 
Next logical step to solve this issue is to have segregated interfaces and test doubles only for the interconnected parts of the logic. 
    * `eth_gasPrice`, `ethMaxPriorityFeePerGas`, `ethFeeHistory` could be a part of a subinterface
    *  methods related to transactions and blocks - in another 
    * general stuff like `ethChainId`, `ethProtocolVersion` could be separate too  

    Having this segregation when you implement the Fake Chain even makes sense internally to segregate a little the logic inside of it.
2. If we use mocks, indeed, only the relevant parts of the logic are mocked, but it comes with all the general disadvantages of the mock usage, which we're trying to avoid in this codebase

jvm-libs/linea/clients/interfaces/src/main/kotlin/linea/ethapi/EthApiClient.kt
```diff
@@ -3,37 +3,119 @@ package linea.ethapi
...
+
+  override fun hashCode(): Int {
+    var result = balance.hashCode()
+    result = 31 * result + nonce.hashCode()
+    result = 31 * result + code.contentHashCode()
+    result = 31 * result + movePrecompileToAddress.contentHashCode()
+    result = 31 * result + state.hashCode()
+    result = 31 * result + stateDiff.hashCode()
+    return result
+  }
```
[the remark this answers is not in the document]
fluentcrafter: Kotlin contentEquals and contentHashCode support nullables

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/mappers/TransactionForEthCallMapper.kt
```diff
@@ -0,0 +1,35 @@
+package linea.web3j.mappers
+
+import linea.domain.TransactionForEthCall
+import linea.kotlin.encodeHex
+import linea.kotlin.toBigInteger
+
+fun TransactionForEthCall.toWeb3j(): org.web3j.protocol.core.methods.request.Transaction {
```
Filter94, line 7: Nit: Why are we using FQN here?

Filter94 (approved):

jvm-libs/linea/clients/interfaces/src/main/kotlin/linea/ethapi/EthLogsClient.kt
```diff
@@ -4,11 +4,25 @@ import linea.domain.BlockParameter
 import linea.domain.EthLog
 import tech.pegasys.teku.infrastructure.async.SafeFuture
 
+data class FilterOptions(
+  val fromBlock: BlockParameter,
+  val toBlock: BlockParameter,
+  val address: String,
+  val topics: List<String?>,
```
gauravahuja, line 11: Does this need to be a list of string nullables? Why not just `List<String>` ?

fluentcrafter: because it's the way filters work, you can confirm on eth_getFilters docs

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/mappers/TransactionForEthCallMapper.kt
fluentcrafter: not sure, likeley an overlook of IDE automation. Will merge as it's worthe 20min wait for a CI run...

PR #1594, opened 2025-10-20
perf: improve smt tree build perf

gbotrel:
This PR improves perf and allocations for `smt.BuildComplete` .

```
benchmark                      old ns/op     new ns/op     delta
BenchmarkBuildComplete-128     282486312     185284604     -34.41%

benchmark                      old allocs     new allocs     delta
BenchmarkBuildComplete-128     8394127        3146893        -62.51%

benchmark                      old bytes     new bytes     delta
BenchmarkBuildComplete-128     369659188     134341529     -63.66%
```

---

> [!NOTE]
> Reworks SMT hashing to reuse a single hasher per task and replaces generic parallel exec with a worker pool and preallocated arena, reducing allocations and speeding up BuildComplete; updates call sites and adds a benchmark.
> 
> - **SMT performance improvements**:
>   - **Hasher reuse**: `hashLR` now accepts a `hashtypes.Hasher` and callers reuse a single hasher across loops in `proof.RecoverRoot`, `Tree.NewEmptyTree`, `Tree.Update`, and `BuildComplete`.
>   - **Parallel build overhaul**: Replace `utils/parallel` with an internal worker pool using `runtime.GOMAXPROCS`, `sync.WaitGroup`, and a task channel; compute each level in parallel.
>   - **Memory optimizations**: Preallocate an arena for all non-root levels and slice per-level ranges to avoid per-level allocations; reuse buffers while building.
>   - **Root/empty nodes hashing**: Compute root and empty nodes using the reused hasher.
> - **Tests/benchmarks**:
>   - Add `BenchmarkBuildComplete` to measure `BuildComplete` with `MiMC` hasher.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2aa4e425809ba68578b3bcdb58479b0dcd310935. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/prover/crypto/state-management/smt/smt_test.go
+++ b/prover/crypto/state-management/smt/smt_test.go
@@ -150,8 +150,7 @@ func BenchmarkBuildComplete(b *testing.B) {
 		leaves[i] = Bytes32(leavesFr[i].Bytes())
 	}
 
-	b.ResetTimer()
-	for i := 0; i < b.N; i++ {
+	for b.Loop() {
 		_ = smt.BuildComplete(leaves, config.HashFunc)
 	}
 }
--- a/prover/crypto/state-management/smt/tree.go
+++ b/prover/crypto/state-management/smt/tree.go
@@ -302,7 +302,7 @@ func BuildComplete(leaves []types.Bytes32, hashFunc func() hashtypes.Hasher) *Tr
 		wg    *sync.WaitGroup
 	}
 
-	workerCount := runtime.NumCPU()
+	workerCount := runtime.GOMAXPROCS(0)
 	if workerCount < 1 {
 		workerCount = 1
 	}
@@ -314,10 +314,8 @@ func BuildComplete(leaves []types.Bytes32, hashFunc func() hashtypes.Hasher) *Tr
 		go func() {
 			hasher := config.HashFunc()
 			for task := range tasks {
-				if task.stop > task.start {
-					for k := task.start; k < task.stop; k++ {
-						task.next[k] = hashLR(hasher, task.curr[2*k], task.curr[2*k+1])
-					}
+				for k := task.start; k < task.stop; k++ {
+					task.next[k] = hashLR(hasher, task.curr[2*k], task.curr[2*k+1])
 				}
 				task.wg.Done()
 			}


prover/crypto/state-management/smt/smt_test.go
```diff
@@ -136,3 +136,22 @@ func TestBuildFromScratch(t *testing.T) {
...
+
+	// Generate random field elements and cast them into Bytes32es
+	leavesFr := vector.Rand(1 << config.Depth)
+	leaves := make([]Bytes32, len(leavesFr))
+	for i := range leaves {
+		leaves[i] = Bytes32(leavesFr[i].Bytes())
+	}
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
```
ivokub: Recommend using `b.Loop()`, it resets the timers/metrics automatically and ensures that nothing is optimized out.

prover/crypto/state-management/smt/tree.go
```diff
@@ -282,26 +285,87 @@ func BuildComplete(leaves []types.Bytes32, hashFunc func() hashtypes.Hasher) *Tr
...
+	// start a worker pool that will parallelize the hashing at each level
+	type workerTask struct {
+		curr  []types.Bytes32
+		next  []types.Bytes32
+		start int
+		stop  int
+		wg    *sync.WaitGroup
+	}
+
+	workerCount := runtime.NumCPU()
```
ivokub: Should use GOMAXPROCS, then honors the env variables

prover/crypto/state-management/smt/tree.go
```diff
@@ -282,26 +285,87 @@ func BuildComplete(leaves []types.Bytes32, hashFunc func() hashtypes.Hasher) *Tr
...
+	}
+
+	tasks := make(chan workerTask)
+	var workersWG sync.WaitGroup
+	workersWG.Add(workerCount)
+	for i := 0; i < workerCount; i++ {
+		go func() {
+			hasher := config.HashFunc()
+			for task := range tasks {
+				if task.stop > task.start {
```
ivokub: Duplicate check, in the loop below we always start from `k := task.start` until `k < task.stop`. If `task.stop < task.start` then we don't loop.

prover/crypto/state-management/smt/tree.go
```diff
@@ -282,26 +285,87 @@ func BuildComplete(leaves []types.Bytes32, hashFunc func() hashtypes.Hasher) *Tr
...
+				}
+				task.wg.Done()
+			}
+			workersWG.Done()
+		}()
+	}
+
+	// build the tree bottom up
 	currLevels := leaves
+	for d := 1; d < depth; d++ {
```
ivokub, line 328: Maybe we could instead in the works goroutine compute full triangles instead of level by level? Then we have less synchronization and less copying between goroutines? And in case we're going use wider compression (a la 4->2 or 8->4 Poseidon) I think it would work more conveniently.

prover/crypto/state-management/smt/tree.go
```diff
@@ -282,26 +285,87 @@ func BuildComplete(leaves []types.Bytes32, hashFunc func() hashtypes.Hasher) *Tr
 	tree := NewEmptyTree(config)
 	tree.OccupiedLeaves = leaves
 
-	// Builds the tree bottom-up
+	// pre-allocate the memory for all levels
+	nbTotalNodes := 0
+	for d := 1; d < depth; d++ {
+		nbTotalNodes += 1 << (depth - d)
+	}
+	arena := make([]types.Bytes32, nbTotalNodes)
```
ivokub, line 293: The naming confuses a bit :). When seeing arena I'm thinking that it is allocated from the memory arena. But as we assign it as a field in `Tree` then we would leak it. But its not an issue as its heap allocated here.

prover/crypto/state-management/smt/tree.go
```diff
@@ -282,26 +285,87 @@ func BuildComplete(leaves []types.Bytes32, hashFunc func() hashtypes.Hasher) *Tr
...
+				start: start,
+				stop:  stop,
+				wg:    &levelWG,
 			}
-		})
-		tree.OccupiedNodes[i] = nextLevel
+		}
+		levelWG.Wait()
+
+		tree.OccupiedNodes[d-1] = nextLevel
```
ivokub, line 352: As per the comment above - we store a lot here, if we store all intermediate nodes here we use 2x the space what we use for leaves.

Maybe - it is more efficient if we only store every 1/2/4/8-th level for example and then compute the particular node whenever we need? Or do we need it at all?

PR #1685, opened 2025-11-04
Refactor periodic polling service to use vertx or jvm timers

gauravahuja:
This PR implements issue(s) #

### Checklist

---

> [!NOTE]
> Adds a new timer/periodic polling framework (Vertx/JVM) and migrates all periodic services to it, moving LongRunningService to linea and adding comprehensive tests.
> 
> - **Core timers + services**:
>   - Introduces `linea.timer` framework: `Timer`, `TimerFactory`, `TimerSchedule` (fixed delay/rate), `JvmTimer`, `VertxTimer`, `PeriodicPollingService`, and `VertxPeriodicPollingService`.
>   - Moves `LongRunningService` to `linea` and adds composable `ServiceAggregator`.
> - **Migrations to VertxPeriodicPollingService** (all now named and using `TimerSchedule.FIXED_DELAY`):
>   - Finalization/monitoring: `FinalizationUpdatePoller`, `FinalizationMonitorImpl`.
>   - State recovery: `StateSynchronizerService`, `SubmissionsFetchingTask`, `SubmissionEventsFetchingTask`, `BlobsFetchingTask`, `BlobDecompressionTask`.
>   - Coordinator: `BlockCreationMonitor`, `ProofAggregationCoordinatorService`, `BlobSubmissionCoordinator`, `AggregationFinalizationCoordinator`.
>   - Gas pricing: `MinMineableFeesPricerService`, `ExtraDataV1PricerService`, `FeeHistoryCachingService`.
>   - Message anchoring: `MessageAnchoringService`, `L1MessageSentEventsPoller`; configurator updated.
>   - Utilities/tests: `BlockEncodingValidator`, `RejectedTransactionCleanupService`.
> - **Package/import updates**:
>   - Replace usages of `net.consensys.zkevm.LongRunningService`/old `PeriodicPollingService` with `linea.LongRunningService` and new timer-based services.
> - **Tests**:
>   - Add `TimerTest` and `GenericPeriodicPollingServiceTest` covering Vertx/JVM timers, fixed delay/rate behavior, idempotent start/stop, error handling, and non-concurrent ticks.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 31747eb102c8595fe92967e6b1a8386541a5da1b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


jvm-libs/linea/core/long-running-service/src/main/kotlin/net/consensys/zkevm/VertxTimer.kt
```diff
@@ -0,0 +1,70 @@
...
+  @Suppress("UNUSED_PARAMETER")
+  private fun taskHandler(_timerId: Long) {
+    val callable = Callable {
+      task.run()
+    }
+    vertx.executeBlocking(callable).onComplete { result ->
+      if (result.cause() != null) {
+        errorHandler(result.cause())
+      }
+      synchronized(this) {
```
jonesho: Why do we need `synchronized` here? How could other threads accessing this block of code simultaneously?

jvm-libs/linea/core/long-running-service/src/main/kotlin/net/consensys/zkevm/VertxTimer.kt
```diff
@@ -0,0 +1,70 @@
...
+  private val vertx: Vertx,
+  override val name: String,
+  override val task: Runnable,
+  override val initialDelay: Duration,
+  override val period: Duration,
+  override val errorHandler: (Throwable) -> Unit,
+) : LineaTimer {
+  private var timerId: Long? = null
+
+  internal fun timerReference(): Long? = timerId
```
jonesho: nit: `timerIdReference`

jvm-libs/linea/core/long-running-service/src/main/kotlin/net/consensys/zkevm/JvmTimer.kt
```diff
@@ -0,0 +1,64 @@
...
+
+  internal fun timerReference(): Timer? = timer
+
+  @Synchronized
+  override fun start() {
+    if (timer != null) {
+      return
+    }
+    timer = Timer(name, true)
+    timer!!.scheduleAtFixedRate(
```
jonesho: This `scheduleAtFixedRate` seems to me not an equivalent to `vertx.setTimer` as it would schedule to run the task periodically at a fixed rate, but the way we use vertx timer is a bit different that we would call `vertx.setTimer` and then when it completes the task, we would explicitly call `vertx.setTimer` again and so on, in a sense that there's no hard gurantee on the time between the start of tasks, as it is: inital_delay -> run_task -> inital_delay  -> run_task -> so on. My main concern is whether the `scheduleAtFixedRate` would fire the multiple same tasks in parallel if the first one is taking too much time.

Btw I saw the unit test in LineaTimerTest `ticks shouldn't run concurrently if execution is blocked for more than polling interval` is the one validating that, and looks like `scheduleAtFixedRate` would also run the task until the current one completed, correct?

jvm-libs/linea/core/long-running-service/src/main/kotlin/linea/timer/PeriodicPollingService.kt
```diff
@@ -0,0 +1,50 @@
...
+  open fun handleError(error: Throwable) {
+    log.warn("Error with periodic polling service: errorMessage={}", error.message, error)
+  }
+
+  @Synchronized
+  override fun start(): SafeFuture<Unit> {
+    if (timer == null) {
+      timer = timerFactory.createTimer(
+        name = this.javaClass.simpleName,
+        task = { action().get() },
```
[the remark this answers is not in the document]
jonesho: I'm also wonderinf why we need to turns async into sync here

jvm-libs/linea/core/long-running-service/src/test/kotlin/linea/timer/GenericPeriodicPollingServiceTest.kt
```diff
@@ -102,7 +148,13 @@ class PeriodicPollingServiceTest {
         result
       }
     }
-    val pollingService = PollingService(vertx, pollingInterval, log, mockAction = action)
+    val pollingService = PollingService(
```
jonesho, line 153: Not related to this PR, but I dun see what's the intention of line:144-148? do u think we can simplify it into:
```
asyncDelay(vertx, 5.milliseconds)
```

jonesho: Same for line:189-193

jvm-libs/linea/core/long-running-service/src/main/kotlin/linea/timer/Timer.kt
fluentcrafter, line 1: We now favour new package `linea.X` instead of `net.consensys.zkevm`,  please move the new interfaces/classes to e.g `linea.timer` ?

jvm-libs/linea/core/long-running-service/src/main/kotlin/linea/timer/Timer.kt
```diff
@@ -0,0 +1,31 @@
...
+ *  * @param errorHandler A function to handle any errors that occur during task execution.
+ *  */
+interface LineaTimer {
+  val name: String
+  val task: Runnable
+  val initialDelay: Duration
+  val period: Duration
+  val errorHandler: (Throwable) -> Unit
+  fun start()
+  fun stop()
```
fluentcrafter, line 48: Did you consider `.pause()`? What was your conclusion?

jvm-libs/linea/core/long-running-service/src/main/kotlin/net/consensys/zkevm/LineaTimer.kt
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ *  * Interface representing a timer that can schedule and execute tasks at fixed intervals.
+ *  * @param name The name of the timer.
+ *  * @param task The task to be executed periodically. Can be blocking
+ *  * @param initialDelay The initial delay before the first execution of the task.
+ *  * @param period The period between successive executions of the task.
+ *  * @param errorHandler A function to handle any errors that occur during task execution.
+ *  */
+interface LineaTimer {
```
fluentcrafter:
```suggestion
interface Timer {
```

The timer can be agnostic. IMO it does not have Linea logic per si.

jvm-libs/linea/core/long-running-service/src/main/kotlin/net/consensys/zkevm/LineaPeriodicPollingService.kt
```diff
@@ -0,0 +1,50 @@
...
+
+  open fun handleError(error: Throwable) {
+    log.warn("Error with periodic polling service: errorMessage={}", error.message, error)
+  }
+
+  @Synchronized
+  override fun start(): SafeFuture<Unit> {
+    if (timer == null) {
+      timer = timerFactory.createTimer(
+        name = this.javaClass.simpleName,
```
fluentcrafter: This is a suboptimal design for the name because we can have multiple instances of the same concrete poller Class with object composition and will yield the same name but poll different things/purposes. We should allow for custom names.

jvm-libs/linea/core/long-running-service/src/main/kotlin/linea/timer/PeriodicPollingService.kt
fluentcrafter, line 33: Could you explain the reasoning behind this design decision: the action function is asynchronous, but we then make a .get() call?

jvm-libs/linea/core/long-running-service/src/main/kotlin/net/consensys/zkevm/VertxTimer.kt
```diff
@@ -0,0 +1,70 @@
...
+  private fun taskHandler(_timerId: Long) {
+    val callable = Callable {
+      task.run()
+    }
+    vertx.executeBlocking(callable).onComplete { result ->
+      if (result.cause() != null) {
+        errorHandler(result.cause())
+      }
+      synchronized(this) {
+        if (timerId != null) {
```
fluentcrafter: What's the purpose of this if statement?

jvm-libs/linea/core/long-running-service/src/main/kotlin/net/consensys/zkevm/JvmTimer.kt
fluentcrafter: @jonesho's point is valid.

There are 3 main types of timers (I don't remember the exact technical terms now) :
- 1. FixedRate - fires every T time, regardless of the result/status of the action;
- 2. "BackOffRate" - fires T time after the previous action returned
- 3. "AtMostRate" - tries to keep T interval, it may take longer if `action` takes longer than T. T = 3 sec, action takes 2s it will fire next `action` 1s after. But if action takes 4s, it fires right away because time has elapsed...

The target behaviour will depend on the use-case; for block building we may use an `AtMostRate` implementation and BackOffRate for L1 Polling. It will depend on each case. We should put these behaviours under a flag in the factory ...  and return the concrete implementation

jvm-libs/linea/core/long-running-service/src/main/kotlin/net/consensys/zkevm/LineaTimer.kt
```diff
@@ -0,0 +1,31 @@
...
+}
+
+interface TimerFactory {
+  fun createTimer(
+    name: String,
+    task: Runnable,
+    initialDelay: Duration,
+    period: Duration,
+    errorHandler: (Throwable) -> Unit,
+  ): LineaTimer
```
fluentcrafter: I recommend putting the tas as the last argument to be more Kotlin idomatic
```suggestion
  fun createTimer(
    name: String,
    initialDelay: Duration,
    period: Duration,
    errorHandler: (Throwable) -> Unit,
    task: Runnable,
  ): LineaTimer
```

jvm-libs/linea/core/long-running-service/src/main/kotlin/net/consensys/zkevm/VertxTimer.kt
```diff
@@ -0,0 +1,70 @@
...
+  override fun createTimer(
+    name: String,
+    task: Runnable,
+    initialDelay: Duration,
+    period: Duration,
+    errorHandler: (Throwable) -> Unit,
+  ): LineaTimer {
+    return VertxTimer(
+      vertx = vertx,
+      name = "$name-${UUID.randomUUID()}",
```
fluentcrafter: Why overriding the name with UUID?

Also, UUID generate ver long thread names which are log noisy :(

jvm-libs/linea/core/long-running-service/src/main/kotlin/net/consensys/zkevm/JvmTimer.kt
gauravahuja: > looks like scheduleAtFixedRate would also run the task until the current one completed, correct?

JVM timer has only one thread so it will execute one task until it is complete before scheduling or running another.

> FixedRate - fires every T time, regardless of the result/status of the action;

This isn't true, if the action takes longer to complete the other tasks will not run until the previous one is complete. However, jvm timer, if set to schedule at fixed rate will try to compensate for it by schueduling subsequent tasks without much delay such that the overall frequency is 1/period. There is only on thread that runs the action and if it is busy it will not concurrently run another task. From the [documentation](https://docs.oracle.com/en/java/javase/23/docs/api/java.base/java/util/Timer.html)

> Corresponding to each Timer object is a single background thread that is used to execute all of the timer's tasks, sequentially. Timer tasks should complete quickly. If a timer task takes excessive time to complete, it "hogs" the timer's task execution thread. This can, in turn, delay the execution of subsequent tasks, which may "bunch up" and execute in rapid succession when (and if) the offending task finally completes.

As of right now the JVM timer is scheduled at Fixed Rate whereas the VertxTimer is Fixed Delay. In Maru we use jvm timer scheduled at fixed rate but none of our maru code is so sensitive to timing. The block builder is invoced within the Besu QBFT library so it has its own timers or scheduler which would be time sensitve.

However, I can make type of timer explicit in the TimerFactory, I will try to have both Fixed Rate and Fixed Delay, but it will be difficult to provide the same guarantees for vertx fixed rate that jvm fixed rate provides.

jvm-libs/linea/core/long-running-service/src/main/kotlin/net/consensys/zkevm/LineaPeriodicPollingService.kt
gauravahuja: I was envisioning the way we currently use it, eg: ElSyncService extends LineaPeriodicPollingService, PeerChainTracker extends LineaPeriodicPollingService. And we only have one object of these concrete implementations.

But it is true it does not work with multiple objects of the same concrete class. So I will move the name argument to the constructor.

jvm-libs/linea/core/long-running-service/src/main/kotlin/linea/timer/PeriodicPollingService.kt
gauravahuja: Something has to wait for the async function to complete, you can't leave the future dangling, it will lead to unexpected results. The timer task is a Runnable which is synchronous. Also take the following scenario, It fails if you do not wait for the action to complete.

```
@Test
  fun `test action get`(vertx: Vertx) {
    val actionComplete = AtomicBoolean(false)

    fun action(delay: Duration): SafeFuture<Unit> {
      val future = SafeFuture<Unit>()
      vertx.setTimer(delay.inWholeMilliseconds) {
        actionComplete.set(true)
        println("End 2")
        future.complete(Unit)
      }
      return future
    }

    val task = Runnable {
      println("Start")
      action( 5.seconds)
      println("End 1")
    }

    val thread = Thread(task)
    thread.start()
    thread.join()
    assertThat(actionComplete.get()).isTrue()
  }
```

jvm-libs/linea/core/long-running-service/src/main/kotlin/linea/timer/Timer.kt
gauravahuja: Stop is pause. But do you mean having an interface method pause and it defaults to stop()?

jvm-libs/linea/core/long-running-service/src/main/kotlin/net/consensys/zkevm/VertxTimer.kt
gauravahuja: By calling stop from another thread. In stop I update timerId

jvm-libs/linea/core/long-running-service/src/main/kotlin/net/consensys/zkevm/VertxTimer.kt
gauravahuja: If another thread called stop we do not want to schedule the next run.

jvm-libs/linea/core/long-running-service/src/main/kotlin/net/consensys/zkevm/VertxTimer.kt
gauravahuja: Was just trying to mimic what we have in maru code. I will remove the uuid so the user is solely responsible for unique names.

jvm-libs/linea/core/long-running-service/src/test/kotlin/linea/timer/GenericPeriodicPollingServiceTest.kt
gauravahuja: I think so it can be simplified.

jvm-libs/linea/core/long-running-service/src/main/kotlin/linea/timer/PeriodicPollingService.kt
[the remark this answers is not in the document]
gauravahuja: https://github.com/Consensys/linea-monorepo/pull/1685#discussion_r2499922873

gauravahuja: the action function is asynchronous is because currently all polling services implement it as async, If i make it sync it will be a lot of refactoring.

jvm-libs/linea/core/long-running-service/src/main/kotlin/linea/timer/VertxTimer.kt
```diff
@@ -0,0 +1,107 @@
...
+
+  private fun nextInvocationDelay(): Duration {
+    return when (timerSchedule) {
+      TimerSchedule.FIXED_DELAY -> period
+      TimerSchedule.FIXED_RATE -> {
+        val firstTime = firstInvocationTime.load()!!
+        val expectedNextInvocationTime = firstTime + period * invocationCounter.get()
+        val now = Clock.System.now()
+        val delay = expectedNextInvocationTime - now
+        if (delay.isNegative() || delay.inWholeMilliseconds == 0L) {
```
jonesho: why not just `delay.inWholeMilliseconds <= 0L`

jvm-libs/linea/core/long-running-service/src/main/kotlin/net/consensys/zkevm/JvmTimer.kt
fluentcrafter: Thanks for the clarification.

> However, I can make type of timer explicit in the TimerFactory, I will try to have both Fixed Rate and Fixed Delay, but it will be difficult to provide the same guarantees for vertx fixed rate that jvm fixed rate provides.

Making it explicit is an advantage IMO and better than what we have. Regarding vertx vs jvm it's an acceptable trade off IMO, the caller made the decision of which one to inject.

jvm-libs/linea/core/long-running-service/src/main/kotlin/linea/timer/Timer.kt
fluentcrafter: My question was more `start, pause, stop`.

So we can call `start() stop() start() stop()` right?

jvm-libs/linea/core/long-running-service/src/main/kotlin/net/consensys/zkevm/VertxTimer.kt
fluentcrafter: You provide a less vebose default value: `"$name-${Random.nextBytes(4).encodeHex(false)}"`

jvm-libs/linea/core/long-running-service/src/main/kotlin/linea/timer/Timer.kt
```diff
@@ -0,0 +1,39 @@
...
+}
+
+interface TimerFactory {
+  fun createTimer(
+    name: String,
+    initialDelay: Duration,
+    period: Duration,
+    timerSchedule: TimerSchedule,
```
fluentcrafter, line 56: IMO, instead of an enum argument, the interface should have 2 methods: `createFixedTimer` `, createFixedDelayTimer` with corresponding documentation of what each one does.

This way, TimerSchedule enum can be package private/internal. 

Regardless, I dont' recommend to make it part of the `Timer` interface. Timer interface IMO can have just the `start()` `stop()` methods.

gauravahuja: How will the GenericPollingService know what type of timer to instantiate if TimerSchedule is not public?

fluentcrafter: aren't they in the same module? If so, internal is enough.

jvm-libs/linea/core/long-running-service/src/main/kotlin/linea/PeriodicPollingService.kt
```diff
@@ -0,0 +1,24 @@
+package linea
+
+import io.vertx.core.Vertx
+import linea.timer.TimerSchedule
+import linea.timer.VertxTimerFactory
+import org.apache.logging.log4j.Logger
+import kotlin.time.Duration.Companion.milliseconds
+
+abstract class PeriodicPollingService(
```
fluentcrafter: This feels just like an alieas to `GenericPeriodicPollingService`.

I think we can remove this `PeriodicPollingService`

jvm-libs/linea/core/long-running-service/src/main/kotlin/linea/timer/Timer.kt
gauravahuja: The concrete implementation of GenericPollingService needs to provide that TimerSchedule. Eg: PollingService1: GenericPollingService(...), it needs to tell what type of timer to use.

jvm-libs/linea/core/long-running-service/src/main/kotlin/linea/PeriodicPollingService.kt
gauravahuja: PeriodicPollingService defaults to vertx and fixed delay timer which was the case before the refactor. So it results in minimum changes to existing code.

PR #1708, opened 2025-11-10
Timer polling service refactor adds test for isolation

fluentcrafter:
This PR implements issue(s) #

### Checklist

---

> [!NOTE]
> Adds a new test ensuring independent timers don’t block each other and simplifies the test PollingService by removing Vertx dependency and adjusting assertions.
> 
> - **Tests**:
>   - Add parameterized test `should not allow action of one timer affect or delay another timer` to verify multiple timers run independently with different intervals.
>   - Update assertions:
>     - Relax error verification to accept any message in `action handles a throw`.
>     - Remove strict error log count verification in `continues to poll after throwing exception`.
> - **Test Helper Refactor** (`PollingService` in `GenericPeriodicPollingServiceTest.kt`):
>   - Remove `Vertx` dependency from constructor and `action()`; now directly executes `mockAction`.
>   - Update all test instantiations to omit `vertx` argument; remove unnecessary mocked `Vertx` in idempotency test.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8c022cc6ebe621d8b50bc690735388990755d209. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/jvm-libs/linea/core/long-running-service/src/test/kotlin/linea/GenericPeriodicPollingServiceTest.kt
+++ b/jvm-libs/linea/core/long-running-service/src/test/kotlin/linea/GenericPeriodicPollingServiceTest.kt
@@ -382,7 +382,7 @@ class GenericPeriodicPollingServiceTest {
     poller1.start()
 
     await()
-      .atMost(5.seconds.toJavaDuration())
+      .atMost(20.seconds.toJavaDuration())
       .untilAsserted {
         assertThat(poller1Calls.get()).isGreaterThanOrEqualTo(1)
         assertThat(poller2Calls.get()).isGreaterThanOrEqualTo(1)


jvm-libs/linea/core/long-running-service/src/test/kotlin/linea/GenericPeriodicPollingServiceTest.kt
```diff
@@ -355,10 +343,55 @@ class GenericPeriodicPollingServiceTest {
...
+      SafeFuture.completedFuture(Unit)
+    }
+
+    val poller2 = PollingService(
+      pollingInterval = 100.milliseconds,
+      log,
+      timerFactory = timerFactory,
+      timerSchedule = timerSchedule,
+    ) {
+      Thread.sleep(500L)
```
jonesho, line 376: @fluentcrafter Is this supposed to be just `50L`?

jvm-libs/linea/core/long-running-service/src/test/kotlin/linea/GenericPeriodicPollingServiceTest.kt
```diff
@@ -355,10 +343,55 @@ class GenericPeriodicPollingServiceTest {
...
+    }
+
+    poller2.start()
+    poller1.start()
+
+    await()
+      .atMost(20.seconds.toJavaDuration())
+      .untilAsserted {
+        assertThat(poller1Calls.get()).isGreaterThanOrEqualTo(1)
+        assertThat(poller2Calls.get()).isGreaterThanOrEqualTo(1)
```
jonesho, line 388: @gauravahuja I tried to make this line-388 to be `isGreaterThanOrEqualTo(4)` and the vertx tests would fail, and what I notice is that, by printing the counter along with the timestamp for each action run, it seems that both poller 1 and 2 were not running in parallel (with FIXED_DELAY and decrease poller2 sleep time to `50L`):
```
poller1Calls: 1 time: 1762803832388
poller1Calls: 2 time: 1762803832409
poller1Calls: 3 time: 1762803832427
poller1Calls: 4 time: 1762803832445
poller1Calls: 5 time: 1762803832462
poller2Calls: 1 time: 1762803832515
poller1Calls: 6 time: 1762803832522
poller1Calls: 7 time: 1762803832539
poller1Calls: 8 time: 1762803832557
poller1Calls: 9 time: 1762803832575
poller1Calls: 10 time: 1762803832592
poller1Calls: 11 time: 1762803832610
poller2Calls: 2 time: 1762803832670
poller1Calls: 12 time: 1762803832677
```
As you can see from the log above, e.g. during poller2 second sleep time (at `1762803832670`), there were no poller1 thread being called, and vice versa, it looks like poller1 and 2 were being called one by one which is not expected, but indeed Java timer didn't have this issue as poller1 and 2 were running in parallel:
```
poller1Calls: 13 time: 1762803756373
poller1Calls: 14 time: 1762803756383
poller2Calls: 1 time: 1762803756385
poller1Calls: 15 time: 1762803756395
poller1Calls: 16 time: 1762803756405
poller1Calls: 17 time: 1762803756416
poller1Calls: 18 time: 1762803756427
poller1Calls: 19 time: 1762803756437
poller1Calls: 20 time: 1762803756450
poller1Calls: 21 time: 1762803756459
poller1Calls: 22 time: 1762803756469
poller1Calls: 23 time: 1762803756482
poller2Calls: 2 time: 1762803756489
poller1Calls: 24 time: 1762803756493
```
can you help check what causes this behavior in Vertx?

gauravahuja: I have a feeling that the vertx instantiated for the test has only one thread in the event loop and it is blocked by the sleep.

jonesho: @gauravahuja oh I found out that it was because we're using the default `ordered` as true in `executeBlocking`, can you help correct that to false and verify on your side?

> I have a feeling that the vertx instantiated for the test has only one thread in the event loop and it is blocked by the sleep.

This is not true, vertx uses a thread pool to handle even blocking operation like `Thread.sleep`, the reason for the sequential execution is simply because we're using the default `ordered` as true, and it means all the tasks would be running on the same worker thread in order to guarantee the ordering

jvm-libs/linea/core/long-running-service/src/test/kotlin/linea/GenericPeriodicPollingServiceTest.kt
fluentcrafter: No, it was 500 on purpose

PR #1754, opened 2025-11-19
Chore: Update Besu in Linea Besu Package

fab-10:
This PR implements issue(s) #

### Checklist

---

> [!NOTE]
> Update `LINEA_BESU_TAR_GZ` to `25.11.0-linea1` in `linea-besu-package/versions.env`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit eaee9a2b4e57f2e5b13c515b0b2aafede7410e71. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


linea-besu-package/versions.env
```diff
@@ -1,4 +1,4 @@
-LINEA_BESU_TAR_GZ=25.11.0-RC1-linea2
+LINEA_BESU_TAR_GZ=25.11.0-linea1
 LINEA_SEQUENCER_PLUGIN_VERSION=4.0-rc23.1
 LINEA_TRACER_PLUGIN_VERSION=beta-v4.0-rc23
```
ahamlat, line 3: There should be a new tracer version, right ?

fab-10: in this specific case I have checked that the new Besu does not introduce incompatibilities, so we can reuse the Tracer and Sequencer versions

ahamlat: How do we make sure that the code that will be executed on linea_estimateGas is the one you specify in this PR and not the one that is referenced by the sequencer ?

fab-10: Because the new version of Besu contains the fix, and neither the Sequencer or the Tracer use an uber jar, so the Sequencer will call the new code in Besu

PR #1756, opened 2025-11-19
Fix: interrupt bundle selection on timeout

fab-10:
This PR implements issue(s) #

### Checklist

---

> [!NOTE]
> Interrupt bundle evaluation when selection yields timeout/cancel results and add multi-bundle tests to verify interruption behavior.
> 
> - **Sequencer** (`LineaTransactionSelectorFactory`):
>   - Interrupts selection on `TransactionSelectionResult` of `PLUGIN_SELECTION_TIMEOUT`, `PLUGIN_SELECTION_TIMEOUT_INVALID_TX`, or `SELECTION_CANCELLED`.
>   - Replaces thread-interrupt checks with `AtomicBoolean isSelectionInterrupted`; updates `.takeWhile(...)` gating and logging; clears flag in `finally`.
>   - Distinguishes interrupted vs failed bundle logs; commits on success, rollbacks on failure/interruption.
> - **Tests** (`BundleSelectionTimeoutTest`):
>   - Adds multi-bundle timeout scenario: constructs small and large bundles; verifies only first bundle is included and subsequent bundles are not selected.
>   - Minor refactor to build raw tx arrays and adjust ranges (`1..11`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f9586d481182b47b649258ef97d9f6628dea326a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorFactory.java
+++ b/besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorFactory.java
@@ -146,9 +146,9 @@ public class LineaTransactionSelectorFactory implements PluginTransactionSelecto
                 final var currentBundleSelectionTime = now - bundleStartedAt;
 
                 if (maybeBadBundleRes.isPresent()) {
-                  final var failure = maybeBadBundleRes.get();
+                  final var notSelectedReason = maybeBadBundleRes.get();
 
-                  if (isSelectionInterrupted(failure)) {
+                  if (isSelectionInterrupted(notSelectedReason)) {
                     isSelectionInterrupted.set(true);
                     log.atDebug()
                         .setMessage(
@@ -163,7 +163,7 @@ public class LineaTransactionSelectorFactory implements PluginTransactionSelecto
                         .setMessage(
                             "Failed bundle {}, reason {}, elapsed time: current bundle {}ms, cumulative {}ms")
                         .addArgument(bundle::bundleIdentifier)
-                        .addArgument(failure)
+                        .addArgument(notSelectedReason)
                         .addArgument(() -> nanosToMillis(currentBundleSelectionTime))
                         .addArgument(() -> nanosToMillis(cumulativeBundleSelectionTime))
                         .log();
@@ -209,16 +209,17 @@ public class LineaTransactionSelectorFactory implements PluginTransactionSelecto
               .findFirst();
 
       if (badBundleRes.isPresent()) {
-        final var failure = badBundleRes.get();
+        final var notSelectedReason = badBundleRes.get();
 
-        if (isSelectionInterrupted(failure)) {
+        if (isSelectionInterrupted(notSelectedReason)) {
           isSelectionInterrupted.set(true);
           log.debug(
               "Bundle selection interrupted while processing liveness bundle {}, reason {}",
               livenessBundle.get(),
-              failure);
+              notSelectedReason);
         } else {
-          log.debug("Failed liveness bundle {}, reason {}", livenessBundle.get(), failure);
+          log.debug(
+              "Failed liveness bundle {}, reason {}", livenessBundle.get(), notSelectedReason);
         }
         livenessService.get().updateUptimeMetrics(false, headBlockTimestamp);
         rollback(bts);


besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorFactory.java
```diff
@@ -109,8 +115,6 @@ public void selectPendingTransactions(
       // check and send liveness bundle if any
       checkAndSendLivenessBundle(bts, pendingBlockHeader.getNumber());
```
jonesho, line 117: @fab-10 can we make sure the liveness bundle will always get included regardless of the block tx selection time configs (e.g. plugin-block-txs-selection-max-time), because in the recent mainnet incident, we encounted the issue that liveness bundle was not included because of other bundles that were unselected due to timeout

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorFactory.java
```diff
@@ -121,20 +125,18 @@ public void selectPendingTransactions(
...
+          .takeWhile(unused -> !isSelectionInterrupted.get())
           .forEach(
               bundle -> {
                 final var bundleStartedAt = System.nanoTime();
                 log.trace("Starting evaluation of bundle {}", bundle.bundleIdentifier());
 
                 var maybeBadBundleRes =
                     bundle.pendingTransactions().stream()
-                        .takeWhile(unused -> !isSelectionInterrupted())
+                        .takeWhile(unused -> !isSelectionInterrupted.get())
```
jonesho: @fab-10 I don't see why we need to check `isSelectionInterrupted` here, given that it had already been checked at line-131 and `isSelectionInterrupted` only gets set to true after we called `evaluatePendingTransaction` on all pending transaction of the bundle.
Unless other thread might call this `selectPendingTransactions` in parallel?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorFactory.java
fab-10: Liveness bundle is always the first and is not affected by other bundles, so if it was not included is due to itself going in timeout.

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorFactory.java
fab-10: Good catch, this check is redundant, because in case of one tx in the bundle fails (including for a timeout) the stream immediately return with that failure as result.

jonesho: Thanks @fab-10, to play safe, can you help double-check or add a test to verify if that was the case?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorFactory.java
```diff
@@ -231,9 +234,8 @@ private long nanosToMillis(final long nanos) {
     return TimeUnit.NANOSECONDS.toMillis(nanos);
   }
 
-  private boolean isSelectionInterrupted() {
-    // returns if the thread is interrupted without resetting the state
-    // so it can be called many times without changing the interrupt state
-    return Thread.currentThread().isInterrupted();
+  private boolean isSelectionInterrupted(final TransactionSelectionResult selectionResult) {
```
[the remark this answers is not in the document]
fab-10: good catch, done

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorFactory.java
fab-10: `BundleSelectionTimeoutTest::multipleBundleSelectionTimeout` covers that case, it sends 3 bundles for the same block:

1. bundle 1 is small and quick to evaluated and is selected 
2. bundle 2 is big and its entire evaluation takes more that the plugin selection timeout, so only a subset of its tx are evaluated before the timeout kicks in, txs after the timeout are not evaluated
3. bundle 3 is not evaluated at all, since  `isSelectionInterrupted` is true

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorFactory.java
```diff
@@ -143,45 +144,46 @@ public void selectPendingTransactions(
...
-                          "Bundle selection interrupted while processing bundle {},"
-                              + " elapsed time: current bundle {}ms, cumulative {}ms")
-                      .addArgument(bundle::bundleIdentifier)
-                      .addArgument(() -> nanosToMillis(currentBundleSelectionTime))
-                      .addArgument(() -> nanosToMillis(cumulativeBundleSelectionTime))
-                      .log();
-                  rollback(bts);
-                } else {
-                  if (maybeBadBundleRes.isPresent()) {
+                if (maybeBadBundleRes.isPresent()) {
```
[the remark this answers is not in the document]
fab-10: The stream stops the evaluation as soon as there is a tx not selected, due to the `findFirst`, and since a timeout is one of the reason a tx is not selected, then no other txs after the one that goes in timeout is evaluated, so the `takeWhile` was redundant, do you agree?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorFactory.java
jonesho: @fab-10 ic, so is it possible to set liveness bundle to not timeout at all, or if its not viable to do so, how could we have a better guarantee that liveness bundle would not be affected by other bundles timeout, maybe indeed the fix of this PR could help?

fab-10: not possible at the moment to make it uninterruptible, but for sure it is not affect by other bundles since it is always the first one. It is only affected by the plugin max selection time.

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorFactory.java
```diff
@@ -206,7 +209,17 @@ private void checkAndSendLivenessBundle(
...
+        final var failure = badBundleRes.get();
+
+        if (isSelectionInterrupted(failure)) {
+          isSelectionInterrupted.set(true);
+          log.debug(
+              "Bundle selection interrupted while processing liveness bundle {}, reason {}",
+              livenessBundle.get(),
+              badBundleRes);
+        } else {
+          log.debug("Failed liveness bundle {}, reason {}", livenessBundle.get(), badBundleRes);
```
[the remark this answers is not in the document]
fab-10: done

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorFactory.java
jonesho: But what I cannot understand is why liveness bundle was not included **before** we changed the plugin max selection time to 1%, it should get included as being the first one to be evaluated but seems like it was preempted by those heavy bundles and eventually timeout together

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorFactory.java
```diff
@@ -56,6 +62,7 @@ public class LineaTransactionSelectorFactory implements PluginTransactionSelecto
   private final AtomicReference<LineaTransactionSelector> currSelector = new AtomicReference<>();
   private final AtomicReference<Map<Address, Set<TransactionEventFilter>>> deniedEvents;
   private final AtomicReference<Map<Address, Set<TransactionEventFilter>>> deniedBundleEvents;
+  private final AtomicBoolean isSelectionInterrupted = new AtomicBoolean(false);
```
fluentcrafter, line 65: I lack context, is the factory instantiated A) only once or B) 1 factory for each block building  

Reason asking is that, if A)  maybe isSelectionInterrupted should be set false at the begining of tx selectcion, no?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorFactory.java
fab-10: let's take this offline checking the logs

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorFactory.java
```diff
@@ -143,45 +145,46 @@ public void selectPendingTransactions(
...
+                  commit(bts);
                 }
               });
     } finally {
       currSelector.set(null);
-      if (isSelectionInterrupted()) {
-        // finally consume the interrupt
-        Thread.currentThread().interrupt();
-      }
+      isSelectionInterrupted.set(false);
```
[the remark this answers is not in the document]
fab-10: this code is not meant to be called by multiple threads concurrently

besu-plugins/linea-sequencer/acceptance-tests/src/test/java/linea/plugin/acc/test/rpc/linea/BundleSelectionTimeoutTest.java
```diff
@@ -81,9 +81,15 @@ public void multipleBundleSelectionTimeout() throws Exception {
...
-        new SendBundleRequest(new BundleParams(rawTxs, Integer.toHexString(2)));
-    final var sendBundleResponseBig = sendBundleRequestBig.execute(minerNode.nodeRequests());
+    final var sendBundleRequestBig1 =
+        new SendBundleRequest(
+            new BundleParams(Arrays.copyOfRange(rawTxs, 0, 9), Integer.toHexString(2)));
+    final var sendBundleResponseBig1 = sendBundleRequestBig1.execute(minerNode.nodeRequests());
+
+    final var sendBundleRequestBig2 =
+        new SendBundleRequest(
+            new BundleParams(Arrays.copyOfRange(rawTxs, 0, 10), Integer.toHexString(2)));
```
ahamlat: As @fluentcrafter suggested, review this to use other transactions with different nonces, just to avoid the confusion around the nonces and timeout

besu-plugins/linea-sequencer/acceptance-tests/src/test/java/linea/plugin/acc/test/rpc/linea/BundleSelectionTimeoutTest.java
ahamlat, line 73: We need to make the test more deterministic and rely only on `mulmodOperation` performance

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorFactory.java
```diff
@@ -121,20 +126,17 @@ public void selectPendingTransactions(
...
+          .takeWhile(unused -> !isSelectionInterrupted.get())
           .forEach(
               bundle -> {
                 final var bundleStartedAt = System.nanoTime();
                 log.trace("Starting evaluation of bundle {}", bundle.bundleIdentifier());
 
                 var maybeBadBundleRes =
                     bundle.pendingTransactions().stream()
-                        .takeWhile(unused -> !isSelectionInterrupted())
                         .map(bts::evaluatePendingTransaction)
```
ahamlat, line 140: This means that we can not stop the evaluation of the bundle if there is a timeout. It still needs to evaluate all the transactions, correct ?

ahamlat: As there is findFirst(), that should not evaluate other transactions.

besu-plugins/linea-sequencer/acceptance-tests/src/test/java/linea/plugin/acc/test/rpc/linea/BundleSelectionTimeoutTest.java
fab-10: merged the PR the from @fluentcrafter that improves the clarity of the test

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorFactory.java
fab-10: it is instantiated once, and reset at the end of the `selectPendingTransactions` method in the finally statement

ahamlat (dismissed):

ahamlat (approved):

besu-plugins/linea-sequencer/acceptance-tests/src/test/java/linea/plugin/acc/test/rpc/linea/BundleSelectionTimeoutTest.java
```diff
@@ -74,16 +74,23 @@ public void multipleBundleSelectionTimeout() throws Exception {
...
 
-    final var rawTxs = Arrays.stream(calls).skip(1).map(MulmodCall::rawTx).toArray(String[]::new);
+    final var sendBundleRequestBig1 =
+        new SendBundleRequest(
+            new BundleParams(Arrays.copyOfRange(rawTxs, 1, 10), Integer.toHexString(2)));
+    // second bundle contains one tx only to be fast to execute,
+    // and ensure timeout occurs on the 2nd bundle and 3rd is not event considered
+    final var sendBundleRequestBig2 =
+        new SendBundleRequest(
+            new BundleParams(Arrays.copyOfRange(rawTxs, 1, 2), Integer.toHexString(2)));
```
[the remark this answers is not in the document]
fab-10: this is intended

PR #1810, opened 2025-11-28
coordinator: support for stop conflation at block timestamp

fluentcrafter:
This PR implements issue(s) #

### Checklist

---

> [!NOTE]
> Adds support to stop conflation at a target L2 block timestamp and migrates block monitoring to the new EthApi client, with TOML parsing for Instants and updated tests.
> 
> - **Conflation config and TOML**:
>   - Add `conflation.forceStopConflationAtBlockTimestampInclusive: Instant?` in `ConflationConfig` and `ConflationToml`, and propagate to runtime.
>   - Use `kotlinx.datetime.Instant` across configs; remove conversions.
> - **Config loading**:
>   - Add `TomlKotlinInstantDecoder` to parse ISO-8601 strings or epoch seconds and register it in `ConfigLoaderHelper`.
> - **Block monitoring/runtime**:
>   - `BlockCreationMonitor` now uses `EthApiBlockClient` instead of `ExtendedWeb3J` and fetches blocks via `ethGetBlockByNumberFullTxs`.
>   - Support stopping when reaching `lastL2BlockTimestampToProcessInclusive` or `lastL2BlockNumberToProcessInclusive`; stop service on reorg or fetch-limit overflow.
>   - `ConflationApp` creates `EthApiClient` and passes new timestamp stop config to `BlockCreationMonitor`.
> - **Tests**:
>   - Extend parsing tests to cover new timestamp fields and epoch-seconds parsing.
>   - Update `BlockCreationMonitorTest` to use `EthApiClient`, generate timestamps, and add test to stop at timestamp boundary.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 57967d262275fce453dcefa9770fd8233c194f27. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/decoders/Decoders.kt
+++ b/coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/decoders/Decoders.kt
@@ -3,8 +3,8 @@ package linea.coordinator.config.v2.toml.decoders
 import com.sksamuel.hoplite.ConfigFailure
 import com.sksamuel.hoplite.ConfigResult
 import com.sksamuel.hoplite.DecoderContext
+import com.sksamuel.hoplite.LongNode
 import com.sksamuel.hoplite.Node
-import com.sksamuel.hoplite.NumberNode
 import com.sksamuel.hoplite.StringNode
 import com.sksamuel.hoplite.decoder.Decoder
 import com.sksamuel.hoplite.fp.invalid
@@ -74,8 +74,8 @@ class TomlKotlinInstantDecoder : Decoder<Instant> {
         { it.valid() },
         { ConfigFailure.DecodeError(node, type).invalid() },
       )
-      is NumberNode -> runCatching {
-        Instant.fromEpochSeconds(node.value as Long)
+      is LongNode -> runCatching {
+        Instant.fromEpochSeconds(node.value)
       }.fold(
         { it.valid() },
         { ConfigFailure.DecodeError(node, type).invalid() },


coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/blockcreation/BlockCreationMonitor.kt
```diff
@@ -120,12 +123,25 @@ class BlockCreationMonitor(
             config.lastL2BlockNumberToProcessInclusive,
             nexBlockNumberToFetch,
           )
+          this.stop()
```
gauravahuja, line 126: why call this.stop here vs super.stop on line 143 and 164?

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/blockcreation/BlockCreationMonitor.kt
```diff
@@ -109,6 +111,7 @@ class BlockCreationMonitor(
             _nexBlockNumberToFetch.get() - lastProvenBlockNumber,
             config.blocksFetchLimit,
           )
+          this.stop()
```
gauravahuja, line 114: stop is an async method, so you can return the safefuture it returns instead of Safefuture.COMPLETE

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/blockcreation/BlockCreationMonitor.kt
fluentcrafter: we must favour this.stop over super.stop, because if stop is overrided it will not work properly

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/blockcreation/BlockCreationMonitor.kt
fluentcrafter: That would fail the InterruptedException. so it should be this way

Issue #1790, opened 2025-11-26
Replace Web3J Direct Usage by EthAPI Client


fluentcrafter:
🔐 🔐 `Should this be a security advisory instead?` 🔐 🔐

## Description
Replace the usage of Web3J by EthAPI client 

## Motivation
- Code consistency
- **Reduce Web3J surface in the code base**
- **Facilitate upgrades of Web3j Lib** 
- **Improve testabilitt by usage of a Fake ethApi Client**

PR #1895, opened 2025-12-12
Add blob fee method to eth api fee client

gauravahuja:
This PR implements issue(s) #1790 

### Checklist

---

> [!NOTE]
> Implements extended eth_feeHistory via custom request and refactors EthApi client/factory to build from rpcUrl/Web3jService, updating apps and tests to reuse shared clients.
> 
> - **Eth API / Web3j extensions**:
>   - `Web3jEthApiClient`: now holds `web3jService`; implements `eth_feeHistory` via manual `Request` using `EthFeeHistoryBlobExtended` (blob-aware).
>   - `createEthApiClient` (factory): new flow builds `Web3jService` with `createWeb3jHttpService`, then `Web3j` via `createWeb3jHttpClient`; signatures updated to accept `Web3jService`.
> - **Coordinator apps**:
>   - `MessageAnchoringAppConfigurator`: create `l1EthApiClient`/`l2EthApiClient` from `rpcUrl`; reuse `l2EthApiClient` for `Web3JL2MessageServiceSmartContractClient`.
>   - `ConflationApp`: create `l2EthClient` from `rpcUrl` with retries; pass same client to providers and read-only message service; remove redundant client creations.
> - **Contract client**:
>   - `Web3JL2MessageServiceSmartContractClient`: use provided `EthApiClient` for deployment block provider (no internal `Web3jEthApiClient`).
> - **Tests**:
>   - `LogsSearcherIntTest`: stop constructing `Web3j`/`HttpService` directly; use `createEthApiClient(rpcUrl, ...)`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 893695d6160f9571cecd4249c2e23f77ba74c771. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClient.kt
+++ b/jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClient.kt
@@ -88,23 +88,17 @@ class Web3jEthApiClient(
     newestBlock: BlockParameter,
     rewardPercentiles: List<Double>,
   ): SafeFuture<FeeHistory> {
-    return SafeFuture.of(
-      Request(
-        "eth_feeHistory",
-        listOf(
-          Numeric.encodeQuantity(BigInteger.valueOf(blockCount.toLong())),
-          newestBlock.toWeb3j().value,
-          rewardPercentiles,
-        ),
-        this.web3jService,
-        EthFeeHistoryBlobExtended::class.java,
-      ).sendAsync(),
-    ).thenCompose {
-      if (it.hasError()) {
-        SafeFuture.failedFuture(Exception(it.error.message))
-      } else {
-        SafeFuture.completedFuture(it.feeHistory.toLineaDomain())
-      }
+    return Request(
+      "eth_feeHistory",
+      listOf(
+        Numeric.encodeQuantity(BigInteger.valueOf(blockCount.toLong())),
+        newestBlock.toWeb3j().value,
+        rewardPercentiles,
+      ),
+      this.web3jService,
+      EthFeeHistoryBlobExtended::class.java,
+    ).requestAsync {
+      it.feeHistory.toLineaDomain()
     }
   }
 
--- a/coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/MessageAnchoringAppConfigurator.kt
+++ b/coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/MessageAnchoringAppConfigurator.kt
@@ -21,10 +21,6 @@ object MessageAnchoringAppConfigurator {
     }
     configs.messageAnchoring!!
 
-    val l1Web3jClient = createWeb3jHttpClient(
-      rpcUrl = configs.messageAnchoring.l1Endpoint.toString(),
-      log = LogManager.getLogger("clients.l1.eth.message-anchoring"),
-    )
     val l1EthApiClient = createEthApiClient(
       rpcUrl = configs.messageAnchoring.l1Endpoint.toString(),
       log = LogManager.getLogger("clients.l1.eth.message-anchoring"),
--- a/jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClientFactory.kt
+++ b/jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClientFactory.kt
@@ -72,21 +72,19 @@ fun createEthApiClient(
   vertx: Vertx? = null,
   stopRetriesOnErrorPredicate: Predicate<Throwable> = Predicate { _ -> false },
 ): EthApiClient {
-  val web3jClient =
-    createWeb3jHttpClient(
-      rpcUrl,
-      log,
-      pollingInterval,
-      executorService,
-      requestResponseLogLevel,
-      failuresLogLevel,
-    )
   val web3jService = createWeb3jHttpService(
     rpcUrl = rpcUrl,
     log = log,
     requestResponseLogLevel = requestResponseLogLevel,
     failuresLogLevel = failuresLogLevel,
   )
+
+  val web3jClient =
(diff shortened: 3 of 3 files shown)


jvm-libs/linea/clients/interfaces/src/main/kotlin/linea/ethapi/EthApiFeeClient.kt
```diff
@@ -13,4 +13,9 @@ interface EthApiFeeClient {
     newestBlock: BlockParameter,
     rewardPercentiles: List<Double>,
   ): SafeFuture<FeeHistory>
+  fun ethFeeHistoryBlobExtended(
```
fluentcrafter: why do we need a new method in this interface? What was you design reasoning?

gauravahuja: I was just following the pattern we have right now, a different class "BlobExtended" for blob fees. However I'm not sure if it is necessary. So I removed this new method and update the implementation of existing ethFeeHistory to parse blob fees.

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClient.kt
```diff
@@ -82,9 +87,26 @@ class Web3jEthApiClient(
...
-  ): SafeFuture<FeeHistory> = web3jClient
-    .ethFeeHistory(blockCount, newestBlock.toWeb3j(), rewardPercentiles)
-    .requestAsync { it.feeHistory.toLineaDomain() }
+  ): SafeFuture<FeeHistory> {
+    return SafeFuture.of(
+      Request(
+        "eth_feeHistory",
+        listOf(
+          Numeric.encodeQuantity(BigInteger.valueOf(blockCount.toLong())),
+          newestBlock.toWeb3j(),
```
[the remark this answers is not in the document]
fluentcrafter: This looks relevant

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClient.kt
```diff
@@ -82,9 +87,26 @@ class Web3jEthApiClient(
...
+          rewardPercentiles,
+        ),
+        this.web3jService,
+        EthFeeHistoryBlobExtended::class.java,
+      ).sendAsync(),
+    ).thenCompose {
+      if (it.hasError()) {
+        SafeFuture.failedFuture(Exception(it.error.message))
```
fluentcrafter: Pleas do better error handling. We already have JSONRpc error somewhere

gauravahuja: I don't understand what you are referring to. what exactly do you want me to do here?

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClient.kt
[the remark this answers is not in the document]
gauravahuja: fixed

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClient.kt
fluentcrafter: Well, if you would do your diligence on the code, you would see that the previous  implementation with `requestAsync` uses `rejectOnJsonRpcError` that then throws JsonRpcErrorResponseException. this will provide better error handling and logs when it fails... 
Likewise, here it should wrap the error into a `JsonRpcErrorResponseException`, maybe you can just leverage the existing functions.

fluentcrafter: Btw, Cursor is actually suggesting that. Please pay attention to them

fluentcrafter: Btw, this whole code could be simplified as 

```kotlin
return Request(
      "eth_feeHistory",
      listOf(
        Numeric.encodeQuantity(BigInteger.valueOf(blockCount.toLong())),
        newestBlock.toWeb3j().value,
        rewardPercentiles,
      ),
      this.web3jService,
      EthFeeHistoryBlobExtended::class.java,
    ).requestAsync { it.feeHistory.toLineaDomain() }
```

1. less code
2. more cosistent
3. proper error handlig


All you had to do was looking to what already exixted.

gauravahuja: If you look at the existing usage of `eth_feeHistory` 
1. https://github.com/Consensys/linea-monorepo/blob/main/coordinator/ethereum/gas-pricing/static-cap/src/main/kotlin/net/consensys/linea/ethereum/gaspricing/staticcap/FeeHistoryFetcherImpl.kt#L53
2. https://github.com/Consensys/linea-monorepo/blob/main/coordinator/ethereum/gas-pricing/dynamic-cap/src/main/kotlin/net/consensys/linea/ethereum/gaspricing/dynamiccap/GasPriceCapFeeHistoryFetcherImpl.kt#L44

You will clearly notice both used `sendAsync` that 1 does no error handling and 2 throws inside of an async function instead of wrapping the exception with SafeFuture.failedFuture. So what I did was still better than existing implementations in  1 and 2.

I agree JsonRpc abstractions are your creation and pushing for them to be used is your right but can be done in a more friendly and clear way. Those abstractions and their methods are not always on the top of non-creators minds. Your initial review comment  `We already have JSONRpc error somewhere` was ambigous and not helpful in directing me towards that path.

PR #1989, opened 2026-01-07
[ZkTracer] feat: enable blob transactions

letypequividelespoubelles:
This PR implements issue(s) #

### Checklist

### DIFF
--- a/tracer-constraints/blockdata/blockdata_checker.zkasm
+++ b/tracer-constraints/blockdata/blockdata_checker.zkasm
@@ -41,12 +41,12 @@ return
 
 number:
 ;; check that NUMBER == BLOCK_NUMBER
-    if BLOCK_NUMBER != VALUE_CURRENT exit_fail
+    if BLOCK_NUMBER != VALUE_CURRENT goto exit_fail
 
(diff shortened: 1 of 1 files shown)


tracer-constraints/.gitignore
```diff
@@ -1,9 +1,4 @@
 zkevm.bin
-zkevm_london.bin
-zkevm_paris.bin
-zkevm_shanghai.bin
-zkevm_cancun.bin
-zkevm_prague.bin
```
fluentcrafter, line 6: Why removing this?

AFIK, these files are dynamically generated by go-corsert, right?

letypequividelespoubelles: because we don't produce them anymore, we just support OSAKA fork

fluentcrafter: Ok, le me rephrase: why not ignore all bins?

Binaries that are generated/compiled should not be checkin. 
In other words, what we should have is `*.bin` inside the root `.gitignore` add exceptions when strictly neccesary (e.g `!dictionary.bin` for the blob compressor)

PR #2001, opened 2026-01-07
tracer: ci try to install corset automatically

fluentcrafter:
This PR implements issue(s) #

### Checklist

---

> [!NOTE]
> Shifts go-corset management to Gradle and streamlines CI setup.
> 
> - Adds Gradle tasks: `installGoCorset` (versioned via `corsetVersion` property or `CORSET_VERSION` env) and makes `corsetExists` depend on it; `buildZkevmBins` now depends on `corsetExists`
> - Removes `setup-go-corset` action and references from workflows
> - Introduces `setup-java-and-gradle` composite action and uses it in `setup-tracer-environment`
> - Updates Go in CI from `1.23.5` to `1.24.6` and refines tracer workflow path filters
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7bc94b2f5a036dc8a6575c710311aa57a5aa6862. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tracer/gradle/corset.gradle
+++ b/tracer/gradle/corset.gradle
@@ -64,7 +64,7 @@ tasks.register('installGoCorset') {
     // Check if go-corset exists
     def goCorsetExists = false
     try {
-      def checkCmd = isWindows ? "where go-corset" : "which go-corset22"
+      def checkCmd = isWindows ? "where go-corset" : "which go-corset"
       def (exitCode, out, err) = runCommand(checkCmd, false)
       goCorsetExists = (exitCode == 0)
     } catch (Exception e) {


tracer/gradle/corset.gradle
```diff
@@ -24,13 +24,112 @@ tasks.register('corsetExists') {
...
+        def out = new StringBuilder()
+        def err = new StringBuilder()
+        versionProcess.consumeProcessOutput(out, err)
+        def exitCode = versionProcess.waitFor()
+
+        if (exitCode == 0) {
+          def versionOutput = out.toString().trim()
+          logger.lifecycle("Version output: $versionOutput")
+
+          if (versionOutput.contains(corsetVersion)) {
```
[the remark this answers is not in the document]
fluentcrafter: It's a flaw, however not effect in practice:
- CI - imposibble  because it's a fresh environment without previous installation
- locally, users would go from 1.1.{3, ...4,5...31}, invalidating the verification.

To solve this we would need to enforce `vM.m.p` tags not worthy the reade off

tracer/gradle/corset.gradle
```diff
@@ -24,13 +25,117 @@ tasks.register('corsetExists') {
...
+      if (isWindows) {
+        process = ["cmd", "/c", cmd].execute()
+      } else {
+        process = cmd.execute()
+      }
+      return process
+    }
+
+    // Check if go-corset exists
+    def checkCmd = isWindows ? "where go-corset" : "which go-corset"
```
DavePearce: This seems fine to support windows. I don't know if we've done any testing at all on Windows though. So, could be other problems there.

tracer/gradle/corset.gradle
```diff
@@ -24,13 +25,84 @@ tasks.register('corsetExists') {
...
+}
+
 // Allowing overriding the default zkevm.bin files.
 def ZKEVM_OSAKA_BIN = System.getenv("ZKEVM_OSAKA_BIN") ?: "zkevm_osaka.bin"
 
 // Specify package to use for trace file
 def TRACE_PKG = "net.consensys.linea.zktracer"
 
 tasks.register('buildZkevmBins', Exec) {
+  dependsOn('corsetExists')
```
letypequividelespoubelles, line 105: I remember we had issues with this ... Is it needed again ? Thought @amkCha

letypequividelespoubelles, line 105: I remember we had issues with this ... Is it needed again ? Thought @amkCha

tracer/gradle/corset.gradle
```diff
@@ -24,13 +25,84 @@ tasks.register('corsetExists') {
   }
 }
 
+ext {
+  corsetVersion = project.hasProperty('corsetVersion')
+    ? project.property('corsetVersion')
+    : System.getenv().getOrDefault('CORSET_VERSION', "v1.1.31")
```
letypequividelespoubelles, line 31: It has to follow the actions.yaml for corset install right ? Can't we we have a single parameter ?

fluentcrafter: actions.yaml was removed. This task replaces it and works on CI and locally ;)

tracer/gradle/corset.gradle
fluentcrafter: It's wrapped in try catch, so it's not problematic on this flow.

DavePearce:
>  if the installed version matches the expected version; if not, it re-installs the go-corset at the expected version tag. I thought this would solve that problem, no?

Ok, actually I hadn't spotted that `installRequired` is default `true`. Am looking on my phone. 

That won't work for me --- I will need to fix that as per my requirement.  I can do it separately though --- its fine.

Issue #1999, opened 2026-01-07
Sequencer Plugin - Cache Tx Compression


fluentcrafter:
## Description
Transactions are compressed multiple times in the sequencer, causing a performance bottleneck.

The transaction should be compressed **on TX Pool inclusion once**, not at block building time,  and the compression result should be reused during block building time for profitability verification. 

## Motivation
Improve sequencer tx throughput 

## Tasks
- [ ] Adapt Sequencer Plugin

## Acceptance criteria
- [ ] Transaction compression is performed only once and before block building

PR #2051, opened 2026-01-14
Compressed tx cache

Filter94:
This PR implements issue(s) #1999

### Checklist

---

> [!NOTE]
> Introduces a cached transaction compression layer and centralizes profitability calculations across the sequencer.
> 
> - Adds `TransactionCompressor` interface and `CachingTransactionCompressor` (configurable via `--plugin-linea-compressed-tx-cache-size`) and integrates it into `TransactionProfitabilityCalculator`
> - Converts `getCompressedTxSize` to an instance method; removes ad-hoc/static compression calls across RPC, txpool validators/metrics, and tx selection
> - Creates and shares a single `TransactionProfitabilityCalculator` from `AbstractLineaSharedPrivateOptionsPlugin`, passing it to `LineaEstimateGas`, txpool validation, selection, and metrics
> - Updates CLI/config (`LineaProfitabilityCliOptions`/`Configuration`) and tests/acceptance tests to use the new wiring
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bb8963f3c9ad269e17e7892263467541e5e9092d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java
+++ b/besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java
@@ -30,7 +30,7 @@ public class CachingTransactionCompressor implements TransactionCompressor {
     compressedSizeCache =
         CacheBuilder.newBuilder()
             .maximumSize(cacheSize)
-            .expireAfterWrite(30, TimeUnit.MINUTES)
+            .expireAfterAccess(30, TimeUnit.MINUTES)
             .build();
   }
 


besu-plugins/linea-sequencer/sequencer/src/test/java/net/consensys/linea/sequencer/txpoolvalidation/validators/ProfitabilityValidatorTest.java
```diff
@@ -40,8 +42,8 @@ public class ProfitabilityValidatorTest {
       Address.fromHexString("0x0000000000000000000000000000000000001000");
   public static final Address RECIPIENT =
       Address.fromHexString("0x0000000000000000000000000000000000001001");
-  private static Wei PROFITABLE_GAS_PRICE = Wei.of(11_000_000);
-  private static Wei UNPROFITABLE_GAS_PRICE = Wei.of(200_000);
+  private static final Wei PROFITABLE_GAS_PRICE = Wei.of(11_000_000);
+  private static final Wei UNPROFITABLE_GAS_PRICE = Wei.of(200_000);
```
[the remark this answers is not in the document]
Filter94: Not true

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * A caching wrapper around the transaction compressor that caches compressed sizes based on
+ * transaction hash. This allows compression results to be reused across different parts of the
+ * system, improving performance by avoiding redundant compression operations.
+ */
+@Slf4j
+public class CachingTransactionCompressor implements TransactionCompressor {
+  private final Cache<Hash, Integer> compressedSizeCache =
+      CacheBuilder.newBuilder().maximumSize(10000).expireAfterWrite(30, TimeUnit.MINUTES).build();
```
fluentcrafter: Is the limit of compressed TX cache? if so, it should be configurable, eventually with default to 2000 (~80KB of memory)

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java
```diff
@@ -0,0 +1,57 @@
...
+          });
+    } catch (ExecutionException e) {
+      log.atWarn()
+          .setMessage(
+              "Failed to calculate compressed size for transaction {}, calculating directly")
+          .addArgument(transaction::getHash)
+          .setCause(e)
+          .log();
+      final byte[] bytes = transaction.encoded().toArrayUnsafe();
+      return Compressor.instance.compressedSize(bytes);
```
fluentcrafter: I don't understand how this --  exactly same code -- can work after failing before. Is compression undeterministic? 

DRY: shall we carveout the duplication into a method e.g `int calculateCompressedSize(Transaction tx)`?

tracer/linea-constraints
fluentcrafter, line 1: why these changes?

Filter94: I have no idea

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java
Filter94: > can work after failing before

I don't get this. Where was it failing?

tracer/linea-constraints
Filter94: I fixed this conflict now

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/TransactionCompressor.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+public interface TransactionCompressor {
+  /**
+   * Get the compressed size of a transaction. Implementations may cache the result based on the
+   * transaction hash to improve performance.
+   *
+   * @param transaction the transaction for which to get the compressed size
+   * @return the compressed size of the transaction
+   */
+  int getCompressedSize(Transaction transaction);
```
ahamlat, line 26: Why is this interface needed ? if there is only one implementation, there is no need for an interface.

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java
Filter94: > shall we carveout the duplication into a method e.g int calculateCompressedSize(Transaction tx)?

Yes, good point

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java
Filter94: @fab-10  do you know if it's possible to make it configurable with the same configuration key that is respected by multiple plugins all at once?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java
fluentcrafter: I mean, the catch makes exactly the same calls 
```java
final byte[] bytes = transaction.encoded().toArrayUnsafe();
return Compressor.instance.compressedSize(bytes);
```

How can it fail before and succeed after the exact same code?

Filter94: I see. I actually don't see the reason why would it work the second time. @ahamlat Is there a specific reason it's implemented this way?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java
```diff
@@ -0,0 +1,57 @@
...
+      return compressedSizeCache.get(
+          transaction.getHash(), () -> calculateCompressedSize(transaction));
+    } catch (ExecutionException e) {
+      log.atWarn()
+          .setMessage(
+              "Failed to calculate compressed size for transaction {}, calculating directly")
+          .addArgument(transaction::getHash)
+          .setCause(e)
+          .log();
+      return calculateCompressedSize(transaction);
```
fluentcrafter, line 66: What I was trying to say before is:
how can the expression on [L46](https://github.com/Consensys/linea-monorepo/pull/2051/files#diff-54253ff22d805b4e4630733a93817fa8fe1786fc83603f58b4a421121fe27151R46) fail with `compressedSizeCache.get(
          transaction.getHash(), () -> calculateCompressedSize(transaction))` and pass here? 
          
I see only 2 possible cases to make sense:
- fail on `transaction.getHash`
- fail on `cache.get(...)`

Otherwise, it will fail in the catch as well...

fluentcrafter (dismissed):

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java
fab-10: yes you can add a new option to `LineaProfitabilityCliOptions` for example, that like all the configurations that are managed in `AbstractLineaSharedPrivateOptionsPlugin` are visible to all plugins

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java
Filter94: I think even these 2 won't be caught because `ExecutionException` is being caught

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java
Filter94: I implemented this

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java
```diff
@@ -0,0 +1,69 @@
...
+@Slf4j
+public class CachingTransactionCompressor implements TransactionCompressor {
+  private static final long DEFAULT_CACHE_SIZE = 10000;
+  private final Cache<Hash, Integer> compressedSizeCache;
+
+  public CachingTransactionCompressor(long cacheSize) {
+    compressedSizeCache =
+        CacheBuilder.newBuilder()
+            .maximumSize(cacheSize)
+            .expireAfterWrite(30, TimeUnit.MINUTES)
```
fab-10: not blocking: since a pending tx could stay in the pool for a long time and tried for selection multiple times, could be useful to not expire by time but only by size

fab-10 (approved):
just a minor suggestion

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java
ahamlat: It is a retry mechanism, as the exception is of type ExecutionException (in java.util.concurrent)
```
Exception thrown when attempting to retrieve the result of a task that aborted by throwing an exception. This exception can be inspected using the getCause() method.
```

Filter94: But in which case the retry can succeed if it failed the first time?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java
ahamlat: It is not blocking neither but it would be good to enable the metrics to have an idea of efficiency of the cache (hit ratio vs cache miss ratio). I haven't done it myself but it is a good indicator to have.

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/TransactionCompressor.java
Filter94: It's more a future proof change for the testability. With the interface we can inject a test double for whatever is the test logic we want

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java
Filter94: I thought about this and to me it makes the most sense to change `expireAfterWrite` to `expireAfterAccess`. If a record is being accessed constantly, it's as good as removing the `expireAfterWrite`. If it wasn't accessed in the last 30 minutes, we probably don't need this record in memory either

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java
ahamlat: Because of concurrency related issues, it is an ExecutionException that happens in a multithreaded environment when you call get on the shared datastructure that is storing the cache values.

Filter94: But I thought Guava cache is thread safe. Is it not?

ahamlat: It is thread safe, the ExecutionException is to handle an unexpected expection while getting the value from the cache. This is at the origin what is throwing that Exception
```
 @CanIgnoreReturnValue
    @ParametricNullness
    public static <V> V getUninterruptibly(Future<V> future) throws ExecutionException {
        boolean interrupted = false;

        try {
            while(true) {
                try {
                    Object var2 = future.get();
                    return var2;
                } catch (InterruptedException var6) {
                    interrupted = true;
                }
            }
        } finally {
            if (interrupted) {
                Thread.currentThread().interrupt();
            }

        }
    }
 ```

 It happens when something aborts getting the value from the cache, future.get() is what may trigger that exception.

PR #2094, opened 2026-01-20
Sequencer plugin - Forced transactions implementation

Filter94:
This PR implements issue(s) #2089

### Checklist

---

> [!NOTE]
> Introduces forced transactions end-to-end, including RPC, pool, selector integration, metrics, and tests.
> 
> - Adds RPC endpoints: `linea_sendForcedRawTransaction` and `linea_getForcedTransactionInclusionStatus` with request/response models and concurrency handling
> - Implements `ForcedTransactionPoolService` with `LineaForcedTransactionPool` to queue, process, and finalize FTX inclusion/rejection per block; maps selection results to statuses; persists recent statuses via cache
> - Integrates forced TX processing into transaction selection (before bundles, after liveness); exposes new CLI/config `forced-transaction-config` with `--plugin-linea-forced-tx-status-cache-size`
> - Extends metrics with `SEQUENCER_FORCED_TX` category and gauges (pool size, status cache, inclusion results)
> - Enhances trace line limit handling: cache now stores overflowing module names; selection results include module name; validators/tests updated
> - Registers new plugin `LineaForcedTransactionEndpointsPlugin`; updates shared plugin wiring and acceptance tests to enable endpoints and metrics
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4167ef59de3a99dfaa96b59627d1b9b17a76ade3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
+++ b/besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
@@ -67,23 +67,21 @@ public class LineaForcedTransactionPool
       new EnumMap<>(ForcedTransactionInclusionResult.class);
 
   /**
-   * Tracks tentative outcomes during block building, separated by block number for defensive
-   * programming. Outer map is keyed by block number, inner map by forcedTransactionNumber. Outcomes
-   * are only finalized when onBlockAdded confirms the block. This ensures at most one rejection per
-   * block, even if processForBlock is called multiple times for the same block number.
+   * Tracks tentative rejections during block building, separated by block number. Only rejections
+   * are stored - inclusions are determined from actual block content in onBlockAdded. Outer map is
+   * keyed by block number, inner map by forcedTransactionNumber. Rejections are only finalized when
+   * onBlockAdded confirms the block. This ensures at most one rejection per block, even if
+   * processForBlock is called multiple times for the same block number.
    *
    * <p>Separating by block number prevents race conditions between block building and block import
-   * - we only clear outcomes for blocks that have been confirmed, not for blocks still being built.
+   * - we only clear rejections for blocks that have been confirmed, not for blocks still being
+   * built.
    */
-  private final Map<Long, Map<Long, TentativeOutcome>> tentativeOutcomesByBlock =
+  private final Map<Long, Map<Long, TentativeRejection>> tentativeRejectionsByBlock =
       new ConcurrentHashMap<>();
 
-  private record TentativeOutcome(
-      ForcedTransaction transaction, ForcedTransactionInclusionResult inclusionResult) {
-    boolean isSelected() {
-      return inclusionResult == Included;
-    }
-  }
+  private record TentativeRejection(
+      ForcedTransaction transaction, ForcedTransactionInclusionResult inclusionResult) {}
 
   public LineaForcedTransactionPool() {
     this(DEFAULT_STATUS_CACHE_SIZE, null, null);
@@ -163,19 +161,12 @@ public class LineaForcedTransactionPool
       return;
     }
 
-    // Get or create the outcomes map for this specific block
-    final Map<Long, TentativeOutcome> blockOutcomes =
-        tentativeOutcomesByBlock.computeIfAbsent(blockNumber, k -> new ConcurrentHashMap<>());
-
-    // Clear outcomes for this block only (in case processForBlock is called multiple times)
-    if (!blockOutcomes.isEmpty()) {
-      log.atDebug()
-          .setMessage("action=clear_block_tentative_outcomes blockNumber={} count={}")
-          .addArgument(blockNumber)
-          .addArgument(blockOutcomes::size)
-          .log();
-      blockOutcomes.clear();
-    }
+    // Get or create the rejections map for this specific block.
+    // We do NOT clear between iterations - rejections accumulate/overwrite.
+    // This ensures we don't lose rejection reasons if getPayload returns a block from an earlier iteration.
+    // Note: We only track rejections, not selections. Inclusions are determined from actual block content.
+    final Map<Long, TentativeRejection> blockRejections =
+        tentativeRejectionsByBlock.computeIfAbsent(blockNumber, k -> new ConcurrentHashMap<>());
 
     log.atDebug()
         .setMessage("action=process_forced_txs_start blockNumber={} pendingCount={}")
@@ -203,7 +194,7 @@ public class LineaForcedTransactionPool
           blockTransactionSelectionService.evaluatePendingTransaction(pendingTx);
 
       if (result.selected()) {
-        blockOutcomes.put(ftx.forcedTransactionNumber(), new TentativeOutcome(ftx, Included));
+        // Selection is not stored - inclusion will be determined from actual block content
         log.atInfo()
             .setMessage(
                 "action=forced_tx_tentatively_selected forcedTxNumber={} txHash={} blockNumber={} index={}")
@@ -220,8 +211,8 @@ public class LineaForcedTransactionPool
 
         if (isFinalRejection) {
           // Final rejection - track tentatively, finalized on onBlockAdded
-          blockOutcomes.put(
-              ftx.forcedTransactionNumber(), new TentativeOutcome(ftx, inclusionResult));
+          blockRejections.put(
+              ftx.forcedTransactionNumber(), new TentativeRejection(ftx, inclusionResult));
           log.atInfo()
               .setMessage(
                   "action=forced_tx_tentatively_rejected forcedTxNumber={} txHash={} blockNumber={} inclusionResult={}")
@@ -249,10 +240,10 @@ public class LineaForcedTransactionPool
 
     log.atDebug()
         .setMessage(
-            "action=process_forced_txs_end blockNumber={} remainingPending={} blockOutcomes={}")
+            "action=process_forced_txs_end blockNumber={} remainingPending={} rejections={}")
         .addArgument(blockNumber)
         .addArgument(pendingQueue::size)
-        .addArgument(blockOutcomes::size)
+        .addArgument(blockRejections::size)
         .log();
   }
 
@@ -264,53 +255,48 @@ public class LineaForcedTransactionPool
     final long blockNumber = addedBlockContext.getBlockHeader().getNumber();
     final long blockTimestamp = addedBlockContext.getBlockHeader().getTimestamp();
 
-    // First, flush stale outcomes from older blocks (defensive cleanup for fresh outcomes)
-    tentativeOutcomesByBlock.keySet().removeIf(bn -> bn < blockNumber);
+    tentativeRejectionsByBlock.keySet().removeIf(bn -> bn < blockNumber);
 
-    final Map<Long, TentativeOutcome> blockOutcomes = tentativeOutcomesByBlock.remove(blockNumber);
+    final Map<Long, TentativeRejection> blockRejections =
+        tentativeRejectionsByBlock.remove(blockNumber);
 
-    if (blockOutcomes != null && !blockOutcomes.isEmpty()) {
-      final Set<Hash> blockTxHashes =
-          addedBlockContext.getBlockBody().getTransactions().stream()
-              .map(Transaction::getHash)
-              .collect(Collectors.toSet());
+    final Set<Hash> blockTxHashes =
+        addedBlockContext.getBlockBody().getTransactions().stream()
+            .map(Transaction::getHash)
+            .collect(Collectors.toSet());
 
-      final var iterator = pendingQueue.iterator();
-      while (iterator.hasNext()) {
-        final ForcedTransaction ftx = iterator.next();
-        final TentativeOutcome outcome = blockOutcomes.get(ftx.forcedTransactionNumber());
+    final var iterator = pendingQueue.iterator();
+    while (iterator.hasNext()) {
+      final ForcedTransaction ftx = iterator.next();
 
-        if (outcome == null) {
-          break; // No more tentative outcomes for this block
-        }
+      if (blockTxHashes.contains(ftx.txHash())) {
+        iterator.remove();
+        recordStatus(ftx, blockNumber, blockTimestamp, Included);
+        log.atInfo()
+            .setMessage("action=forced_tx_included forcedTxNumber={} txHash={} blockNumber={}")
+            .addArgument(ftx::forcedTransactionNumber)
+            .addArgument(ftx.txHash()::toHexString)
+            .addArgument(blockNumber)
+            .log();
+      } else {
+        final TentativeRejection rejection =
+            blockRejections != null ? blockRejections.get(ftx.forcedTransactionNumber()) : null;
 
-        if (outcome.isSelected()) {
-          if (!blockTxHashes.contains(ftx.txHash())) {
-            break; // Selection not in block, meaning that tentativeOutcomes don't match the
-            // actually selected block
-          }
+        if (rejection != null) {
           iterator.remove();
-          recordStatus(ftx, blockNumber, blockTimestamp, Included);
-          log.atInfo()
-              .setMessage("action=forced_tx_included forcedTxNumber={} txHash={} blockNumber={}")
-              .addArgument(ftx::forcedTransactionNumber)
-              .addArgument(ftx.txHash()::toHexString)
-              .addArgument(blockNumber)
-              .log();
-        } else {
-          // Rejection - finalize
-          iterator.remove();
-          recordStatus(ftx, blockNumber, blockTimestamp, outcome.inclusionResult());
+          recordStatus(ftx, blockNumber, blockTimestamp, rejection.inclusionResult());
           log.atInfo()
               .setMessage(
                   "action=forced_tx_rejected forcedTxNumber={} txHash={} blockNumber={} inclusionResult={}")
               .addArgument(ftx::forcedTransactionNumber)
               .addArgument(ftx.txHash()::toHexString)
               .addArgument(blockNumber)
-              .addArgument(outcome::inclusionResult)
+              .addArgument(rejection::inclusionResult)
               .log();
-          break; // Rejection is always last
         }
+        // Whether rejected or not, stop processing - subsequent FTXs can't be in this block
+        // (since FTXs are processed in order and this one wasn't included)
+        break;
       }
     }
   }


e2e/src/forced-transactions.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+          value: ethers.parseUnits("1000", "wei"),
+          maxPriorityFeePerGas: ethers.parseEther("0.000000001"), // 1 Gwei
+          maxFeePerGas: ethers.parseEther("0.00000001"), // 10 Gwei
+          nonce: senderNonce++,
+        };
+        txs.push({ transaction: await getRawTransactionHex(txRequest, senderWallet) });
+        txHashes.push(await getTransactionHash(txRequest, senderWallet));
+      }
+
+      const resultHashes = await forcedTxClient.lineaSendForcedRawTransaction(txs);
```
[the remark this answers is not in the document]
Filter94: It's there in txs elements

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/rpc/methods/LineaSendForcedRawTransaction.java
```diff
@@ -0,0 +1,211 @@
...
+    @Override
+    public int getCode() {
+      return INVALID_PARAMS_ERROR_CODE;
+    }
+
+    @Override
+    public String getMessage() {
+      return errMessage;
+    }
+  }
```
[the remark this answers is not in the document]
Filter94: It's used

besu-plugins/linea-sequencer/acceptance-tests/src/test/kotlin/linea/plugin/acc/test/rpc/linea/ForcedTransactionTest.kt
```diff
@@ -0,0 +1,395 @@
...
+      .execute(minerNode.nodeRequests())
+    val invalidStatus = GetForcedTransactionInclusionStatusRequest(invalidTxHash)
+      .execute(minerNode.nodeRequests())
+
+    assertThat(validStatus.result).isNotNull
+    assertThat(invalidStatus.result).isNotNull
+    assertThat(validStatus.result!!.inclusionResult).isEqualTo("INCLUDED")
+
+    // The failing tx must be in a different (later) block than the successful one
+    val validBlockNumber = java.lang.Long.decode(validStatus.result!!.blockNumber)
```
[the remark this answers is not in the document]
Filter94: That's too verbose for a test

besu-plugins/linea-sequencer/acceptance-tests/src/test/kotlin/linea/plugin/acc/test/rpc/linea/ForcedTransactionTest.kt
```diff
@@ -0,0 +1,395 @@
...
+    val invalidBlockNumber = java.lang.Long.decode(invalidStatus.result!!.blockNumber)
```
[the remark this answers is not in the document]
Filter94: That's too verbose for a test

besu-plugins/linea-sequencer/acceptance-tests/src/test/kotlin/linea/plugin/acc/test/rpc/linea/ForcedTransactionTest.kt
```diff
@@ -0,0 +1,395 @@
...
+          assertThat(statusResponse.result).isNotNull
+          assertThat(statusResponse.result!!.inclusionResult).isEqualTo("BAD_NONCE")
+        }
+    }
+
+    // Get all block numbers
+    val blockNumbers = txHashes.map { txHash ->
+      val status = GetForcedTransactionInclusionStatusRequest(txHash)
+        .execute(minerNode.nodeRequests())
+      java.lang.Long.decode(status.result!!.blockNumber)
```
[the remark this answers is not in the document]
Filter94: That's too verbose for a test

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
```diff
@@ -0,0 +1,288 @@
...
+
+    // Skip if we already had a failure in this block (only one invalidity proof per block)
+    if (failureBlockNumber == blockNumber) {
+      log.atDebug()
+          .setMessage("action=skip_forced_txs_already_failed blockNumber={} failureBlockNumber={}")
+          .addArgument(blockNumber)
+          .addArgument(failureBlockNumber)
+          .log();
+      return;
+    }
```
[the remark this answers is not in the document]
Filter94: @fab-10 Can you please advise? I don't have enough experience with Besu plugins to clearly know if this is valid concern or not

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+  /**
+   * Maps a Besu TransactionSelectionResult to a ForcedTransactionInclusionResult.
+   *
+   * <p>First checks against known Linea-specific result constants using equality, then falls back
+   * to string matching for standard Besu results. Returns OTHER for unrecognized results, which
+   * triggers a retry in the next block.
+   */
+  private ForcedTransactionInclusionResult mapToInclusionResult(
```
Filter94, line 349: @fab-10 @daniellehrner Please advise if we can make this part more sturdy

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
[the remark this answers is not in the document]
fab-10: Since block creation iterations are sequential, and this is the first thing that is done, before evaluating any txs, I do not see the race condition here. 
The only way there could be 2 concurrent block building threads is when the previous iteration went in timeout, and it is still completing, and the next iteration starts, but as said even in this case the two threads could not overlap on this check.

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
```diff
@@ -0,0 +1,290 @@
...
+
+      final PendingTransaction pendingTx =
+          new org.hyperledger.besu.ethereum.eth.transactions.PendingTransaction.Local(
+              ftx.transaction());
+
+      final TransactionSelectionResult result =
+          blockTransactionSelectionService.evaluatePendingTransaction(pendingTx);
+
+      if (result.selected()) {
+        pendingQueue.pollFirst();
```
fab-10: Would it be safer to remove the tx from the queue only when the block is confirmed? just in case something happens that could invalidate the block after the selection

You can listen to `BesuEvents.BlockAddedListener::onBlockAdded` event that is triggered when the block is validated and imported, and get the list of confirmed txs

besu-plugins/linea-sequencer/acceptance-tests/src/test/kotlin/linea/plugin/acc/test/rpc/linea/ForcedTransactionTest.kt
```diff
@@ -0,0 +1,395 @@
...
+
+    assertThat(sendResponse.hasError()).isFalse()
+    assertThat(sendResponse.result).hasSize(3)
+
+    // Wait for all to be mined
+    minerNode.verify(eth.expectSuccessfulTransactionReceipt(txHash1))
+    minerNode.verify(eth.expectSuccessfulTransactionReceipt(txHash2))
+    minerNode.verify(eth.expectSuccessfulTransactionReceipt(txHash3))
+
+    // Verify all have INCLUDED status
```
fab-10: Does it make sense to verify that they are included in the same block?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorFactory.java
```diff
@@ -123,7 +127,10 @@ public void selectPendingTransactions(
       final ProcessableBlockHeader pendingBlockHeader,
       final List<? extends PendingTransaction> candidatePendingTransactions) {
     try {
-      // check and send liveness bundle if any
+      forcedTransactionPoolService.processForBlock(
+          pendingBlockHeader.getNumber(), pendingBlockHeader.getTimestamp(), bts);
+      commit(bts);
```
fab-10: Since force txs are independent and not part of a bundle, the commit or rollback must be done after the evaluation of each one. 

Doing only the commit here does not work in case there is a not selected tx, since if there is no rollback the world state will be inconsistent for the next txs.

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
fab-10: indeed this is not optimal, I will check if there is a better approach

besu-plugins/linea-sequencer/acceptance-tests/src/test/kotlin/linea/plugin/acc/test/rpc/linea/ForcedTransactionTest.kt
Filter94: I believe it can't be guaranteed because FTX request may come at the end of the block building attempt. In the same block or not, both outcomes are valid

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
Filter94: I've added `tentativeOutcomes`, kind of a staging layer for the cache updates. The queue and the cache is only updated on new blocks now. Let me know if it's an overcomplication or a proper solution

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/selectors/LineaTransactionSelector.java
```diff
@@ -117,24 +122,22 @@ private List<PluginTransactionSelector> createTransactionSelectors(
...
-                transactionProfitabilityCalculator),
-            new BundleConstraintTransactionSelector(),
-            new MaxBundleGasPerBlockTransactionSelector(
-                selectorsStateManager, txSelectorConfiguration.maxBundleGasPerBlock()),
-            traceLineLimitTransactionSelector,
-            new TransactionEventSelector(deniedEvents, deniedBundleEvents));
-
-    return selectors;
+    return List.of(
+        new AllowedAddressTransactionSelector(new DeniedAddressValidator(deniedAddresses)),
```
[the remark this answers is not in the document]
Filter94: This is validated on the contract, so I don't think we need to validate this during transaction selection

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorFactory.java
Filter94: I changed it, so commits and reverts happen in LineaForcedTransactionPool

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/selectors/AllowedAddressTransactionSelector.java
```diff
@@ -0,0 +1,61 @@
...
+      log.atTrace()
+          .setMessage("action=reject_filtered_address txHash={} reason={}")
+          .addArgument(transaction::getHash)
+          .addArgument(validationError::get)
+          .log();
+      return TX_FILTERED_ADDRESSES;
+    }
+
+    return SELECTED;
+  }
```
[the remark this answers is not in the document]
Filter94: Precompile check is done in the contract, so it's not needed here

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/rpc/methods/LineaGetForcedTransactionInclusionStatus.java
```diff
@@ -0,0 +1,191 @@
...
+      throw new PluginRpcEndpointException(new GetInclusionStatusError(e.getMessage()));
+    } catch (final Exception e) {
+      log.atError()
+          .setMessage("action=get_inclusion_status_error logId={} error={}")
+          .addArgument(logId)
+          .addArgument(e::getMessage)
+          .setCause(e)
+          .log();
+      throw new PluginRpcEndpointException(
+          new GetInclusionStatusError("Internal error: " + e.getMessage()));
```
jonesho: In this case, `GetInclusionStatusError` with code `INVALID_PARAMS_ERROR_CODE` doesn't sounds aligning with the internal error, should we consider using the error code `-32603` sth like `INTERNAL_ERROR(-32603, "Internal error")` from `hyperledger/besu`?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/rpc/methods/LineaSendForcedRawTransaction.java
```diff
@@ -0,0 +1,307 @@
...
+          responses.add(
+              ForcedTransactionResponse.error(
+                  param.forcedTransactionNumber,
+                  "Failed to decode transaction: " + e.getMessage()));
+          break;
+        }
+
+        final long deadline;
+        try {
+          deadline = parseDeadline(param.deadline);
```
jonesho: I assume `deadline` is a block number but in the code I was confused whether its a timestamp or block number, can we make it (e.g. the naming) more explicit?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/rpc/methods/LineaSendForcedRawTransaction.java
```diff
@@ -0,0 +1,307 @@
...
+        } catch (final Exception e) {
+          responses.add(
+              ForcedTransactionResponse.error(param.forcedTransactionNumber, e.getMessage()));
+          break;
+        }
+
+        forcedTransactions.add(
+            new ForcedTransaction(param.forcedTransactionNumber, tx.getHash(), tx, deadline));
```
jonesho: How do we handle multiple ftx params with identical `forcedTransactionNumber` in the request?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/rpc/methods/LineaSendForcedRawTransaction.java
```diff
@@ -0,0 +1,307 @@
...
+  public String getName() {
+    return "sendForcedRawTransaction";
+  }
+
+  public List<ForcedTransactionResponse> execute(final PluginRpcRequest request) {
+    final int logId = log.isDebugEnabled() ? LOG_SEQUENCE.incrementAndGet() : -1;
+
+    if (!requestLock.tryLock()) {
+      throw new PluginRpcEndpointException(
+          new SendForcedRawTransactionError("Another request is already being processed"));
```
jonesho: Same here, `SendForcedRawTransactionError` with code `INVALID_PARAMS_ERROR_CODE` doesn't sounds aligning with the actual "request was in lock" error, should we consider using the error code `-32603` from hyperledger/besu?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
```diff
@@ -0,0 +1,365 @@
...
+   * only finalized when onBlockAdded confirms the block. This ensures at most one rejection per
+   * block, even if processForBlock is called multiple times for the same block number.
+   */
+  private final Map<Long, TentativeOutcome> tentativeOutcomes = new ConcurrentHashMap<>();
+
+  private record TentativeOutcome(
+      long blockNumber,
+      ForcedTransaction transaction,
+      ForcedTransactionInclusionResult inclusionResult) {
+    boolean isSelection() {
```
jonesho: nit: can we call it in `isSelected()` or `isInSelection()`?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
```diff
@@ -0,0 +1,365 @@
...
+              .addArgument(blockNumber)
+              .addArgument(index)
+              .addArgument(inclusionResult)
+              .log();
+        }
+        blockTransactionSelectionService.rollback();
+        break; // Stop processing - only one invalidity proof per block
+      }
+    }
+
```
jonesho, line 240: @Filter94 Want to confirm my understanding on these codes, so a block can have either:
- multiple included forced transactions
- one rejected forced transaction

Filter94: That's correct

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/rpc/methods/LineaSendForcedRawTransaction.java
Filter94: Good point. We can and we should validate the uniqueness of forcedTransactionNumber at least per request

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
```diff
@@ -0,0 +1,365 @@
...
+      return;
+    }
+
+    if (!tentativeOutcomes.isEmpty()) {
+      log.atDebug()
+          .setMessage("action=clear_stale_tentative_outcomes count={}")
+          .addArgument(tentativeOutcomes::size)
+          .log();
+      tentativeOutcomes.clear();
+    }
```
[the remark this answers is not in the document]
Filter94: Looking at LineaLimitedBundlePool and assuming it doesn't have concurrency issues, concurrency there is handled only relative to pool's internal logic of flushing the state to disk. `blockIndex` remains unsynchronized relative to block building's cycle. So I think the same logic should apply to LineaForcedTransactionPool's internal state.

@fab-10 Can you please tell if my reasoning is correct? Also it would be great for me to know how exactly block selection / onBlockAdded race condition is handled in plugins?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
Filter94: @fab-10 Do you have an update on this?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
```diff
@@ -0,0 +1,365 @@
...
+              "action=evaluate_forced_tx forcedTxNumber={} txHash={} index={} blockNumber={}")
+          .addArgument(ftx::forcedTransactionNumber)
+          .addArgument(ftx.txHash()::toHexString)
+          .addArgument(index)
+          .addArgument(blockNumber)
+          .log();
+
+      final PendingTransaction pendingTx =
+          new org.hyperledger.besu.ethereum.eth.transactions.PendingTransaction.Local(
+              ftx.transaction());
```
fab-10, line 191: probably FT should have priority, to skip min gas price filter
```suggestion
      final PendingTransaction pendingTx =
          new org.hyperledger.besu.ethereum.eth.transactions.PendingTransaction.Local.Priority(
              ftx.transaction());
```

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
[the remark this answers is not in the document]
fab-10: Now `BesuEvents::onBlockAdded` is called synchronously during block import, so it completes, before the FcU returns (unless the method itself has some asynchronicity internally, to avoid adding latency to FcU, but it is not the case here, since it is fast), and the new payload building process, only starts after the block import is completed, thus there should be no risk of overlapping if Maru FcU+attrs are sequentials and not duplicated.

BTW, for defensive programming, you could keep the `tentativeOutcomes` separated per block, and clear only onBlockAdded the ones with blockNumber less that the current one, so you are not clearing in `processForBlock`

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
fab-10: not yet

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
Filter94: Yes, I just changed that and covered it with an acceptance test

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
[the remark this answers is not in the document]
Filter94: Great suggestion. I'll implement it, just to be on the safe side

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
```diff
@@ -239,52 +246,53 @@ public void processForBlock(
...
+            "action=process_forced_txs_end blockNumber={} remainingPending={} blockOutcomes={}")
         .addArgument(blockNumber)
         .addArgument(pendingQueue::size)
-        .addArgument(tentativeOutcomes::size)
+        .addArgument(blockOutcomes::size)
         .log();
   }
 
   @Override
   public void onBlockAdded(final AddedBlockContext addedBlockContext) {
```
fab-10, line 251: you should filter on `AddedBlockContext::getEventType` since you should be only interested in `HEAD_ADVANCED` and maybe `CHAIN_REORG`, but for sure you should nothing for `STORED_ONLY` and `FORK`

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
```diff
@@ -239,52 +246,53 @@ public void processForBlock(
...
-            .log();
-      } else {
-        // Rejection - finalize if block number matches
-        if (outcome.blockNumber() == blockNumber) {
+
+        if (outcome.isSelected()) {
+          if (!blockTxHashes.contains(ftx.txHash())) {
+            break; // Selection not in block, meaning that tentativeOutcomes don't match the
+            // actually selected block
+          }
```
fab-10: this should not happen right, so probably it is better to log an error

Filter94: What happens if there's a time gap between getPayload and FCU that is long enough so that another block selection attempt starts in between?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
Filter94: `CHAIN_REORG` shouldn't be possible. I think in case it happens, forced transaction pool could be corrupted, so I don't want to handle that case

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
fab-10: you can do something like that, so it is checked at compilation time

```
    if(result.maybeInvalidReason().isPresent()) {
      final String invalidReason = result.maybeInvalidReason().get();

      if(TransactionInvalidReason.NONCE_TOO_HIGH.name().equals(invalidReason) ||
      TransactionInvalidReason.NONCE_TOO_LOW.name().equals(invalidReason)
      ) {
        return BadNonce;
      }
    }
```

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
fab-10: I think we need to create a doc to outline all the possible (even if remote) combinations of calls to avoid missing something.

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
Filter94: I improved the error matching in mapToInclusionResult

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
Filter94: So is it a valid concern? If yes, `tentativeOutcomesByBlock` doesn't work at all. And in that case we probably need proper transaction managements and atomic updates of the `statusCache`

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
```diff
@@ -0,0 +1,401 @@
...
+      }
+
+      if (invalidReason.equals("PRECOMPILE_RIPEMD_BLOCKS")
+          || invalidReason.equals("PRECOMPILE_BLAKE_EFFECTIVE_CALLS")) {
+        return BadPrecompile;
+      }
+
+      if (invalidReason.equals("BLOCK_L2_L1_LOGS")) {
+        return TooManyLogs;
+      }
```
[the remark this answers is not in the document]
Filter94: We don't expect it to overflow anything else, because of its gas limit of 300k

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
fab-10: I mean, that I am not sure that for some _exotic_ combination of events, that would not happen unless we describe all of them. 
But now that I am thinking about what could happen, it comes to mind, that we have multiple iterations for a single block creation, so since there are external factors (not related to internal state) that could affect the inclusion of a tx, like the denylist or phylax, then it is possible that for the same block number a tx is first selected and then not, and here you do not know if the actual confirmed block is the one with or without the tx, so I suggest this approach:
* for each block number, keep track of the incremental iteration, and save the result of the FTX indexed by block number + iteration. You do not clear anything on the `processForBlock`
* when you receive the confirmed block, 
  * if the FTX is present then you should have at least one entry, in the block number + iteration map, saying so
  * if the FTX is not present then you should have at least one non selected entry in the block number + iteration map (what if there are 2 entries with different reasons?)

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
```diff
@@ -358,25 +361,41 @@ private ForcedTransactionInclusionResult mapToInclusionResult(
...
-    if (resultString.contains("UPFRONT_COST_EXCEEDS_BALANCE")) {
-      return BadBalance;
+    if (result.maybeInvalidReason().isPresent()) {
+      final String invalidReason = result.maybeInvalidReason().get();
+
+      final Set<String> nonceErrors = buildReasonsStringSet(
+        TransactionInvalidReason.NONCE_TOO_HIGH,
+        TransactionInvalidReason.NONCE_TOO_LOW,
+        TransactionInvalidReason.NONCE_OVERFLOW,
+        TransactionInvalidReason.NONCE_TOO_FAR_IN_FUTURE_FOR_SENDER);
```
fab-10: nit1: you can pre-compute the set since it is static

nit2: `TransactionInvalidReason.NONCE_TOO_FAR_IN_FUTURE_FOR_SENDER` is only applicable to txpool inclusion and cannot be triggered during block production

besu-plugins/linea-sequencer/sequencer/src/test/java/net/consensys/linea/sequencer/txselection/selectors/TraceLineLimitTransactionSelectorTest.java
```diff
@@ -215,14 +213,14 @@ private void verifyTransactionSelection(
...
+      final String expectedPreProcessingResultPrefix,
+      final String expectedPostProcessingResultPrefix) {
     var preProcessingResult = selector.evaluateTransactionPreProcessing(evaluationContext);
-    assertThat(preProcessingResult).isEqualTo(expectedPreProcessingResult);
+    assertThat(preProcessingResult.toString()).startsWith(expectedPreProcessingResultPrefix);
     if (preProcessingResult.equals(SELECTED)) {
       var postProcessingResult =
           selector.evaluateTransactionPostProcessing(evaluationContext, processingResult);
-      assertThat(postProcessingResult).isEqualTo(expectedPostProcessingResult);
+      assertThat(postProcessingResult.toString()).startsWith(expectedPostProcessingResultPrefix);
```
Filter94, line 223: @fab-10 If we had access to the `TX_MODULE_LINE_COUNT_OVERFLOW` status in the `TransactionSelectionResult`, we wouldn't have to rely on the strings here. [This](https://github.com/Consensys/linea-monorepo/pull/2094#discussion_r2735836472) place is also problematic

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/selectors/TraceLineLimitTransactionSelector.java
```diff
@@ -108,14 +109,16 @@ public TraceLineLimitTransactionSelector(
...
       log.atTrace()
           .setMessage(
-              "Transaction {} was already identified to go over line count limit, dropping it")
-          .addArgument(evaluationContext.getPendingTransaction().getTransaction()::getHash)
+              "Transaction {} was already identified to go over line count limit for module {}, dropping it")
+          .addArgument(txHash)
+          .addArgument(cachedModule::get)
           .log();
-      return TX_MODULE_LINE_COUNT_OVERFLOW_CACHED;
+      return txModuleLineCountOverflowCached(cachedModule.get());
```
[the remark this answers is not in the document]
Filter94: I've implemented workaround that relies on `startsWith` instead of the strict equality

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/forced/LineaForcedTransactionPool.java
```diff
@@ -0,0 +1,402 @@
...
+      }
+
+      if (balanceErrors.contains(invalidReason)) {
+        return BadBalance;
+      }
+
+      if (invalidReason.equals("PRECOMPILE_RIPEMD_BLOCKS")
+          || invalidReason.equals("PRECOMPILE_BLAKE_EFFECTIVE_CALLS")) {
+        return BadPrecompile;
+      }
```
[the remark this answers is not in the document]
Filter94: I don't think we care about non-limitless mode anymore

besu-plugins/linea-sequencer/sequencer/src/test/java/net/consensys/linea/sequencer/txselection/selectors/TraceLineLimitTransactionSelectorTest.java
fab-10: So it seems that we need a way to add custom attributes to `TransactionSelectionResult` right?

Filter94: I think `InvalidReason` somewhat works, but we can't access `TransactionSelectionResult`'s `Status` and I think that's the main issue here

With `Status` exposed, we could typecheck and compare it as `LineaStatus` enum, for example. With a new custom attributes, I believe we'd also have to do the runtime type check

fab-10: I will look into this

Issue #1932, opened 2025-12-17
Refactor ConflationApp for backtesting


gauravahuja:
- Backtesting should be isolated from the actual conflation 
  - Logs, metrics and databases should be isolated.
- In order for rolling blob shnarf calculator to fetch the parent blob data, use genesis shnarf.
- Add the ability to stop ConflationApp when all proofs between [start block number, end block number] are created.

ProofGeneratingConflationHandlerImpl
- Create a similar class for backtesting purposes that handles Conflated Batches and simply creates proof requests and finishes the Batch instead of waiting for proof.

Issue #1740, opened 2025-11-17
Conflation Backtesting Endpoint


fluentcrafter:
## Description
The coordinator should implement an endpoint to reconflate and regenerate prover and tracer requests for a given block range. 

This would allow the prover and arithmatization team to perform backtesting of new tracer/prover versions.

## Motivation
- allow back testing of conflation
- generate prover tracer requests quickly and fast

## Tasks
- [ ] implement JSON-RPC endpoint that reconflates a given block range, by regenerating prover requests files

## Acceptance criteria
- [ ] List tests that need to be implemented or checks to consider this task completed. 

## Risks
- [ ] If called in a production environment, using the same prod Shomei and Besu Tracer, it can overwhelm the nodes. It's a known risk, and the endpoint should be used carefully depending on the test scope.

### Request conflation_createProverRequests
```json
{
    "jsonrpc": "2.0",
    "id": "1",
    "method": "conflation_createProverRequests",
    "params": [{
        "startBlockNumber": 234,
        "endBlockNumber": 999,
        "parentShnarf": "0x00...",
        "l1MessageNumber": 12312,
        "l1MessageRollingHash": "0x0...",
        "compressorVersion": "V3",
        "batchFixedSize (optional)": "when defined, batches have fixed size - allows skipping trace counting, which is a bottleneck", 
        "tracesLimits (optional) - overrides coordinator's current traces limits": {""},
        "aggregationProofLimit": 123,
        "aggregationBlobsLimit (optional)": 6,
        "proverRequestsDestinationDir": "/path/to/prover/ - cannot be same as current prover, otherwise will with prod prover", 
        "conflatedTracesDestinationDir": "(trace: override the destination path)",
        "ethApi(optional)": {
            "endpoint": "http://besu-full:8545/",
            "polling-interval": "PT1S",
        },
        "tracesApi": {
            "endpoint": "http://traces-api:8545/",
            "version": "V1.2.3",
            "requests-limit-per-endpoint": 10
        },
        "shomeiApi": {
            "endpoint": "http://traces-api:8545/",
            "version": "V1.2.3",
            "requests-limit-per-endpoint": 10
        }
    }]
}
```

It should generate all pover requests (execution, compression, aggregation) into the specified `proverRequestsDestinationDir/{execution, compression,aggregation}`. 
`conflatedTracesDestinationDir` shall be sent as parameter to `linea_generateConflatedTracesToFile` NOTE: tracer needs development to support this feature.

[...]

PR #2138, opened 2026-01-25
[Coordinator] Add conflation backtesting app

gauravahuja:
This PR implements issue(s) #1932 

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Introduces a new backtesting runtime that spins up separate conflation/prover/traces/shomei clients per job and adds new lifecycle hooks in core coordination code, so regressions could affect proving/conflation flow and resource usage. No auth or persistence schema changes, but concurrency and external-client configuration paths increase operational risk.
> 
> **Overview**
> Adds a **conflation backtesting** capability that can be triggered via JSON-RPC, running conflation/proof generation for a specified block range using dedicated Traces/Shomei endpoints and per-job prover file directories.
> 
> Extends conflation configuration with `backtestingDirectory`, adds richer request DTOs for `conflation_createProverRequests`, and replaces the stub `ConflationBacktestingService` with a polling service that tracks in-flight/completed jobs and stops per-job apps.
> 
> Refactors several coordinator/prover components to better support backtesting (injectable loggers, per-job prover client logging, `BlobCompressionProofCoordinator` callback on proof-request creation, and minor simplification of genesis shnarf wiring), and adds focused tests for DTO parsing and prover path rewriting.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6f920e0fde9b00488238d1f44f151029818c8b8e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/CoordinatorApp.kt
+++ b/coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/CoordinatorApp.kt
@@ -184,8 +184,8 @@ class CoordinatorApp(private val configs: CoordinatorConfig) {
     return runCatching {
       SafeFuture.allOf(
         l1App.stop(),
-        conflationBacktestingService.stop(),
         api.stop(),
+        conflationBacktestingService.stop(),
       ).thenApply {
         LoadBalancingJsonRpcClient.stop()
       }.thenCompose {
--- a/coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
+++ b/coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
@@ -59,12 +59,15 @@ class ConflationBacktestingApp(
 ) : LongRunningService {
 
   init {
+    require(mainCoordinatorConfig.conflation.backtestingDirectory != null) {
+      "Backtesting requests parent directory must be set in conflation config"
+    }
     mainCoordinatorConfig.traces.common?.endpoints?.contains(conflationBacktestingAppConfig.tracesApi.endpoint)
-      ?.let { require(!it) { "Cannot use prod traces endpoint for backtesting" } }
+      ?.let { require(!it) { "Cannot use same traces endpoint for backtesting and main conflation" } }
     mainCoordinatorConfig.traces.counters?.endpoints?.contains(conflationBacktestingAppConfig.tracesApi.endpoint)
-      ?.let { require(!it) { "Cannot use prod traces endpoint for backtesting" } }
+      ?.let { require(!it) { "Cannot use same traces endpoint for backtesting and main conflation" } }
     mainCoordinatorConfig.traces.conflation?.endpoints?.contains(conflationBacktestingAppConfig.tracesApi.endpoint)
-      ?.let { require(!it) { "Cannot use prod traces endpoint for backtesting" } }
+      ?.let { require(!it) { "Cannot use same traces endpoint for backtesting and main conflation" } }
   }
 
   private val log = LogManager.getLogger("conflation_backtesting.job_${conflationBacktestingAppConfig.jobId()}")
@@ -101,21 +104,20 @@ class ConflationBacktestingApp(
       ),
       proofAggregation = mainCoordinatorConfig.conflation.proofAggregation.copy(
         targetEndBlocks = (mainCoordinatorConfig.conflation.proofAggregation.targetEndBlocks ?: emptyList())
-          .toMutableList().let {
-            it.add(conflationBacktestingAppConfig.endBlockNumber)
-            it
-          },
+          .toMutableList().also { it.add(conflationBacktestingAppConfig.endBlockNumber) },
       ),
     ),
     proversConfig = mainCoordinatorConfig.proversConfig.copy(
       proverA = getUpdatedProverConfig(
-        mainCoordinatorConfig.proversConfig.proverA,
-        "conflation-backtesting-${conflationBacktestingAppConfig.jobId()}",
+        proverConfig = mainCoordinatorConfig.proversConfig.proverA,
+        backtestingDirectory = mainCoordinatorConfig.conflation.backtestingDirectory!!,
+        conflationBacktestingJobId = conflationBacktestingAppConfig.jobId(),
       ),
       proverB = mainCoordinatorConfig.proversConfig.proverB?.let { proverB ->
         getUpdatedProverConfig(
-          proverB,
-          "conflation-backtesting-${conflationBacktestingAppConfig.jobId()}",
+          proverConfig = proverB,
+          backtestingDirectory = mainCoordinatorConfig.conflation.backtestingDirectory,
+          conflationBacktestingJobId = conflationBacktestingAppConfig.jobId(),
         )
       },
     ),
@@ -362,25 +364,36 @@ class ConflationBacktestingApp(
   }
 
   companion object {
-    fun getUpdatedPath(path: Path, suffix: String): Path {
-      val lastDirectory = path.getName(path.nameCount - 1)
-      val updatedDirectory = "$lastDirectory-$suffix"
-      return path.parent.resolve(updatedDirectory)
-    }
-
-    fun getUpdatedProverConfig(proverConfig: ProverConfig, suffix: String): ProverConfig {
+    fun getUpdatedProverConfig(
+      proverConfig: ProverConfig,
+      backtestingDirectory: Path,
+      conflationBacktestingJobId: String,
+    ): ProverConfig {
+      val jobDirectory = backtestingDirectory.resolve(conflationBacktestingJobId)
       return proverConfig.copy(
         execution = proverConfig.execution.copy(
-          requestsDirectory = getUpdatedPath(proverConfig.execution.requestsDirectory, suffix),
-          responsesDirectory = getUpdatedPath(proverConfig.execution.responsesDirectory, suffix),
+          requestsDirectory = jobDirectory
+            .resolve("execution")
+            .resolve("requests"),
+          responsesDirectory = jobDirectory
+            .resolve("execution")
+            .resolve("responses"),
         ),
         blobCompression = proverConfig.blobCompression.copy(
-          requestsDirectory = getUpdatedPath(proverConfig.blobCompression.requestsDirectory, suffix),
-          responsesDirectory = getUpdatedPath(proverConfig.blobCompression.responsesDirectory, suffix),
+          requestsDirectory = jobDirectory
+            .resolve("compression")
+            .resolve("requests"),
+          responsesDirectory = jobDirectory
+            .resolve("compression")
+            .resolve("responses"),
         ),
         proofAggregation = proverConfig.proofAggregation.copy(
-          requestsDirectory = getUpdatedPath(proverConfig.proofAggregation.requestsDirectory, suffix),
-          responsesDirectory = getUpdatedPath(proverConfig.proofAggregation.responsesDirectory, suffix),
+          requestsDirectory = jobDirectory
+            .resolve("aggregation")
+            .resolve("requests"),
+          responsesDirectory = jobDirectory
+            .resolve("aggregation")
+            .resolve("responses"),
         ),
       )
     }
--- a/coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingService.kt
+++ b/coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingService.kt
@@ -40,7 +40,7 @@ class ConflationBacktestingService(
   fun submitConflationBacktestingJob(conflationBacktestingConfig: ConflationBacktestingConfig): String {
     val jobId = conflationBacktestingConfig.jobId()
     if (completedJobs.contains(jobId)) {
-      throw IllegalArgumentException("Conflation backtesting job with id $jobId already completed")
+      throw IllegalArgumentException("Given Conflation backtesting Request with jobId=$jobId is already completed")
     }
     val app = ConflationBacktestingApp(
       vertx = vertx,
@@ -50,7 +50,9 @@ class ConflationBacktestingService(
       metricsFacade = metricsFacade,
     )
     if (conflationBackTestingApps.putIfAbsent(jobId, app) != null) {
-      throw IllegalArgumentException("Conflation backtesting job with id $jobId already exists")
+      throw IllegalArgumentException(
+        "Given Conflation backtesting Request with jobId=$jobId is already being processed",
+      )
     }
     app.start().thenPeek {
       log.info("Conflation backtesting job started: jobId={}", jobId)
--- a/coordinator/clients/prover-client/file-based-client/src/main/kotlin/net/consensys/zkevm/coordinator/clients/prover/ProverClientFactory.kt
+++ b/coordinator/clients/prover-client/file-based-client/src/main/kotlin/net/consensys/zkevm/coordinator/clients/prover/ProverClientFactory.kt
@@ -45,72 +45,53 @@ class ProverClientFactory(
   fun executionProverClient(
     tracesVersion: String,
     stateManagerVersion: String,
-    log: Logger? = null,
+    log: Logger = FileBasedExecutionProverClientV2.LOG,
   ): ExecutionProverClientV2 {
     return createClient(
       proverAConfig = config.proverA.execution,
       proverBConfig = config.proverB?.execution,
       switchBlockNumberInclusive = config.switchBlockNumberInclusive,
     ) { proverConfig ->
-      if (log == null) {
-        FileBasedExecutionProverClientV2(
-          config = proverConfig,
-          vertx = vertx,
-          tracesVersion = tracesVersion,
-          stateManagerVersion = stateManagerVersion,
-        )
-      } else {
-        FileBasedExecutionProverClientV2(
-          config = proverConfig,
-          vertx = vertx,
-          tracesVersion = tracesVersion,
-          stateManagerVersion = stateManagerVersion,
-          log = log,
-        )
-      }.also { executionWaitingResponsesMetric.addReporter(it) }
+      FileBasedExecutionProverClientV2(
+        config = proverConfig,
+        vertx = vertx,
+        tracesVersion = tracesVersion,
+        stateManagerVersion = stateManagerVersion,
+        log = log,
+      ).also { executionWaitingResponsesMetric.addReporter(it) }
     }
   }
 
-  fun blobCompressionProverClient(log: Logger? = null): BlobCompressionProverClientV2 {
+  fun blobCompressionProverClient(
+    log: Logger = FileBasedBlobCompressionProverClientV2.LOG,
+  ): BlobCompressionProverClientV2 {
     return createClient(
       proverAConfig = config.proverA.blobCompression,
       proverBConfig = config.proverB?.blobCompression,
       switchBlockNumberInclusive = config.switchBlockNumberInclusive,
     ) { proverConfig ->
-      if (log == null) {
-        FileBasedBlobCompressionProverClientV2(
-          config = proverConfig,
-          vertx = vertx,
-        )
-      } else {
-        FileBasedBlobCompressionProverClientV2(
-          config = proverConfig,
-          vertx = vertx,
-          log = log,
-        )
-      }
+      FileBasedBlobCompressionProverClientV2(
+        config = proverConfig,
+        vertx = vertx,
+        log = log,
+      )
         .also { blobWaitingResponsesMetric.addReporter(it) }
     }
   }
 
-  fun proofAggregationProverClient(log: Logger? = null): ProofAggregationProverClientV2 {
+  fun proofAggregationProverClient(
+    log: Logger = FileBasedProofAggregationClientV2.LOG,
+  ): ProofAggregationProverClientV2 {
     return createClient(
       proverAConfig = config.proverA.proofAggregation,
       proverBConfig = config.proverB?.proofAggregation,
       switchBlockNumberInclusive = config.switchBlockNumberInclusive,
     ) { proverConfig ->
-      if (log == null) {
-        FileBasedProofAggregationClientV2(
-          config = proverConfig,
-          vertx = vertx,
-        )
-      } else {
-        FileBasedProofAggregationClientV2(
-          config = proverConfig,
-          vertx = vertx,
-          log = log,
-        )
-      }
+      FileBasedProofAggregationClientV2(
+        config = proverConfig,
+        vertx = vertx,
+        log = log,
+      )
         .also { aggregationWaitingResponsesMetric.addReporter(it) }
     }
   }
--- a/coordinator/app/src/main/kotlin/linea/coordinator/config/v2/ConflationConfig.kt
+++ b/coordinator/app/src/main/kotlin/linea/coordinator/config/v2/ConflationConfig.kt
@@ -5,6 +5,7 @@ import linea.blob.BlobCompressorVersion
 import linea.domain.RetryConfig
 import net.consensys.linea.traces.TracesCounters
 import java.net.URL
+import java.nio.file.Path
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.seconds
 
@@ -29,6 +30,7 @@ data class ConflationConfig(
   val blobCompression: BlobCompression = BlobCompression(),
   val proofAggregation: ProofAggregation = ProofAggregation(),
   val tracesLimits: TracesCounters,
+  val backtestingDirectory: Path? = null,
 ) : FeatureToggle {
   data class BlobCompression(
     val blobSizeLimit: UInt = 102400u,
--- a/coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/ConflationToml.kt
+++ b/coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/ConflationToml.kt
@@ -6,6 +6,7 @@ import linea.coordinator.config.v2.ConflationConfig
 import net.consensys.linea.traces.TracesCountersV2
 import net.consensys.linea.traces.TracesCountersV4
 import java.net.URL
+import java.nio.file.Path
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.seconds
 
@@ -25,6 +26,7 @@ data class ConflationToml(
   val l2LogsEndpoint: URL? = null,
   val blobCompression: BlobCompressionToml = BlobCompressionToml(),
   val proofAggregation: ProofAggregationToml = ProofAggregationToml(),
+  val backtestingDirectory: Path? = null,
 ) {
   init {
     require(proofAggregation.proofsLimit >= (blobCompression.batchesLimit ?: 1u) + 1u) {
@@ -109,6 +111,7 @@ data class ConflationToml(
       blobCompression = this.blobCompression.reified(),
       proofAggregation = this.proofAggregation.reified(),
       tracesLimits = tracesCountersLimitsV2 ?: tracesCountersLimitsV4!!,
+      backtestingDirectory = backtestingDirectory,
     )
   }
 }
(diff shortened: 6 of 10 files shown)


coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
```diff
@@ -0,0 +1,384 @@
+package net.consensys.zkevm.coordinator.app.conflationbacktesting
```
gauravahuja, line 1: This is the main file for review. Github hides the diff in default view.

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
```diff
@@ -0,0 +1,360 @@
...
+          requestRetryConfig = updatedCoordinatorConfig.conflation.l2RequestRetries,
+          vertx = vertx,
+        ),
+        log = log,
+      ),
+      batchProofHandler = { _ -> SafeFuture.completedFuture(Unit) },
+      vertx = vertx,
+      config = ProofGeneratingConflationHandlerImpl.Config(
+        conflationAndProofGenerationRetryBackoffDelay = 5.seconds,
+        executionProofPollingInterval = 100.seconds,
```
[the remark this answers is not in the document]
gauravahuja: I don't want the poller as no proof response will be generated.

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
```diff
@@ -0,0 +1,384 @@
...
+) : LongRunningService {
+
+  init {
+    prodCoordinatorConfigs.traces.common?.endpoints?.contains(conflationBacktestingAppConfig.tracesApi.endpoint)
+      ?.let { require(!it) { "Cannot use prod traces endpoint for backtesting" } }
+    prodCoordinatorConfigs.traces.counters?.endpoints?.contains(conflationBacktestingAppConfig.tracesApi.endpoint)
+      ?.let { require(!it) { "Cannot use prod traces endpoint for backtesting" } }
+    prodCoordinatorConfigs.traces.conflation?.endpoints?.contains(conflationBacktestingAppConfig.tracesApi.endpoint)
+      ?.let { require(!it) { "Cannot use prod traces endpoint for backtesting" } }
+  }
```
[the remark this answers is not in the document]
gauravahuja: I don't think using prod shomei will be an issue. Using prod traces is an issue because it will overwrite traces files in prod directory.

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingService.kt
```diff
@@ -1,34 +1,47 @@
...
+    val app = ConflationBacktestingApp(
+      vertx = vertx,
+      conflationBacktestingAppConfig = conflationBacktestingConfig,
+      prodCoordinatorConfigs = configs,
+      httpJsonRpcClientFactory = httpJsonRpcClientFactory,
+      metricsFacade = metricsFacade,
+    )
+    conflationBackTestingApps[jobId] = app
+    app.start()
+    return jobId
```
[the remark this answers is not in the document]
gauravahuja: I think this is fine for now. There is also no way to figure out if the backtesting failed after start. Thus no need for a FAILED status right now.

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
```diff
@@ -0,0 +1,384 @@
...
+    log = log,
+    requestRetryConfig = updatedCoordinatorConfig.conflation.l2RequestRetries,
+    vertx = vertx,
+  )
+
+  private val lastProcessedBlockNumber = conflationBacktestingAppConfig.startBlockNumber - 1uL
+  private val lastProcessedBlock = l2EthClient.ethGetBlockByNumberTxHashes(
+    BlockParameter.fromNumber(lastProcessedBlockNumber),
+  ).get()
+  private val lastProcessedTimestamp = Instant.fromEpochSeconds(lastProcessedBlock.timestamp.toLong())
```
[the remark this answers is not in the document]
gauravahuja: ethGetBlockByNumberTxHashes does not return a nullable block

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingService.kt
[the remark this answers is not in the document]
gauravahuja: Updated now, the service polls for jobs that have been completed and removes the app object from the map.

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/api/dto/ConflationCreateProverRequestJsonDto.kt
```diff
@@ -1,14 +1,44 @@
...
+import net.consensys.zkevm.coordinator.app.conflationbacktesting.ConflationBacktestingConfig
+import net.consensys.zkevm.coordinator.app.conflationbacktesting.ShomeiApiConfig
+import net.consensys.zkevm.coordinator.app.conflationbacktesting.TracesApiConfig
 
 data class ConflationCreateProverRequestJsonDto(
   val startBlockNumber: Long,
   val endBlockNumber: Long,
+  val blobCompressorVersion: String,
+  val batchesFixedSize: Int?,
+  val parentBlobShnarf: String?,
```
Filter94, line 16: Is null value possible only in case startBlockNumber is 1? If so, I think we better add a validation here

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflation/ConflationApp.kt
```diff
@@ -197,10 +196,6 @@ class ConflationApp(
         maxProvenBlobCache,
       ),
     )
-    val genesisStateProvider = GenesisStateProvider(
```
Filter94, line 200: Does it have any other usages? Can we remove the `GenesisStateProvider` from the codebase?

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
```diff
@@ -0,0 +1,383 @@
...
+import java.nio.file.Path
+import kotlin.concurrent.atomics.AtomicBoolean
+import kotlin.concurrent.atomics.ExperimentalAtomicApi
+import kotlin.time.Duration.Companion.seconds
+
+@OptIn(ExperimentalAtomicApi::class)
+class ConflationBacktestingApp(
+  val vertx: Vertx,
+  val conflationBacktestingAppConfig: ConflationBacktestingConfig,
+  prodCoordinatorConfigs: CoordinatorConfig,
```
Filter94: I'd suggest `main` over `prod`, because sometimes prod isn't a prod one

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
```diff
@@ -0,0 +1,383 @@
...
+      ?.let { require(!it) { "Cannot use prod traces endpoint for backtesting" } }
+  }
+
+  private val log = LogManager.getLogger("conflation_backtesting.job_${conflationBacktestingAppConfig.jobId()}")
+
+  private val conflationBacktestingComplete = AtomicBoolean(false)
+
+  fun isConflationBacktestingComplete(): Boolean = conflationBacktestingComplete.load()
+
+  fun onConflationProgress(
```
Filter94, line 79: This might be a good place to log the progress on the Info level, unless it's logged on Info somewhere else. In other places logging the requests / block number might be on lower logging levels, so we'll have less visibility by default, but I didn't check

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
```diff
@@ -0,0 +1,383 @@
...
+  }
+
+  val l2EthClient: EthApiClient = createEthApiClient(
+    rpcUrl = updatedCoordinatorConfig.conflation.l2Endpoint.toString(),
+    log = log,
+    requestRetryConfig = updatedCoordinatorConfig.conflation.l2RequestRetries,
+    vertx = vertx,
+  )
+
+  private val lastProcessedBlockNumber = conflationBacktestingAppConfig.startBlockNumber - 1uL
```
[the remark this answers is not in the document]
Filter94: I believe startBlockNumber = 0 is impossible, because it's a genesis block and it's not executed by the tracer

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
```diff
@@ -0,0 +1,383 @@
...
+    return if (proofIndex.endBlockNumber == conflationBacktestingAppConfig.endBlockNumber) {
+      conflationBacktestingComplete.store(true)
+      log.info("Conflation backtesting complete")
+      this.stop()
+    } else {
+      SafeFuture.completedFuture(Unit)
+    }
+  }
+
+  val updatedCoordinatorConfig: CoordinatorConfig = prodCoordinatorConfigs.copy(
```
Filter94: I suggest to rename it to `backtestingCoordinatorConfig`

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
```diff
@@ -0,0 +1,383 @@
...
+    vertx = vertx,
+  )
+
+  private val lastProcessedBlockNumber = conflationBacktestingAppConfig.startBlockNumber - 1uL
+  private val lastProcessedBlock = l2EthClient.ethGetBlockByNumberTxHashes(
+    BlockParameter.fromNumber(lastProcessedBlockNumber),
+  ).get()
+  private val lastProcessedTimestamp = Instant.fromEpochSeconds(lastProcessedBlock.timestamp.toLong())
+
+  private val conflationCalculator: TracesConflationCalculator = run {
```
Filter94, line 161: What would it take to reuse this code block between the main app and conflation backtesting app?

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingConfig.kt
```diff
@@ -0,0 +1,30 @@
...
+  val startBlockNumber: ULong,
+  val endBlockNumber: ULong,
+  val blobCompressorVersion: BlobCompressorVersion,
+  val batchesFixedSize: UInt? = null,
+  val parentBlobShnarf: String? = null,
+  val tracesApi: TracesApiConfig,
+  val shomeiApi: ShomeiApiConfig,
+) {
+  fun jobId(): String {
+    return "$startBlockNumber-$endBlockNumber-${this.hashCode()}"
```
Filter94, line 16: Is there a case when we want to run 2 separate backtesting jobs for the same block range at the same time?

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/blockcreation/BlockCreationMonitor.kt
```diff
@@ -42,6 +42,7 @@ class BlockCreationMonitor(
     val startingBlockWaitTimeout: Duration = 14.days,
     val lastL2BlockNumberToProcessInclusive: ULong? = null,
     val lastL2BlockTimestampToProcessInclusive: Instant? = null,
+    val skipGapCheck: Boolean = false,
```
Filter94: I suggest we stick to the default MAX_VALUE of l2FetchBlocksLimit instead, rather than creating a new config

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingConfig.kt
gauravahuja: Use case could be to test different trace nodes or other parameters which are currently not implemented like traces limits, conflation parameters etc.

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflation/ConflationApp.kt
gauravahuja: I can do that in a follow up pr, this one is already big

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
gauravahuja: We need a builder for conflation calculator. For now major differenrces in the conflation calculator in conflationapp and conflationbacktesting app are:
- No Deadline based conflation in backtesting
- One logger used everywhere in backtesting
- No metrics tracking in backtesting

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
```diff
@@ -0,0 +1,388 @@
...
+  val vertx: Vertx,
+  val conflationBacktestingAppConfig: ConflationBacktestingConfig,
+  mainCoordinatorConfig: CoordinatorConfig,
+  httpJsonRpcClientFactory: VertxHttpJsonRpcClientFactory,
+  val metricsFacade: MetricsFacade,
+) : LongRunningService {
+
+  init {
+    mainCoordinatorConfig.traces.common?.endpoints?.contains(conflationBacktestingAppConfig.tracesApi.endpoint)
+      ?.let { require(!it) { "Cannot use prod traces endpoint for backtesting" } }
```
jonesho: I dun see why we hardcoded the error message with `prod` for those traces endpoints, they can be anything given by the users

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
```diff
@@ -0,0 +1,388 @@
...
+      conflationDeadline = null,
+      blobCompression = mainCoordinatorConfig.conflation.blobCompression.copy(
+        blobCompressorVersion = conflationBacktestingAppConfig.blobCompressorVersion,
+      ),
+      proofAggregation = mainCoordinatorConfig.conflation.proofAggregation.copy(
+        targetEndBlocks = (mainCoordinatorConfig.conflation.proofAggregation.targetEndBlocks ?: emptyList())
+          .toMutableList().let {
+            it.add(conflationBacktestingAppConfig.endBlockNumber)
+            it
+          },
```
jonesho: Suggestion:
```
.toMutableList().also { it.add(conflationBacktestingAppConfig.endBlockNumber)},
```

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingService.kt
```diff
@@ -1,34 +1,96 @@
...
   }
 
-  fun submitConflationBacktestingJob(conflationBacktestingRequest: ConflationBacktestingRequest): String {
-    return conflationBacktestingRequest.hashCode().toString()
+  private val completedJobs: MutableSet<String> = ConcurrentHashSet()
+  private val conflationBackTestingApps: MutableMap<String, ConflationBacktestingApp> = ConcurrentHashMap()
+
+  fun submitConflationBacktestingJob(conflationBacktestingConfig: ConflationBacktestingConfig): String {
+    val jobId = conflationBacktestingConfig.jobId()
+    if (completedJobs.contains(jobId)) {
```
jonesho, line 42: I'm not sure if its a good idea for the above check, because there could be cases that the user want to submit the exact same request multiple times for different prover versions, wdyt? Would that be a valid use case?

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
```diff
@@ -0,0 +1,388 @@
...
+    ).thenApply {
+      log.info("Conflation backtesting stopped successfully")
+    }
+  }
+
+  companion object {
+    fun getUpdatedPath(path: Path, suffix: String): Path {
+      val lastDirectory = path.getName(path.nameCount - 1)
+      val updatedDirectory = "$lastDirectory-$suffix"
+      return path.parent.resolve(updatedDirectory)
```
jonesho: @gauravahuja is prover team aware of the request and response folder paths for backtesting with the suffix containing the jobId of each job? i.e. the prover would need to know the request/response folder path for each jobId

gauravahuja: I did speak with @gusiri about this. The other option was to have the request response directory be part of the conflation backtesting request but we felt it will be easier to test if they didn't have to provide multiple request attributes.

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingService.kt
gauravahuja: The thing to note is that the prover is not looking at the backtesting proof requests. We are simply generating the proof requests and no prover is creating the responses.

One thing I can do is once the job is complete and another request comes in for the same job I can delete the proof request directories and restart the run. If the request directory and request files are not deleted, coordinator will reuse the existing requests.

gauravahuja: Also HashCode of the request is part of the jobId so even if you keep the same start-end block number but  change any other parameter like traces api or shomei endpoint or tracer version etc it will be different run.

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
gusiri: Yes, as we discussed before, I think passing only `start, end` block ranges is the best. 
The path or name of the folder doesn't really matter. 
As long as there’s a way to identify the folder path, we’re good. The goal is todownload the folder to a local environment and run the integration tests locally.

Q. Same question. How does the prover team know the name of the folder where the requests were generated?
Will we be notified as soon as we make the RPC call as a return value, or some other way?

Please note that the prover team doesn’t have read access to the prod EFS, so we can’t find the path manually.
The ops team will likely create a cron job where the prover team can specify the target path to sync a folder to the devnet S3.

gauravahuja: > How does the prover team know the name of the folder

eg: `/shared/v3/prover-execution/requests` is the main requests directory configured in coordinator. For backtesting the requests will be generated in `/shared/v3/prover-execution/requests-conflation-backtesting-$jobId` where `jobId = <startBlockNubmer>-<endBlockNumber>-<backtestingRequestHashCode>`. JobId is returned as response to the backtesting request api

gusiri: Sounds good to me.
I know that was just an example :), but let’s use a path that doesn’t share parent directories with the prod environment, to be safe and avoid confusion.
e.g., `/shared/prover-backtesting/requests-conflation-backtesting-$jobId`

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/CoordinatorApp.kt
```diff
@@ -179,8 +184,8 @@ class CoordinatorApp(private val configs: CoordinatorConfig) {
     return runCatching {
       SafeFuture.allOf(
         l1App.stop(),
-        api.stop(),
         conflationBacktestingService.stop(),
+        api.stop(),
```
fluentcrafter, line 187: why not the other way around. can the IP work without the servicer working?

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
fluentcrafter: >eg: /shared/v3/prover-execution/requests is the main requests directory configured in coordinator. For backtesting the requests will be generated in /shared/v3/prover-execution/requests-conflation-backtesting-$jobId where jobId = <startBlockNubmer>-<endBlockNumber>-<backtestingRequestHashCode>. JobId is returned as response to the backtesting request api

This should change. The back testing directories **most not mix** with production directories. 

The original [issue](https://github.com/Consensys/linea-monorepo/issues/1740) defines it 
> It should generate all pover requests (execution, compression, aggregation) into the specified proverRequestsDestinationDir/{execution, compression,aggregation}.

The reason for this segregation an paren directory level is that:
- prover auto scaling looks at the quests under `/shared/v3/prover-{execution,}/requests` and current design will mess with that
- the clean up task is cumbersome as we cannot just do  `rm -fr /shared/backtesting/prover/ `
- we will have prod and not-prod files mixed up same directory.

So, we should do something like `/shared/backtesting/2026-01-26T1200-jobid/prover/{execution/compression/aggregation}`

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
```diff
@@ -0,0 +1,388 @@
...
+        proofIndex.endBlockNumber,
+        conflationBacktestingAppConfig.endBlockNumber,
+      )
+      SafeFuture.completedFuture(Unit)
+    }
+  }
+
+  val backtestingCoordinatorConfig: CoordinatorConfig = mainCoordinatorConfig.copy(
+    conflation = mainCoordinatorConfig.conflation.copy(
+      l2FetchBlocksLimit = UInt.MAX_VALUE,
```
fluentcrafter, line 98: why not endBlockNumber - startBlockNumber + 1 ?

coordinator/clients/prover-client/file-based-client/src/main/kotlin/net/consensys/zkevm/coordinator/clients/prover/ProverClientFactory.kt
```diff
@@ -41,44 +42,76 @@ class ProverClientFactory(
...
       switchBlockNumberInclusive = config.switchBlockNumberInclusive,
     ) { proverConfig ->
-      FileBasedExecutionProverClientV2(
-        config = proverConfig,
-        vertx = vertx,
-        tracesVersion = tracesVersion,
-        stateManagerVersion = stateManagerVersion,
-      ).also { executionWaitingResponsesMetric.addReporter(it) }
+      if (log == null) {
+        FileBasedExecutionProverClientV2(
```
fluentcrafter: This repeated if/else for the logger can be improved IMO.

We can do something like
```kotlin
class FileBasedExecutionProverClientV2(
...
 log: Logger = FileBasedExecutionProverClientV2.log
) {

  companion object {
    val log = LogManager.getLogger(FileBasedExecutionProverClientV2::class.java)
  }
} 

fun executionProverClient(
    tracesVersion: String,
    stateManagerVersion: String,
    log: Logger = FileBasedExecutionProverClientV2.log,
  ): ExecutionProverClientV2 {

```

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingService.kt
```diff
@@ -1,34 +1,96 @@
...
+      throw IllegalArgumentException("Conflation backtesting job with id $jobId already completed")
```
fluentcrafter: this message wont' be very friendly ti the user. We should tell that the given request is arleady being executed/processed with $jobId so they can retrieve the progress

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/CoordinatorApp.kt
gauravahuja: CompletableFuture.allOf() does not guarantee the order of completion of individual futures. So the order does not really matter here. Once CoordinatorApp.stop is called any api invocation should be expected to return unexpected ressults.

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
gauravahuja: Yes that does work because i set the last proven block to startBlockNumber - 1. I don't want to play with boundaries for the fetch limit.

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
gauravahuja: Updated! I introduced a new config parameter for backtesting directory and the directories are now set as
`<backtesting-directory>/jobId/{execution|compression|aggregation}/{requests|responses}`

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/api/dto/ConflationCreateProverRequestJsonDto.kt
```diff
@@ -20,12 +50,44 @@ data class ConflationCreateProverRequestJsonDto(
...
-        startBlockNumber = dto.startBlockNumber.toULong(),
-        endBlockNumber = dto.endBlockNumber.toULong(),
-      )
-    }
+data class TracesApiDto(
+  val endpoint: String,
+  val version: String,
+  val requestLimitPerEndpoint: Int,
+) {
+  constructor() : this(
```
fluentcrafter, line 61: why do we need these constructors?

gauravahuja: I thought jackson requires default constructor when parsing json from the http request to dto.

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
fluentcrafter: It's much better. Thanks

Issue #2130, opened 2026-01-23
Persistence Retryer should display WARN log messages after certain amount of time during endless retries


jonesho:
🔐 🔐 `Should this be a security advisory instead?` 🔐 🔐

## Description
Persistence Retryer should display WARN log messages after certain amount of time X during retries. X should be a configurable Duration value. 

## Motivation
Ops team can trigger an alert on the persistence retry WARN logs and investigate to avoid silent endless retries

## Tasks
- [ ] Medium-level task that needs to be implemented for this feature request to be considered complete.

## Acceptance criteria
- [ ] List tests that need to be implemented or checks to consider this task completed. 

## Risks
- [ ] These are risks that may prevent completion of the task as well as areas to pay special attention to

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #2151, opened 2026-01-27
feat: add config option exceptionLogDelay for PersistenceRetryer

jonesho:
This PR implements issue(s) #2130

This PR adds the `exceptionConsumerDelay` config variable to the `AsyncRetryer` to delay the calling to exception consumer for certain amount of time since the retry started, the motivation is to mute the `PersistenceRetryer` warn logs by delaying the calling of its exception consumer which is stateless just for logging the exceptions.

It'd be tricky to do this feature on the `PersistenceRetryer` class because the object of this class is shared among different repository instances.

Therefore here we go for adding an optional config to delay the calling of exception consumer in AsyncRetryer, the cons is that this assumes PersistenceRetryer's exception consumer would only do stateless operations like logging, if later-on, it adds exception handling logic that couldn't be skipped, then we would need to disable the delay.

OR we can go for a more generic logging approach by something similar to `DebouncingFilter`, lets say `DelayFilter` for displaying certain level (e.g. WARN) logs only after some specific delay, where the log-level and delay values would be defined on `log4j.xml` instead of via coordinator configs. cc @fluentcrafter 

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches shared retry/persistence infrastructure by changing when exceptions are surfaced to consumers (and logs), which could mask early failures or alter observability if misconfigured. Behavior is gated by an optional duration and covered by new tests, limiting blast radius.
> 
> **Overview**
> Introduces `ignoreFirstExceptionsUntilTimeElapsed` across retry configuration (`RetryConfig`, TOML parsing) and threads it into `CoordinatorApp` persistence retry settings.
> 
> Updates `AsyncRetryer` to accept this optional duration and **skip calling `exceptionConsumer`** for failures occurring before the elapsed threshold, with new unit coverage.
> 
> Updates `PersistenceRetryer` to use the new delay option and changes persistence retry logging to `WARN`, plus a dev `log4j2` logger entry for `PersistenceRetryer` and tests asserting logs are muted during the configured window.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d001c1ebf51108bdcb416528f0e816bb9ce1b82c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


jvm-libs/generic/extensions/futures/src/test/kotlin/net/consensys/linea/async/AsyncRetryerTest.kt
```diff
@@ -379,6 +380,39 @@ class AsyncRetryerTest {
...
+    assertThat(future)
+      .isCompletedExceptionally
+      .isNotCancelled
+      .failsWithin(2.milliseconds.toJavaDuration())
+      .withThrowableOfType(ExecutionException::class.java)
+      .withCauseInstanceOf(IndexOutOfBoundsException::class.java)
+      .withMessageContaining("Error 11")
+    assertThat(
+      exceptionConsumerCallTimes.find {
+        (it.toEpochMilli() - startTime.toEpochMilli()) < 100.milliseconds.inWholeMilliseconds
```
fluentcrafter: this is quite tricky to read/understand. 

easier to understand 
```kotlin
@Test
  fun `retry should retry until maxRetries are reached and consumer not being called before delay`(vertx: Vertx) {
    val callCount = AtomicInteger(0)
    val startTime = kotlin.time.Clock.System.now()
    val exceptionConsumerCallDelays: MutableList<Duration> = mutableListOf()
    val future =
      AsyncRetryer.retry<Int>(
        vertx = vertx,
        backoffDelay = 20.milliseconds,
        maxRetries = 10,
        exceptionConsumer = { exceptionConsumerCallDelays.add(kotlin.time.Clock.System.now().minus(startTime)) },
        exceptionConsumerDelay = 100.milliseconds,
      ) {
        throw IndexOutOfBoundsException("Error ${callCount.incrementAndGet()}")
      }

    runCatching { future.get() }

    assertThat(callCount.get()).isEqualTo(11)
    assertThat(future)
      .isCompletedExceptionally
      .isNotCancelled
      .failsWithin(2.milliseconds.toJavaDuration())
      .withThrowableOfType(ExecutionException::class.java)
      .withCauseInstanceOf(IndexOutOfBoundsException::class.java)
      .withMessageContaining("Error 11")

    assertThat(exceptionConsumerCallDelays.filter { it > 100.milliseconds }).isNotEmpty()
    assertThat(exceptionConsumerCallDelays.filter { it < 100.milliseconds }).isEmpty()
  }
```

fluentcrafter:
Hi @jonesho

> Should we go for `ignoreExceptionsInitialWindow` ?

I think `ignoreFirstExceptionsUntilTimeElapsed` is slightly more suggestive easier to understand for external users. But I am ok with your name as well.

I would like to have @Filter94 @gauravahuja feecback.

fluentcrafter (dismissed):

fluentcrafter (approved):

PR #2164, opened 2026-01-28
[ZkTracer] Feat/blake with defcall

amkCha:
This PR implements issue(s) #

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Updates the BLAKE2f constraint wiring and removes the Java-side `Blake2f` tracing module, which can affect trace/circuit consistency for precompile execution. Also bumps the `go-corset` version to an RC, which may change generated traces/validation behavior.
> 
> **Overview**
> Routes BLAKE2f precompile data from `blake2fmodexpdata` into the `blake2f` constraints using a new `defcall` + selector, replacing the previous explicit lookup and reordering the `blake2f` zkasm function arguments to match the new call convention.
> 
> Removes the Java `Blake2f` module/operation tracing and stops `BlakeModexpData.callBlake()` from emitting separate `blake2f` rows, relying solely on the combined `BLAKE_MODEXP_DATA` path for BLAKE accounting/limits.
> 
> Build/test plumbing is updated by bumping the default `go-corset` version to `v1.2.0-rc6` and standardizing JUnit default timeouts for `test`/`nightlyTests`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b79704c07c6af92ccdc50e1f5dd6ae561ce5bf1f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tracer/gradle/tests.gradle
+++ b/tracer/gradle/tests.gradle
@@ -168,6 +168,7 @@ tasks.test.configure {
     excludeTags("prc-calltests")
   }
   finalizedBy(jacocoTestReport)
+  systemProperty("junit.jupiter.execution.timeout.default", "20 m")
 }
 
 tasks.register("fastReplayTests", Test) {
@@ -182,7 +183,6 @@ tasks.register("nightlyReplayTests", Test) {
   useJUnitPlatform {
     includeTags("nightly & replay")
   }
-  systemProperty("junit.jupiter.execution.timeout.default", "20 m")
 }
 
 tasks.register("nightlyTests", Test) {
@@ -191,6 +191,7 @@ tasks.register("nightlyTests", Test) {
     excludeTags("replay")
   }
   finalizedBy("jacocoNightlyTestsReport")
+  systemProperty("junit.jupiter.execution.timeout.default", "20 m")
 }
 
 tasks.register("weeklyTests", Test) {


tracer-constraints/blake2fmodexpdata/columns.lisp
```diff
@@ -14,5 +14,16 @@
...
+  (h6h7_be              :i128 :display :bytes)
   )
 
+(defun (blake2f-selector)
+  (== 1 ( * (- 1 (prev blake2fmodexpdata.IS_BLAKE_DATA)) blake2fmodexpdata.IS_BLAKE_DATA)))
+
+(defcall (h0h1_be h2h3_be h4h5_be h6h7_be) blake2f ( (i64 (shift LIMB 13)) (shift LIMB 0) (shift LIMB 1) (shift LIMB 2) (shift LIMB 3)
+                                       (i1 (shift LIMB 4)) (shift LIMB 5) (shift LIMB 6) (shift LIMB 7)
+                                         (shift LIMB 8) (shift LIMB 9) (shift LIMB 10)
+                                        (i1 (shift LIMB 11)) (shift LIMB 12) (i1 (shift LIMB 14))) (blake2f-selector))
```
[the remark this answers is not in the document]
amkCha: In this case, the cast will just not work, but it's incorrect to try that cast

tracer-constraints/blake2fmodexpdata/columns.lisp
```diff
@@ -14,5 +14,18 @@
...
 
+(defun (blake2f-selector)
+  (== 1 ( * (- 1 (prev blake2fmodexpdata.IS_BLAKE_DATA)) blake2fmodexpdata.IS_BLAKE_DATA)))
+
+(defcall
+ (h0h1_be h2h3_be h4h5_be h6h7_be)
+  blake2f
+ ( (i64 (shift LIMB 13)) (shift LIMB 0) (shift LIMB 1) (shift LIMB 2) (shift LIMB 3)
+   (shift LIMB 4) (shift LIMB 5) (shift LIMB 6) (shift LIMB 7) (shift LIMB 8)
+   (shift LIMB 9) (shift LIMB 10) (shift LIMB 11) (shift LIMB 12) (i1 (shift LIMB 14))) (blake2f-selector))
```
OlivierBBB: Potentially silly question: you include the 13 data limbs (`0..13`), which include the original values of `h0 .. h7`, and the two parameters (`13..15`) `r` and `f` and I take it that the following
```rust
( h0h1_be h2h3_be h4h5_be h6h7_be )
```
is meant to contain the result (what the EYP calls `h'0 .. h'7`).

In the present module the result is contained in rows `15..19`. At which point do we compare the predicted result (`LIMB[i + 15], ..., LIMB[i + 18]`) from `BLK_MDX` to the results of the `defcall` ? And why not directly include the outputs as the above limbs ?

amkCha: So the way I understood it, in the java part of the tracer, I feed the result columns
```
( h0h1_be h2h3_be h4h5_be h6h7_be )
```
image

And the `defcall` replaces the lookup by computing the results and checking that the values filled in the trace match with the zkasm module computation. So it looks like the `defcall` is filling columns, but it's more like a constraint that needs to hold

Now, to feed the result columns, I must admit that I take my "external" result from the return data
image
Vs the limbs are fed with values from `computeBlake2fResult()` 
image

Should I change that ?

tracer-constraints/blake2fmodexpdata/columns.lisp
```diff
@@ -14,5 +14,18 @@
...
+  (h6h7_be              :i128 :display :bytes)
   )
 
+(defun (blake2f-selector)
+  (== 1 ( * (- 1 (prev blake2fmodexpdata.IS_BLAKE_DATA)) blake2fmodexpdata.IS_BLAKE_DATA)))
+
+(defcall
+ (h0h1_be h2h3_be h4h5_be h6h7_be)
+  blake2f
+ ( (i64 (shift LIMB 13)) (shift LIMB 0) (shift LIMB 1) (shift LIMB 2) (shift LIMB 3)
```
lorenzogentile404: The lookup looks go. Maybe it is more readable if we change the signature of `blake2f` function to have `r u64` in the corresponding spot of the limb (13), so it is more readable?
Also, I did not know downcasting was possible, interesting!

lorenzogentile404: Also minot change, `(shift LIMB 0) (shift LIMB 1)` can be just `LIMB` and `(next LIMB)`

tracer-constraints/blake2fmodexpdata/columns.lisp
OlivierBBB: I don't think you need to change that _per se_ but my understanding is that we are missing a "linking" constraint. You have these columns `hxhy_be` that contain the provably correct output of the BLAKE2f hash function, and you also have the `LIMB[i + 15], ..., LIMB[i + 18]` cells that are supposed to contain the result.

These cells are where our RAM (i.e. `MMIO`) copies the result from. The impression I get is that there is no connection at all between your `hxhy_be` columns and the `LIMB[i + k]` cells (i.e. the two aren't connected by e.g. an equality constraint that forces your values computed in zkASM to flow into the expected cells of the trace).

---------------

The main point here is that `LIMB[i + 15], ..., LIMB[i + 18]` are where the `BLK_MDX` and `MMIO` module interface. These are the cells where the `MMIO -> BLK_MDX` lookup lands on when it is time to copy the return data of the prc call to blake to memory.

If my understanding is correct then we are currently correctly proving the value of BLAKE2f but not feeding it into the part of the trace that interfaces with `MMIO`. In other words one could just lie about the `IS_RESULT` portion of the trace and write whatever we wanted, and the `MMIO` would just use those values.

amkCha: The `defcall` now directly checks the shifted LIMBs, which make way more sense indeed 
image

tracer-constraints/blake2fmodexpdata/columns.lisp
amkCha: Good point for the order of the params, I changed the zkasm function, it's easier to read !
And yes, if casting possible, it works - else it just doesn't do anything :ok_hand: 
In the end, I didn't use `LIMB` and `next LIMB` to make it easier to read - you understand that it's an incremental shifting more easily

tracer-constraints/blake2fmodexpdata/columns.lisp
```diff
@@ -16,3 +16,18 @@
...
+  (shift LIMB 15) (shift LIMB 16) (shift LIMB 17) (shift LIMB 18)
+ )
+  blake2f
+ (
+  (shift LIMB 0)  (shift LIMB 1)  (shift LIMB 2)  (shift LIMB 3) (shift LIMB 4)
+  (shift LIMB 5)  (shift LIMB 6)  (shift LIMB 7)  (shift LIMB 8) (shift LIMB 9)
+  (shift LIMB 10) (shift LIMB 11) (shift LIMB 12) (i64 (shift LIMB 13)) (i1 (shift LIMB 14))
+  )
+  (blake2f-selector)
+ )
```
OlivierBBB, line 33: LGTM

OlivierBBB: Very minor point: the `r` parameter, `(shift LIMB 13)`, is a 4 byte integer so could be cast to `i32`

image

tracer-constraints/blake2fmodexpdata/columns.lisp
```diff
@@ -16,3 +16,18 @@
   (IS_BLAKE_RESULT      :binary@prove)
   )
 
+(defun (blake2f-selector)
+  (== 1 ( * (- 1 (prev blake2fmodexpdata.IS_BLAKE_DATA)) blake2fmodexpdata.IS_BLAKE_DATA)))
```
OlivierBBB, line 20: Another micro comment: why make the selector an equality ? Directly setting it as the product
```
(defun (blake2f-selector)   ( * (- 1 (prev blake2fmodexpdata.IS_BLAKE_DATA)) blake2fmodexpdata.IS_BLAKE_DATA)))
```
will work just as well (unless I'm overlooking something in the syntax).

Issue #1997, opened 2026-01-07
[Epic] Forced Transactions


fluentcrafter:
🔐 🔐 `Should this be a security advisory instead?` 🔐 🔐

## Description
Implement Forced Transactions according to the specification

## Tasks
- [ ] Coordinator - support to prove forced transaction inclusion
- [ ] Contracts - 
- [ ] Prover - 
- [ ] Shomei - 
- [ ] Tracer - 
- [ ] Maru - 
- [ ] Besu Plugin ? 

## Acceptance criteria
- [ ]

PR #2171, opened 2026-01-28
Coordinator - adds EthLogsfilter subscription through polling

fluentcrafter:
This PR implements issue(s) #1997

### Checklist

---

> [!NOTE]
> Introduces a polling-based Ethereum logs subscription with exactly-once delivery and updates forced transaction interfaces.
> 
> - Adds `EthLogsFilterSubscriptionFactoryPollingBased` and `EthLogsFilterPoller` to poll `eth_getLogs` in chunks, maintain progress (`lastSearchedBlock`), and ensure retries without duplicates; integrates with `VertxPeriodicPollingService`
> - New subscription interfaces (`EthLogConsumer`, `EthLogsFilterSubscriptionManager`, `EthLogsFilterSubscriptionFactory`) and build dep on `long-running-service`
> - Comprehensive tests for ordering, retries, chunking, filtering, lifecycle, and tag-based polling
> - BREAKING: `ForcedTransactionInclusionStatus` now includes `blockTimestamp: Instant`; `ForcedTransactionInclusionResult` splits `FilteredAddresses` into `FilteredAddressFrom` and `FilteredAddressTo`; equals/hashCode/toString updated accordingly
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b3de9aa81d41b7cc4d61edb943d2e01c46e07a91. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/jvm-libs/linea/clients/interfaces/src/main/kotlin/linea/ethapi/extensions/EthLogsFilterSubscription.kt
+++ b/jvm-libs/linea/clients/interfaces/src/main/kotlin/linea/ethapi/extensions/EthLogsFilterSubscription.kt
@@ -4,7 +4,9 @@ import linea.LongRunningService
 import linea.domain.EthLog
 import linea.ethapi.EthLogsFilterOptions
 
-typealias EthLogConsumer = (EthLog) -> Unit
+fun interface EthLogConsumer {
+  fun accept(ethLog: EthLog)
+}
 
 interface EthLogsFilterSubscriptionManager : LongRunningService {
   fun setConsumer(logsConsumer: EthLogConsumer)
--- a/jvm-libs/linea/clients/eth-logs-searcher/src/main/kotlin/linea/ethapi/EthLogsFilterSubscriptionPollingBased.kt
+++ b/jvm-libs/linea/clients/eth-logs-searcher/src/main/kotlin/linea/ethapi/EthLogsFilterSubscriptionPollingBased.kt
@@ -47,7 +47,7 @@ class EthLogsFilterSubscriptionFactoryPollingBased(
  * PENDING tag is not supported.
  */
 class EthLogsFilterPoller(
-  private val vertx: Vertx,
+  vertx: Vertx,
   private val ethApiClient: EthApiClient,
   private val filterOptions: EthLogsFilterOptions,
   private val l1FtxLogsPollingInterval: Duration,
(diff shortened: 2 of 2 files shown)


jvm-libs/linea/clients/interfaces/src/main/kotlin/linea/ethapi/extensions/EthLogsFilterSubscription.kt
```diff
@@ -0,0 +1,18 @@
+package linea.ethapi.extensions
+
+import linea.LongRunningService
+import linea.domain.EthLog
+import linea.ethapi.EthLogsFilterOptions
+
+typealias EthLogConsumer = (EthLog) -> Unit
```
gauravahuja: We do use this typealias pattern in a few places in the code. My experience with this is that it is difficult to find invocations of such methods using IDE tools like click to navigate etc. Personally I prefer a functional interface.

jvm-libs/linea/clients/eth-logs-searcher/src/test/kotlin/linea/ethapi/EthLogsFilterPollerTest.kt
```diff
@@ -0,0 +1,478 @@
...
+    val consumedLogs = mutableListOf<EthLog>()
+    poller = createPoller(
+      fromBlock = BlockParameter.BlockNumber(20UL),
+      toBlock = BlockParameter.Tag.FINALIZED,
+      pollingInterval = 100.milliseconds,
+      consumer = { log -> consumedLogs.add(log) },
+    )
+
+    poller.start().get()
+    Thread.sleep(250)
```
gauravahuja, line 257: I feel that these Thread.sleep can be repalced with `Awaitability.withPollDelay`

fluentcrafter: I can't find `Awaitability.withPollDelay`, can you share link to the doc?

Also, what's the advantage of that given that there are no state changes to observe so we cannot do an reliable "until"?

gauravahuja: It is `Awaitility.await().pollDelay(<duration>)`. 
https://github.com/awaitility/awaitility/wiki/Usage#advanced

I was suggesting that because I thought Awaitility might not block the OS thread the way Thread.sleep would, but I'm not sure about the internal implementation. Otherwise I do not see any use of Awaitility.

PR #2189, opened 2026-01-29
Coordinator: Forced Transactions Module

fluentcrafter:
This PR implements issue(s) #1997 

Adds `ForcedTransactionApp` without connection to the conflation yet

this PR depends on preceding 
- #2174

---

> [!NOTE]
> **Medium Risk**
> Introduces new long-running services that ingest L1 logs and drive sequencer RPC calls with ordering/queueing logic; mistakes could lead to missed/duplicated forced transactions or incorrect persistence state.
> 
> **Overview**
> Adds a new `coordinator:ethereum:forced-transactions` module implementing a `ForcedTransactionsApp` that resumes from the latest finalized forced-tx number, subscribes to L1 `ForcedTransactionAdded` logs, queues events strictly in order, and periodically sends batches to the sequencer.
> 
> Adds status tracking/persistence via new `ForcedTransactionRecord` and `ForcedTransactionsDao`, including logic to query the sequencer for inclusion status to mark-and-remove already-processed entries (preventing gaps) while allowing resends when status is unavailable.
> 
> Build/test plumbing is updated to support the new module (Gradle deps, `settings.gradle`, persistence test fixtures) and adds test fixtures/fakes plus a basic integration-style test for ordered sending; minor doc/comment tweaks are included.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bbf9f2ade3a64a5b62dfa0db9872f83aeb9037ff. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsL1EventsFetcher.kt
+++ b/coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsL1EventsFetcher.kt
@@ -130,7 +130,8 @@ class ForcedTransactionsL1EventsFetcher(
           nextExpectedFtx.decrementAndFetch()
         }.getOrThrow()
     } else {
-      val message = "event ForcedTransactionAdded is out of order: expectedFtx=$nextExpectedFtx, gotFtx=${event.event}"
+      val message =
+        "event ForcedTransactionAdded is out of order: expectedFtx=${nextExpectedFtx.load()}, gotFtx=${event.event}"
       log.error(message)
       this.stop()
       throw IllegalStateException(message)


coordinator/core/src/main/kotlin/net/consensys/zkevm/domain/ForcedTransactionRecord.kt
```diff
@@ -0,0 +1,45 @@
...
+import kotlin.time.Instant
+
+/**
+ *  linea_getForcedTransactionInclusionStatus
+ * {
+ *     "blockNumber": "0xeff35f", // On which transaction was tried
+ *     "from": "0x6221a9c005f6e47eb398fd867784cacfdcfff4e7",
+ *     // inclusion result / the type of invalidity for each forced transaction;
+ *     // for the executed valid transaction it is set to [invalidity.BadNonce]
+ *     // cases: Included, BadNonce, BadBalance, BadPrecompile, TooManyLogs, FilteredAddresses, Phylax)
```
Filter94: Minor: `FilteredAddresses` was split into 2

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsExecutionStatusUpdater.kt
```diff
@@ -0,0 +1,70 @@
+package linea.ftx
+
+import linea.contract.events.ForcedTransactionAddedEvent
+import linea.forcedtx.ForcedTransactionsClient
+import linea.persistence.ftx.ForcedTransactionsDao
+import net.consensys.zkevm.domain.ForcedTransactionRecord
+import tech.pegasys.teku.infrastructure.async.SafeFuture
+
+interface ProcessedTransactionsFilter {
+  fun isAlreadyProcessed(ftx: ForcedTransactionAddedEvent): SafeFuture<Boolean>
```
Filter94: Nit: This method isn't really meant to be used externally. It would be the cleanest to split this interface into 2, but for practicality it's also fine to leave as is

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsExecutionStatusUpdater.kt
```diff
@@ -0,0 +1,70 @@
...
+    return SafeFuture
+      .collectAll(unprocessed.stream())
+      .thenApply { it.filterNotNull() }
+  }
+}
+
+/**
+ * Responsible for getting Forced Transactions status from the sequencer and update local DB.
+ */
+class ForcedTransactionsStatusUpdater(
```
Filter94, line 20: I'm concerned that filterOutAlreadyProcessed persists the status as a side effect. Wouldn't it be clearer to make the persistence call explicit in the ForcedTransactionsSenderForExecution?

The name ForcedTransactionsStatusUpdater also poorly matches a class which main purpose is to expose filterOutAlreadyProcessed so if we leave it as is, I think `ForcedTransactionStatusCache` would suite better

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsL1EventsFetcher.kt
```diff
@@ -0,0 +1,121 @@
...
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.Logger
+import tech.pegasys.teku.infrastructure.async.SafeFuture
+import java.util.concurrent.CompletableFuture
+
+class ForcedTransactionsL1EventsFetcher(
+  private val address: String,
+  private val finalizedStateProvider: LineaRollupSmartContractClientReadOnlyFinalizedStateProvider,
+  private val ethLogsClient: EthLogsClient,
+  private val ethLogsFilterSubscriptionFactory: EthLogsFilterSubscriptionFactoryPollingBased,
```
Filter94: Minor
```suggestion
  private val ethLogsFilterSubscriptionFactory: EthLogsFilterSubscriptionFactory,
```

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsSenderForExecution.kt
```diff
@@ -0,0 +1,75 @@
...
+      .whenException { exception ->
+        log.warn(
+          "Failed sending forced transactions to sequencer, will retry in {}. errorMessage={} ftx={}",
+          pollingInterval,
+          exception.message,
+          unprocessedTransactions,
+          exception,
+        )
+      }
+  }
```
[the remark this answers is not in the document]
fluentcrafter: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/Consensys/linea-monorepo/pull/2189#discussion_r2741277524)

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsL1EventsFetcher.kt
```diff
@@ -0,0 +1,121 @@
...
+  private val l1EarliestBlock: BlockParameter = BlockParameter.Tag.EARLIEST,
+  private val l1HighestBlock: BlockParameter = BlockParameter.Tag.FINALIZED,
+  private val ftxQueue: MutableMap<ULong, ForcedTransactionAddedEvent>,
```
Filter94: It doesn't seem to be used in a meaningful way

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsL1EventsFetcher.kt
```diff
@@ -0,0 +1,121 @@
...
+        } else {
+          throw th
+        }
+      }
+  }
+
+  fun getStartBlockNumberToListenForLogs(): SafeFuture<BlockParameter> {
+    return lastFinalizedForcedTransaction()
+      .thenCompose { finalizedForcedTransactionNumber ->
+        this.log.info(
```
Filter94: Minor:
Why all the `this.`?

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsL1EventsFetcher.kt
```diff
@@ -0,0 +1,121 @@
...
+          topics = listOf(ForcedTransactionAddedEvent.topic),
+        )
+        this.log.info("start listening to ForcedTransactionAdded filter={}", filterOptions)
+        this.eventsSubscription = ethLogsFilterSubscriptionFactory.create(filterOptions)
+        this.eventsSubscription.setConsumer(::onNewForcedTransaction)
+        this.eventsSubscription.start()
+      }
+  }
+
+  fun onNewForcedTransaction(ethLog: EthLog) {
```
Filter94: I don't see why it has to be public

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsSenderForExecution.kt
```diff
@@ -0,0 +1,75 @@
...
+import kotlin.time.Duration
+
+/**
+ * Takes ForcedTransactionEventAdded from the queue and sends them to the sequencer.
+ */
+class ForcedTransactionsSenderForExecution(
+  vertx: Vertx,
+  val alreadyProcessed: ProcessedTransactionsFilter,
+  val ftxClient: ForcedTransactionsClient,
+  val ftxQueue: MutableMap<ULong, ForcedTransactionAddedEvent>,
```
Filter94: I see a few issues using `MutableMap` as a `Channel`:
1. It needs sorting in `ftxQueue.values.toList()` or it has to be a `LinkedHashMap`
1. It's more unconventional than `BlockingQueue`, for example, as a shared communication buffer

I suggest to use the `BlockingQueue` for communication that would receive items in strict order guaranteed by the `ForcedTransactionsL1EventsFetcher`. Also, if `EthLogsFilterPoller` provides strong guarantees of "exactly once" and in order delivery semantics, we shouldn't need a Map

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsSenderForExecution.kt
```diff
@@ -0,0 +1,75 @@
...
+    log.trace("all ftxs in queue={}", allFtx.map { it.forcedTransactionNumber })
+    if (allFtx.isEmpty()) {
+      return SafeFuture.completedFuture(Unit)
+    }
+
+    return alreadyProcessed
+      .filterOutAlreadyProcessed(allFtx)
+      .thenCompose { unprocessedTxs ->
+        log.debug("unprocessed ftxs={}", unprocessedTxs.map { it.forcedTransactionNumber })
+        this.sendTransactions(unprocessedTransactions = unprocessedTxs.take(txLimitToSendPerTick))
```
Filter94: `EthLogsFilterPoller` should guarantee the order and that there are no events skipped, but just in case, I'd like to take precaution measures to guarantee FTXs aren't skipped and go in order. Either here or in the `ForcedTransactionsL1EventsFetcher`

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsSenderForExecution.kt
```diff
@@ -0,0 +1,75 @@
...
+      }
```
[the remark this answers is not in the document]
Filter94: Yes, the items should be taken out by `ForcedTransactionsSenderForExecution` from the channel

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsL1EventsFetcher.kt
fluentcrafter: on the log is irrelevant, but on other fields I like to see what's object's state (`this.state`) vs methods temporal state

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsSenderForExecution.kt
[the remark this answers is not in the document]
fluentcrafter: I did an overall redesign to address multiple things. 

Queue's cleanup is responsibility of `ForcedTransactionsStatusUpdater` that already provides unprocessed transactions and in order.

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsSenderForExecution.kt
fluentcrafter: I added an invariant checker inside `ForcedTransactionsL1EventsFetcher` it logs an error and stops processing in case it's broken...

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsSenderForExecution.kt
fluentcrafter: Moved the approach to a queue. Please take a look...

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsExecutionStatusUpdater.kt
```diff
@@ -0,0 +1,156 @@
...
+
+  /**
+   * Process transactions sequentially in order using iteration.
+   * For each processed transaction, remove it from the queue.
+   * Stop at the first unprocessed transaction.
+   */
+  private fun processConsecutiveTransactions(
+    ftxs: List<ForcedTransactionAddedEvent>,
+  ): SafeFuture<List<ForcedTransactionAddedEvent>> {
+    if (ftxs.isEmpty()) {
```
jonesho: nit: I dun think this "empty list check and return" is even needed, the code below already handles that

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsL1EventsFetcher.kt
```diff
@@ -0,0 +1,145 @@
...
+  private val log: Logger = LogManager.getLogger(ForcedTransactionsL1EventsFetcher::class.java),
+) : LongRunningService {
+  private lateinit var eventsSubscription: EthLogsFilterSubscriptionManager
+  private var nextExpectedFtx = AtomicLong(0)
+
+  fun getStartBlockNumberToListenForLogs(): SafeFuture<BlockParameter> {
+    return resumePointProvider
+      .lastFinalizedForcedTransaction()
+      .thenCompose { finalizedForcedTransactionNumber ->
+        nextExpectedFtx.exchange(finalizedForcedTransactionNumber.toLong() + 1)
```
jonesho: nit: Why not just use `store` if we're not concerning to get the `nextExpectedFtx` old value

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsExecutionStatusUpdater.kt
```diff
@@ -0,0 +1,156 @@
...
+      return SafeFuture.completedFuture(emptyList())
+    }
+
+    var currentFuture = SafeFuture.completedFuture(ftxs)
+
+    for (index in ftxs.indices) {
+      currentFuture = currentFuture.thenCompose { remaining ->
+        if (remaining.isEmpty() || remaining.first().forcedTransactionNumber != ftxs[index].forcedTransactionNumber) {
+          // Already stopped processing in previous iteration
+          SafeFuture.completedFuture(remaining)
```
jonesho: In this case, why don't we just break the outer "for loop"?

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsResumePointProvider.kt
```diff
@@ -0,0 +1,36 @@
...
+          )
+          0UL
+        } else {
+          throw th
+        }
+      }
+  }
+
+  fun getLastProcessedForcedTransactionNumber(): SafeFuture<ULong> {
+    return lastFinalizedForcedTransaction()
```
jonesho: Should we further optimize by returning the max of "highest ftx number from the db" and "lastFinalizedForcedTransaction()" as the `LastProcessedForcedTransactionNumber`?

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsResumePointProvider.kt
```diff
@@ -0,0 +1,36 @@
...
+  }
```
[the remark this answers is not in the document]
Filter94: Yes, I don't get its purpose either

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsExecutionStatusUpdater.kt
fluentcrafter: Because this is the Final Iteration of the loop anyway and we shall return Future.

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsApp.kt
```diff
@@ -0,0 +1,100 @@
...
+  private lateinit var ftxSender: ForcedTransactionsSenderForExecution
+
+  override fun start(): CompletableFuture<Unit> {
+    log.debug("starting ForcedTransactionsApp")
+    // getLastProcessedForcedTransactionNumber relies on APi
+    // that should have request retries
+    return ftxResumePointProvider
+      .getLastProcessedForcedTransactionNumber()
+      .thenCompose { lastProcessedForcedTransactionNumber ->
+        val lastProcessedFtxProvider = ForcedTransactionsResumePointProvider {
```
Filter94, line 54: Either this is a mistake or I don't get the idea. This `lastProcessedFtxProvider` that always returns the same thing is then passed to `ForcedTransactionsL1EventsFetcher`. This means that `ForcedTransactionsL1EventsFetcher` will always return events starting from the same FTX number. They will be filtered out later on, but it still looks like a not intended way

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsExecutionStatusUpdater.kt
```diff
@@ -0,0 +1,150 @@
...
+    log.debug(
+      "filtering FTXs: nextExpectedFtxNumber={}, queuedFtxs={}",
+      nextExpectedFtxNumber,
+      ftxs.map { it.forcedTransactionNumber },
+    )
+
+    // Sort by ftxNumber and only consider transactions >= expectedNum
+    val sortedFtxs = ftxs
+      .filter { it.forcedTransactionNumber >= nextExpectedFtxNumber }
+      .sortedBy { it.forcedTransactionNumber }
```
Filter94, line 47: Since we're using a queue and `ForcedTransactionsL1EventsFetcher` is quite strict on the order, we shouldn't need the sorting here

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsExecutionStatusUpdater.kt
```diff
@@ -0,0 +1,150 @@
...
+
+    // Find consecutive transactions starting from expectedNum (no gaps)
```
Filter94, line 49: Is this just a precaution or you foresee some cases in which `ForcedTransactionsL1EventsFetcher` can return gaps?

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsExecutionStatusUpdater.kt
```diff
@@ -0,0 +1,150 @@
...
+                )
+                remaining
+              }
+            }
+        }
+      }
+    }
+
+    return currentFuture
+  }
```
Filter94: Minor: 
I feel like we're manually managing the call stack here in `currentFuture`
```suggestion
  private fun processConsecutiveTransactions(
    ftxs: List<ForcedTransactionAddedEvent>,
  ): SafeFuture<List<ForcedTransactionAddedEvent>> {
    fun processRecursively(remaining: List<ForcedTransactionAddedEvent>): SafeFuture<List<ForcedTransactionAddedEvent>> {
      if (remaining.isEmpty()) {
        return SafeFuture.completedFuture(emptyList())
      }

      return processTransaction(remaining.first()).thenCompose { wasProcessed ->
        if (wasProcessed) {
          processRecursively(remaining.drop(1))
        } else {
          SafeFuture.completedFuture(remaining)
        }
      }
    }

    return processRecursively(ftxs)
  }

  private fun processTransaction(ftx: ForcedTransactionAddedEvent): SafeFuture<Boolean> {
    return isAlreadyProcessed(ftx).thenApply { alreadyProcessed ->
      if (alreadyProcessed) {
        ftxQueue.remove(ftx)
        nextExpectedFtxNumber = ftx.forcedTransactionNumber + 1uL
        log.debug(
          "FTX #{} already processed, removed from queue. Next expected: {}",
          ftx.forcedTransactionNumber,
          nextExpectedFtxNumber,
        )
        true
      } else {
        log.debug(
          "FTX #{} not yet processed, stopping sequential check",
          ftx.forcedTransactionNumber,
        )
        false
      }
    }
  }
```
Note that it doesn't really build up the stack on any thread as long as `isAlreadyProcessed` is asynchronous. It does build up `SafeFuture` objects in memory though

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsL1EventsFetcher.kt
```diff
@@ -0,0 +1,147 @@
...
+  private val ethLogsFilterSubscriptionFactory: EthLogsFilterSubscriptionFactory,
+  private val l1EarliestBlock: BlockParameter = BlockParameter.Tag.EARLIEST,
+  private val l1HighestBlock: BlockParameter = BlockParameter.Tag.FINALIZED,
+  private val ftxQueue: Queue<ForcedTransactionAddedEvent>,
+  private val log: Logger = LogManager.getLogger(ForcedTransactionsL1EventsFetcher::class.java),
+) : LongRunningService {
+  private lateinit var eventsSubscription: EthLogsFilterSubscriptionManager
+  private var nextExpectedFtx = AtomicLong(0)
+
+  fun getStartBlockNumberToListenForLogs(): SafeFuture<BlockParameter> {
```
Filter94: Minor: I think it can be private

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsL1EventsFetcher.kt
```diff
@@ -0,0 +1,149 @@
...
+          nextExpectedFtx.decrementAndFetch()
+        }
+    } else {
+      log.error(
+        "event ForcedTransactionAdded: is out of order. will stop polling for events, expected={}, ftx={}",
+        nextExpectedFtx,
+        event.event,
+      )
+      this.stop()
+    }
```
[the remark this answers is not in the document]
Filter94: It does throw now

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsSenderForExecution.kt
```diff
@@ -0,0 +1,81 @@
...
+        )
+        if (unprocessedFtx.isEmpty()) {
+          SafeFuture.completedFuture(null)
+        } else {
+          this.sendTransactions(unprocessedTransactions = unprocessedFtx.take(txLimitToSendPerTick))
+        }
+      }
+  }
+
+  fun sendTransactions(unprocessedTransactions: List<ForcedTransactionAddedEvent>): SafeFuture<*> {
```
Filter94: Minor: This can be private

coordinator/ethereum/forced-transactions/src/test/kotlin/linea/ftx/FakeForcedTransactionsClient.kt
```diff
@@ -0,0 +1,99 @@
...
+      ftxNumber = ftxNumber,
+      blockNumber = l2BlockNumber,
+      blockTimestamp = Clock.System.now(),
+      inclusionResult = inclusionResult,
+      ftxHash = ByteArray(0),
+      from = ByteArray(0),
+    )
+  }
+
+  fun setFtxInclusionResult(
```
Filter94, line 48: Since `ftxReceived`, `ftxInclusionResults` and `ftxInclusionResultsAfterReception` are public, do we really need dedicated setters?

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsApp.kt
fluentcrafter: The idea is that all the subcomponent (ForcedTransactionsStatusUpdater &ForcedTransactionsL1EventsFetcher) are `start()`ed from same resume point. Otherwise, becasue of async code, if there was a finalization on L1 between the start of either one they would have different `lastProcessedForcedTransactionNumber`. 

In ForcedTransactionsL1EventsFetcher, that field is only used once on `.start`, to subscribe for the events, not on evey tick..

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsExecutionStatusUpdater.kt
fluentcrafter: It's a precaution: I want to ensure the invariant of order from the protocol of ftx are always sent by order

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsExecutionStatusUpdater.kt
fluentcrafter: In practice, we don't need sorting here yes. However, queue guaranties the FIFO order, not priority order and this component has no visibility of insertion order will match `ftxNumber`. So I rather keep it for precaution, sinse performance it's not a concern at all for FTXs

coordinator/ethereum/forced-transactions/src/test/kotlin/linea/ftx/FakeForcedTransactionsClient.kt
fluentcrafter: "need" no, but it makes the code less verbose/duplicated at caller site. That's the goal

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsExecutionStatusUpdater.kt
Filter94: I'm not highly against these safety nets, but I think that it would make things simpler if we'd just stick to supplier-side ordering guarantees, which we already have. Or we can make it safer by using PriorityQueue at some performance cost

> this component has no visibility

This kind of reasoning is a sign of low cohesion, which isn't nice, because the new components in this PR are meant for the same goal and to be used in the same module. I don't think it's practical to treat components using the same `ftxQueue` as decoupled things unaware of each other.

fluentcrafter: Ok, I will come back ti later if I have time. However, I would the to keep order invariant assurance at one single class and this is the best one given it recociles events from L1 with DB and Sequencer

PR #2202, opened 2026-01-30
fix: merge issues with 532e3403

ivokub:
Fixes a few issues created during merging 532e3403095c001d1613175079640e0eb256878a which prevents compilation and running of tests.

### Checklist

* [] I wrote new tests for my new core changes.

---

> [!NOTE]
> **Medium Risk**
> Touches full proving output construction (`execution.MakeProof` IOP argument) and gnark circuit compilation in tests; if mismatched, proofs or verification could fail at runtime. Also removes step-level verifier logging, reducing visibility when debugging constraint generation.
> 
> **Overview**
> Fixes full execution proof construction to pass `fullZkEvm.InitialCompiledIOP` into `execution.MakeProof` (instead of `WizardIOP`), aligning the proof’s spec with the prover’s compiled IOP.
> 
> Updates the Vortex gnark verifier test to compile KoalaBear circuits using `gnarkutil.NewMockBuilder(scs.NewBuilder)`.
> 
> Removes per-step `logrus`/timing/constraint-count logging from `VerifierCircuit.Verify`, leaving verification logic unchanged but quieter.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5bab6bc5522db83ba597fddf61c655a860834b76. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/prover/protocol/wizard/gnark_verifier.go
+++ b/prover/protocol/wizard/gnark_verifier.go
@@ -3,7 +3,6 @@ package wizard
 import (
 	"encoding/json"
 	"fmt"
-	"time"
 
 	"github.com/consensys/gnark/frontend"
 	"github.com/consensys/linea-monorepo/prover/crypto/fiatshamir"
@@ -17,7 +16,6 @@ import (
 	"github.com/consensys/linea-monorepo/prover/utils"
 	"github.com/consensys/linea-monorepo/prover/utils/collection"
 	"github.com/consensys/linea-monorepo/prover/utils/gnarkutil"
-	"github.com/sirupsen/logrus"
 )
 
 // GnarkRuntime is the interface implemented by the struct [VerifierCircuit]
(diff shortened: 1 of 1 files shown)


prover/protocol/compiler/vortex/gnark_verifier_test.go
```diff
@@ -174,7 +175,8 @@ func runVortexGnarkVerifier(t *testing.T, isBLS bool) {
 		}
 
 	} else {
-		cs, err := frontend.CompileU32(koalabear.Modulus(), scs.NewBuilder,
```
AlexandreBelling, line 177: Do you know why the normal builder was no fit here? Do we use PlonkInWizard in that test?

AlexandreBelling: Ah, is it because it tries to use the range-checks?

ivokub: scs.NewBuilder returns a builder which doesn't have `WideCommit`. So it already fails at compile time otherwise.

ivokub: Hmm, but now thinking I'm not sure why we would need it though if we're doing only native operations. Let me see further

prover/protocol/wizard/gnark_verifier.go
```diff
@@ -371,13 +369,8 @@ func (c *VerifierCircuit) Verify(api frontend.API) {
 
 		c.GenerateCoinsForRound(api, round)
 
-		for k, step := range roundSteps {
```
AlexandreBelling, line 374: I removed the attempt to compare constraints (it did not work and still got committed unfortunately). But the the time thing was helpful actually as it helped me track the slow parts of the gnark compilation and that hinted me to the best parts to optimize. We could demote the log to Debug though

ivokub: The issue is that `GetNbConstraints()` is not defined on the api instance. It is a method on compiled constraint system. So the assertion cannot hold in-circuit.

PR #2204, opened 2026-01-30
Prover/koala optimize lagrange batch evaluation

AlexandreBelling:
This PR implements issue(s) #

### Checklist

---

> [!NOTE]
> **High Risk**
> High risk because it changes core cryptographic circuit/prover code paths (Fiat-Shamir/Poseidon2 hashing, global-constraint compilation, MPTS evaluation, and recursion public inputs), which can affect proof correctness and verification compatibility.
> 
> **Overview**
> **Main impact:** Updates several core prover/circuit building blocks (Koala gnark hashing, Lagrange evaluation, global-constraint compilation, and recursion wiring) to reduce constraint cost and enable new optimizations.
> 
> Introduces a new Koala gnark hasher abstraction (`GnarkKoalaHasher`/`GnarkOctuplet`) and threads it through Fiat-Shamir, Poseidon2, multiset hashing, SMT/vortex circuits, and hasher factories (including external-hasher support); removes older WV-specific wrappers and updates tests accordingly.
> 
> Optimizes `fastpolyext.BatchEvaluateLagrangeGnark` by switching to pointer-based polynomial inputs, lazily computing barycentric terms, and aggressively skipping constant-zero terms; updates MPTS verifier/evaluation mapping to avoid copies (new `GetColAssignmentGnarkExtAsPtr` helpers) and runs `splitextension` immediately after MPTS.
> 
> Adds a standalone degree-reduction compiler pass (`protocol/compiler/degreereduction`) and integrates it into `compiler.Arcane` (and adjusts global constraint compilation/evaluation inputs to handle base-vs-extension values more directly).
> 
> Recursion/proving is updated to handle mixed base/extension public inputs by flattening extension elements into 4 coordinates, to optionally use an external Fiat-Shamir hasher factory, and to rename/retarget some recursion witness/state fields; `prove` now builds proofs using `RecursionCompiledIOP`.
> 
> Smaller changes include registering a SIGUSR1 stacktrace dump handler in `prove`/`setup` commands and tweaking a sample config (`prover_mode`, module name).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c47db42872b035a06227e45a19c99245dc068513. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/prover/protocol/compiler/globalcs/degree_reduction.go
+++ b/prover/protocol/compiler/globalcs/degree_reduction.go
@@ -117,11 +117,11 @@ func degreeReduce(comp *wizard.CompiledIOP, degreeBound int) *DegreeReductionSte
 
 		sem <- struct{}{}
 		exprCompilationWG.Add(1)
-		go func() {
+		go func(i int) {
 			degRedStep.NewColumnsExpressions[i].Compile()
 			exprCompilationWG.Done()
 			<-sem
-		}()
+		}(i)
 
 		degRedStep.DegreeReducedExpression = append(
 			degRedStep.DegreeReducedExpression,
--- a/prover/protocol/compiler/recursion/circuit.go
+++ b/prover/protocol/compiler/recursion/circuit.go
@@ -29,7 +29,6 @@ type RecursionCircuit struct {
 	Commitments        [][blockSize]frontend.Variable `gnark:",public"`
 	Pubs               []frontend.Variable            `gnark:",public"`
 	WizardVerifier     *wizard.VerifierCircuit
-	withoutGkr         bool                       `gnark:"-"`
 	withExternalHasher bool                       `gnark:"-"`
 	PolyQuery          query.UnivariateEval       `gnark:"-"`
 	MerkleRoots        [][blockSize]ifaces.Column `gnark:"-"`
@@ -55,7 +54,7 @@ func Ext4FV(x koalagnark.Ext) ExtFrontendVariable {
 
 // AllocRecursionCircuit allocates a new RecursionCircuit with the
 // given parameters.
-func AllocRecursionCircuit(comp *wizard.CompiledIOP, withoutGkr bool, withExternalHasher bool) *RecursionCircuit {
+func AllocRecursionCircuit(comp *wizard.CompiledIOP, withExternalHasher bool) *RecursionCircuit {
 	var (
 		pcsCtx      = comp.PcsCtxs.(*vortex.Ctx)
 		polyQuery   = pcsCtx.Query
@@ -84,7 +83,6 @@ func AllocRecursionCircuit(comp *wizard.CompiledIOP, withoutGkr bool, withExtern
 	}
 
 	return &RecursionCircuit{
-		withoutGkr:         withoutGkr,
 		withExternalHasher: withExternalHasher,
 		PolyQuery:          polyQuery,
 		MerkleRoots:        merkleRoots,
--- a/prover/protocol/compiler/recursion/recursion.go
+++ b/prover/protocol/compiler/recursion/recursion.go
@@ -109,9 +109,6 @@ type Parameters struct {
 	// and queries of the recursion.
 	Subscript string
 
-	// WithoutGkr is a flag that indicates if the recursion is run without GKR
-	WithoutGkr bool
-
 	// MaxNumProof controls the number of proofs to be recursed at once. It
 	// is a mandatory parameter and must be a positive integer.
 	MaxNumProof int
@@ -171,12 +168,8 @@ func DefineRecursionOf(comp, inputComp *wizard.CompiledIOP, params Parameters) *
 		plonkOpts = append(plonkOpts, plonkinternal.WithSubscript(params.Subscript))
 	}
 
-	if !params.WithoutGkr && params.WithExternalHasherOpts {
-		utils.Panic("inconsistent choice, cannot use GKR and external hasher together")
-	}
-
 	var (
-		plonkCircuit = AllocRecursionCircuit(inputComp, params.WithoutGkr, params.WithExternalHasherOpts)
+		plonkCircuit = AllocRecursionCircuit(inputComp, params.WithExternalHasherOpts)
 		plonkCtx     = plonkinternal.PlonkCheck(comp, params.Name, 0, plonkCircuit, params.MaxNumProof, plonkOpts...)
 		vortexCtxs   = make([]*vortex.Ctx, params.MaxNumProof)
 		numYs        = len(plonkCircuit.Ys)
--- a/prover/protocol/compiler/recursion/recursion_test.go
+++ b/prover/protocol/compiler/recursion/recursion_test.go
@@ -101,7 +101,6 @@ func TestLookup(t *testing.T) {
 			define2 := func(build2 *wizard.Builder) {
 				recCtx = DefineRecursionOf(build2.CompiledIOP, comp1, Parameters{
 					Name:        "test",
-					WithoutGkr:  true,
 					MaxNumProof: 1,
 				})
 			}
@@ -163,7 +162,6 @@ func TestWithMaxNumProofNotFilled(t *testing.T) {
 	define2 := func(build2 *wizard.Builder) {
 		recCtx = DefineRecursionOf(build2.CompiledIOP, comp1, Parameters{
 			Name:        "test",
-			WithoutGkr:  true,
 			MaxNumProof: 2,
 		})
 	}
@@ -224,7 +222,6 @@ func TestWithMaxNumProofFilled(t *testing.T) {
 	define2 := func(build2 *wizard.Builder) {
 		recCtx = DefineRecursionOf(build2.CompiledIOP, comp1, Parameters{
 			Name:        "test",
-			WithoutGkr:  true,
 			MaxNumProof: 2,
 		})
 	}
@@ -317,7 +314,6 @@ func TestWithPublicInputs(t *testing.T) {
 	define2 := func(build2 *wizard.Builder) {
 		recCtx = DefineRecursionOf(build2.CompiledIOP, comp1, Parameters{
 			Name:        "test",
-			WithoutGkr:  true,
 			MaxNumProof: 1,
 		})
 	}
--- a/prover/protocol/distributed/conglomeration_hierarchical.go
+++ b/prover/protocol/distributed/conglomeration_hierarchical.go
@@ -314,7 +314,6 @@ func (c *ModuleConglo) Compile(comp *wizard.CompiledIOP, moduleMod *wizard.Compi
 
 	c.Recursion = recursion.DefineRecursionOf(comp, moduleMod, recursion.Parameters{
 		Name:                   "conglomeration",
-		WithoutGkr:             true,
 		MaxNumProof:            2,
 		WithExternalHasherOpts: true,
 	})
--- a/prover/protocol/distributed/segment_compilation.go
+++ b/prover/protocol/distributed/segment_compilation.go
@@ -283,7 +283,6 @@ func CompileSegment(mod any, params CompilationParams) *RecursedSegmentCompilati
 			modIOP,
 			recursion.Parameters{
 				Name:                   "wizard-recursion",
-				WithoutGkr:             true,
 				MaxNumProof:            1,
 				FixedNbRowPlonkCircuit: params.FixedNbRowPlonkCircuit,
 				WithExternalHasherOpts: true,
--- a/prover/protocol/serde/test/io_test.go
+++ b/prover/protocol/serde/test/io_test.go
@@ -422,7 +422,6 @@ func getSerdeTestCases() []serdeTestCase {
 				rec := wizard.NewCompiledIOP()
 				recursion.DefineRecursionOf(rec, wiop, recursion.Parameters{
 					MaxNumProof: 1,
-					WithoutGkr:  true,
 					Name:        "recursion",
 				})
 
--- a/prover/protocol/serde/test/serde_test.go
+++ b/prover/protocol/serde/test/serde_test.go
@@ -408,7 +408,6 @@ func TestSerdeValue(t *testing.T) {
 				rec := wizard.NewCompiledIOP()
 				recursion.DefineRecursionOf(rec, wiop, recursion.Parameters{
 					MaxNumProof: 1,
-					WithoutGkr:  true,
 					Name:        "recursion",
 				})
 
--- a/prover/protocol/serde/test/zkevm_iop_test.go
+++ b/prover/protocol/serde/test/zkevm_iop_test.go
@@ -436,7 +436,6 @@ var serdeScenarios = []serdeScenario{
 			define2 := func(build2 *wizard.Builder) {
 				recursion.DefineRecursionOf(build2.CompiledIOP, comp1, recursion.Parameters{
 					Name:        "test",
-					WithoutGkr:  true,
 					MaxNumProof: 1,
 				})
 			}
--- a/prover/protocol/serialization/compiled_iop_test.go
+++ b/prover/protocol/serialization/compiled_iop_test.go
@@ -366,7 +366,6 @@ var serdeScenarios = []serdeScenario{
 			define2 := func(build2 *wizard.Builder) {
 				recursion.DefineRecursionOf(build2.CompiledIOP, comp1, recursion.Parameters{
 					Name:        "test",
-					WithoutGkr:  true,
 					MaxNumProof: 1,
 				})
 			}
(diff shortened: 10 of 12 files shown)


prover/circuits/execution/builder.go
```diff
@@ -23,6 +24,9 @@ func (b *builder) Compile() (constraint.ConstraintSystem, error) {
 // Makes the constraint system for the execution circuit
 func makeCS(z *zkevm.ZkEvm) constraint.ConstraintSystem {
 
+	p := profile.Start(profile.WithPath("execution-circuit.pprof"))
```
ivokub: Should omit

prover/zkevm/full.go
```diff
@@ -83,6 +83,7 @@ var (
 		compiler.Arcane(
 			compiler.WithStitcherMinSize(16),
 			compiler.WithTargetColSize(1<<22),
+			compiler.GenCSVAfterExpansion("zkevm_first_compilation.csv"),
```
ivokub: Should omit?

prover/cmd/prover/cmd/prove.go
```diff
@@ -25,6 +26,10 @@ type ProverArgs struct {
 // Prove orchestrates the proving process based on the job type
 func Prove(args ProverArgs) error {
 
+	// This allows the user to dump stacktraces by sending a SIGUSR1 to the
+	// current process.
+	signal.RegisterStackTraceDumpHandler()
```
AlexandreBelling, line 31: Thanks to that, if you send a SIGUSR2 to the process. It will generate a 10 sec timeslice of pprof CPU profiling. Very handy, when the prover is stuck doing something somewhere and you would like to know where and why

prover/zkevm/full.go
```diff
@@ -122,22 +129,32 @@ var (
...
+	// This is the compilation suite in use for the full prover *after* the
+	// recursion step.
 	fullSecondCompilationSuite = CompilationSuite{
 		cleanup.CleanUp,
 		poseidon2.CompilePoseidon2,
-		compiler.Arcane(compiler.WithTargetColSize(1 << 18)),
+		compiler.Arcane(
+			compiler.WithTargetColSize(1<<22),
+			compiler.WithStitcherMinSize(16),
+			compiler.GenCSVAfterExpansion("zkevm_second_compilation.csv"),
```
ivokub: Should omit?

prover/crypto/fiatshamir_koalabear/fiatshamir_circuit.go
```diff
@@ -4,33 +4,42 @@ import (
...
 	"github.com/consensys/gnark/frontend"
+	"github.com/consensys/linea-monorepo/prover/crypto/hasherfactory_koalabear"
 	"github.com/consensys/linea-monorepo/prover/crypto/poseidon2_koalabear"
 	"github.com/consensys/linea-monorepo/prover/maths/field/koalagnark"
 	"github.com/consensys/linea-monorepo/prover/utils"
 )
 
-type GnarkFS struct {
-	hasher poseidon2_koalabear.GnarkMDHasher
+type GnarkFSWV struct {
```
AlexandreBelling, line 13: I noticed that GnarkFS was never used. So I deleted it

prover/config/config-mainnet-limitless.toml
```diff
@@ -80,7 +80,7 @@ modules = [
     {module = "RLP_TXN", limit = 131072, limit_large = 262144},
     {module = "RLP_TXN_RCPT", limit = 65536, limit_large = 131072},
     {module = "ROM", limit = 8388608, limit_large = 8388608},
-    {module = "ROM_LEX", limit = 1024, limit_large = 2048},
+    {module = "ROMLEX", limit = 1024, limit_large = 2048},
```
AlexandreBelling, line 83: Without it, it gives a very large column size for ROMLEX and this implies a lot of expensive hashing

prover/backend/execution/prove.go
```diff
@@ -163,7 +163,7 @@ func mustProveAndPass(
 		}
 
 		// TODO: implements the collection of the functional inputs from the prover response
-		return execution.MakeProof(traces, setup, fullZkEvm.WizardIOP, proof, *w.FuncInp), setup.VerifyingKeyDigest()
+		return execution.MakeProof(traces, setup, fullZkEvm.RecursionCompiledIOP, proof, *w.FuncInp), setup.VerifyingKeyDigest()
```
AlexandreBelling, line 166: The fix to the compile error

prover/crypto/poseidon2_koalabear/poseidon2_circuit.go
```diff
@@ -8,7 +8,19 @@ import (
 	"github.com/consensys/gnark/frontend"
 )
 
-type Octuplet [8]frontend.Variable
+// GnarkKoalaHasher is an interface implemented by structures that can compute
+// Poseidon2 hashes based on koalabear.
+type GnarkKoalaHasher interface {
```
AlexandreBelling, line 13: I needed an interface to be able to interchange the various type of poseidon hashers in FS

prover/crypto/vortex/vortex_koalabear/verifier_circuit.go
```diff
@@ -10,7 +10,7 @@ import (
 // Check the merkle proof opening (merkleProofs[i][j], root[i]) for columns[i][j].
 // The leaves are poseidon2_koalabear(columns[i][j])
 func GnarkCheckColumnInclusionNoSis(api frontend.API, columns [][][]koalagnark.Element,
-	merkleProofs [][]smt_koalabear.GnarkProof, roots []poseidon2_koalabear.Octuplet) error {
+	merkleProofs [][]smt_koalabear.GnarkProof, roots []poseidon2_koalabear.GnarkOctuplet) error {
```
AlexandreBelling, line 13: I found that poseidon2_koalabear.Octuplet was misleading. And still think that we should move GnarkOctuplet to a different package as it is not necessarily a poseidon thing

prover/maths/field/koalagnark/api.go
```diff
@@ -120,6 +120,25 @@ func (a *API) Add(x, y Element) Element {
...
+		}
+		return Element{V: res}
+	}
+
+	toSum := make([]*emulated.Element[emulated.KoalaBear], len(xs))
+	for i := range xs {
+		toSum[i] = &xs[i].EV
+	}
+
+	sum := a.emulatedAPI.Sum(toSum...)
```
ivokub, line 138: If this approach works now, then good, but I actually I found a bug in the implementation inside the field emulation package. In gnark the method never reduces internally - if the initial overflows are small then it shouldn't be an issue as the overflow only grows by `log_2(nbInputs)`, but if we're already close to maximum overflow, then may overflow and we have failed checks.

prover/maths/field/koalagnark/api.go
```diff
@@ -251,6 +279,15 @@ func (a *API) AssertIsLessOrEqual(x, y Element) {
 	}
 }
 
+// AssertIsBoolean constrains x == 0 or x == 1.
+func (a *API) AssertIsBoolean(x Element) {
+	if a.IsNative() {
+		a.nativeAPI.AssertIsBoolean(x.Native())
+	} else {
+		a.emulatedAPI.AssertIsEqual(a.emulatedAPI.Mul(&x.EV, &x.EV), &x.EV)
```
ivokub, line 278: Should use MulNoReduce. I still haven't made the default to use `MulNoReduce` when emulating a small field.

This method seems to be used, but dunno if it is the most optimal way of doing it? If only a few times, then I think we can leave as is.

prover/maths/field/koalagnark/api.go
AlexandreBelling: I have not tested on supermassive columns =/. I am not sure how much overflow overhead does [MulExtNoReduceIncur]. IIRC, each step of the tower gives a little above ~31 bits of overhead so we should be fine. I don't think this should be commented in the code as this is mainly a matter of letting gnark add tighter overflow control.

AlexandreBelling: Maybe stating what the expected bound on the overflow are would be helpful though

prover/maths/field/koalagnark/api.go
AlexandreBelling: Not used a lot but still good to optimize as this is easy

prover/maths/field/koalagnark/api.go
ivokub: Yep, here we can leave as is. It is bug on gnark side which needs to be fixed. But I have encountered by using this approach that standard benchmark tests start to fail. There we do already several multiplications in a row, but the inputs imo are already with bigger overflow i.e. 30->90->220 (values 31 bits -> 61 bit -> 128 bit -> 256 bit)

prover/maths/common/fastpolyext/gnark_lagrange.go
```diff
@@ -118,25 +155,41 @@ func BatchEvaluateLagrangeGnark(api frontend.API, polys [][]koalagnark.Ext, x ko
...
 		for k := 0; k < n; k++ {
-			// this saves constraints when the constant term is zero.
-			if polyKConst, isConst := koalaAPI.ConstantValueOfExt(poly[k]); isConst && polyKConst.IsZero() {
+
+			if poly[k] == lastSeenZeroPtr {
 				continue
 			}
 
+			// This saves constraints when the constant term is zero.
+			if polyKConst, isConst := koalaAPI.ConstantValueOfExt(*poly[k]); isConst && polyKConst.IsZero() {
```
ivokub: I don't completely understand why we can do this thing - the variables are supposed to be non-constants. And if they are constants, then why mix with variables?

There is similar implementation in `prover/crypto/vortex/verifier_circuit_common.go`. And the point is that we actually already know how many zeros we're expected to have as a tail. We can just omit evaluating the tail and this is sufficient.

prover/maths/field/koalagnark/element.go
```diff
@@ -109,28 +110,58 @@ func (o Octuplet) NativeArray() [8]frontend.Variable {
...
+// In case the
+var (
+	zeroKoalaFr = field.Zero()
+	oneKoalaFr  = field.One()
+	oneBigInt   = big.NewInt(1)
+)
+
 // ConstantValueOf returns true if the variable represent a constant value and
 // returns the non-nil value if so.
 func (api API) ConstantValueOfElement(v Element) (*field.Element, bool) {
```
ivokub: If we remove the constant check in fast Lagrange batch eval then can also remove this. Imo we shouldn't go too much between constant->Variable/Element->constant. If we know something is constant then lets use it directly instead. Otherwise I am afraid it just can cause soundness issues (we use constants in place of variables)

prover/maths/field/koalagnark/api.go
```diff
@@ -144,6 +163,15 @@ func (a *API) Mul(x, y Element) Element {
 	return Element{EV: *a.emulatedAPI.MulNoReduce(x.Emulated(), y.Emulated())}
 }
 
+// MulNoReduce returns x * y but not reduced (in case of emulated mode). In
+// native mode it does the same as Mul.
+func (a *API) MulNoReduce(x, y Element) Element {
```
ivokub: We shouldn't need it imo. Right now `API.Mul` above uses already `Field.MulNoReduce`. Normally - gnark should choose most optimal implementation based on overflow natively, we just haven't been able to fix it there yet when we emulate a small field.

prover/maths/field/koalagnark/ext.go
```diff
@@ -187,6 +187,24 @@ func (a *API) e2Mul(x, y E2) E2 {
 	return E2{A0: a0, A1: a1}
 }
 
+// e2MulNoReduce returns x * y but not reduced (in case of emulated mode). In
+// native mode it does the same as Mul.
+func (a *API) e2MulNoReduce(x, y E2) E2 {
```
ivokub: I think can remove. `API.Mul` already uses `MulNoReduce` right now.

prover/protocol/column/verifiercol/from_accessors.go
```diff
@@ -96,8 +96,24 @@ func (f FromAccessors) GetColAssignmentGnarkExt(run ifaces.GnarkRuntime) []koala
...
+	padding := koalagnark.NewExt(f.Padding)
 	for i := len(f.Accessors); i < f.Size_; i++ {
-		res[i] = koalagnark.NewExt(f.Padding)
+		res[i] = padding
+	}
+
+	return res
+}
+
+func (f FromAccessors) GetColAssignmentGnarkExtAsPtr(run ifaces.GnarkRuntime) []*koalagnark.Ext {
```
ivokub, line 107: I have been also noted down that we should by default use reference to `koalagnark.Ext` everywhere to ensure we don't accidentally copy the embedded `emulated.Element` which causes sub-optimality with overflow tracking etc.

So - lets keep this method right now, but for future reference we should remove `GetColAssignmentGnarkExt` which returns slice of values

prover/protocol/compiler/globalcs/compile.go
```diff
@@ -22,10 +21,14 @@ const (
...
 
-	logrus.Trace("started global constraint compiler")
-	defer logrus.Trace("finished global constraint compiler")
+	degreeRed := DegreeReduce(comp, 5)
+	comp.RegisterProverAction(degreeRed.MaxRound, degreeRed)
 
-	merging, anyCs := accumulateConstraints(comp)
+	if degreeRed == nil {
+		return
+	}
```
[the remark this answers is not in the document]
ivokub: Yup, should move nil check a few lines up

prover/maths/common/fastpolyext/gnark_lagrange.go
AlexandreBelling: The reason is that columns can have specific shapes and sometime this is statically enforced in the wizard. For instance, the columns can [xxxxx0000000] and this is in fact the overwhelming majority of the columns we have to manually evaluate during the MPTS part.

An example of such columns are the PI columns of Plonk in Wizard that are resized to 4M but actually have ~100 elements

ivokub: I see. But don't we know before calling the Lagrange Eval how big zero tail we have? We could just evaluate where we slice the tail off.

prover/symbolic/evaluation.go
```diff
@@ -26,7 +27,7 @@ func (b *ExpressionBoard) ListVariableMetadata() []Metadata {
 // Degree returns the overall degree of the expression board. It admits a custom
 // function `getDeg` which is used to assign a degree to the [Variable] leaves
 // of the ExpressionBoard.
-func (b *ExpressionBoard) Degree(getdeg GetDegree) int {
+func (b ExpressionBoard) Degree(getdeg GetDegree) int {
```
ivokub: Mixed receivers usually not recommended

prover/maths/field/koalagnark/api.go
AlexandreBelling: Perhaps, we could consider rolling back when it is done?

ivokub: Yup, lets do it like this.

Actually I'm thinking to implement all the koalagnark stuff natively in gnark's `std/math/emulated` directly. It should simplify significantly the implementation here. But there are still some other ongoing optimizations going on which has delayed it.

AlexandreBelling: Ah, you mean in koalagnark. Ok

prover/protocol/column/verifiercol/from_accessors.go
AlexandreBelling: Yes, but to be honest. The migration to []*koalagnark.Ext is going to extra spicy. So I don't see it being prioritized

AlexandreBelling: In particular, I found that copying koala.Element was costly for the garbage collector

prover/protocol/compiler/globalcs/compile.go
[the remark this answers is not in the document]
AlexandreBelling: Implemented just that for the fix

prover/maths/common/fastpolyext/gnark_lagrange.go
AlexandreBelling: We could know that inside of [Column.GetColAssignmentGnarkExt]. Ideally, we could have return a "sparse" representation of the column. Like [(12, A), (20, B)] to skip the zero values. But that would add complexity as well so I think this is not too bad

AlexandreBelling: But as soon as it returns, we loose this information without "asking" gnark. Perhaps, we would manipulate interface object that would be more explicit though. But that would also add an extra type in the mix

ivokub: Perhaps - we define `IsConstantZero() bool` and `IsConstantExtZero()` methods instead in koalagnark.API which looks at the limb count and returns if it is zero. Then we don't have to construct constants i.e. And imo it would be cleaner.

Or - on gnark side we have isStrictZero method https://github.com/Consensys/gnark/blob/master/std/math/emulated/element.go#L171 and could just make it public?

PR #2220, opened 2026-02-02
fix: add empty data check in Poseidon2

VGau:
This PR implements issue(s) #

### Checklist

---

> [!NOTE]
> **Low Risk**
> Low risk: adds a simple input validation guard and updates tests; only behavior change is for zero-length inputs which now revert instead of returning a zero hash.
> 
> **Overview**
> `Poseidon2.hash` now **rejects empty calldata** by reverting with a new custom error `DataIsEmpty` (via a shared `revert_with_error` helper and selector constant), instead of returning a zero hash.
> 
> Hardhat tests were updated to expect this revert for `0x` inputs while keeping existing mod-32 alignment checks unchanged.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 294f972fa3c11d834b6afebff0754b1e28f7a52b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/contracts/src/libraries/Poseidon2.sol
+++ b/contracts/src/libraries/Poseidon2.sol
@@ -68,11 +68,11 @@ library Poseidon2 {
     assembly {
       let len := _msg.length
       if iszero(len) {
-        error_size_data(DATA_IS_EMPTY_SELECTOR)
+        revert_with_error(DATA_IS_EMPTY_SELECTOR)
       }
 
       if and(len, 31) {
-        error_size_data(DATA_IS_NOT_MOD32_SELECTOR)
+        revert_with_error(DATA_IS_NOT_MOD32_SELECTOR)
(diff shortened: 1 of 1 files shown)


contracts/src/libraries/Poseidon2.sol
```diff
@@ -573,9 +581,9 @@ library Poseidon2 {
         rx := mulmod(x, mulmod(x, x, R_MOD), R_MOD)
       }
 
-      function error_size_data() {
+      function error_size_data(selector) {
```
thedarkjester: I suppose the `size_data` is not so applicable actually with the not mod 32.

One is length, the other is format.

I would rename to `revert_with_error`

Issue #2102, opened 2026-01-21
Coordinator - Prover Client for Invalidity Proof


fluentcrafter:
## Description
Implement Prover client to request Invalidity proofs (check pec document)

Extend aggregations proof client to have list of invalidity proofs

PR #2228, opened 2026-02-02
[Coordinator] Refactor ProofIndex

gauravahuja:
This PR implements issue(s) #2102 

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Medium risk due to a broad API refactor across prover clients and coordination flow; mistakes could break prover routing or file name/path matching for proofs.
> 
> **Overview**
> Refactors `ProofIndex` from a single data class into a marker interface with dedicated types (`ExecutionProofIndex`, `CompressionProofIndex`, `AggregationProofIndex`, `InvalidityProofIndex`) and updates all call sites to use the correct index type.
> 
> Updates the prover client API and file-based prover implementations to be generic over the proof-index type, including `ABProverClientRouter` switching logic and `ProverFileNameProvider` becoming a typed interface with explicit request/response filename generation per proof type.
> 
> Adjusts aggregation, blob compression polling/handlers, backtesting progress reporting, and unit/integration tests to pass/compare the new typed indexes (notably fixing invalidity indexing to use `ftxNumber` and ensuring compression indexes always include a hash).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 37c350836fd035371e4064fa73ccaeafb3e78c2b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/coordinator/core/src/main/kotlin/net/consensys/zkevm/domain/ProofIndex.kt
+++ b/coordinator/core/src/main/kotlin/net/consensys/zkevm/domain/ProofIndex.kt
@@ -1,26 +1,26 @@
 package net.consensys.zkevm.domain
 
-import linea.domain.BlockInterval
+interface ProofIndex
 
-data class ProofIndex(
-  override val startBlockNumber: ULong,
-  override val endBlockNumber: ULong,
-  val hash: ByteArray? = null,
-) : BlockInterval {
+data class ExecutionProofIndex(
+  val startBlockNumber: ULong,
+  val endBlockNumber: ULong,
+) : ProofIndex
+
+data class CompressionProofIndex(
+  val startBlockNumber: ULong,
+  val endBlockNumber: ULong,
+  val hash: ByteArray,
+) : ProofIndex {
   override fun equals(other: Any?): Boolean {
     if (this === other) return true
     if (javaClass != other?.javaClass) return false
 
-    other as ProofIndex
+    other as CompressionProofIndex
 
     if (startBlockNumber != other.startBlockNumber) return false
     if (endBlockNumber != other.endBlockNumber) return false
-    if (hash != null) {
-      if (other.hash == null) return false
-      if (!hash.contentEquals(other.hash)) return false
-    } else if (other.hash != null) {
-      return false
-    }
+    if (!hash.contentEquals(other.hash)) return false
 
     return true
   }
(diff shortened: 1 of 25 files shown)


coordinator/core/src/main/kotlin/net/consensys/zkevm/domain/ProofIndex.kt
```diff
@@ -1,34 +1,65 @@
 package net.consensys.zkevm.domain
 
-import linea.domain.BlockInterval
+interface ProofIndex
```
fluentcrafter, line 3: I recommend a sealed interface. 

Also, I think we could have made this PR/change significantly simpler without compromising on semantic correctness.

```kotlin
sealed interface ProofIndex: BlockInterval

data class BlockIntervalProofIndex(
  override val startBlockNumber: ULong,
  override val endBlockNumber: ULong,
) : ProofIndex

// for CompressionProofIndex and AggregationProofIndex
data class BlockIntervalWithHashProofIndex(
  override val startBlockNumber: ULong,
  override val endBlockNumber: ULong,
  val hash: ByteArray,
) : ProofIndex

data class InvalidityProofIndex(
  val simulatedExecutionBlockNumber: ULong,
  val ftxNumber: ULong,
) : ProofIndex {
  override val startBlockNumber: ULong = simulatedExecutionBlockNumber
  override val endBlockNumber: ULong  = simulatedExecutionBlockNumber
}
```

WDYT?

gauravahuja: > I recommend a sealed interface.

I had that initially but I need a proof Index provider for [GenericFileBasedProverClientTest](coordinator/clients/prover-client/file-based-client/src/test/kotlin/net/consensys/zkevm/coordinator/clients/prover/GenericFileBasedProverClientTest.kt) and it did not feel right for it to  return one of the defined proof indexes for the test. So I created a test specific ProofIndex and thus needed to remove sealed interface.

 

> Also, I think we could have made this PR/change significantly simpler without compromising on semantic correctness.

By having `InvalidityProofIndex` extend `ProofIndex` which is a `BlockInterval`, we are saying that `InvalidityProofIndex` `is a` `BlockInterval` which is not true. Thus I did not go this route even though it would have been easier.

Also, a ProofIndex does not have to a block interval, we could also chose to simply have a hash in any of the proof requests without block numbers. So removing that dependency made sense to me.

fluentcrafter: Your points are valid as well. I am ok with it. will approve it.

fluentcrafter (approved):

PR #2235, opened 2026-02-03
feat: add Solidity smart contract development skill for AI agents

kyzooghost:
This skill codifies the Linea smart contract style guide into structured rule files that AI agents can reference when writing, reviewing, or refactoring Solidity contracts.

| Priority | Category       | Impact   | Rule File                |
| -------- | -------------- | -------- | ------------------------ |
| 1        | Gas optimization    | CRITICAL | `rules/gas-optimization.md`   |
| 2        | NatSpec        | HIGH     | `rules/natspec.md`       |
| 3        | File Layout    | HIGH     | `rules/file-layout.md`   |
| 4        | Naming         | HIGH     | `rules/naming-conventions.md` |
| 5        | Imports        | MEDIUM   | `rules/imports.md`       |
| 6        | Visibility     | MEDIUM   | `rules/visibility.md`    |
| 7        | General Rules  | MEDIUM      | `rules/general-rules.md` |

### Checklist

---

> [!NOTE]
> **Low Risk**
> Primarily adds documentation/agent-skill content with a small `.gitignore` tweak; minimal runtime impact aside from potential changes to what `.claude` files can be committed.
> 
> **Overview**
> Introduces a new `.agents/skills/developing-smart-contracts` skill package (with `SKILL.md`, `metadata.json`, and rule docs) that codifies Linea Solidity conventions around **NatSpec**, **file layout**, **naming**, **imports**, **visibility**, **gas optimization**, plus common commands/checklists for use by AI agents.
> 
> Updates `.gitignore` to stop ignoring the `.claude/` directory broadly and instead ignore only `.claude/settings.local.json`, allowing other `.claude` files to be tracked while keeping local settings uncommitted.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ece79dc33f9ddd762d4dbc3c258c7678f8f0baf6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.agents/skills/developing-smart-contracts/rules/general-rules.md
+++ b/.agents/skills/developing-smart-contracts/rules/general-rules.md
@@ -24,26 +24,22 @@ function getVersion() public pure override returns (string memory) {
 
 ### `public` vs `external` for Overridden Functions
 
-When overriding a base function and calling `super`, the function must be `public` (not `external`) since `super` is an internal call. This is safe when the function has reentrancy guards. Avoid this for functions that check state before modifying it (TOCTOU risk - the state may change between the check and the use).
+Use `public virtual` for overridable functions. `public` allows children to both replace and extend (via `super`), whereas `external virtual` only allows replacement. Ensure functions that can be called internally have reentrancy guards. Avoid making functions `public virtual` if they check state before modifying it (TOCTOU risk - the state may change between the check and the use).
 
 ```solidity
-// Safe: reentrancy guard protects the internal call path
-function deposit() public override nonReentrant {
-  super.deposit();
-  // additional logic
+// Base contract: public virtual allows children to replace or extend
+function abc() public virtual nonReentrant {
+  // base logic
 }
 
-// Unsafe TOCTOU: internal call could bypass stale state checks
-function withdraw(uint256 _amount) public override {
-  require(balances[msg.sender] >= _amount, "Insufficient"); // check
-  super.withdraw(_amount); // super modifies state
-  balances[msg.sender] -= _amount; // use - balance already changed by super
+// Child contract: safe because nonReentrant protects the internal call path
+function abc() public override nonReentrant {
+  super.abc();
+  // additional logic
 }
 ```
 
-If the function is never called internally via `super`, keep it `external` for gas efficiency.
-
-**Note**: Any modifications from audited code should be independently audited.
+Reserve `external` for functions that are not intended to be overridden.
 
 ## OpenZeppelin Dependencies
 


.agents/skills/developing-smart-contracts/rules/file-layout.md
```diff
@@ -0,0 +1,178 @@
+# File Layout
+
+**Impact: HIGH (ensures consistent structure across all contracts)**
+
+## Pragma Version
+
+Choose **one** pragma style and use it consistently across all contracts in the project:
+
+- **Exact version**: `pragma solidity 0.8.33;` — pins to a specific compiler version
+- **Caret version**: `pragma solidity ^0.8.33;` — allows compatible patch updates
```
thedarkjester: it should probably look into overrides before changing everything?

.agents/skills/developing-smart-contracts/rules/file-layout.md
```diff
@@ -0,0 +1,178 @@
...
+// File B: pragma solidity ^0.8.33;
+```
+
+## Interface Structure
+
+```solidity
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.33;
+
+import { ImportType } from "../ImportType.sol";
```
thedarkjester: is there a rule to use the explicit type being imported
`{ ImportType }`

.agents/skills/developing-smart-contracts/rules/gas-optimization.md
```diff
@@ -0,0 +1,126 @@
...
+```
+
+## Use Custom Errors
+
+Custom errors are cheaper than revert strings and should be preferred.
+
+```solidity
+// Correct
+error Unauthorized();
+if (msg.sender != owner) revert Unauthorized();
```
thedarkjester: we should opt for `require` semantics for line count and readabilty

.agents/skills/developing-smart-contracts/rules/gas-optimization.md
```diff
@@ -0,0 +1,126 @@
...
+
+// Incorrect
+require(msg.sender == owner, "Unauthorized");
+```
+
+## Tight Loops and Unchecked Math
+
+- Use `unchecked` only when you can prove the arithmetic cannot overflow.
+- Keep the invariant in a short comment when using `unchecked`.
```
thedarkjester, line 84: this is nuanced and could cause issue, maybe we need to add something to validate against potential complications. The `can prove the arithmetic cannot overflow.` is nice, but I would prefer visibility (a TODO?) to sanity check it maybe

.agents/skills/developing-smart-contracts/rules/gas-optimization.md
```diff
@@ -0,0 +1,126 @@
...
+```
+
+## Short-Circuit Expensive Checks
+
+Order checks to fail early and avoid unnecessary work.
+
+```solidity
+// Correct: cheap check first
+if (_to == address(0)) revert ZeroAddressNotAllowed();
+if (!_isEligible(_to)) revert NotEligible();
```
thedarkjester: we should adopt the require pattern here for consistency

.agents/skills/developing-smart-contracts/rules/imports.md
```diff
@@ -0,0 +1,71 @@
+# Import Conventions
+
+**Impact: MEDIUM (improves code clarity and prevents naming conflicts)**
```
thedarkjester, line 3: we need a rule here to validate imports at the top level still make it clear about some of the underlying imports - e.g. even if I import the LineaRollupBase in the EIP4844 data acceptor that inherits it, the LineaRollup should show it. This helps a developer/AI understand the code quicker.

.agents/skills/developing-smart-contracts/rules/natspec.md
```diff
@@ -0,0 +1,100 @@
...
+ * @notice Thrown when the provided fee is insufficient.
+ * @param provided The fee amount provided.
+ * @param required The minimum fee required.
+ */
+error InsufficientFee(uint256 provided, uint256 required);
+```
+
+## Deprecation
+
+Use `DEPRECATED` in NatSpec for deprecated items:
```
thedarkjester: we can also change visibility and do this for variables being deprecated e.g. 

`uint256 private _someVar_DEPRECATED;`

.agents/skills/developing-smart-contracts/rules/visibility.md
```diff
@@ -0,0 +1,55 @@
...
+Refactor to use internal calls instead:
+
+```solidity
+// Incorrect: external self-call
+function process() external {
+  this.validate();  // Wastes gas, breaks internal invariants
+}
+
+// Correct: internal call
+function process() external {
```
thedarkjester: depending on what it is a `virtual` modifier should be here - worth being consistent across the rules

.agents/skills/developing-smart-contracts/AGENTS.md
```diff
@@ -0,0 +1,289 @@
+# Linea Smart Contracts - Development Guidelines
+
+**Version 1.0.0**
```
thedarkjester: this should be updated with any other changes made

kyzooghost: hmm, I might delete this file because it's not really needed

.agents/skills/developing-smart-contracts/rules/file-layout.md
```diff
@@ -0,0 +1,182 @@
...
+
+  function _validateMessage(bytes calldata _data) internal pure returns (bool) {
+    // internal helper
+  }
+}
+```
+
+## Using for Library Extensions
+
+The `using A for B` directive attaches library functions to a type. Place these directives immediately after the contract declaration, before any other declarations.
```
thedarkjester, line 129:
```suggestion
The `using A for B` directive attaches library functions to a type. Place these directives immediately after the contract declaration, before any other declarations.

These should only be added if the code uses the `extension` method functionality vs. `LibraryName.Function(..)`
```

.agents/skills/developing-smart-contracts/rules/gas-optimization.md
```diff
@@ -0,0 +1,127 @@
...
+// Incorrect: repeated storage reads
+require(fee != 0, FeeNotSet());
+_charge(fee);
+```
+
+## Memory Usage
+
+- Avoid copying `calldata` to memory unless required.
+- Avoid `abi.encode`/`abi.encodePacked` in tight loops unless measured and necessary.
+- Use storage pointers when updating multiple fields in a struct.
```
thedarkjester: I don't know if we can stipulate here, but there is the case of a struct being read many times from the calldata being more expensive than just putting it in memory.

.agents/skills/developing-smart-contracts/rules/general-rules.md
```diff
@@ -0,0 +1,181 @@
+# General Rules
+
+**Impact: MEDIUM (improves code quality and maintainability)**
+
+## Inheritance & Customization
+
+When extending Linea contracts:
+
+- Use `virtual`/`override` keywords
+- Override `CONTRACT_VERSION()` for custom versions
```
thedarkjester, line 10:
```suggestion
- Override `CONTRACT_VERSION()` for custom versions
- Consider `public` vs. `external` for some functions that could be safely reused without TOCTOU issues

```

.agents/skills/developing-smart-contracts/rules/general-rules.md
```diff
@@ -0,0 +1,181 @@
...
+  require(amount <= MAX_WITHDRAWAL, "Too much");
+}
+```
+
+## Reinitializer Pattern
+
+Both `initialize` (for new chain deployments) and `reinitializeVN` (for canonical chain upgrades) MUST use the same `reinitializer(N)` modifier. Do not use the `initializer` modifier.
+
+```solidity
+// Correct: both functions use reinitializer(8)
```
thedarkjester:
```suggestion
// Correct: both functions use reinitializer(8) and `initialize` also uses `onlyInitializedVersion(0)` for new deployments
```

.agents/skills/developing-smart-contracts/rules/general-rules.md
```diff
@@ -0,0 +1,181 @@
...
+function initialize(...) external reinitializer(8) { ... }
```
thedarkjester:
```suggestion
function initialize(...) external onlyInitializedVersion(0) reinitializer(8) { ... }
```

.agents/skills/developing-smart-contracts/rules/general-rules.md
```diff
@@ -0,0 +1,181 @@
...
+
+// Incorrect: initialize uses initializer
+function initialize(...) external initializer { ... }
+function reinitializeV8(...) external reinitializer(8) { ... }
+```
+
+Why:
+- `_getInitializedVersion()` returns the version number even on fresh deployments, giving a consistent version indicator
+- No access control or proxy admin needed on the implementation contract - `reinitializeV8` naturally fails on new deployments of this version
+- In `reinitializeVN`, you can check `_getInitializedVersion() == N-1` to prevent upgrades from the wrong version
```
thedarkjester, line 128:
```suggestion
- `onlyInitializedVersion(0)` enforces the contract is not accidentally being reinitialized from a lower version
- `_getInitializedVersion()` returns the version number even on fresh deployments, giving a consistent version indicator
- No access control or proxy admin needed on the implementation contract - `reinitializeV8` naturally fails on new deployments of this version
- In `reinitializeVN`, you can check `_getInitializedVersion() == N-1` to prevent upgrades from the wrong version
```

.agents/skills/developing-smart-contracts/rules/general-rules.md
kyzooghost: Could you clarify what you mean here? If you have a concrete example in mind, I think that would make it clearer :)

.agents/skills/developing-smart-contracts/rules/gas-optimization.md
kyzooghost: changed to 

```
Avoid copying `calldata` to memory unless required (e.g., a struct whose fields are read many times may be cheaper to copy to memory once).
```

.agents/skills/developing-smart-contracts/rules/general-rules.md
thedarkjester: if you have a function let's say you have functionality that is `external virtual` and you want to inherit, override and just call the external `super.`, then `public` is needed. If it has reentry checks then it is ok, not as gas efficient, but ok/safe.

If however let's say you do something where you expect or use a certain state, it calls something else, modifies state, calls it again then the state checks might be different and behavior is different.

kyzooghost: added this section

```
### `public` vs `external` for Overridden Functions

When overriding a base function and calling `super`, the function must be `public` (not `external`) since `super` is an internal call. This is safe when the function has reentrancy guards. Avoid this for functions that check state before modifying it (TOCTOU risk - the state may change between the check and the use).

```solidity
// Safe: reentrancy guard protects the internal call path
function deposit() public override nonReentrant {
  super.deposit();
  // additional logic
}

// Unsafe TOCTOU: internal call could bypass stale state checks
function withdraw(uint256 _amount) public override {
  require(balances[msg.sender] >= _amount, "Insufficient"); // check
  super.withdraw(_amount); // super modifies state
  balances[msg.sender] -= _amount; // use - balance already changed by super
}
```
```

PR #2256, opened 2026-02-04
Cleanup koalagnark and Debug tests

YaoJGalteland:
This PR implements issue(s) #

Refactored the `koalagnark` package to enforce a clear separation between witness assignment constructors and in-circuit constant constructors.  Added necessary APIs through the GnarkRuntime interface so circuit methods can use the correct constructors. Replaced api with koalaAPI in gnark algebraic computations to avoid NewAPI allocation in every call.

Previously, the same constructors (e.g. `NewElement`, `NewExt`, `WrapFrontendVariable`) were used in both witness assignment and circuit definition contexts. This caused subtle bugs:

- VerifyingKey columns (compile-time constants) were being wrapped with witness constructors, which set both the native V and emulated EV fields unnecessarily and incorrectly
- GnarkRuntime methods like GetColumn didn't receive frontend.API, so they couldn't construct proper circuit constants
- Extension field lifting (LiftToExt) used NewElementFromValue(0) for zero components, which is a witness constructor

Additionally, fix broken tests

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches core circuit/verifier plumbing and field construction paths (including recursion, vortex, and Fiat–Shamir), so subtle API or constant-vs-witness mistakes could affect proof verification despite being largely mechanical refactors.
> 
> **Overview**
> Refactors `koalagnark` to **separate witness-assignment constructors from in-circuit constant/variable constructors**, moving extension-field circuit operations into a new `extapi.go` and tightening APIs (e.g., `API.Const(field.Element)`, `API.WrapFrontendVariable`, `NewElementFromBase`, `NewElementFromValue`).
> 
> Propagates `frontend.API` through the keccak PI-interconnection column interfaces (`GetColAssignmentGnark*`) and updates recursion/vortex/verifier logic to pass `api` (and often a shared `*koalagnark.API`) so verifier-visible columns and public inputs are constructed with the correct in-circuit constructors.
> 
> Updates Fiat-Shamir gnark implementations so `RandomManyIntegers` returns `[]koalagnark.Element` (not raw `frontend.Variable`), and adjusts many crypto/math polynomial helpers to accept `*koalagnark.API` instead of creating new wrappers internally; fixes/updates related tests and expected Poseidon2 accumulator vectors, and registers gnark hash getters where needed.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 828196840bbb96d7b3680beda4a3b5957b730670. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


prover/maths/field/koalagnark/element.go
```diff
@@ -66,6 +109,10 @@ func WrapFrontendVariable(v frontend.Variable) Element {
 	return res
```
ivokub, line 51: We shouldn't really create manually `emulated.Element`, it is always better to do through the constructor. But imo the whole method doesn't make sense - we seem to be using it in-circuit, but then it should be a method on `koalaAPI`

YaoJGalteland: you are right.

During witness assignment, we wouldn't have a frontend.Variable. NewElement and NewExt should only have values like int, field.Element, string, etc. as input.  

frontend.Variable is a circuit variable that only exists inside the circuit.

YaoJGalteland: we cannot completely remove `WrapFrontendVariable`, as `GetRandomCoinIntegerVec` lacks access to a frontend.API but still requires some function like `WrapFrontendVariable` to wrap frontend.Variable types. We also need this to wrap the `GnarkHornerParams`. Are you okay with keeping it, or do you have ideas on how to remove it and fix those two calling locations?

YaoJGalteland: no problem anymore, WrapFrontendVariable is completely removed!

prover/maths/field/koalagnark/ext.go
```diff
@@ -31,643 +34,68 @@ type Ext struct {
...
+	}
+}
+
+// NewExt creates an Ext for witness assignment from various input types:
+//   - fext.Element: full 4-component extension element
+//   - extensions.E2: quadratic extension in B0, B1 is zero
+//   - Element, field.Element: base field element in B0.A0, all others zero
+//   - int, int64, uint32, *big.Int: numeric constants in B0.A0
+//   - string: decimal representation of a field element in B0.A0
+func NewExt(v any) Ext {
```
ivokub: For me it would make sense more if we have more concrete constructors instead of relying on runtime detection of types. I think we could have two approaches:
- either we have more concrete constructors `NewExt(v fext.Element)`, `NewExtFromBase(v field.Element)`, `NewExtFromE2(v extensions.E2)` and then for fallback `NewExtFromAny(v any)` which handles int, int64, uint32, *big.Int and string
- or we use generics:
```go
func NewExt(interface {field.Element | extensions.E2 | fext.Element} T) (v T) Ext
```
to have already compile-time type safety.

And I would also avoid accepting inputs of type `Ext/*Ext/Element/*Element` as it indicates something strange going on (we assign a witness from witness etc.)

prover/maths/field/koalagnark/api.go
```diff
@@ -72,42 +73,69 @@ func (a *API) GetFrontendVariable(v Element) frontend.Variable {
...
-// ConstBig creates a circuit constant from a big.Int.
-func (a *API) ConstBig(c *big.Int) Element {
+// elementFromFrontendVar wraps an existing frontend.Variable as an Element.
+func (a *API) elementFromFrontendVar(v frontend.Variable) Element {
 	if a.IsNative() {
-		return Element{V: c}
+		return Element{V: v}
 	}
-	return Element{EV: *a.emulatedAPI.NewElement(c)}
+	return Element{EV: emulated.Element[emulated.KoalaBear]{Limbs: []frontend.Variable{v}}}
```
ivokub: Here I would avoid constructing the `emulated.Element` manually. There is a method `Field.NewElement` which you should use. I think right now it works as expected, but direct manipulation of `emulated.Element` can lead to some issues (as we want to add overflow/range check tracking etc).

prover/maths/field/koalagnark/api.go
```diff
@@ -72,42 +73,69 @@ func (a *API) GetFrontendVariable(v Element) frontend.Variable {
...
+		return a.elementFromValue(int64(v))
+	case int64:
+		return a.elementFromValue(v)
+	case *big.Int:
+		return a.elementFromValue(v)
+	case field.Element:
+		return a.elementFromValue(int64(v.Uint64()))
+	case *field.Element:
+		return a.elementFromValue(int64(v.Uint64()))
+	case frontend.Variable:
```
ivokub: Instead of handling `frontend.Variable` in this method, I think it would be better to keep it separately to distinguish between creating a in-circuit koalabear element from a constant from packing a `frontend.Variable` into in-circuit koalabear element.

prover/maths/polynomials/lagrange.go
```diff
@@ -49,8 +49,7 @@ func GnarkEvaluateLagrangeExt(api frontend.API, p []koalagnark.Ext, z koalagnark
 		weightedP := koalaAPI.MulConstExt(p[i], bwi)
 
 		// z - ωⁱ
-		wOmegai := koalagnark.NewElementFromKoala(accOmega)
-		omegaiExt := koalaAPI.FromBaseExt(wOmegai)
+		omegaiExt := koalaAPI.ExtFrom(koalaAPI.ElementFrom(accOmega))
```
ivokub: I think in principle you can do `koalaAPI.ExtFrom(accOmega)` as well? Right now it should handle `field.Element` as input.

And that imo reflects the issue with using `any` as inputs everywhere, not enforcing any kind of compile time safety and we need to just try out and to see if stuff fails at runtime. It would be way safer if we just have more concrete methods (a la as I recommended having `ExtFromBase` etc.)

prover/protocol/accessors/from_local_opening.go
```diff
@@ -51,8 +51,10 @@ func (l *FromLocalOpeningYAccessor) GetFrontendVariableBase(api frontend.API, c
 
 func (l *FromLocalOpeningYAccessor) GetFrontendVariableExt(api frontend.API, c ifaces.GnarkRuntime) koalagnark.Ext {
 	p := c.GetParams(l.Q.ID).(query.GnarkLocalOpeningParams)
+
 	if p.IsBase {
-		return koalagnark.FromBaseVar(p.BaseY)
+		// Use NewExt which doesn't require API - works when called with nil API
+		return koalagnark.NewExt(p.BaseY)
```
ivokub: Or we could just create an API instead and if there is a method `BaseAsExt` etc.

prover/protocol/compiler/recursion/circuit.go
```diff
@@ -140,10 +134,17 @@ func (r *RecursionCircuit) Define(api frontend.API) error {
 	}
 
 	polyParams := w.GetUnivariateParams(r.PolyQuery.Name())
-	koalaAPI.AssertIsEqualExt(E4Gen(r.X), polyParams.ExtX)
+
+	api.AssertIsEqual(r.X[0], polyParams.ExtX.B0.A0.Native())
```
ivokub: For me this is API leakage as we have inlined the `koalaAPI.AssertIsEqualExt` method here.

prover/protocol/compiler/selfrecursion/column_opening.go
```diff
@@ -524,7 +524,7 @@ func (a *FoldPhaseVerifierAction) RunGnark(api frontend.API, run wizard.GnarkRun
 	yActual := poly.EvaluateUnivariateGnarkExt(api, dcollapse, rfold)
 
 	one := koalaAPI.OneExt()
-	two := koalagnark.FromBaseVar(koalagnark.NewElement(2))
+	two := koalaAPI.ExtFrom(koalagnark.NewElement(2))
```
ivokub: Currently `koalaAPI.ExtFrom(2)` should work imo as well.

prover/crypto/fiatshamir_bls12377/fiatshamir_circuit.go
```diff
@@ -50,6 +50,17 @@ func (fs *GnarkFS) UpdateFrElmt(vec ...frontend.Variable) {
 	fs.hasher.Write(vec...)
 }
 
+// UpdateFrVariable updates a vector of frontend.Variable
+// This is used for Horner query parameters.
+func (fs *GnarkFS) UpdateFrVariable(vec ...frontend.Variable) {
+	koalaAPI := koalagnark.NewAPI(fs.api)
```
ivokub: We shouldn't create APIs all the time. Just embed it into `GnarkFS`

prover/crypto/poseidon2_bls12377/poseidon2_test.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/consensys/gnark-crypto/ecc/bls12-377/fr"
 	"github.com/consensys/gnark/frontend"
 	"github.com/consensys/gnark/frontend/cs/scs"
+	_ "github.com/consensys/gnark/std/hash/mimc" // Register MIMC hash function
```
ivokub: Better import `github.com/consensys/gnark/std/hash/all` which imports all hash function getters.

prover/crypto/fiatshamir/fs.go
```diff
@@ -14,11 +14,12 @@ import (
 // circuit
 type GnarkFS interface {
 	Update(vec ...koalagnark.Element)
+	UpdateFrVariable(vec ...frontend.Variable)
```
ivokub: We don't need to add the interface, see comment below.

prover/protocol/query/gnark_params.go
```diff
@@ -153,6 +153,6 @@ func (p GnarkHornerParams) UpdateFS(fs fiatshamir.GnarkFS) {
 	fs.UpdateExt(p.FinalResult)
 
 	for _, part := range p.Parts {
-		fs.Update(koalagnark.WrapFrontendVariable(part.N0), koalagnark.WrapFrontendVariable(part.N1))
+		fs.UpdateFrVariable(part.N0, part.N1)
```
ivokub: The issue here why you added UpdateFrVariable was that `part.N0` and `part.N1` are `frontend.Variable` in `GnarkHornerParams`.

But actually if you follow the implementation then it is `koalagnark.Element`. So you can just change the type definition to match.

Keep in mind that `frontend.Variable` is actually `any`! This is for historical reasons and now we keep it for backwards compatibility. But in practice it could be anything -- constants, in-circuit variables, terms, whatever. We usually need to build around that to have something concretely typed.

prover/crypto/fiatshamir_bls12377/fiatshamir_circuit.go
```diff
@@ -50,6 +50,17 @@ func (fs *GnarkFS) UpdateFrElmt(vec ...frontend.Variable) {
 	fs.hasher.Write(vec...)
 }
 
+// UpdateFrVariable updates a vector of frontend.Variable
+// This is used for Horner query parameters.
+func (fs *GnarkFS) UpdateFrVariable(vec ...frontend.Variable) {
```
ivokub: Don't need to add

prover/crypto/fiatshamir_koalabear/fiatshamir_circuit.go
```diff
@@ -35,6 +35,12 @@ func NewGnarkFSFromFactory(api frontend.API, factory hasherfactory_koalabear.Has
 	}
 }
 
+// UpdateFrVariable updates a vector of frontend.Variable
+// This is used for Horner query parameters.
+func (fs *GnarkFSWV) UpdateFrVariable(vec ...frontend.Variable) {
```
ivokub: Don't need to add

prover/crypto/state-management/accumulator/testvectors_poseidon2_test.go
```diff
@@ -117,20 +117,20 @@ func TestInsertionRootHashPoseidon2(t *testing.T) {
 
 		// also check its hash value
 		leaf := leafOpening.Hash()
-		require.Equal(t, "0x42bee0052f812f13511ced262f95e62d5e572dc971f68fa8759914a441332ff1", leaf.Hex())
+		require.Equal(t, "0x05f2b1de703588c454c171b1380e4ca0096466bc0b9b18b7741976ff291c699e", leaf.Hex())
```
ivokub, line 120: Why changed? Because of hash function change?

prover/crypto/state-management/smt_bls12377/gnark_test.go
```diff
@@ -7,8 +7,8 @@ import (
 	"github.com/consensys/gnark-crypto/ecc"
 	"github.com/consensys/gnark-crypto/ecc/bls12-377/fr"
 	"github.com/consensys/gnark/frontend"
-
 	"github.com/consensys/gnark/frontend/cs/scs"
+	_ "github.com/consensys/gnark/std/hash/mimc" // Register MIMC hash function
```
ivokub: import `github.com/consensys/gnark/std/hash/all` instead

ivokub (changes requested):
I had a look - I think it is in better direction but imo could be still simplified a lot. What I mostly would avoid is having these assignment/init methods taking `any` as input. Even though right now it seems it reduces the clutter by having less methods, then long-term it only causes problems down the road when we are having some odd assertion bugs at runtime. And then we need to debug and trace what are the actual types etc. It can also cause issues when we assume something should be extension element but is base etc.

And we also don't have to change the FiatShamir instance.

prover/crypto/state-management/accumulator/testvectors_poseidon2_test.go
YaoJGalteland: The root cause is that the poseidon2_koalabear hash function's behavior changed at some earlier point in the prover/dev-small-fields branch history (likely the state update, buffer reset or the WriteInt64On64Bytes encoding were adjusted), and the test vectors were not updated.

prover/crypto/vortex/vortex_koalabear/verifier_circuit_test.go
```diff
@@ -105,7 +106,7 @@ func TestGnarkVerifier(t *testing.T) {
 		}
 	}
 
-	ccs, err := frontend.CompileU32(koalabear.Modulus(), scs.NewBuilder, &circuit)
+	ccs, err := frontend.CompileU32(koalabear.Modulus(), gnarkutil.NewMockBuilder(scs.NewBuilder), &circuit)
```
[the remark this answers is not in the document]
YaoJGalteland: following the discussion here: https://github.com/Consensys/linea-monorepo/pull/2202/changes#diff-08b8c6a8481499ea3ae630284ac3a0fb41363425df9a13802838870fdc18ab6cR166 

@ivokub what do you think? can we ignore these tests, or should we fix the range check issue properly?

PR #2319, opened 2026-02-08
fix: use Go 1.24.13 and regenerate

ivokub:
Go 1.24.13 fixed a bug where it was possible to define additional code in Go comments vs cgo. It also allowed to run the external code at build time. See:
* https://groups.google.com/g/golang-announce/c/K09ubi9FQFk
* https://github.com/golang/go/issues/76697
* https://github.com/golang/go/issues/77103

The fix for now is to not generate documentation for generated headers built using cgo. **Linea is unaffected, we don't use cgo from external packages (only libcompress/libdecompress which is internal).**

But as we generate the cgo headers during Go generate and we use generate extensively (e.g. in serde and type-specific code generation), then this means that the generated headers now differ from committed variant. We check in our CI that the generated files wouldn't be modified from committed version (to avoid stale implementation) and the lack of headers broke CI.

It was tentatively fixed in https://github.com/Consensys/linea-monorepo/pull/2313 but it is not full long-term solution as it only hardcoded the CI image used for running the tests to an older Go version. But Go toolchain internally upgrades itself to whatever version is defined in `go.mod` so it means that in case `go.mod` is updated the tests would break again. And we also don't want to rely on old Go version.

This PR implements complete mitigation:
* change CI images to non-hardcoded version (i.e. 1.24.x)
* update Go tool to 1.24.13 in `go.mod`
* regenerate the headers built with cgo to apply the Go mitigations (i.e. remove comments in `.h` headers)

### Checklist

---

> [!NOTE]
> **Low Risk**
> Primarily a toolchain/CI pin update plus regenerated cgo header output; low behavioral risk but may affect build/test reproducibility across environments.
> 
> **Overview**
> Updates the prover CI workflow to use the floating `1.24.x` Go toolchain (instead of a pinned patch version) and aligns the 32-bit/race test gate with the new matrix.
> 
> Bumps `prover/go.mod` to Go `1.24.13` and regenerates the cgo-exported `libcompressor.h`/`libdecompressor.h` headers, removing the embedded doc comments so `go generate` no longer dirties the repo under the newer Go mitigation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit eab56eea0bf36ea700b4cbdc454d2a1783af7f54. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


eloi010 (approved):
LGTM. I'd consider updating the `README.md` to indicate that the Prover uses Go `1.24.x`.

PR #2347, opened 2026-02-10
feat: Compression aware block building

Filter94:
This PR implements issue(s) #

### Checklist

---

> [!NOTE]
> **High Risk**
> Changes core transaction selection and profitability compression behavior and makes previous calldata-size limits optional (behavior change) while introducing new native-compressor-backed, async block-size enforcement; misconfiguration or compressor/threading issues could impact block production and tx acceptance.
> 
> **Overview**
> Adds **compression-aware block size enforcement** for the sequencer via a new `CompressionAwareTransactionSelector` (enabled only when `--plugin-linea-blob-size-limit` is set) that uses per-tx compressed-size accumulation plus an async full-block compression check, and surfaces a new rejection reason `BLOCK_COMPRESSED_SIZE_OVERFLOW`.
> 
> Updates configuration/CLI to make block/tx calldata size limits *optional* (and logs deprecations for `--plugin-linea-max-block-calldata-size` / `--plugin-linea-max-tx-calldata-size`), reworks compression to use a shared `GoBackedBlobCompressor` v2 with new `TxEncodingUtils` encoding, and adds unit/acceptance tests (including a test-only `RecordingTransactionSelectorPlugin`) plus docs and minor CI workflow artifact-download path fixes.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 51334aba9aeebffb634fe1e517dc585d7524daf5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/besu-plugins/linea-sequencer/acceptance-tests/src/test/kotlin/linea/plugin/acc/test/CompressionAwareBlockBuildingTest.kt
+++ b/besu-plugins/linea-sequencer/acceptance-tests/src/test/kotlin/linea/plugin/acc/test/CompressionAwareBlockBuildingTest.kt
@@ -56,6 +56,8 @@ class CompressionAwareBlockBuildingTest : LineaPluginPoSTestBase() {
   override fun getRequestedPlugins(): List<String> =
     DEFAULT_REQUESTED_PLUGINS + "RecordingTransactionSelectorPlugin"
 
+  override fun getAdditionalRpcApis(): Set<String> = setOf("TEST")
+
   override fun getTestCliOptions(): List<String> {
     return TestCommandLineOptionsBuilder()
       .set("--plugin-linea-blob-size-limit=", BLOB_SIZE_LIMIT.toString())
--- a/besu-plugins/linea-sequencer/acceptance-tests/src/test/kotlin/linea/plugin/acc/test/LineaPluginPoSTestBase.kt
+++ b/besu-plugins/linea-sequencer/acceptance-tests/src/test/kotlin/linea/plugin/acc/test/LineaPluginPoSTestBase.kt
@@ -107,13 +107,16 @@ abstract class LineaPluginPoSTestBase : LineaPluginTestBase() {
 
   protected open fun getRequestedPlugins(): List<String> = DEFAULT_REQUESTED_PLUGINS
 
+  protected open fun getAdditionalRpcApis(): Set<String> = emptySet()
+
   @BeforeEach
   override fun setup() {
+    val rpcApis = setOf("LINEA", "MINER", "PLUGINS") + getAdditionalRpcApis()
     minerNode = createCliqueNodeWithExtraCliOptionsAndRpcApis(
       "miner1",
       getCliqueOptions(),
       getTestCliOptions(),
-      setOf("LINEA", "MINER", "PLUGINS"),
+      rpcApis,
       true,
       getRequestedPlugins(),
     )


besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/selectors/CompressionAwareBlockTransactionSelector.java
```diff
@@ -0,0 +1,230 @@
...
+    // Slow path: conservative estimate exceeded the limit.
+    // Build full block RLP and check with canAppendBlock for actual compression.
+    final ProcessableBlockHeader pendingHeader = evaluationContext.getPendingBlockHeader();
+    final List<Transaction> tentativeTxs = new ArrayList<>(state.selectedTransactions());
+    tentativeTxs.add(transaction);
+    final byte[] blockRlp = buildBlockRlp(pendingHeader, tentativeTxs);
+
+    blobCompressor.reset();
+    final boolean fits = blobCompressor.canAppendBlock(blockRlp);
+
```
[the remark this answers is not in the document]
Filter94: That is a valid concern, but I don't think I can't solve it, because it comes down to the JNA being non-thread safe I believe

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/selectors/CompressionAwareTransactionSelector.java
```diff
@@ -0,0 +1,246 @@
...
+              "Fast path: tx {} compressed={}, cumulative would be {} < effectiveLimit {} "
+                  + "(blobLimit={} - headerOverhead={})")
+          .addArgument(transaction::getHash)
+          .addArgument(txCompressedSize)
+          .addArgument(newCumulative)
+          .addArgument(fastPathLimit)
+          .addArgument(blobSizeLimit)
+          .addArgument(compressedBlockHeaderOverhead)
+          .log();
+      return SELECTED;
```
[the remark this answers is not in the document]
Filter94: That's why we're using the `.addArgument(compressedBlockHeaderOverhead)`. And we're not using `WorstCompressedTxSize`, because we want to leverage the cache

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/selectors/CompressionAwareTransactionSelector.java
```diff
@@ -0,0 +1,239 @@
...
+    // Build full block RLP and check with canAppendBlock for actual compression.
+    final ProcessableBlockHeader pendingHeader = evaluationContext.getPendingBlockHeader();
+    final List<Transaction> tentativeTxs = new ArrayList<>(state.selectedTransactions());
+    tentativeTxs.add(transaction);
+    final byte[] blockRlp = buildBlockRlp(pendingHeader, tentativeTxs);
+
+    blobCompressor.reset();
+
+    final BlobCompressor.AppendResult appendResult = blobCompressor.appendBlock(blockRlp);
+    if (!appendResult.getBlockAppended()) {
```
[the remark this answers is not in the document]
Filter94: Unfortunately `canAppendBlock` doesn't return the compressed size, so it can't be used

besu-plugins/linea-sequencer/acceptance-tests/src/test/kotlin/linea/plugin/acc/test/CompressionAwareBlockBuildingTest.kt
```diff
@@ -0,0 +1,204 @@
...
+      .withFailMessage { "Small tx submission failed: ${smallTxResponse.error?.message}" }
+      .isFalse()
+
+    val largeTxHash = largeTxResponse.transactionHash
+    val smallTxHash = smallTxResponse.transactionHash
+
+    buildNewBlockAndWait()
+
+    minerNode.verify(eth.expectSuccessfulTransactionReceipt(smallTxHash))
+    minerNode.verify(eth.expectNoTransactionReceipt(largeTxHash))
```
fab-10, line 111: The test can pass even if the `largeTx` fails to be included for any other reason and not only for its compressed size.
One easy way to ensure the exact failure is matching the expected failure in the logs, but I if you do not like that solution, then another options is to build a plugin, for use in AT, that implements `PluginTransactionSelector::onTransactionNotSelected` and collects in memory the the results so you can later query them

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/selectors/CompressionAwareTransactionSelector.java
```diff
@@ -0,0 +1,239 @@
...
+   */
+  private static byte[] buildBlockRlp(
+      final ProcessableBlockHeader pendingHeader, final List<Transaction> transactions) {
+    // Use a new Random instance with fixed seed for each call to ensure deterministic output
+    final Random random = new Random(PLACEHOLDER_SEED);
+
+    final BlockHeader header =
+        BlockHeaderBuilder.create()
+            .parentHash(randomHash(random))
+            .ommersHash(randomHash(random))
```
fab-10: Can the random stuffs be statically pre-computed once for better performance?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/selectors/CompressionAwareTransactionSelector.java
```diff
@@ -0,0 +1,239 @@
...
+
+    return SELECTED;
+  }
+
+  /**
+   * Builds an RLP-encoded block from the pending block header and the list of selected
+   * transactions, using Besu's standard {@link Block}/{@link BlockHeader}/{@link BlockBody}
+   * encoding. Placeholder header fields use deterministic pseudo-random bytes (seeded with a fixed
+   * value) to ensure reproducible compression behavior while still providing varied data that
+   * compresses similarly to real headers.
```
fab-10: should not be better to assume these fields are not _compress-able_ to avoid worst case scenario? or the `compressedBlockHeaderOverhead` already count for that?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/selectors/CompressionAwareTransactionSelector.java
```diff
@@ -0,0 +1,239 @@
...
+ * Stateful transaction selector that enforces a compressed block size limit (blob fit) using a
+ * two-phase strategy:
+ *
+ * <p><b>Fast path:</b> accumulates {@code RawCompressedSize} per individual transaction. While the
+ * cumulative sum is below {@code blobSizeLimit}, the block is guaranteed to fit because compressing
+ * all transactions together always yields a smaller result than the sum of individually compressed
+ * transactions.
+ *
+ * <p><b>Slow path:</b> once the cumulative sum reaches {@code blobSizeLimit}, builds a full block
+ * RLP (using the pending block header data + all selected transactions + the candidate) and checks
```
fab-10: the fast/slow path is not covered by ATs yet

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/AbstractLineaSharedPrivateOptionsPlugin.java
```diff
@@ -293,9 +293,20 @@ private void performSharedStartTasksOnce(final ServiceManager serviceManager) {
...
+            : DEFAULT_LIMIT;
+    blobCompressor =
+        GoBackedBlobCompressor.getInstance(BlobCompressorVersion.V1_2, effectiveBlobLimit);
+
     final LineaProfitabilityConfiguration profitabilityConfiguration = profitabilityConfiguration();
-    final TransactionCompressor transactionCompressor =
-        new CachingTransactionCompressor(profitabilityConfiguration.compressedTxCacheSize());
+    transactionCompressor =
+        new CachingTransactionCompressor(
+            profitabilityConfiguration.compressedTxCacheSize(), blobCompressor);
```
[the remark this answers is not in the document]
Filter94: If

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/selectors/CompressionAwareTransactionSelector.java
```diff
@@ -0,0 +1,239 @@
...
+            .mixHash(randomHash(random))
+            .nonce(random.nextLong())
+            .baseFee(Wei.of(random.nextLong(Long.MAX_VALUE)))
+            .blockHeaderFunctions(new MainnetBlockHeaderFunctions())
+            .buildBlockHeader();
+
+    final BlockBody body = new BlockBody(transactions, Collections.emptyList());
+    final Block block = new Block(header, body);
+    return block.toRlp().toArray();
+  }
```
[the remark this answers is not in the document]
Filter94: In the header only the timestamp actually goes into the blob as of v2

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/AbstractLineaSharedPrivateOptionsPlugin.java
```diff
@@ -82,16 +84,14 @@ public abstract class AbstractLineaSharedPrivateOptionsPlugin
   protected static MetricCategoryRegistry metricCategoryRegistry;
   protected static RpcEndpointService rpcEndpointService;
   protected static InvalidTransactionByLineCountCache invalidTransactionByLineCountCache;
+  protected static BlobCompressor blobCompressor;
+  protected static TransactionCompressor transactionCompressor;
   protected static TransactionProfitabilityCalculator transactionProfitabilityCalculator;
 
+  public static final int DEFAULT_LIMIT = 128 * 1024;
```
fab-10: `DEFAULT_LIMIT` of what?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/selectors/CompressionAwareTransactionSelector.java
Filter94: In practice only the timestamp is being put inside a blob. So even the variety doesn't matter

besu-plugins/linea-sequencer/acceptance-tests/src/test/kotlin/linea/plugin/acc/test/CompressionAwareBlockBuildingTest.kt
Filter94: I'd avoid going these lengths in the AT tests given the coverage in the unit tests. I think it's good enough

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/selectors/CompressionAwareTransactionSelector.java
Filter94: I don't think it should be covered by the AT. Also because fast/slow path is an implementation detail for the sake of performance, not a behavioral change

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/TxEncodingUtils.java
```diff
@@ -0,0 +1,202 @@
...
+  private static UInt256 toUInt256(final Quantity quantity) {
+    return UInt256.valueOf(quantity.getAsBigInteger());
+  }
+
+  private static Address getTo(final Transaction transaction) {
+    return transaction
+        .getTo()
+        .map(addr -> Address.wrap(Bytes.wrap(addr.toArrayUnsafe())))
+        .orElse(Address.ZERO);
+  }
```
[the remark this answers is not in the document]
Filter94: I believe that's correct, because for the contract creation, to has to be zero address

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/selectors/CompressionAwareTransactionSelector.java
Filter94: I replaced all values with an empty hash

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/TxEncodingUtils.java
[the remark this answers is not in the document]
Filter94: You might be right, but compression will also take care of empty bytes to some extend. I don't know how to easily fix this in a type safe way

besu-plugins/linea-sequencer/acceptance-tests/src/test/kotlin/linea/plugin/acc/test/CompressionAwareBlockBuildingTest.kt
fab-10: I invite you to reconsider adding such check, or removing this AT, in favor of unit tests only, because as already happened in the past, this could give a false sense of confidence that the code is working, even if in future the code can change and this could test a completely different non selection reason

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/AbstractLineaSharedPrivateOptionsPlugin.java
```diff
@@ -293,9 +293,20 @@ private void performSharedStartTasksOnce(final ServiceManager serviceManager) {
...
 
+    // Initialise the native compressor once with the authoritative limit so that
+    // CachingTransactionCompressor and CompressionAwareTransactionSelector share
+    // the same instance. Fall back to the default when no blob size limit is configured.
+    final int effectiveBlobLimit =
+        transactionSelectorConfiguration().blobSizeLimit() != null
+            ? transactionSelectorConfiguration().blobSizeLimit()
+            : DEFAULT_COMPRESSED_SIZE_LIMIT;
+    blobCompressor =
+        GoBackedBlobCompressor.getInstance(BlobCompressorVersion.V1_2, effectiveBlobLimit);
```
jonesho: How to configure to use later version like `BlobCompressorVersion.V3`?

Filter94: I updated it everywhere to V3 in the Sequencer, thanks!

jonesho: But why not making the compression version configurable instead? E.g. to support environment that's still using the old provers with compressor V2

Filter94: I think we discussed switchable compression based on the block number, but that's a whole separate task

besu-plugins/linea-sequencer/acceptance-tests/src/test/kotlin/linea/plugin/acc/test/CompressionAwareBlockBuildingTest.kt
Filter94: I got intimidated by the idea of a plugin solely for this purpose, but I gave it a try and it turned out to be concise and self contained. I think it will make a nice addition to our testing toolkit

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/AbstractLineaSharedPrivateOptionsPlugin.java
Filter94: I will roll it back to V2 because for compatibility sake

besu-plugins/linea-sequencer/acceptance-tests/src/main/kotlin/linea/plugin/acc/test/RecordingTransactionSelectorPlugin.kt
```diff
@@ -0,0 +1,110 @@
...
+import org.hyperledger.besu.plugin.services.txselection.TransactionEvaluationContext
+import java.util.concurrent.ConcurrentHashMap
+
+/**
+ * Acceptance-test-only plugin that observes transaction selection outcomes without affecting them.
+ *
+ * Registers a [PluginTransactionSelector] whose sole purpose is to record which transactions were
+ * not selected and why. The recorded rejections can be queried via the companion object so that
+ * acceptance tests can assert the exact rejection reason instead of relying on indirect signals
+ * such as receipt absence.
```
fluentcrafter, line 31: Why a companion/static object instead of an instance object? It feels like an anti-pattern...

I am afraid of side-effects on parallel test running inside the same JVM instance, were the same txHash could yield different results if we have different Besu instances... 

"different Besu instances", I know it's not possible atm in same JVM. So, if not a big effort, I would favour instance over static.

besu-plugins/linea-sequencer/acceptance-tests/src/test/kotlin/linea/plugin/acc/test/CompressionAwareBlockBuildingTest.kt
```diff
@@ -0,0 +1,228 @@
...
+   * 1. Disable background block building
+   * 2. Submit a large transaction that exceeds the compressed size limit (~4000 bytes calldata)
+   * 3. Submit a small transaction that fits easily (~500 bytes calldata)
+   * 4. Build a block — verify the small transaction is included, the large one is not
+   * 5. Submit a medium transaction slightly below the threshold (~3400 bytes calldata)
+   * 6. Build another block — verify the medium transaction is now included
+   */
+  @Test
+  fun largeTransactionExceedingCompressedLimitIsNotIncluded() {
+    val newAccounts = createAccounts(3, 10)
```
fluentcrafter: named arguments would help with readbility as we can't tell what 3 and 10 are for.

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/selectors/CompressionAwareTransactionSelector.java
```diff
@@ -111,35 +157,29 @@ public TransactionSelectionResult evaluateTransactionPreProcessing(
...
         .log();
+
+    pendingSlowPathFuture =
+        COMPRESSION_EXECUTOR.submit(
+            () -> {
+              blobCompressor.reset();
+              return blobCompressor.appendBlock(blockRlp);
+            });
+
     setWorkingState(new CompressionState(fastPathLimit, state.selectedTransactions));
```
jonesho: Why not provide `newConservativeCumulative` but `fastPathLimit` here? but anyway it seems to me either values are not very relevant as they would be overwritten later.

jonesho (approved):

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/selectors/CompressionAwareTransactionSelector.java
Filter94: Good point. That line is useless, I removed it

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txpoolvalidation/LineaTransactionPoolValidatorFactory.java
```diff
@@ -77,6 +80,13 @@ public LineaTransactionPoolValidatorFactory(
     this.transactionProfitabilityCalculator = transactionProfitabilityCalculator;
 
     this.deniedAddresses = new AtomicReference<>(txPoolValidatorConf.deniedAddresses());
+
+    if (txPoolValidatorConf.maxTxCalldataSize() != null) {
+      log.warn(
+          "DEPRECATION: --plugin-linea-max-tx-calldata-size is deprecated and will be removed "
```
fluentcrafter, line 86: @copilot create an issue to address this in 3 months time. Assign to @fluentcrafter

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/selectors/CompressionAwareTransactionSelector.java
```diff
@@ -0,0 +1,323 @@
...
+import org.hyperledger.besu.plugin.data.TransactionSelectionResult;
+import org.hyperledger.besu.plugin.services.txselection.AbstractStatefulPluginTransactionSelector;
+import org.hyperledger.besu.plugin.services.txselection.SelectorsStateManager;
+import org.hyperledger.besu.plugin.services.txselection.TransactionEvaluationContext;
+
+/**
+ * Stateful transaction selector that enforces a compressed block size limit (blob fit) using a
+ * two-phase strategy:
+ *
+ * <p><b>Fast path:</b> accumulates {@code RawCompressedSize} per individual transaction. While the
```
fluentcrafter: nit: fast/slow panth suggest that we may doing one or the other, whereas we do both one after the other. So I suggest calling it "Stage 1", "Stage 2"

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/selectors/CompressionAwareTransactionSelector.java
```diff
@@ -0,0 +1,325 @@
...
+
+    pendingSlowPathFuture =
+        COMPRESSION_EXECUTOR.submit(
+            () -> {
+              blobCompressor.reset();
+              return blobCompressor.appendBlock(blockRlp);
+            });
+
+    setWorkingState(new CompressionState(fastPathLimit, state.selectedTransactions));
+    return SELECTED;
```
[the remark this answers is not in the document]
fluentcrafter: I think this is a valid concern. I rather have `BLOCK_HEADER_TEAMPLATE` with more representative values.

besu-plugins/linea-sequencer/acceptance-tests/src/main/kotlin/linea/plugin/acc/test/RecordingTransactionSelectorPlugin.kt
Filter94: No reason. It's an undersight

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/selectors/CompressionAwareTransactionSelector.java
Filter94: I'll rename them to be execution path-s, because that's what I really mean there

besu-plugins/linea-sequencer/acceptance-tests/src/main/kotlin/linea/plugin/acc/test/RecordingTransactionSelectorPlugin.kt
Filter94: When I started to fix this I realized that the reason was that it was the simplest approach. Now this test plugin exposes an RPC method to expose the captured rejection reasons

PR #2370, opened 2026-02-12
feat: callDelegation testing for different targets - no collision

### LINKED ISSUE #2132: [ZkTracer] Test CALL's wrt EIP-7702
# Main point
We need to test `CALL`'s in light of [EIP-7702](https://eips.ethereum.org/EIPS/eip-7702). We should re-use the current `CALL` testing, but extend it so that we can configure the target of the `CALL` to be delegated. Specifically we need to test all `CALL`-scenarios on the cases below pertaining to the **delegation status of the target**:
- [x] (target) isn't delegated
- [ ] (target) is delegated
  - [ ] to smc (on Amelie's branch)
  - [ ] to eoa
    - [ ] to other eoa
    - [ ] to self
  - [ ] to account that is itself delegated (amounting to an EOA call)
  - [ ] to prc (which should already be covered by the "call EOA that is delegated to other EOA" case) => (pending to be added on Amelie's branch)

For the SMC we would like very simple bytecode:
```rust
// smc byte code
// designed to test environment variables after call to delegated account
ADDRESS
SELFBALANCE
CALLER
CALLVALUE
GAS
CODESIZE
PUSH0
PUSH0
CODECOPY
```
(no need to `POP` anything)

So in essence there are 3 cases of interest in the **target is delegated** case.

Image

In the above
- ${\color{blue}\text{blue arrows}} = \text{CALL target}$
- ${\color{green}\text{green arrows}} = \text{delegation target}$

The simplest way to implement this would be to have 4 accounts pre-added to the state and then to have 3 scenarios for the call target:
- 1x delegated EOA
- 2x delegated EOA
- self delegated EOA

And we could add a 5th account and an extra scenario for the call target (not pictured) with
- delegated to PRC
```
addr5
code = 0x ef 01 00 <some prc address>
```

amkCha:
Closes #2132 

---

> [!NOTE]
> **Low Risk**
> Test-only additions plus a small Gradle test-timeout tweak; no production logic changes.
> 
> **Overview**
> Adds **disabled** `CallDelegation` JUnit coverage for EIP-7702-style code delegation across several target types (delegation to a contract, to another EOA, to self, chained delegation, and to a precompile) using the toy execution environment.
> 
> Extends test `Utils` with a delegation bytecode prefix constant and helpers to build delegation code for `ToyAccount`/`Address`, and increases `fastReplayTests` default JUnit timeout to `30 m`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e24f62d54a2ff372fb4f8e21a17d8411bd807895. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


tracer/arithmetization/src/test/java/net/consensys/linea/zktracer/instructionprocessing/callTests/CallDelegation.java
```diff
@@ -0,0 +1,257 @@
...
+        .nonce(99)
+        .code(
+          Bytes.concatenate(Bytes.fromHexString(delegationCodeToSmc)))
+        .build();
+
+    Transaction tx = ToyTransaction.builder()
+        .sender(senderAccount)
+        .keyPair(keyPair)
+        .to(eoaDelegatedToSmc).build();
+
```
lorenzogentile404, line 112: Here I believe we need a type 4 transaction where:

`authority/delegator` = `eoaDelegatedToSmc`
`delegatee` = `smcAccount`

Moverover, `authority/delegator` is also the `recipient` of the transaction as you correctly already set.

So you get something like:

```
 .addCodeDelegation(BigInteger.valueOf(LINEA_CHAIN_ID), smcAccountAddress, eoaDelegatedToSmcNonce, eoaDelegatedToSmcKeyPair)
```

Similarly for the tests below.

OlivierBBB: @lorenzogentile404 These are "Call tests" i.e. we care not about delegation transactions but about CALLs to accounts that are delegated. In particular any transaction type will do.

tracer/arithmetization/src/test/java/net/consensys/linea/zktracer/instructionprocessing/callTests/CallDelegation.java
```diff
@@ -0,0 +1,257 @@
...
+  final String smcBytecode = "0x30473833345a";
+  final ToyAccount smcAccount =
+    ToyAccount.builder()
+      .address(Address.fromHexString("1234"))
+      .nonce(90)
+      .code(
+        Bytes.concatenate(Bytes.fromHexString(smcBytecode)))
+      .build();
+  final String delegationCodeToSmc = "0xef0100" + smcAccount.getAddress().toHexString().substring(2);
```
lorenzogentile404: I sugest to turn `0xef0100` in a constant (e.g., `DELEGATION_PREFIX`) and extract a method to compute this string, e.g., `codeToDelegateTo(Address address)`

tracer/arithmetization/src/test/java/net/consensys/linea/zktracer/instructionprocessing/callTests/CallDelegation.java
```diff
@@ -0,0 +1,257 @@
...
+          Bytes.concatenate(Bytes.fromHexString(delegationCodeToEoa2)))
+        .build();
+
+    Transaction tx = ToyTransaction.builder()
+      .sender(senderAccount)
+      .keyPair(keyPair)
+      .to(eoa1DelegatedToEoa2).build();
+
+    ToyExecutionEnvironmentV2.builder(chainConfig, testInfo)
+      .accounts(List.of(senderAccount, smcAccount, eoa1DelegatedToEoa2, eoa2))
```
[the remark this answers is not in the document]
lorenzogentile404: This is relevant, even if not harmful to leave it as it is.

tracer/arithmetization/src/test/java/net/consensys/linea/zktracer/instructionprocessing/callTests/CallDelegation.java
amkCha: Good idea, I pushed the utils function and constant :+1:

lorenzogentile404 (dismissed):

Issue #1247, opened 2025-07-15
E2E tests are failing in local environment on the main branch


mkrielza:
I'm running the `e2e/bridge-tokens.spec.ts` test from the main branch and encountering a consistent failure during the L2 → L1 bridge step. 

The L2 → L1 log shows:

timestamp=2025-07-15T11:36:14.293Z level=debug test=Bridge ERC20 Tokens L1 -> L2 and L2 -> L1 Bridge a token from L1 to L2 | message=Message claimed on L2. event={"_type":"log","address":"0xe537D669CA013d86EBeF1D64e40fC74CADC91987","blockHash":"0x822576e11c03decf55d39579b95cac509008f25cc7a108d9ed57883309c1843b","blockNumber":26,"data":"0x","index":5,"removed":false,"topics":["0xa4c827e719e911e8f19393ccdb85b5102f08f0910604d340ba38390b7ff2ab0e","0xf03cd24ca1625e83800477d5fa51218059304151531b49e2c336c7df4971fbe5"],"transactionHash":"0xc4b2fb3fb6c3b347e6c28a1859e4ff083c09fb394ef8d52552f41b64c2bee510","transactionIndex":0}.

The L2 → L1 logs shows:

timestamp=2025-07-15T11:37:09.637Z level=debug test=Bridge ERC20 Tokens L1 -> L2 and L2 -> L1 Bridge a token from L2 to L1 | message=Message claimed on L1. event=undefined

It fails to get a mapping for the L1 bridged token. It is the zero address when I manually query this.

This results in a failing test on the main branch:

 FAIL  src/bridge-tokens.spec.ts (260.69 s)
  Bridge ERC20 Tokens L1 -> L2 and L2 -> L1
    ✓ Bridge a token from L1 to L2 (82312 ms)
    ✕ Bridge a token from L2 to L1 (175729 ms)

Note that I have not modified anything.

It feels to me that this is the most basic functionality of the bridge functionality and it is not working locally. 

I am testing on Ubuntu 24.0.2 LTS. It is important to us to get a local environment working to be able to complete our assessment of the readiness of this L2 solution.

Can you please point me to a developer branch where this is in working state or help me fix the tests that is there.

PR #2371, opened 2026-02-12
fix: start postman listeners from latest in local env

Othryades:
Use -1 for L1/L2 listener initial block in local postman config so e2e bridge claims are not delayed by replaying full chain history.

This PR fixes issue #1247 

### Checklist

---

> [!NOTE]
> **Low Risk**
> Only changes local Docker environment defaults/comments; no production code paths or security-sensitive logic are modified.
> 
> **Overview**
> Local Postman Docker env config now defaults `L1_LISTENER_INITIAL_FROM_BLOCK` and `L2_LISTENER_INITIAL_FROM_BLOCK` to `-1` (start from latest) instead of `0` (genesis replay), with streamlined comments clarifying the intended local e2e behavior.
> 
> This reduces local startup/backfill time and avoids delays caused by replaying full chain history during end-to-end bridge flows.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 83b2e6ac9c69d2b1a0b1f101b0a2469e44e355cd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


docker/config/postman/env
```diff
@@ -3,11 +3,10 @@ L1_CONTRACT_ADDRESS=0xCf7Ed3AccA5a467e9e704C703E8D87F634fB0Fc9
...
 L1_LISTENER_INTERVAL=2000
-# Starting block for L1 event fetching. Controls from which block the Postman service starts fetching events.
-# Options: -1 = current latest block (default), 0 = from genesis (all historical events),
-# specific number = custom starting block (e.g., 18500000)
-# Note: This value is always prioritized when set, even if database contains previously processed messages
-L1_LISTENER_INITIAL_FROM_BLOCK=0
+# Starting block for L1 event fetching.
+# -1 = start from latest block (fast local e2e/default)
+#  0 = replay from genesis (historical backfill, slower startup)
+L1_LISTENER_INITIAL_FROM_BLOCK=-1
```
jonesho, line 9: @Othryades I'm keen on setting this to `-1`, but do you see it really helps the timeout issue? Set to -1 can speed up the event fetching, but iiuc, the impact could be quite miminal given the fresh L1 chain without too many blocks, same for  `L2_LISTENER_INITIAL_FROM_BLOCK`

VGau: @Othryades We can put -`1` but it won't  fix the issue in the e2e tests because this param is useful when we restart the postman.
We always use a fresh local stack and the postman only start once so this change won't have any effect.

Othryades: @jonesho @VGau  Why I proposed this: in my first reproduction of the issue, the failure was exactly a MessageClaimed timeout after MessageSent/anchoring, and Postman was behind. 
After setting L1/L2_LISTENER_INITIAL_FROM_BLOCK=-1, that same path passed in my next test.

Now, after you comment, I did additional fresh-stack reruns, and the fix was indeed not consistent. I keep investigating.

Othryades: @jonesho @VGau Following your feedback I've done a deeper investigation by examining the actual container logs during a test run and I think was able to identify the root cause with opus 4.6.

The real root cause seems to be a "Known transaction" retry loop in the coordinator's blob submission path. The coordinator finalizes blocks 1-15 during initial startup, but when **BlobSubmitterAsEIP4844MultipleBlobsPerTx** attempts to submit blobs for blocks 16+, the L1 node returns Known transaction. The coordinator doesn't handle this as "tx already in mempool, wait for inclusion" but instead retries with the identical signed transaction, creating an infinite loop. lastFinalizedBlockNumber never advances past 15.

Since the L2→L1 bridge message lands in block 28 (above the finalized watermark), it's never marked CLAIMABLE on L1, so the postman cannot claim it. The postman also independently hits a "Known transaction" loop on its own claim tx (nonce 0, 66+ retries before giving up).

This seems to be a coordinator service bug in EIP-4844 blob tx nonce/retry handling, not a configuration issue.
 
I'd like your guidance on whether to: 

1. File this as a separate issue against the coordinator service
2. Attempt a fix in the **BlobSubmitterAsEIP4844MultipleBlobsPerTx** retry logic (treating "Known transaction" as a success/pending state rather than a failure)


One Caveat
This may be a known issue internally that the team hasn't prioritized for local dev environments. The "Known transaction" pattern in EIP-4844 blob transactions seems to be a recognized challenge in nonce management.

Othryades: The team is apparently aware of the issue with "Known transaction" with the coordinator. I'm closing this PR.
cc @fluentcrafter @Filter94

PR #2375, opened 2026-02-13
[ZkTracer] fix conversion bytes to int/long

letypequividelespoubelles:
This PR implements issue(s) #

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Small, localized change, but it alters low-level byte-to-number conversion semantics and could change runtime behavior or throw on edge-case values where previous signed `toLong()` decoding differed.
> 
> **Overview**
> Fixes byte-slice numeric decoding by making `Conversions.bytesToLong` interpret input as *unsigned* via `toUnsignedBigInteger().longValueExact()` and enforcing the 8-byte limit.
> 
> Updates `RomLex` delegation-code detection (EIP-7702) to use `bytesToLong` and `long` intermediates, and removes the unused `Conversions.bytesToInt` helper to prevent accidental signed/overflow-prone conversions.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b9f96b3507919d92f8870e40d56fc379790bfbda. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tracer/arithmetization/src/main/java/net/consensys/linea/zktracer/types/Conversions.java
+++ b/tracer/arithmetization/src/main/java/net/consensys/linea/zktracer/types/Conversions.java
@@ -133,12 +133,6 @@ public class Conversions {
     return sb.toString().trim();
   }
 
-  public static int bytesToInt(Bytes bytes) {
-    final Bytes trimmedBytes = bytes.trimLeadingZeros();
-    checkArgument(trimmedBytes.size() <= 4, "Input bytes must be at most 4 bytes long");
-    return Math.toIntExact(bytes.trimLeadingZeros().toLong());
-  }
-
   /**
    * This method expects a "small-ish" long value and returns the corresponding int value.
    *
--- a/tracer/arithmetization/src/main/java/net/consensys/linea/zktracer/module/romlex/RomLex.java
+++ b/tracer/arithmetization/src/main/java/net/consensys/linea/zktracer/module/romlex/RomLex.java
@@ -21,7 +21,7 @@ import static net.consensys.linea.zktracer.module.ModuleName.ROM_LEX;
 import static net.consensys.linea.zktracer.types.AddressUtils.getDeploymentAddress;
 import static net.consensys.linea.zktracer.types.AddressUtils.highPart;
 import static net.consensys.linea.zktracer.types.AddressUtils.lowPart;
-import static net.consensys.linea.zktracer.types.Conversions.bytesToInt;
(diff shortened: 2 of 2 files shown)


tracer/arithmetization/src/main/java/net/consensys/linea/zktracer/types/Conversions.java
```diff
@@ -134,7 +134,9 @@ public static String bytesToHex(byte[] bytes) {
   }
 
   public static int bytesToInt(Bytes bytes) {
-    return bytes.trimLeadingZeros().toInt();
+    final Bytes trimmedBytes = bytes.trimLeadingZeros();
+    checkArgument(trimmedBytes.size() <= 4, "Input bytes must be at most 4 bytes long");
+    return Math.toIntExact(bytes.trimLeadingZeros().toLong());
```
OlivierBBB: I solved the same issue I think on the 7702 HUB PR by letting `LEAD_DELEGATION_BYTES` be a long rather than an int. Probably best to use the same approach in both PR's.

letypequividelespoubelles: it could work as well. But the issue is that this functiun can be called elsewhere and has to be fixed anyway. And once it's fixed, there is no need to have the leading bytes as long, as it's an (unsigned) int. Or the best would be to have leadingBytes a Bytes of length 4 ...

DavePearce: Ok, I resolved this by removing `bytesToInt()` and just using `bytesToLong()` instead.

tracer/arithmetization/src/main/java/net/consensys/linea/zktracer/types/Conversions.java
```diff
@@ -109,7 +109,7 @@ public static Bytes booleanToBytes(boolean x) {
   public static long bytesToLong(final Bytes input) {
     final Bytes trimmedBytes = input.trimLeadingZeros();
     checkArgument(trimmedBytes.size() <= 8, "Input bytes must be at most 8 bytes long");
-    return trimmedBytes.toLong();
+    return trimmedBytes.toUnsignedBigInteger().longValueExact();
```
OlivierBBB, line 112: Did this blow up ?

DavePearce: Not that I am aware of.  But it looks reasonable enough to me.

Issue #2097, opened 2026-01-21
Streamline the release process of linea-besu-package


jonesho:
🔐 🔐 `Should this be a security advisory instead?` 🔐 🔐

## Description
With the recent migration of `linea-tracer` (#1723) and `linea-constraints`(#1990) into `linea-monorepo`, we're now ready to proceed for streamlining the release process of `linea-besu-package` image.

Currently, the release process is not straightforward with a lot of manual works, for instance, the process of making a new `linea-besu-package` image with a `hyperledger/besu` version update goes as the followings:

1. Create a `linea-besu-upstream` release by merging a PR to specify `hyperledger/besu` commit hash and the desired linea-besu version (example [PR](https://github.com/Consensys/linea-besu-upstream/pull/65/) for the `25.12.0-linea4` release)
2. Update the `besu` field in `libs.versions.toml` with the desired `linea-besu-upstream` version, probably along with changes to make sure the e2e tests have passed (example [PR](https://github.com/Consensys/linea-monorepo/pull/2081/) for update `besu` to `25.12.0-linea4`)
3. Manually trigger the specific [workflow](https://github.com/Consensys/linea-monorepo/actions/workflows/linea-tracer-plugin-release.yml) to release a new tracer plugin and maven package (e.g. this [workflow](https://github.com/Consensys/linea-monorepo/actions/runs/21149217535) for releasing [linea-tracer-beta-v4.4-rc7.4](https://github.com/Consensys/linea-monorepo/releases/tag/linea-tracer-beta-v4.4-rc7.4))
4. Update the `arithmetization` field in `libs.versions.toml` with the desired version (e.g. [PR](https://github.com/Consensys/linea-monorepo/pull/2082/) for `beta-v4.4-rc7.4`) and then manually trigger the specific [workflow](https://github.com/Consensys/linea-monorepo/actions/workflows/linea-sequencer-plugin-release.yml) to release a new sequencer plugin release with a desired version (e.g. this [workflow](https://github.com/Consensys/linea-monorepo/actions/runs/21161842894) for releasing [linea-sequencer-v4.4-rc7.4.1](https://github.com/Consensys/linea-monorepo/releases/tag/linea-sequencer-v4.4-rc7.4.1)) [see this [doc](https://www.notion.so/consensys/Besu-4-Linea-HowTo-Release-4ae7ea9b2f0e4ad29a7c7485c2571336#237fc61a326e8064b57ae37797163987) from Fabio for the detailed instructions]
[...]

PR #2396, opened 2026-02-16
Feat/support besu local build for tracer sequencer

jonesho:
This PR implements issue(s) #2097

What have been added in this PR:
- Added a field called `besuCommit` in `gradle/libs.versions.toml` to denote the desired besu commit hash from [hyperledger/besu](https://github.com/hyperledger/besu)
- Added gradle tasks to locally build the besu distribution and publish besu (the version would be `[latest-release-tag]-[7-digit-commit-hash]`, e.g. `25.12.0-d68679d` ) to public/local maven at `besuCommit`, and automatically update the `besu` in `gradle/libs.versions.toml`
- Added Makefile commands in `linea-besu-package/Makefile` to build the `linea-besu-package` image locally with locally-built besu, tracer, and sequencer plugins (e.g. `make build` will do the trick) 
- Added a GitHub workflow to publish the besu tar.gz distribution to [cloudsmith](https://cloudsmith.io/~consensys/repos/linea-besu/packages/) from pull requests after e2e test passes to avoid repeated local build in pipelines **AND** automatically push a commit to update the `besu` in `gradle/libs.versions.toml` if needed
- Revised manual release [workflow](https://github.com/Consensys/linea-monorepo/actions/workflows/linea-besu-package-release.yml) to have an user-input `prefix` for the release version (given that the default prefix which was the `arithmetization` in `gradle/releases.versions.toml` has been removed in this PR)

Impacts:
- Get rid of the need to manually build and publish `linea-besu` from https://github.com/Consensys/linea-besu-upstream when upgrading the besu version for `linea-besu-package` release
- E.g. when the besu version needs to be updated for a linea-besu-package release, simply update the `besuCommit` in `gradle/libs.versions.toml` to denote the desired besu commit hash in the PR, push the commit, and the workflow would build the corresponding `linea-besu-package` image to run e2e tests, and after the tests passed, it will publish the new `linea-besu` on [maven](https://cloudsmith.io/~consensys/repos/linea-besu/packages/), and auto-push a commit to update the `besu` field in `gradle/libs.versions.toml`. Alternatively, to run e2e tests locally with new `besuCommit`, simply run `cd linea-besu-package && make build && make run-e2e-tests` 
- Remove the besu, tracer, and sequencer versions in `linea-besu-package/versions.env` for downloading the artifacts as they would always be building from source
- Remove sequencer manual release workflow
- Remove `releases.versions.toml`
- Remove the auto-update of coordinator config `expected-traces-api-version` in workflows and in makefile command `run-e2e-test`

### Checklist

---

> [!NOTE]
> **High Risk**
> High risk because it rewires CI/release pipelines to build Besu from a pinned commit, publish artifacts (including Cloudsmith) and auto-push `gradle/libs.versions.toml` updates, which can impact build/release integrity and permissions.
> 
> **Overview**
> **Moves Besu packaging to a commit-pinned, source-built workflow.** Adds `besuCommit` to `gradle/libs.versions.toml`, introduces a new Gradle plugin (`linea.besu-plugin`) with tasks to clone `besu-eth/besu`, resolve a version (`tag` + 7-char commit), build/publish it (optionally to Maven/Cloudsmith), and auto-update the `besu` entry.
> 
> **Reworks linea-besu-package assembly and CI metadata/tagging.** The package build now always assembles from locally built artifacts (Besu `.tar.gz` + tracer/sequencer `installDist` outputs) and removes the separate `releases.versions.toml`/arithmetization-based versioning; tracer/sequencer versions become `linea-<commit>`, release tags are driven by a new `release_tag_prefix` input, and workflows stop rewriting `expected-traces-api-version` during e2e.
> 
> **Updates CI/release workflows and triggers.** Adds a `linea-besu-build-and-publish` workflow that runs after e2e on same-repo PRs to publish Besu and push the `libs.versions.toml` update, updates path filters and lib-version checks to key off `besuCommit`, removes the manual sequencer release workflow, and bumps Postman native libs tag plus a small transaction signing fix (`chainId` set non-zero).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3f1509dd3bc7c3ecab3b3b69740558a99a83d283. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tracer/gradle/corset.gradle
+++ b/tracer/gradle/corset.gradle
@@ -28,7 +28,7 @@ tasks.register('corsetExists') {
 ext {
   corsetVersion = project.hasProperty('corsetVersion')
     ? project.property('corsetVersion')
-    : System.getenv().getOrDefault('CORSET_VERSION', "v1.2.6")
+    : System.getenv().getOrDefault('CORSET_VERSION', "v1.2.7")
 }
 
 tasks.register('installGoCorset') {


linea-besu-package/Makefile
```diff
@@ -2,39 +2,42 @@ include versions.env
...
+build-tracer-and-sequencer:
+	cd ..; \
+	echo "Build versions: TRACER_PLUGIN_VERSION=${TRACER_PLUGIN_VERSION} SEQUENCER_PLUGIN_VERSION=$(SEQUENCER_PLUGIN_VERSION)" ; \
 	echo "Clean and build zip distribution for tracer" ; \
 	./gradlew :tracer:clean && ./gradlew :tracer:artifacts -PreleaseVersion=$(TRACER_PLUGIN_VERSION) ; \
 	echo "Clean and build zip distribution for sequencer" ; \
 	./gradlew besu-plugins:linea-sequencer:sequencer:clean && ./gradlew besu-plugins:linea-sequencer:sequencer:artifacts -PreleaseVersion=$(SEQUENCER_PLUGIN_VERSION)
 
+assemble: BESU_VERSION := $(shell grep -E '^\s*besu\s*=' ../gradle/libs.versions.toml | sed -n 's/.*"\([^"]*\)".*/\1/p')
+assemble: LOCAL_BESU_ZIP_PATH := $(LOCAL_BESU_DIST_FOLDER)/besu-$(BESU_VERSION).tar.gz
```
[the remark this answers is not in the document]
jonesho: Not valid. It parses fine.

.github/actions/linea-besu-package/assemble/action.yml
```diff
@@ -17,40 +17,36 @@ inputs:
...
   steps:
-    - name: Set local zip paths (build from source)
-      if: ${{ inputs.local_besu_version != '' && inputs.local_arithmetization_version != '' && inputs.local_sequencer_version != '' }}
+    - name: Set local zip paths
       id: set_local_zip_paths
       shell: bash
       run: |
         echo "local_tracer_zip_path=$GITHUB_WORKSPACE/tracer/plugins/build/distributions/linea-tracer-${{ inputs.local_arithmetization_version }}.zip" >> $GITHUB_OUTPUT
         echo "local_sequencer_zip_path=$GITHUB_WORKSPACE/besu-plugins/linea-sequencer/sequencer/build/distributions/linea-sequencer-${{ inputs.local_sequencer_version }}.zip" >> $GITHUB_OUTPUT
+        echo "local_besu_zip_path=$GITHUB_WORKSPACE/tmp/hyperledger-besu/build/distributions/besu-${{ inputs.local_besu_version }}.tar.gz" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
jonesho: Not relevant

.github/workflows/linea-besu-build-and-publish.yml
```diff
@@ -0,0 +1,91 @@
...
+          LINEA_BESU_VERSION: ${{ steps.release_versions.outputs.besu_version }}
+
+      - name: Commit and push gradle/libs.versions.toml besu version update to PR
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          if git diff --quiet gradle/libs.versions.toml; then
+            echo "No changes to gradle/libs.versions.toml, skipping commit"
+          else
+            git pull
```
Filter94: Do we need a pull here? In case this job is triggered from a branch, on the checkout step it will be on a merge commit already. If it's triggered from main, IIUC, it will also be on the latest commit, from which the job is triggered, so no need to pull

.github/workflows/linea-besu-build-and-publish.yml
```diff
@@ -0,0 +1,91 @@
...
+          ORG_GRADLE_PROJECT_cloudsmithApiKey: ${{ secrets.CLOUDSMITH_API_KEY }}
+      
+      - name: Parse besu version from lib.versions.toml
+        id: release_versions
+        uses: ./.github/actions/parse-release-versions
+
+      - name: Check besu distribution exists
+        id: check_besu_dist
+        run: |
+          BESU_DIST="tmp/hyperledger-besu/build/distributions/besu-${{ steps.release_versions.outputs.besu_version }}.tar.gz"
```
Filter94: I get why we'd need to publish besu jar-s to Cloudsmith, but where `tar.gz` distribution is used exactly?

build.gradle
```diff
@@ -175,6 +175,186 @@ task jacocoRootReport(type: JacocoReport) {
...
+    def resolvedBesuVer = rootProject.ext.resolvedBesuVer
+    def shouldSkip = false
+    if (publishToMaven) {
+      shouldSkip = isBesuAvailableInMaven.call(resolvedBesuVer) && downloadBesuDistributionFromMaven.call(resolvedBesuVer, true)
+    } else {
+      if (isBesuAndDistributionAvailableInMavenLocal.call(resolvedBesuVer)) shouldSkip = true
+      else if (isBesuAvailableInMaven.call(resolvedBesuVer) && downloadBesuDistributionFromMaven.call(resolvedBesuVer)) shouldSkip = true
+    }
+    if (shouldSkip) {
+      throw new org.gradle.api.tasks.StopExecutionException("Skipping checkoutAndBuildBesu: Besu ${resolvedBesuVer} already available")
```
Filter94: 1. Why is this an exception?
2. Is it possible to leverage gradle input and outputs for this so it manages it as a cache seamlessly?

build.gradle
```diff
@@ -175,6 +175,186 @@ task jacocoRootReport(type: JacocoReport) {
...
+      destFile.delete()
+      logger.lifecycle("downloadBesuDistributionFromMaven: Removed partial/corrupt file so next run can retry: ${destFile}")
+    }
+    return false
+  }
+}
+
+def besuCommit = libs.versions.besuCommit.get()
+def besuCommitShort = besuCommit.take(7)
+def checkoutAndResolveVersionScript = """
```
Filter94: Those chunks of bash code inside gradle sources look very alien and clutter the main build file. Consider moving this to a makefile, also to make it reusable manually

build.gradle
```diff
@@ -175,6 +175,186 @@ task jacocoRootReport(type: JacocoReport) {
   onlyIf = { true }
 }
 
+/** Returns true if the given Besu version is available in Maven local and the distribution exists */
```
Filter94: I think all of this new gradle code needs to be moved away from here into a custom gradle plugin and used here as complete gradle tasks only

tracer/build.gradle
```diff
@@ -76,14 +75,33 @@ sonar {
...
 }
 
 distributions {
   main {
     distributionBaseName = distibutionBaseName
+    version = calculateVersion()
+  }
+}
+
+def buildAndUpdateBesuVersionInLibsVersions = rootProject.tasks.named('buildAndUpdateBesuVersionInLibsVersions')
```
Filter94: Maybe this can also be moved to a gradle plugin and reused?

build.gradle
jonesho: Switched to `doLast` to avoid throwing StopExecutionException

.github/workflows/linea-besu-build-and-publish.yml
```diff
@@ -0,0 +1,83 @@
...
+  build:
+    environment: publish
+    runs-on: gha-runner-scale-set-ubuntu-22.04-amd64-med
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.ref }}
+
+      - name: Setup Java and Gradle
```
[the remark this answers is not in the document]
fluentcrafter: @jonesho this seems relevant and shall be addressed.

besu-plugins/linea-sequencer/sequencer/build.gradle
```diff
@@ -44,11 +44,17 @@ distZip {
   exclude '**/arithmetization-*.jar'
 }
 
-// Simple artifacts task that runs jar and distPlugin
+// Simple artifacts task that runs buildAndUpdateBesuVersionInLibsVersions, then jar and distZip
 task artifacts {
   group = 'Build'
-  description = 'Creates all build artifacts (JAR and distribution ZIP)'
-  dependsOn 'jar', 'distZip'
+  description = 'Creates all build artifacts (JAR and distribution ZIP). Ensures hyperledger/besu is checked out and built first, then syncs besu version to libs.versions.toml.'
```
fluentcrafter: What do we need the ZIP for?

If not necessary, let's remove it. Assembling artefacts take not negligible time...

build.gradle
```diff
@@ -175,6 +175,8 @@ task jacocoRootReport(type: JacocoReport) {
   onlyIf = { true }
 }
 
+apply plugin: 'linea.besu-plugin'
```
fluentcrafter, line 178: why do we apply this here on the root project to all projects?

.github/workflows/linea-besu-build-and-publish.yml
jonesho: The tar.gz distribution would be used when assembling the linea-besu-package during image build, see https://github.com/Consensys/linea-monorepo/blob/feat/support-besu-local-build-for-tracer-sequencer/linea-besu-package/scripts/assemble-packages.sh#L36

.github/workflows/linea-besu-build-and-publish.yml
jonesho: Yes, we need a `git pull` here, the workflow is checkout with the pull request source branch (`github.head_ref`) and hence we need to pull the latest changes before we push the commit; otherwise, we'd get push commit rejected.

linea-besu-package/versions.env
```diff
@@ -1,11 +1,3 @@
-# LINEA_BESU_TAR_GZ, LINEA_SEQUENCER_PLUGIN_VERSION, and LINEA_TRACER_PLUGIN_VERSION 
-# will only take effect on manual linea-besu-package release
-LINEA_BESU_TAR_GZ=25.12.0-linea4 
-LINEA_SEQUENCER_PLUGIN_VERSION=4.4-rc7.4.1
-LINEA_TRACER_PLUGIN_VERSION=beta-v4.4-rc7.4
 LINEA_STATERECOVERY_PLUGIN_VERSION=1.0.4
```
fluentcrafter, line 1: This plugin should also be built locally and used. But it can be done on a follow-up PR

linea-besu/scripts/build-dist-and-publish.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e
+
+PUBLISH_TASK="${1:?publish task name required (e.g. publish or publishToMavenLocal)}"
+
+echo "Injecting Cloudsmith publish to $BESU_DIR/build.gradle"
+java ./linea-besu/scripts/InjectLines.java "$BESU_DIR"/build.gradle
```
fluentcrafter, line 7: Why do we need this? This feels like an anti-pattern/code smell to edit the build scripts.

Build scripts, are scripts what support logic, so they should not need this "hacky" bash scripts.

linea-besu-package/scripts/assemble-packages.sh
```diff
@@ -8,49 +8,51 @@ source ./versions.env
...
 echo "copying the versions.env to the container as versions.txt"
 cp ../versions.env ./besu/versions.txt
 
 mkdir -p ./besu/plugins
 cd ./besu/plugins
 
-echo "downloading the plugins and copying if the following local artifact paths are not empty:"
-echo "LOCAL_SEQUENCER_ZIP_PATH: $LOCAL_SEQUENCER_ZIP_PATH"
-echo "LOCAL_TRACER_ZIP_PATH: $LOCAL_TRACER_ZIP_PATH"
+echo "using local sequencer zip: $LOCAL_SEQUENCER_ZIP_PATH"
```
fluentcrafter: why do we use ZIP as asset "carry overs"  when we have access deirectly to the distributions folder? 

Wouldn't this be more efficient?

gradle/libs.versions.toml
```diff
@@ -36,7 +36,12 @@ besuShomeiPluginVersion = "v0.8.4"
 undercouchDownload = "5.6.0"
 
 # Runtime
-besu = "25.12.0-linea4.1"
+# Besu commit hash from https://github.com/hyperledger/besu
+# when updates it, should run "gradlew buildAndUpdateBesuVersionInLibsVersions" to
+# checkout and build Besu from source, and will auto-update the "besu" version below
+besuCommit = "d68679d6887abf8c407f99456cf2502004fa2bad"
+
+besu = "25.12.0-linea-d68679d"
```
fluentcrafter: IMO, besu version should be without `linea-`. Why because it's the vanila Besu without any Linea custom logic/plugin.

buildSrc/src/main/groovy/LineaBesuPlugin.groovy
```diff
@@ -0,0 +1,186 @@
...
+        libsVersionsFile.text = content
+        project.logger.lifecycle("Updated gradle/libs.versions.toml: besu = \"${localBesuVersion}\"")
+      }
+    }
+  }
+
+  private static boolean isBesuAndDistributionAvailableInMavenLocal(Project project, String version) {
+    if (!version) return false
+    def mavenLocalPom = new File(System.getProperty('user.home'),
+        ".m2/repository/org/hyperledger/besu/bom/${version}/bom-${version}.pom")
```
fluentcrafter: @copilot review this

fluentcrafter: My gut feeling told me the hardcoded values could a problem, calude code confirmed.

# Claude Review


  

## Bug: Regex can match unintended keys (line 91)

```groovy
content = content.replaceFirst(/besu\s*=\s*"[^"]*"/, "besu = \"${localBesuVersion}\"")
```

The pattern `besu\s*=` has no word boundary, so it can match `besuCommit = "..."` or `besuRelease = "..."` if they appear before the `besu = "..."` entry in the file. A safer pattern:

```groovy
content = content.replaceFirst(/(?m)^besu\s*=\s*"[^"]*"/, "besu = \"${localBesuVersion}\"")
```

The `(?m)` flag with `^` anchors the match to the start of a line, preventing partial-key matches.

---

## Resource Leak: `HttpURLConnection` never disconnected

Both `isBesuAvailableInMaven` (line 127) and `downloadBesuDistributionFromMaven` (line 153) open connections that are never explicitly closed:

```groovy
def conn = (java.net.HttpURLConnection) new URL(pomUrl).openConnection()
// ... conn.disconnect() is never called
```

The fix is to call `conn.disconnect()` in a `finally` block or use a try-with-resources pattern.

---

## Code Smell: Dual-mode function via boolean flag (line 142)

`downloadBesuDistributionFromMaven` accepts a `checkOnly` parameter that radically changes its behavior — it either does a `HEAD` check or a full `GET` download. This violates single responsibility and makes call sites harder to read (e.g., line 58: `downloadBesuDistributionFromMaven(project, resolvedBesuVer, true)`).

**Suggestion:** Split into two methods: `canDownloadBesuDistributionFromMaven` and `downloadBesuDistributionFromMaven`.

---

## Hardcoded Maven Local Repository Path (line 100)

```groovy
def mavenLocalPom = new File(System.getProperty('user.home'),
    ".m2/repository/org/hyperledger/besu/...")
```

The Maven local repo can be overridden via `~/.m2/settings.xml` or the `maven.repo.local` system property. This hardcodes `~/.m2/repository` unconditionally. At minimum, respect the `maven.repo.local` system property:

```groovy
def mavenLocalBase = System.getProperty('maven.repo.local')
    ?: "${System.getProperty('user.home')}/.m2/repository"
```

---

## Verbose Intermediate Logging in `isBesuAndDistributionAvailableInMavenLocal` (lines 102–112)

The method logs a message for every intermediate check even when called as part of a larger decision. The individual existence logs for POM and tar.gz files are noisy and fire regardless of the outcome. Consider logging only the final result, or using `project.logger.debug` for intermediate checks.

---

## `shouldSkip` Logic Is More Verbose Than Needed (lines 55–66)

```groovy
def shouldSkip = false
if (publishToMaven) {
  shouldSkip = isBesuAvailableInMaven(...) && downloadBesuDistributionFromMaven(...)
} else {
  if (isBesuAndDistributionAvailableInMavenLocal(...)) {
    shouldSkip = true
  } else if (isBesuAvailableInMaven(...) && downloadBesuDistributionFromMaven(...)) {
    shouldSkip = true
  }
}
```

The `else` branch can be flattened:

```groovy
def shouldSkip = publishToMaven
    ? isBesuAvailableInMaven(project, resolvedBesuVer) && downloadBesuDistributionFromMaven(project, resolvedBesuVer, true)
    : isBesuAndDistributionAvailableInMavenLocal(project, resolvedBesuVer) ||
      (isBesuAvailableInMaven(project, resolvedBesuVer) && downloadBesuDistributionFromMaven(project, resolvedBesuVer))
```

---

## Minor: Stale `distTar` Reference in Task Description (line 46)

```groovy
description = 'Build Besu at the resolved version (distTar publish/publishToMavenLocal)'
```

The actual artifact handled is `besu-${version}.tar.gz`, not a Gradle `distTar` task output. The description should be corrected.

---

## Summary

| Severity | Issue |
|----------|-------|
| Bug      | Regex on line 91 can match unintended keys (e.g., `besuCommit`) |
| Medium   | `HttpURLConnection` never disconnected — resource leak in two methods |
| Medium   | Hardcoded Maven local path ignores `maven.repo.local` override |
| Low      | `checkOnly` flag in `downloadBesuDistributionFromMaven` — split into two methods |
| Low      | Noisy intermediate `lifecycle` logging in `isBesuAndDistributionAvailableInMavenLocal` |
| Low      | `shouldSkip` logic can be simplified |
| Trivial  | Stale `distTar` reference in task description |

```

Comment (jonesho) [reply buildSrc/src/main/groovy/LineaBesuPlugin.groovy]: The bug is legit, fixed along with revisions for other copilot comments

File: linea-besu-package/versions.env (hunk 12 above)
Comment (jonesho) [reply linea-besu-package/versions.env:1]: Sure, created a ticket to follow up: https://github.com/Consensys/linea-monorepo/issues/2574

File: linea-besu-package/scripts/assemble-packages.sh (hunk 14 above)
Comment (jonesho) [reply linea-besu-package/scripts/assemble-packages.sh]: Let me use run `installDist` to build the tracer and sequencer jar files and use them directly when assembling `linea-besu-package`

File: build.gradle (hunk 11 above)
Comment (jonesho) [reply build.gradle:178]: It only applies to the root project — not to all subprojects. It's not inside an allprojects {} or subprojects {} block.

File: linea-besu/scripts/build-dist-and-publish.sh (hunk 13 above)
Comment (jonesho) [reply linea-besu/scripts/build-dist-and-publish.sh:7]: This script is originated from `linea-besu-upstream` and we need this because `hyperledger/besu` doesn't support publishing to cloudsmith `linea-besu` repo during the `publish` gradle task

File: linea-besu-package/scripts/assemble-packages.sh [hunk 17]
```diff
@@ -8,49 +8,47 @@ source ./versions.env
...
-  unzip -j -o $(basename "$LOCAL_TRACER_ZIP_PATH")
-  rm $(basename "$LOCAL_TRACER_ZIP_PATH")
-else
-  echo "getting linea_tracer_plugin_version: $LINEA_TRACER_PLUGIN_VERSION"
-  wget -nv https://github.com/Consensys/linea-monorepo/releases/download/linea-tracer-$LINEA_TRACER_PLUGIN_VERSION/linea-tracer-$LINEA_TRACER_PLUGIN_VERSION.zip
-  unzip -j -o linea-tracer-$LINEA_TRACER_PLUGIN_VERSION.zip
-  rm linea-tracer-$LINEA_TRACER_PLUGIN_VERSION.zip
-fi
+echo "using JAR files under local tracer distribution folder: $LOCAL_TRACER_DIST_FOLDER"
+cp -a $LOCAL_TRACER_DIST_FOLDER/. .
```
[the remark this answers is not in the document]
jonesho: This is not valid, as the tracer and sequencer jars are all placed under `tracer/plugins/build/install/linea-tracer` and `besu-plugins/linea-sequencer/sequencer/build/install/sequencer` without any subfolders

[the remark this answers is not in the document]
jonesho: This is not valid, as the tracer and sequencer jars are all placed under tracer/plugins/build/install/linea-tracer and besu-plugins/linea-sequencer/sequencer/build/install/sequencer without any subfolders

besu-plugins/linea-sequencer/sequencer/build.gradle
```diff
@@ -41,7 +41,6 @@ apply from: lineaSequencerProject.file("gradle/dist.gradle")
 
 distZip {
```
fluentcrafter: We don't need distZip/distTar anymore

gradle/libs.versions.toml
```diff
@@ -36,7 +36,12 @@ besuShomeiPluginVersion = "v0.8.4"
 undercouchDownload = "5.6.0"
 
 # Runtime
-besu = "25.12.0-linea4.1"
+# Besu commit hash from https://github.com/hyperledger/besu
+# when updates it, should run "gradlew buildAndUpdateBesuVersionInLibsVersions" to
+# checkout and build Besu from source, and will auto-update the "besu" version below
+besuCommit = "d68679d6887abf8c407f99456cf2502004fa2bad"
+
+besu = "25.12.0-d68679d"
```
fluentcrafter, line 44: add comment that besu is updated autmatically.

PR #2466, opened 2026-02-22
refactor: combine notify attempt increment with state transition

kyzooghost:
This PR implements issue(s) # N/A

## Summary

- Remove the separate `incrementNotifyAttempt` DB write that occurred before the Slack send in `NotificationService`
- Combine the attempt count increment atomically with the state transition in `markNotified` (success path) and a new `markNotifyFailed` method (failure path)
- Add `attemptMarkNotifyFailed` as a best-effort wrapper for the catch block
- Remove now-unused `incrementNotifyAttempt` from `IProposalRepository` interface and `ProposalRepository` implementation

## Problem

The notification flow performed a separate DB write to increment `notifyAttemptCount` before sending the Slack notification, then performed another DB write for the state transition. This created an unnecessary intermediate round-trip and a window where the counter and state could be inconsistent (e.g., counter incremented but state not yet transitioned if the process crashed between the two writes).

## Solution

The attempt count is now incremented in the same Prisma `update` call as the state transition:
- **Success:** `markNotified` sets state=NOTIFIED, notifiedAt, and increments `notifyAttemptCount`
- **Failure:** New `markNotifyFailed` sets state=NOTIFY_FAILED and increments `notifyAttemptCount`
- **Error (catch):** `attemptMarkNotifyFailed` replaces `attemptUpdateState` for best-effort recovery

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches core notification persistence and retry behavior; while the logic is simpler and better aligned, any mistake could impact retry counts or state transitions for alerts.
> 
> **Overview**
> Refactors the notification pipeline to remove the separate `incrementNotifyAttempt` write and instead **atomically** increment `notifyAttemptCount` during state transitions.
> 
> `markNotified` now increments attempts on success, and a new `markNotifyFailed`/`attemptMarkNotifyFailed` path handles validation failures, Slack send failures, and catch-block recovery; the repository interface/implementation and unit/integration tests are updated accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bbe66607680f7f08a7c3f664a07bb51cb3d891af. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/native-yield-operations/lido-governance-monitor/src/services/NotificationService.ts
+++ b/native-yield-operations/lido-governance-monitor/src/services/NotificationService.ts
@@ -51,8 +51,7 @@ export class NotificationService implements INotificationService {
       // Validate assessment exists
       if (!proposal.assessmentJson) {
         this.logger.error("Proposal missing assessment data", { proposalId: proposal.id });
-        await this.proposalRepository.incrementNotifyAttempt(proposal.id);
-        await this.proposalRepository.updateState(proposal.id, ProposalState.NOTIFY_FAILED);
+        await this.proposalRepository.markNotifyFailed(proposal.id);
         return;
       }
 
@@ -62,8 +61,7 @@ export class NotificationService implements INotificationService {
           proposalId: proposal.id,
           errors: parseResult.error.errors,
         });
-        await this.proposalRepository.incrementNotifyAttempt(proposal.id);
-        await this.proposalRepository.updateState(proposal.id, ProposalState.NOTIFY_FAILED);
+        await this.proposalRepository.markNotifyFailed(proposal.id);
         return;
       }
       const assessment = parseResult.data;
@@ -90,9 +88,6 @@ export class NotificationService implements INotificationService {
         return;
       }
 
-      // Increment attempt count before notification
-      const updated = await this.proposalRepository.incrementNotifyAttempt(proposal.id);
-
       // Send Slack notification
       const result = await this.slackClient.sendProposalAlert(proposal, assessment);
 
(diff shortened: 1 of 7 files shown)


native-yield-operations/lido-governance-monitor/src/services/NotificationService.ts
```diff
@@ -90,9 +88,6 @@ export class NotificationService implements INotificationService {
         return;
       }
 
-      // Increment attempt count before notification
-      const updated = await this.proposalRepository.incrementNotifyAttempt(proposal.id);
```
alainncls, line 94: Because the attempt count now increments only after Slack succeeds, a DB failure in `markNotified()` (and then in the fallback) leaves the proposal in `ANALYZED` with the same `notifyAttemptCount`. That means the same alert can be sent again on every polling cycle after recovery. We should persist the attempt before the external side effect, otherwise `maxNotifyAttempts` no longer bounds duplicate sends on this path.

kyzooghost: > Because the attempt count now increments only after Slack succeeds

This is incorrect, if Slack fails then it also increments the attempt count through L101
```
const updated = await this.proposalRepository.markNotifyFailed(proposal.id);
```

> a DB failure in markNotified() (and then in the fallback) leaves the proposal in ANALYZED with the same notifyAttemptCount

So this is the dual-write dilemma:
- We perform an external side effect (Slack send) and then persist the outcome in the DB, but those two writes are not atomic. If Slack succeeds and the DB update fails, the system forgets that the alert was already sent and may retry, causing duplicates.
- Practically we have to choose between "at most once" and "at least once". 
    - In our situation we can tolerate duplicate Slack messages so we choose "at least once"
- There is an academic "2 phase-commit" solution for the dual-write dilemma, but it's not practical

PR #2499, opened 2026-02-26
Prover/eip7702-setup-verifier-contracts

srinathln7:
This PR implements issue(s) #

### Checklist

---

> [!NOTE]
> **High Risk**
> High risk because it rewires how `traces_limits` are represented/loaded (impacting setup/verifier compatibility) and alters distributed constraint compilation by rewriting queries involving shifted columns, which can affect proof correctness and sizing.
> 
> **Overview**
> Refactors `traces_limits` from a large struct (plus separate `traces_limits_large`) into a prefix-matched `modules[]` table with per-module `limit`/`limit_large`, scaling support, and large-mode toggling via `SetLargeMode()`; all configs are migrated to the new format and defaults for trace limits are effectively removed.
> 
> Updates prover/circuit code to consume limits via accessor methods (e.g. `BlockL2L1Logs()`), bumps `go-corset`/constraints versions, and tweaks several environment configs (notably sepolia execution mode/paths and mainnet/sepolia version).
> 
> Fixes distributed proving with shifted columns by adding a `manual_shifter` compiler pass that ignores inclusion/permutation/projection queries containing `column.Shifted`, replays equivalent queries using dedicated manually-shifted natural columns, and registers an `AssignManualShifts` prover action; auditing now skips ignored queries and module discovery tightens size-changing rules when precomputed/VK columns are present.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0c2d32ed76e54deced2e074b5e384f123f4ed9d5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


prover/zkevm/arithmetization/files.go
```diff
@@ -78,16 +79,18 @@ func UnmarshalZkEVMBin(buf []byte) (*binfile.BinaryFile, typed.Map, error) {
...
 	// Lower to mixed micro schema
 	uasmSchema := asm.LowerMixedMacroProgram(asmConfig.Vectorize, binf.Schema)
 	// Apply register splitting for field agnosticity
-	mirSchema, mapping := asm.Concretize[bls12_377.Element, bls12_377.Element](asmConfig.Field, uasmSchema)
+	nasmSchema, mapping := asm.Concretize[bls12_377.Element](asmConfig.Field, uasmSchema)
+	// Compile
+	mirSchema := asm.Compile(nasmSchema)
 	// Lower to AIR
-	airSchema := mir.LowerToAir(mirSchema, *optConfig)
+	airSchema := mir.LowerToAir(mirSchema, 30, *optConfig)
```
DavePearce: You should not be using a direct constant `30` here.  It could come from the field configuration, which is in `asmConfig.Field`.  Since you are using `BLS12_377` here ... then `30` is not a correct value for the bandwidth and will cause significant issues.

prover/zkevm/prover/ecdsa/ecdsa.go
```diff
@@ -50,7 +50,7 @@ func getEcdataArithmetization(comp *wizard.CompiledIOP) *ecDataSource {
 
 func getTxnDataArithmetization(comp *wizard.CompiledIOP) *txnData {
 	td := &txnData{
-		FromHi:   comp.Columns.GetHandle("txndata.hubFROM_ADDRESS_HI_xor_rlpCHAIN_ID"),
+		FromHi:   comp.Columns.GetHandle("txndata.hubFROM_ADDRESS_HI_xor_rlpNUMBER_OF_NONZERO_BYTES"),
```
DavePearce, line 53: Using direct column names is not recommended as they are subject to change.  This problem has been solved in the `prover/dev-small-fields` branch and you should look at that.  Specifically, in `arithmetization.go` the function `LimbsOf` takes a source-level module and column name, and returns its actual underlying column name.  So, for example, in this case you would request e.g. `LimbsOf("txndata.hub", "FROM_ADDRESS_HI")` and it would return `"txndata.hubFROM_ADDRESS_HI_xor_rlpNUMBER_OF_NONZERO_BYTES"`.  The point being that it figures out the raw name for you which protects you when the name changes.

gusiri (approved):

Issue #2503, opened 2026-02-26
Switchable compressor version


Filter94:
🔐 🔐 `Should this be a security advisory instead?` 🔐 🔐

## Description
Currently in Sequencer plugin Compressor version is hardcoded. For the future releases it would be nice to be able to switch it without restarting the coordinator

## Motivation
It's useful so we can simplify the upgrade process and not have to manage multiple linea-besu-package releases during the upgrade process

## Tasks
- [ ] Make the Compressor switchable based on timestamp without a restart
- [ ] Do we need it to be switchable based on a block number?
- [ ] Make sure the Compression aware selector is aligned with the V3 compressor
- [ ] Make sure the Compressor cache is completely invalidated on the switch

## Acceptance criteria
- [ ] Compressor inside linea-besu-package is switchable without restart

## Risks
- [ ] Compressor cache may retain records from an incompatible version
- [ ] The current assumptions about V2 compressor might not hold for V3. E.g. we don't include signatures into compression, we include `from` separately, BlobMaker only puts a timestamp inside the compressed blob, etc.

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #2538, opened 2026-03-04
[jvm-libs][blob-compressor] Add timestamp based selector for blob compressor

gauravahuja:
This PR implements issue(s) #2503 

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> Introduces a new `BlobCompressor` API surface (`version`) and new selection logic that instantiates and reuses multiple native-backed compressors, which could affect consumers and runtime behavior if misconfigured (e.g., timestamp mapping).
> 
> **Overview**
> Adds `BlobCompressor.version` and wires it through `GoBackedBlobCompressor`, enabling callers (including Java) to introspect the active compressor version.
> 
> Introduces `BlobCompressorSelector` and a timestamp-based implementation (`BlobCompressorSelectorByTimestamp`) that chooses the compressor version based on the latest switch time at or before a given `Instant`, pre-instantiating one `GoBackedBlobCompressor` per configured version.
> 
> Adds tests covering version exposure, timestamp selection/validation, multi-version compression behavior, and basic concurrent usage across two compressor instances.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 310e78cff04e1f2761c3973253fdaa69176713cd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/jvm-libs/linea/blob-compressor/src/test/kotlin/linea/blob/BlobCompressorSelectorTest.kt
+++ b/jvm-libs/linea/blob-compressor/src/test/kotlin/linea/blob/BlobCompressorSelectorTest.kt
@@ -114,13 +114,13 @@ class BlobCompressorSelectorTest {
     val compressorV3 = selector.getBlobCompressor(t3)
     assertTrue(compressorV2 != compressorV3)
 
-    val sampleBlocks = CompressorTestData.blocksRlpEncoded.take(100)
+    val sampleBlocks = CompressorTestData.blocksRlpEncoded
     assertThat(sampleBlocks).isNotEmpty
     val compressedV2 = compressBlocks(compressorV2, sampleBlocks)
     val compressedV3 = compressBlocks(compressorV3, sampleBlocks)
 
-    assertThat(compressedV2).isNotEqualTo(compressBlocks(compressorV2, sampleBlocks))
-    assertThat(compressedV3).isNotEqualTo(compressBlocks(compressorV3, sampleBlocks))
+    assertThat(compressedV2).isEqualTo(compressBlocks(compressorV2, sampleBlocks))
+    assertThat(compressedV3).isEqualTo(compressBlocks(compressorV3, sampleBlocks))
 
     val compressedV2Parallel: AtomicReference<ByteArray> = AtomicReference(ByteArray(0))
     val compressedV3Parallel: AtomicReference<ByteArray> = AtomicReference(ByteArray(0))
@@ -154,10 +154,10 @@ class BlobCompressorSelectorTest {
     v2Thread.start()
     v3Thread.start()
     startBarrier.await()
+    assertTrue(doneLatch.await(10, TimeUnit.SECONDS), "compression threads did not complete in time")
+
     assertThat(compressedV2Parallel.load()).isEqualTo(compressedV2)
     assertThat(compressedV3Parallel.load()).isEqualTo(compressedV3)
-
-    assertTrue(doneLatch.await(10, TimeUnit.SECONDS), "compression threads did not complete in time")
     assertThat(errors).isEmpty()
   }
 


jvm-libs/linea/blob-compressor/src/test/kotlin/linea/blob/BlobCompressorSelectorTest.kt
```diff
@@ -0,0 +1,59 @@
...
+      mapOf(v1 to t1, v2 to t2),
+      dataLimit,
+    )
+    // Should select v2 for t2, v1 for t1 and mid-timestamps
+    val compressorV2 = selector.getBlobCompressor(t2)
+    val compressorV1 = selector.getBlobCompressor(t1)
+    val compressorMid = selector.getBlobCompressor(Instant.parse("2025-06-01T00:00:00Z"))
+    assertNotNull(compressorV2)
+    assertNotNull(compressorV1)
+    assertNotNull(compressorMid)
```
jonesho, line 38: Can we check if the `compressorV2` is really the one for `BlobCompressorVersion.V2`, same for `compressorV1` and `compressorMid`, not just assert for non-null?

jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/BlobCompressorSelector.kt
```diff
@@ -0,0 +1,30 @@
...
+
+  private val blobCompressors = blobCompressorVersionSwitchTimestampMap.mapValues { (blobCompressorVersion, _) ->
+    GoBackedBlobCompressor.getInstance(blobCompressorVersion, dataLimit)
+  }
+
+  override fun getBlobCompressor(selector: Instant): BlobCompressor {
+    val blobCompressorVersion = blobCompressorVersionSwitchTimestampMap.entries
+      .filter { selector >= it.value }
+      .maxByOrNull { it.value }
+      ?.key ?: throw IllegalStateException("No BlobCompressor version found for timestamp $selector")
```
jonesho, line 34: What happens for multiple `blobCompressor` with the same timestamps? Rejected during instantiation or any rule to pick one of them? Always pick the highest version?

jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/BlobCompressorSelector.kt
```diff
@@ -0,0 +1,30 @@
+package linea.blob
+
+import kotlin.time.Instant
+
+interface BlobCompressorSelector<T> {
+  /**
+   * Returns a BlobCompressor instance for the given timestamp.
```
jonesho: nit: `for the given selector`, `timestamp` is too specific for this generic interface

jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/BlobCompressorSelector.kt
gauravahuja: Technically this should not be configured such that we have multiple blobcompressor versions with the same timestamps. I will add a check for that.

jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/BlobCompressorSelector.kt
```diff
@@ -0,0 +1,30 @@
...
+}
+
+class BlobCompressorSelectorByTimestamp(
+  private val blobCompressorVersionSwitchTimestampMap: Map<BlobCompressorVersion, Instant>,
+  private val dataLimit: Int,
+) : BlobCompressorSelector<Instant> {
+
+  private val blobCompressors = blobCompressorVersionSwitchTimestampMap.mapValues { (blobCompressorVersion, _) ->
+    GoBackedBlobCompressor.getInstance(blobCompressorVersion, dataLimit)
+  }
```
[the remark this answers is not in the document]
fluentcrafter: IIRC, the underlying compressor is not designed to have multiple versions running that the same time.
@tabaie can you confirm?

If so, Compressor instantiation must be lazzy loaded evaluated when calling `getBlobCompressor` and keep track current instantiated one and return it.

jvm-libs/linea/blob-compressor/src/test/kotlin/linea/blob/BlobCompressorSelectorTest.kt
```diff
@@ -0,0 +1,61 @@
...
+    assertEquals(compressorV1, compressorMid)
+    assert(compressorV2 != compressorV1)
```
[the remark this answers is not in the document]
fluentcrafter: @gauravahuja please adress this and keep consistency in mind

jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/BlobCompressorSelector.kt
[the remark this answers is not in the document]
Filter94: If it does, we need a test proving that

jvm-libs/linea/blob-compressor/src/test/kotlin/linea/blob/BlobCompressorSelectorTest.kt
```diff
@@ -0,0 +1,93 @@
...
+    compressorV2.appendBlock(sampleBlock)
+    val compressedV2 = compressorV2.getCompressedDataAndReset()
+    assertNotNull(compressedV2)
+
+    compressorV3.startNewBatch()
+    compressorV3.appendBlock(sampleBlock)
+    val compressedV3 = compressorV2.getCompressedDataAndReset()
+    assertNotNull(compressedV3)
+
+    assertThat(compressedV2).isNotEqualTo(compressedV3) // Different versions should produce different compressed data
```
jonesho, line 97: I think we also want to test whether multiple compressor instances of different version could simultaneously doing block compression, e.g. two threads using different version of compressor to compress a block at the same time

fluentcrafter: Yes, we do need that test.

jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/BlobCompressorSelector.kt
gauravahuja: Added a check that all timestamps must be unique

jvm-libs/linea/blob-compressor/src/test/kotlin/linea/blob/BlobCompressorSelectorTest.kt
gauravahuja: Can you see if the test below `compresses multiple blocks concurrently with v2 and v3 without exception` is what you expected

fluentcrafter: I am afraid test is weak and can yield a false positive:

- `CompressorTestData.blocksRlpEncoded.take(100)` - `blocksRlpEncoded` has only 19 small blocks, not 100 so compressions will be fast and not even run in parallel?
- Before in the test, we should do compression with V2 and V3, save the data, make the parallel compression and at the end compar the prallel compressions yeld the same compression to ensure no side effects...

fluentcrafter: Like I suspected, this test is incomplete and yields false positives. I improved the test [here](https://github.com/Consensys/linea-monorepo/pull/2604), please have a look

fluentcrafter: I see I made a mistake on the test. Thanks for the fix...

Issue #2537, opened 2026-03-04
Manual CI workflow for Arithmetization with Partial Prover


jonesho:
🔐 🔐 `Should this be a security advisory instead?` 🔐 🔐

## Description
We should have a GHA manual workflow to run the local stack with a partial prover for certain e2e tests, essentially to verify the prover version is compatible with that of tracer and Shomei by checking finalizations on L1.

## Motivation
The goal is to detect incompatibility between prover and tracer/shomei before we deployed to any environment (especially sepolia and mainnet), givent that the dev mode provers aren't able to detect the incompatibility on devnet and our e2e local stack

## Tasks
- [ ] Medium-level task that needs to be implemented for this feature request to be considered complete.

## Acceptance criteria
- [ ] List tests that need to be implemented or checks to consider this task completed. 

## Risks
- [ ] These are risks that may prevent completion of the task as well as areas to pay special attention to

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #2553, opened 2026-03-05
2537 manual ci workflow e2e with partial prover

jonesho:
This PR implements issue(s) #2537

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> Mostly CI/dev-infra changes, but it alters `docker compose` service env var names/commands and introduces new image-tagging/build paths that could break local/CI environment startup if misconfigured.
> 
> **Overview**
> Adds a new manually-triggered GitHub Actions workflow to build (or reuse provided tags for) `linea-besu-package` and `linea-prover`, start a tracing-v2 stack with a *partial prover* override, run a targeted e2e suite, and upload container/shared-dir logs on failure.
> 
> Introduces `start-env-with-tracing-v2-partialprover` and a new `compose-tracing-v2-partialprover-override.yml` plus `prover-config-partial.toml` to run the prover in `partial` mode with higher memory limits and coordinator conflation overrides (including optional Shomei version injection via `CONFIG_OVERRIDE_OPTS`).
> 
> Refactors `linea-besu-package/Makefile` to use `BESU_PACKAGE_TAG`/`PROVER_TAG`, adds `build-prover`/`build-all` and a partial-prover e2e target, updates prover docker builds to tag images (`consensys/linea-prover:${VERSION}`), and tweaks prover trace limits in the default config; also adds an e2e script and conditional test that verifies the latest L2 block reaches `finalized` when running with `PARTIAL_PROVER=true`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 491b38e1ccc23937d29cd2c9b30823d1e55b450a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docker/compose-spec-l2-services.yml
+++ b/docker/compose-spec-l2-services.yml
@@ -335,7 +335,7 @@ services:
       config__override__l1_submission__blob__signer__type: ${LINEA_COORDINATOR_SIGNER_TYPE:-web3j}
       config__override__l1_submission__aggregation__signer__type: ${LINEA_COORDINATOR_SIGNER_TYPE:-web3j}
       config__override__message_anchoring__signer__type: ${LINEA_COORDINATOR_SIGNER_TYPE:-web3j}
-      config__override__l1_submission__data-availability: ${LINEA_COORDINATOR_DATA_AVAILABILITY:-ROLLUP}
+      config__override__l1_submission__data_availability: ${LINEA_COORDINATOR_DATA_AVAILABILITY:-ROLLUP}
       dev_LINEA_COORDINATOR_DISABLED: ${dev_LINEA_COORDINATOR_DISABLED:-false}
       CONFIG_OVERRIDE_OPTS: ''
     command:


docker/config/prover/v3/prover-config.toml
```diff
@@ -52,13 +52,15 @@ modules = [
     {module = "EXT", limit = 524288, limit_large = 1048576},
     {module = "GAS", limit = 65536, limit_large = 131072},
     {module = "HUB", limit = 2097152, limit_large = 4194304},
+    {module = "HUB×4", limit = 8388608, limit_large = 16777216},
```
jonesho, line 55: @gusiri do we still need the trace-limit update for regular prover here?

docker/compose-spec-l2-services.yml
```diff
@@ -331,14 +331,31 @@ services:
       retries: 120
     restart: on-failure
     environment:
-      config__override__type2-state-proof-provider__disabled: ${LINEA_COORDINATOR_DISABLE_TYPE2_STATE_PROOF_PROVIDER:-true}
```
jonesho, line 334: These environment variables are broken for a long while, it's because the we switch to use shell in the `command` section below, and shell would not accept environment variable names contain '-', e.g. `config__override__type2-state-proof-provider__disabled`

The root cause is that POSIX shell silently drops environment variables whose names aren't valid shell identifiers — here hyphens (-) make a name invalid.
when shell starts, it filters its environment to only expose variables with names matching [A-Za-z_][A-Za-z0-9_]*. Variables with hyphens are dropped before any child process is spawned.

Previously, we calls java directly in `command` section without invoking any shell, and hence no issue and the env vars works fine

docker/config/prover/v3/prover-config.toml
gusiri: Yes. we need to use the same trace limits here vs prod.

docker/compose-tracing-v2.yml
```diff
@@ -60,11 +60,6 @@ services:
     extends:
       file: compose-spec-l2-services.yml
       service: prover-v3
-    volumes:
```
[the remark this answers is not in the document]
jonesho: Not relevant,  compose-spec-l2-services.yml defined volumes for prover-v3

PR #2569, opened 2026-03-09
[Besu-Plugin] Use blob compressor selector by timestamp

gauravahuja:
This PR implements issue(s) #2503 

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> Changes core transaction compression/selection plumbing and introduces timestamp-based compressor version switching, which could affect block building and profitability decisions if misconfigured. Dependency resolution changes may also impact build/runtime classpaths.
> 
> **Overview**
> Switches the sequencer from a single `GoBackedBlobCompressor` instance to `BlobCompressorSelectorByTimestamp`, allowing **blob compressor versions to change by block timestamp** and propagating that timestamp through `TransactionCompressor`/`CachingTransactionCompressor` (cache keys now include compressor version).
> 
> Adds a hidden CLI option `--plugin-linea-blob-compressor-version-timestamps` (wired into `LineaTransactionSelectorConfiguration`) plus tests for parsing it, and updates selectors (`CompressionAwareTransactionSelector`, factory/plugin wiring) and acceptance/unit tests accordingly.
> 
> Bumps `blob-compressor` to `3.0.2` and `lineaKotlin` to `0.2.0`, and enforces a single `kotlin-stdlib` version via Gradle resolution strategy (also explicitly depends on `kotlin-stdlib` in sequencer and acceptance tests).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit af88cf15c679c745acec6fe345be66f09f3d59b6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/TransactionCompressor.java
+++ b/besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/TransactionCompressor.java
@@ -8,7 +8,6 @@
  */
 package net.consensys.linea.utils;
 
-import kotlin.time.Clock;
 import kotlin.time.Instant;
 import org.hyperledger.besu.datatypes.Transaction;
 
@@ -29,6 +28,7 @@ public interface TransactionCompressor {
   int getCompressedSize(Transaction transaction, Instant blockTimestamp);
 
   default int getCompressedSize(Transaction transaction) {
-    return getCompressedSize(transaction, Clock.System.INSTANCE.now());
+    return getCompressedSize(
+        transaction, Instant.Companion.fromEpochMilliseconds(System.currentTimeMillis()));
   }
 }


besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java
```diff
@@ -25,25 +26,30 @@
...
+      final BlobCompressorSelectorByTimestamp blobCompressorSelectorByTimestamp) {
+    this(DEFAULT_CACHE_SIZE, blobCompressorSelectorByTimestamp);
   }
 
   private int calculateCompressedSize(final Transaction transaction) {
     final byte[] encoded = TxEncodingUtils.encodeForCompressor(transaction);
-    return blobCompressor.compressedSize(encoded);
+    return blobCompressorSelectorByTimestamp
+        .getBlobCompressor(Clock.System.INSTANCE.now())
+        .compressedSize(encoded);
```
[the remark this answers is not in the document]
Filter94: I think it has to depend on the situation. For P2P we have no better way other than taking the current wall clock time. For Block building logic IMO, we have to use pending header's timestamp

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java
```diff
@@ -25,25 +26,30 @@
 @Slf4j
 public class CachingTransactionCompressor implements TransactionCompressor {
```
Filter94, line 29: It looks like once we're past a certain timestamp:
1. The caller of `CachingTransactionCompressor` may get a cached compression size, compressed by an old compressor, because the key is only tx's hash
2. Cache is not invalidated after a switch timestamp is reached. However we don't need old cached entries for an old compressor version, so they will linger in memory for the expiry timeout

gauravahuja: CachingTransactionCompressor is not aware of the switch timestamps, neither does the BlobCompressorSelectorByTimestamp exposes this detail.So I don't know how cace invalidation could be accomplished. However there is TTL for the cache entry so it should automatically be invalidated after that.

I see two options:
1. We might need to change the interface of BlobCompressorSelectorByTimestamp to be able to tell the caller that it has switched or
2. The caller can always compare the previous BlobCompressor object with the one returned by BlobCompressorSelectorByTimestamp each time and if the object is different, invalidate the cache.

Which option do you prefer?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/selectors/CompressionAwareTransactionSelector.java
```diff
@@ -204,6 +206,10 @@ public TransactionSelectionResult evaluateTransactionPreProcessing(
     pendingSlowExecutionPathFuture =
         COMPRESSION_EXECUTOR.submit(
             () -> {
+              Instant timestamp =
+                  Instant.Companion.fromEpochSeconds(pendingHeader.getTimestamp(), 0L);
+              BlobCompressor blobCompressor =
+                  blobCompressorSelectorByTimestamp.getBlobCompressor(timestamp);
```
Filter94: I think that's wrong. Since blobCompressor can be injected into `CompressionAwareTransactionSelector`'s constructor, why would we need to get it by timestamp here?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/selectors/LineaTransactionSelector.java
Filter94, line 1: Since this is the first class in the call stack which has access to evaluationContext, I think it makes the most sense to get a specific transaction compressor here. It would make even more sense in `LineaTransactionSelectorFactory`, but it seems that you can't access pending block's header from there

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java
Filter94: I think we need a deeper restructuring.

* We can introduce a cache with a new key: `(Blob compressor version, tx hash)`. Since the cache key is different, this cache can't conform to `TransactionCompressor`, it needs to accept `Blob compressor version` along with every transaction. This shared cache will be created in `AbstractLineaSharedPrivateOptionsPlugin` and will be referenced by the other instances as backend
* This new cache can be used by the new implementation of `TransactionCompressor` which will accept the blob version on instantiation. This new implementation of `TransactionCompressor` will abstract users from the switch logic and will pass a constant blob compressor version, which was set in the constructor to the shared cache
* Then we'll need custom implementations of `BlobCompressorSelectorByTimestamp` that would be aware of Blob compressor versions, the switch logic and would instantiate the `TransactionCompressor` implementation I described above

Instantiation 
* For tx selection it can be done in `LineaTransactionSelector`, because it's the first class that is aware of `TransactionEvaluationContext` which has the pending block's header, which has the timestamp
* For txpool validation, instantiation will be in `LineaTransactionPoolValidatorFactory::createTransactionValidator`, by the current timestamp. It's called for every transaction before validation

`BlobCompressorSelectorByTimestamp` can be used in `LineaTransactionSelector` to instantiate `BlobCompressor` and pass it to `CompressionAwareTransactionSelector`'s constructor

besu-plugins/linea-sequencer/acceptance-tests/build.gradle
```diff
@@ -91,6 +91,7 @@ dependencies {
   testImplementation 'org.junit.jupiter:junit-jupiter-params'
   testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
   testImplementation "org.wiremock:wiremock:${libs.versions.wiremock.get()}"
+  implementation "org.jetbrains.kotlin:kotlin-stdlib:${libs.versions.kotlin.get()}"
```
fluentcrafter: please follow common best practice: declare `implementation` before testImplementation

besu-plugins/linea-sequencer/acceptance-tests/src/test/kotlin/linea/plugin/acc/test/rpc/linea/EstimateGasTest.kt
```diff
@@ -79,7 +81,12 @@ open class EstimateGasTest : LineaPluginPoSTestBase() {
         .build()
     profitabilityCalculator = TransactionProfitabilityCalculator(
       profitabilityConf,
-      CachingTransactionCompressor(GoBackedBlobCompressor.getInstance(BlobCompressorVersion.V2, 128 * 1024)),
+      CachingTransactionCompressor(
+        BlobCompressorSelectorByTimestamp(
+          mapOf(BlobCompressorVersion.V2 to Instant.MIN.toKotlinInstant()),
```
fluentcrafter: nit: Use kotlin.time.Instant, no need to use Java one
```suggestion
          mapOf(BlobCompressorVersion.V2 to Instant.DISTANT_PAST),
```

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java
gauravahuja: > BlobCompressorSelectorByTimestamp can be used in LineaTransactionSelector to instantiate BlobCompressor and pass it to CompressionAwareTransactionSelector's constructor

It is not clear to me how `CompressionAwareTransactionSelector` be reinstantiated with new version of `BlobCompressor` when the new compressor version timestamp is reached.

Filter94: You're right. We can't do it with the API we have, but in practice the timestamp we need is 1 frame away from the `PluginTransactionSelectorFactory.create(...)` call. So the way you create `BlobCompressor` inside CompressionAwareTransactionSelector is already optimal
 
@fab-10 Do you think it makes sense to pass some pending block metadata to `PluginTransactionSelectorFactory`? It would allow making the switch logic like that cleaner

fab-10: I need a bit of context, is the timestamp related to a hard fork?

Filter94: Could be a hard fork, could be a soft one. In this case it's a ground to switch a blob compressor. The idea is similar to the hard forks. Starting from a certain block timestamp we change the blob compressor version

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/config/LineaTransactionSelectorCliOptions.java
```diff
@@ -233,4 +248,29 @@ public Map<Address, Set<TransactionEventFilter>> parseTransactionEventDenyList(
...
+    return result;
+  }
+
+  public Map<BlobCompressorVersion, Instant> getBlobCompressorVersionTimestamps() {
+    if (blobCompressorVersionTimestampsRaw != null
+        && !blobCompressorVersionTimestampsRaw.isEmpty()) {
+      return parseBlobCompressorVersionTimestamps(blobCompressorVersionTimestampsRaw);
+    }
+    return Map.of(
+        BlobCompressorVersion.V2, InstantConversionsJDK8Kt.toKotlinInstant(java.time.Instant.MIN));
```
fluentcrafter: Instead of this if/else, I recommend to add a default value for the BLOB_COMPRESSOR_VERSION_TIMESTAMPS. Beside being simpler, it will show on the helper CLI the default value.

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java
fab-10: I think that's feasible, since selectors are created for each new block, they can take advantage of the pending block header. 
Let me propose a quick PR

Filter94: In any case we won't be able to use your change in this PR, because we first need to finish the work on upgrading the Besu version

fab-10: https://github.com/besu-eth/besu/pull/10034

fab-10: The new API is now on Besu main

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/TransactionCompressor.java
```diff
@@ -21,7 +23,12 @@ public interface TransactionCompressor {
...
    * @param transaction the transaction for which to get the compressed size
+   * @param blockTimestamp the timestamp of the block which is expected to include this transaction
    * @return the compressed size of the transaction
    */
-  int getCompressedSize(Transaction transaction);
+  int getCompressedSize(Transaction transaction, Instant blockTimestamp);
+
+  default int getCompressedSize(Transaction transaction) {
+    return getCompressedSize(transaction, Clock.System.INSTANCE.now());
+  }
```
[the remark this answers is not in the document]
gauravahuja: TransactionProfitabilityCalculator does not know the block timestamp so `Clock.System.INSTANCE.now()` is the best that can be done right now.

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/config/LineaTransactionSelectorCliOptions.java
```diff
@@ -132,7 +138,14 @@ public class LineaTransactionSelectorCliOptions implements LineaCliOptions {
...
 
-  private LineaTransactionSelectorCliOptions() {}
+  @CommandLine.Option(
+      names = {BLOB_COMPRESSOR_VERSION_TIMESTAMPS},
+      hidden = true,
+      paramLabel = "<MAP>",
+      description =
+          "Comma-separated map of BlobCompressorVersion to Instant, e.g. V1_2=2025-01-01T00:00:00Z,V2=2026-01-01T00:00:00Z")
+  private String blobCompressorVersionTimestampsRaw =
+      String.format("%s=%s", BlobCompressorVersion.V2.name(), java.time.Instant.MIN);
```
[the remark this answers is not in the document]
gauravahuja: I do not know what issue this comment is referring to.

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/config/LineaTransactionSelectorCliOptions.java
```diff
@@ -132,7 +138,14 @@ public class LineaTransactionSelectorCliOptions implements LineaCliOptions {
       description = "Path to the file containing the events deny list for bundles")
   private String eventsBundleDenyListPath;
 
-  private LineaTransactionSelectorCliOptions() {}
+  @CommandLine.Option(
+      names = {BLOB_COMPRESSOR_VERSION_TIMESTAMPS},
+      hidden = true,
+      paramLabel = "<MAP>",
+      description =
+          "Comma-separated map of BlobCompressorVersion to Instant, e.g. V1_2=2025-01-01T00:00:00Z,V2=2026-01-01T00:00:00Z")
```
Filter94: It would be nice to include all supported versions in the description so user knows what is available and how to write versions exactly

besu-plugins/linea-sequencer/sequencer/build.gradle
```diff
@@ -20,10 +20,19 @@ apply from: lineaSequencerProject.file("gradle/java.gradle")
 apply from: lineaSequencerProject.file("gradle/dependency-management.gradle")
 apply from: lineaSequencerProject.file("gradle/lint.gradle")
 
+configurations.all {
```
Filter94, line 29: Can you please add a clarifying comment on why we need this and what's the resolution strategy we do here?

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/config/LineaTransactionSelectorCliOptions.java
```diff
@@ -52,6 +58,10 @@ public class LineaTransactionSelectorCliOptions implements LineaCliOptions {
   public static final String EVENTS_DENY_LIST_PATH = "--plugin-linea-events-deny-list-path";
   public static final String EVENTS_BUNDLE_DENY_LIST_PATH =
       "--plugin-linea-events-bundle-deny-list-path";
+  public static final String BLOB_COMPRESSOR_VERSION_TIMESTAMPS =
+      "--plugin-linea-blob-compressor-version-timestamps";
+  public static final String BLOB_COMPRESSOR_VERSION_TIMESTAMPS_DEFAULT =
+      String.format("%s=%s", BlobCompressorVersion.V2.name(), Instant.Companion.getDISTANT_PAST());
```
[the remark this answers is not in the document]
Filter94: Does it actually work?

besu-plugins/linea-sequencer/sequencer/build.gradle
```diff
@@ -20,10 +20,25 @@ apply from: lineaSequencerProject.file("gradle/java.gradle")
...
+    if (details.requested.group == 'org.jetbrains.kotlin' && details.requested.name.startsWith('kotlin-stdlib')) {
+      details.useVersion "${libs.versions.kotlin.get()}"
+    }
+  }
+}
+
 dependencies {
   implementation project(":tracer:arithmetization")
   implementation "build.linea:blob-compressor:${libs.versions.blobCompressor.get()}"
   implementation "build.linea.internal:kotlin:${libs.versions.lineaKotlin.get()}"
```
Filter94, line 40: I believe this can be just a project dependency now. Same might apply to compressor, but I'm not 100% sure

besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/config/LineaTransactionSelectorCliOptions.java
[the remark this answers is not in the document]
gauravahuja: It works. Bot is now making up scenarios. "However, if DISTANT_PAST.toString() ever contains a comma or equals sign, or if the format differs across Kotlin stdlib versions, parsing will break" Timestamp string does not contain comma or equals sign.

PR #2598, opened 2026-03-16
[ZkTracer] feat: split AddressCollisionDelegationWarmingAndDeploymentTests per d…

amkCha:
…elegation type

This PR implements issue(s) #

### Checklist

  PR.

---

> [!NOTE]
> **Low Risk**
> Test-only refactor that changes how parameterized cases are generated/executed; primary risk is increased runtime or altered coverage if the new method split misconfigures parameters.
> 
> **Overview**
> Refactors `AddressCollisionDelegationWarmingAndDeploymentTests` so delegation type is no longer part of the `@MethodSource` parameter set, and instead the suite is split into multiple `@ParameterizedTest` methods that each run the same input matrix with a fixed `AccountDelegationType`.
> 
> This keeps the underlying `runWithParameters(...)` logic the same while changing test enumeration/structure (and likely increasing the number of executed test cases via separate test methods per delegation type).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8fb92b7b8c0660b8f66f743f45dbcaae40507223. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


tracer/arithmetization/src/test/java/net/consensys/linea/zktracer/module/hub/AddressCollisionDelegationWarmingAndDeploymentTests.java
```diff
@@ -89,43 +89,35 @@ public class AddressCollisionWarmingAndDeploymentTests extends TracerTestBase {
...
+                        skip == 1, collision, isDeployment == 1, warming1, warming2, warming3));
               }
             }
           }
         }
       }
     }
+
     return arguments.stream();
   }
```
lorenzogentile404, line 122: I believe the input function is invoked for each of the tests below (you can check with debugging), while we could compute it only once, store it in the class and feed it to each test.

lorenzogentile404 (approved):

letypequividelespoubelles (changes requested):
You just splitted the test in 5 different tests to not have a single test timeout right ? I'm wondering if it's not better to just do a sampling of the test ...

Issue #2393, opened 2026-02-16
Migrate postman from ethers.js to viem


VGau:
## Description
Migrate the postman package from ethers.js v6 to viem for all blockchain interactions (providers, contracts, encoding, error handling, scripts).
The core interfaces are already library-agnostic (generic type parameters, no ethers imports). The work is in the implementation layer and dependencies.

## Motivation
The postman is the last major ethers.js consumer in the monorepo. Migrating unifies the codebase on a single blockchain library, removes dual maintenance, and enables sharing utilities across packages.

## Tasks
- [ ] Provide viem-based implementations of all blockchain client interfaces (contract clients, log clients, providers, gas providers)
- [ ] Migrate all processors, pollers, validation services, and persistence layer to viem types
- [ ] Rewrite error parser for viem error types (preserve retry/mitigation semantics)
- [ ] Migrate transaction size calculator to viem equivalents
- [ ] Migrate CLI scripts
- [ ] Update all test mocks and fixtures
- [ ] Remove ethers from dependencies, update build config

## Acceptance criteria
- [ ] Zero ethers imports in the postman package
- [ ] All unit tests pass with updated implementations
- [ ] Error retry/mitigation behavior preserved
- [ ] Docker image builds; e2e tests pass without regression
- [ ] Smoke test: postman claims messages in both directions in local environment

PR #2617, opened 2026-03-18
chore(postman): refactor using Viem

VGau:
This PR implements issue(s) #2393 

### Checklist

  PR.

---

> [!NOTE]
> **High Risk**
> High risk because it replaces core blockchain interaction/wiring (signing, nonce/claiming lifecycle, pollers) and changes runtime entrypoints and dependencies, which can break cross-chain message processing in production.
> 
> **Overview**
> Refactors `postman` to a new application/container architecture (`PostmanApp`, `PostmanContainer`, `L1ToL2App`, `L2ToL1App`) and switches blockchain interactions from `ethers`/`@consensys/linea-sdk` to **Viem** clients, including updated claiming/anchoring/persisting poller wiring and transaction lifecycle handling.
> 
> Updates packaging and ops to match the new runtime: Docker entrypoint now runs `dist/src/main.js`, local run docs point to `src/main.ts`, CI builds `@consensys/linea-postman` and runs `sdk-viem` tests, and the compose dependency for `postman` changes to `l2-node-besu`.
> 
> Expands and documents configuration: adds per-chain `*_SIGNER_TYPE` with Web3Signer + optional mTLS settings, receipt polling intervals, gas bumping/retry cycle limits (`MAX_BUMPS_PER_CYCLE`, `MAX_RETRY_CYCLES`), optional claim-via address, Postgres SSL flags, and `LOG_LEVEL`; adds/updates Jest tests to cover the new config/env loader and app wiring, and updates scripts to use Viem for sending/anchoring test messages.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 794b0c003758adeecacef84fb051e19be47aecd4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


postman/src/infrastructure/blockchain/viem/ViemTransactionRetrier.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    this.logger.debug("Bumped fees computed.", {
+      transactionHash,
+      attempt,
+      maxFeePerGas: maxFeePerGas.toString(),
+      maxPriorityFeePerGas: maxPriorityFeePerGas.toString(),
+    });
+
+    const txHash = await this.walletClient.sendTransaction({
+      account: transaction.from,
```
alainncls: Passing a bare address as `account` makes viem take the JSON-RPC account path (`eth_sendTransaction`) instead of reusing the hoisted local/Web3Signer account created in `createChainContext`. On nodes that do not manage the signer account locally, fee bumps and cancellations will fail even though the initial claim path can sign locally. Please reuse the wallet clients hoisted account object here, and in `cancelTransaction`, instead of passing an address string.

postman/src/services/processors/MessageClaimingProcessor.ts
```diff
@@ -121,75 +81,52 @@ export class MessageClaimingProcessor implements IMessageClaimingProcessor {
...
-    const claimTxFn = async () =>
-      await this.messageServiceContract.claim(
+    const previousStatus = message.status;
+
+    // Step 1: Reserve nonce in DB (pure persistence)
+    message.edit({
+      claimTxNonce: nonce,
+      status: MessageStatus.PENDING,
+      ...(message.status === MessageStatus.FEE_UNDERPRICED
+        ? { claimNumberOfRetry: message.claimNumberOfRetry + 1, claimLastRetriedAt: new Date() }
```
alainncls: This increments `claimNumberOfRetry` and updates `claimLastRetriedAt` before `claim()` has actually returned a transaction. If the RPC/Web3Signer call fails before broadcast, the catch below only restores `status`, so an already `FEE_UNDERPRICED` message loses retry budget and enters backoff without any on-chain attempt. Please move the retry counter update to the post-send persistence step, or restore these fields in the failure path.

postman/src/services/processors/MessageClaimingProcessor.ts
```diff
@@ -200,61 +135,55 @@ export class MessageClaimingProcessor implements IMessageClaimingProcessor {
...
+    message.edit({
+      claimTxCreationDate,
+      claimTxGasLimit: Number(tx.gasLimit),
+      claimTxMaxFeePerGas: tx.maxFeePerGas ?? undefined,
+      claimTxMaxPriorityFeePerGas: tx.maxPriorityFeePerGas ?? undefined,
+      claimTxHash: tx.hash,
+      ...(isRetry ? { claimNumberOfRetry: message.claimNumberOfRetry + 1, claimLastRetriedAt: new Date() } : {}),
+    });
+    await this.messageRepository.updateMessage(message);
   }
```
kyzooghost, line 142: From what I understand, this used to be wrapped in a single DB transaction.

There can be a race condition with `MessageClaimingPersister.process()` here if step 2 is stuck

See below AI report

---

### 🟠 HIGH: In-flight claims can be reset back to `SENT` before the transaction hash exists

**File**: `postman/src/services/processors/MessageClaimingProcessor.ts:115`  
**File**: `postman/src/services/processors/MessageClaimingPersister.ts:33`  
**Commit**: `f129903a0`  
**Blast Radius**: 2 production pipelines (`L1ToL2App` and `L2ToL1App`)  
**Test Coverage**: PARTIAL

**Description**:  
The new claim path writes `status = PENDING` and `claimTxNonce` to the database before the RPC/Web3Signer submission returns a transaction hash, then the persister treats any `PENDING` row without `claimTxHash` as broken state and immediately resets it to `SENT`.

**Historical Context**:
- Baseline code used `updateMessageWithClaimTxAtomic(...)` and only exposed the `PENDING` row after the claim call returned and `claimTxHash` was stored in the same repository transaction: `postman/src/application/postman/persistence/repositories/TypeOrmMessageRepository.ts` at baseline lines 275-305.
- This branch replaces that with an explicit three-step flow where the pre-hash `PENDING` state is externally visible: `MessageClaimingProcessor.ts:115-160`.
- The new reset behavior is now codified by test coverage instead of guarded against: `postman/src/services/processors/__tests__/MessageClaimingPersister.test.ts:345-360`.

**Attack / Failure Scenario**:
1. `MessageClaimingProcessor` updates a message to `PENDING` with a reserved nonce at `MessageClaimingProcessor.ts:115-123`.
2. The subsequent `messageServiceContract.claim(...)` call blocks on a slow RPC, remote signer, or transient network issue at `MessageClaimingProcessor.ts:128-139`.
3. Before that call returns, `MessageClaimingPersister` polls the same row, sees `claimTxHash` is still empty, and resets the message to `SENT` at `MessageClaimingPersister.ts:33-39`.
4. The claim pipeline is now free to pick the same bridge message again while the original submission is still in flight, creating duplicate claim attempts or state oscillation between retry loops and on-chain success.

**Exploitability:** MEDIUM  
**Justification:** No privileged access is required, but the window depends on normal RPC latency or signer slowness rather than a single direct call.

**Concrete Impact**:
- Duplicate claim submissions against the same bridge message
- False retries and unnecessary gas spend
- Higher chance of inconsistent retry bookkeeping when one attempt lands after the row has already been reset

**Proof of Concept**:
```text
Claiming processor:
1. update message -> PENDING, claimTxNonce = N
2. await slow claim RPC

Persister in parallel:
1. getFirstPendingMessage() returns same row
2. claimTxHash is empty
3. reset row to SENT

Result:
The same message becomes eligible again while the original claim submission is still unresolved.
```

**Recommendation**:
- Do not expose `PENDING` to the persister until `claimTxHash` has been persisted.
- If the split write is required, add an explicit intermediate status such as `SUBMITTING` and make the persister ignore it.
- Add a concurrency test where `claim()` is delayed longer than `receiptPollingInterval` and assert the message is not reset mid-submission.

postman/src/services/processors/TransactionLifecycleManager.ts
```diff
@@ -0,0 +1,129 @@
...
+      if (message.claimTxNonce !== undefined) {
+        await this.transactionRetrier.cancelTransaction(message.claimTxNonce);
+      }
+
+      message.edit({
+        status: MessageStatus.SENT,
+        claimNumberOfRetry: 0,
+        claimCycleCount: message.claimCycleCount + 1,
+        claimTxHash: undefined,
+        claimTxNonce: undefined,
```
kyzooghost: IIUC `message.edit` will ignore undefined fields - so the stale txHash and txNonce will stay in the DB

See AI finding below:

---

### 🟠 HIGH: Retry-cycle reset does not actually clear stale claim identifiers

**File**: `postman/src/services/processors/TransactionLifecycleManager.ts:105`  
**File**: `postman/src/core/entities/Message.ts:119`  
**Commit**: `f129903a0`  
**Blast Radius**: Shared `Message.edit()` helper has 20 call sites; immediate impact is both retry-cycle paths  
**Test Coverage**: PARTIAL

**Description**:  
`cancelAndResetMessage()` tries to clear `claimTxHash` and `claimTxNonce` by passing `undefined`, but `Message.edit()` only writes fields when the new value is not `undefined`. The message therefore keeps the old claim identifiers even after being "reset".

**Historical Context**:
- `TransactionLifecycleManager.cancelAndResetMessage()` explicitly intends to clear the fields at `TransactionLifecycleManager.ts:105-112`.
- `Message.edit()` silently ignores `undefined` for optional fields at `Message.ts:119-137`.
- The branch test suite already documents this helper behavior for fee fields, which confirms the semantics are intentional and known: `postman/src/services/processors/__tests__/MessageClaimingProcessor.test.ts:469-474`.
- The reset-path test does not assert that the stale transaction identifiers are actually removed: `postman/src/services/processors/__tests__/MessageClaimingPersister.test.ts:285-315`.

**Attack / Failure Scenario**:
1. A pending claim exhausts `maxBumpsPerCycle`.
2. `cancelAndResetMessage()` marks the message back to `SENT` and intends to clear `claimTxHash` / `claimTxNonce`.
3. Because `edit()` skips `undefined`, the old transaction hash and nonce remain on the message object and in the database.
4. On the next cycle, downstream logic can observe stale transaction metadata for a message that is no longer supposed to be tied to that old claim attempt.

**Exploitability:** MEDIUM  
**Justification:** This requires the retry-cycle path, but once reached it is deterministic.

**Concrete Impact**:
- Stale transaction metadata survives across retry cycles
- The service can poll, log, or reason about the wrong claim transaction after a reset
- Follow-up retries operate on a logically "fresh" message that still carries old claim state

**Proof of Concept**:
```text
message.edit({
  status: SENT,
  claimTxHash: undefined,
  claimTxNonce: undefined,
})

// Message.edit only assigns when value !== undefined
// Old claimTxHash and claimTxNonce remain unchanged
```

**Recommendation**:
- Change `Message.edit()` to support explicit clearing for optional fields, or use `null` as the persisted clear value.
- Add a regression test that asserts `claimTxHash` and `claimTxNonce` are absent after `cancelAndResetMessage()`.
- Audit other `edit({ ...: undefined })` usages before merge.

postman/src/application/postman/app/config/schema.ts
```diff
@@ -0,0 +1,115 @@
...
+const hexString = z
+  .string()
+  .regex(/^0x[0-9a-fA-F]+$/, "Must be a hex string starting with 0x") as z.ZodType<`0x${string}`>;
+
+const privateKeySchema = z.custom<`0x${string}`>((val) => {
+  return isHex(val) && size(val) === 32;
+}, "Invalid private key");
+
+const ethAddress = z.string().regex(/^0x[0-9a-fA-F]{40}$/, "Must be a valid Ethereum address") as z.ZodType<Address>;
+
```
kyzooghost, line 15: this looks like we could re-use in the shared-utils project

postman/src/services/processors/TransactionLifecycleManager.ts
```diff
@@ -0,0 +1,129 @@
...
+      }
+
+      this.logger.warn("Max fee bumps exhausted, cancelling stuck transaction and resetting message.", {
+        messageHash: message.messageHash,
+        claimTxNonce: message.claimTxNonce,
+        claimCycleCount: message.claimCycleCount,
+      });
+
+      if (message.claimTxNonce !== undefined) {
+        await this.transactionRetrier.cancelTransaction(message.claimTxNonce);
```
kyzooghost: IIUC this does not wait for the cancel tx to be mined, but instead waits for the tx to be accepted to the node txpool.

So there can be a race condition (if tx is stuck in txpool) that the reset message has a new claim tx competing with the cancel tx.

See AI finding

---

### 🟠 HIGH: Exhausted fee bumps re-queue the message before the cancel transaction is confirmed

**File**: `postman/src/services/processors/TransactionLifecycleManager.ts:95-115`  
**File**: `postman/src/services/processors/MessageAnchoringProcessor.ts:38-83`  
**Commit**: `9d6116c487`  
**Blast Radius**: 2 production pipelines (`L1ToL2App` and `L2ToL1App`) through the shared retry-cycle path  
**Test Coverage**: PARTIAL

**Description**:  
`cancelAndResetMessage()` resets the message back to `SENT` immediately after broadcasting the cancel transaction. It does not wait for the cancellation to be mined. `MessageAnchoringProcessor.process()` actively scans `SENT` messages and promotes any still-`CLAIMABLE` message back to `ANCHORED`, so the same bridge message can be selected for a fresh claim while both the original claim and the cancel transaction are still pending.

**Historical Context**:
- `cancelAndResetMessage()` calls `cancelTransaction(...)` and immediately resets the row to `SENT` at `TransactionLifecycleManager.ts:101-113`.
- `MessageAnchoringProcessor.process()` fetches `SENT` messages and promotes `CLAIMABLE` ones to `ANCHORED` at `MessageAnchoringProcessor.ts:38-74`.
- Existing coverage codifies the immediate reset behavior by asserting `status = SENT` right after `cancelTransaction(...)` returns, without any receipt wait or intermediate state: `postman/src/services/processors/__tests__/MessageClaimingPersister.test.ts:281-315`.

**Attack / Failure Scenario**:
1. A pending claim reaches `maxBumpsPerCycle`.
2. `cancelAndResetMessage()` broadcasts a cancel transaction for the stuck nonce.
3. Before that cancel transaction is mined, the row is already reset to `SENT`.
4. `MessageAnchoringProcessor` sees the message is still `CLAIMABLE` on-chain, promotes it back to `ANCHORED`, and the claim processor can submit a fresh claim while the old claim and its cancellation are both still in flight.

**Exploitability:** MEDIUM  
**Justification:** No privileged access is required; normal mempool congestion or delayed mining is enough to trigger the overlap.

**Concrete Impact**:
- Duplicate live transactions for the same bridge message
- Extra nonce consumption and unnecessary gas spend
- Retry-cycle guarantees are defeated because a new cycle can begin before the previous one has actually been cancelled

**Proof of Concept**:
```text
1. pending claim tx T0 reaches max bumps
2. cancelAndResetMessage() sends cancel tx C0 for the same nonce
3. row is immediately reset to SENT
4. anchoring processor sees message still CLAIMABLE and moves it back to ANCHORED
5. claiming processor submits a fresh claim tx T1 before C0 is mined
```

**Recommendation**:
- Do not reset the message to `SENT` until the cancel transaction has been confirmed on-chain.
- If waiting inline is too expensive, introduce an intermediate status such as `CANCELLING` and make `MessageAnchoringProcessor` and claim selectors ignore it.
- Add a regression test where `cancelTransaction()` succeeds but the cancel receipt is still pending, and assert the message cannot be re-anchored or re-claimed during that window.

postman/src/infrastructure/blockchain/viem/gas/ViemLineaGasProvider.ts
kyzooghost, line 1: is this a reusable utility?

postman/src/infrastructure/blockchain/viem/providers/ViemProvider.ts
kyzooghost, line 1: is this a reusable utility?

postman/README.md
kyzooghost, line 1: Should we update .env.sample with the new env variables?

postman/src/core/types/primitives.ts
```diff
@@ -0,0 +1,3 @@
+export type Address = `0x${string}`;
+export type Hash = `0x${string}`;
+export type Hex = `0x${string}`;
```
thedarkjester, line 3: hex.Hex feels weird, I might call this DynamicBytes or similar - same with Hash - it is assumed to be 32 bytes vs. explicit

postman/src/infrastructure/blockchain/viem/ViemTransactionRetrier.ts
```diff
@@ -0,0 +1,111 @@
...
+    if (maxFeePerGas > this.maxFeePerGasCap) maxFeePerGas = this.maxFeePerGasCap;
+    if (maxPriorityFeePerGas > this.maxFeePerGasCap) maxPriorityFeePerGas = this.maxFeePerGasCap;
+
+    return { maxFeePerGas, maxPriorityFeePerGas };
+  }
+
+  private async getAggressiveFees(): Promise<{ maxFeePerGas: bigint; maxPriorityFeePerGas: bigint }> {
+    const fees = await this.publicClient.estimateFeesPerGas();
+    let maxFeePerGas = fees.maxFeePerGas * 2n;
+    let maxPriorityFeePerGas = fees.maxPriorityFeePerGas * 2n;
```
kyzooghost, line 127: With a flat multiplier of 2 - is it possible for the exponential scaling of `computeBumpedFees` to exceed this, then the cancel tx (which uses `getAggressiveFees`) is stuck?

postman/src/infrastructure/blockchain/viem/gas/ViemEthereumGasProvider.ts
```diff
@@ -0,0 +1,93 @@
...
+      maxFeePerGas: this.gasFeesCache.maxFeePerGas.toString(),
+      maxPriorityFeePerGas: this.gasFeesCache.maxPriorityFeePerGas.toString(),
+      blockNumber: currentBlockNumber.toString(),
+    });
+    return this.gasFeesCache;
+  }
+
+  private async fetchFeeHistory() {
+    return this.client.getFeeHistory({
+      blockCount: 4,
```
thedarkjester: this is a magic number - might be worth making a constant with a name like `feeHistoryBlockCount` - I am not sure consumers would want to configure this along with `gasEstimationPercentile`

postman/src/services/processors/ReceiptStatusResolver.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    return { processingTimeInSeconds, infuraConfirmationTimeInSeconds };
+  }
+
+  private async handleRevertedReceipt(
+    message: Message,
+    receipt: TransactionReceipt,
+    timingMetrics: Record<string, number>,
+  ): Promise<void> {
+    const isRateLimitExceeded = await this.messageServiceContract.isRateLimitExceededError(receipt.hash);
```
kyzooghost: IIUC this simulates the tx at the current block 

```ViemLineaRollupClient.ts
public async parseTransactionError(...
 ...
 const tx = await this.publicClient.getTransaction({ hash: transactionHash });
```

Is there a race condition where:
- In the current block `isRateLimitExceeded = false`
- However in the original block, `isRateLimitExceeded = true`
- Occurring due to rate limit change over time

Then this message falls through to `CLAIMED_REVERTED` status -> the Postman abandons this message

postman/src/core/metrics/MessageMetricsAttributes.ts
kyzooghost, line 1: As a future enhancement, is it useful to track failed transactions from the Postman and the gas burned on failed tx attempts?

postman/src/services/processors/TransactionLifecycleManager.ts
kyzooghost, line 1: do we have unit tests for this file?

postman/src/services/processors/TransactionLifecycleManager.ts
```diff
@@ -0,0 +1,129 @@
...
+        claimTxCreationDate: retryCreationDate,
+        claimTxGasLimit: Number(tx.gasLimit),
+        claimTxMaxFeePerGas: tx.maxFeePerGas ?? undefined,
+        claimTxMaxPriorityFeePerGas: tx.maxPriorityFeePerGas ?? undefined,
+        claimTxHash: tx.hash,
+        claimNumberOfRetry: attempt,
+        claimLastRetriedAt: new Date(),
+        claimTxNonce: tx.nonce,
+      });
+      await this.messageRepository.updateMessage(message);
```
kyzooghost: making a note that there is a dual write dilemma here - tx can succeed but DB update subsequently fails

AI analysis of the consequences:
```
If tx succeeds but DB update fails
-> DB holds stale claimTxHash (old) and stale claimNumberOfRetry (N-1)
-> On the next polling cycle, getTransactionReceipt(oldHash) returns null
   (old tx was replaced/dropped from mempool)
-> Timeout fires -> claimNumberOfRetry (N-1) < maxBumpsPerCycle -> calls retryWithBump
-> retryWithHigherFee(oldHash) calls getTransaction(oldHash) -> null
   -> throws "Transaction with hash X not found"
-> catch block swallows the error, returns null
-> DB claimNumberOfRetry never increments (write never succeeded)
-> The check claimNumberOfRetry >= maxBumpsPerCycle is evaluated against the same stale N-1 every cycle
-> cancelAndResetMessage is never reached; the message spins indefinitely
   burning timeout intervals on each cycle
```

postman/src/services/processors/TransactionLifecycleManager.ts
VGau: added one :)

postman/src/infrastructure/blockchain/viem/ViemTransactionRetrier.ts
VGau: Fixed in 51abb3ecc — `ViemTransactionRetrier` now accepts a `LocalAccount` (typed) and passes it directly to both `sendTransaction` calls, so viem uses local signing throughout.

postman/src/services/processors/MessageClaimingProcessor.ts
VGau: Fixed — `claimNumberOfRetry` and `claimLastRetriedAt` are now updated only in the post-send persistence step, together with `claimTxHash`. If `claim()` throws, only the nonce is rolled back — DB state is untouched.

postman/src/services/processors/MessageClaimingProcessor.ts
VGau: Fixed — `claim()` is now called before any DB write. A single atomic update writes `status=PENDING`, `claimTxNonce`, `claimTxHash`, and all gas fields together. There is no longer a window where a `PENDING` row exists without a `claimTxHash`.

postman/src/services/processors/TransactionLifecycleManager.ts
VGau: Fixed — `claimTxHash` and `claimTxNonce` are now cleared via direct property assignment (`message.claimTxHash = undefined`) before calling `edit()`, bypassing the undefined-skip behaviour.

postman/src/services/processors/TransactionLifecycleManager.ts
VGau: Fixed — `cancelAndResetMessage` now polls for the cancel tx receipt before resetting the message to `SENT`, so the anchoring processor cannot re-queue the message while the cancel is still in-flight.

postman/src/services/processors/TransactionLifecycleManager.ts
VGau: Fixed — if `updateMessage` fails after a successful `retryWithHigherFee`, the error is now caught and logged (with the new tx hash), but execution continues to `receiptPoller.poll(tx.hash)`. The persister receives the receipt and recovers the final state, breaking the infinite spin. Added a regression test for this path.

postman/src/services/processors/ReceiptStatusResolver.ts
VGau: Fixed — `ReceiptStatusResolver` now passes `BigInt(receipt.blockNumber)` to `isRateLimitExceededError`, which threads it through to `parseTransactionError`. The `call` simulation is replayed at the original block so state-dependent checks reflect the conditions at the time of the revert.

postman/README.md
VGau: Done — `.env.sample` has been updated with web3signer options, receipt polling intervals, SSL config, `MAX_BUMPS_PER_CYCLE`/`MAX_RETRY_CYCLES` replacing `MAX_TX_RETRIES`, and `L1/L2_CLAIM_VIA_ADDRESS`.

postman/src/infrastructure/blockchain/viem/gas/ViemEthereumGasProvider.ts
VGau: Fixed — extracted as `FEE_HISTORY_BLOCK_COUNT = 4` in `core/constants/blockchain.ts` and used in both `ViemEthereumGasProvider` and `ViemLineaGasProvider`.

postman/src/core/types/primitives.ts
VGau: Renamed `hex.ts` → `primitives.ts` to remove the module/type name redundancy. Kept `Hash` and `Hex` — they follow Ethereum ecosystem convention (used in viem, ethers, EIPs). `core/` is intentionally viem-free so we define our own aliases rather than re-exporting from viem.

postman/src/core/metrics/MessageMetricsAttributes.ts
VGau: Good idea for observability. Tracked as a future enhancement, out of scope for this PR.

postman/src/infrastructure/blockchain/viem/ViemTransactionRetrier.ts
VGau: The cancel tx already handles this via the stuckFees parameter — after each bump, updateMessage persists the new claimTxMaxFeePerGas/claimTxMaxPriorityFeePerGas. When cancelAndResetMessage is called, it reads those persisted values and passes them as stuckFees. The cancel tx then uses max(aggressiveFees, stuckFees x 110%), so no matter how high the exponential bump has gone, the cancel always bids at least 10% above the latest stuck tx.

postman/src/infrastructure/blockchain/viem/gas/ViemLineaGasProvider.ts
VGau: It uses linea_estimateGas and postman-specific config types so it is postman-specific in its current form. Could be generalised in a follow-up, out of scope here.

postman/src/infrastructure/blockchain/viem/providers/ViemProvider.ts
VGau: Same answer as ViemLineaGasProvider - wraps viem primitives with postman domain types. Could be extracted to shared-utils in a follow-up.

postman/src/application/postman/app/config/schema.ts
VGau: Good point - Zod helpers for Ethereum primitives (addressSchema etc.) could live in shared-utils. Out of scope for this PR, will track as a follow-up.

thedarkjester (approved):

Issue #2103, opened 2026-01-21
Coordinator - Forced Transactions Module


fluentcrafter:
## Description
implement module that responsible for the business logic of:

- relaying forced transactions from L1 to the sequencer, by their order
- detect forced transactions inclusion status
- request prover's for transaction's inclusion/valid in L2
- aggregate and finalize on L1
- export forced transactions metrics
- log when forced transactions:
  - are observed on L1
  - invalid and rejected by sequencer (e.g call to unsupported pre-compiles)

PR #2658, opened 2026-03-26
feat: add cleanup ftx db

jonesho:
This PR implements issue(s) #2103

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> Changes the finalization monitor data source and extends finalized-update payloads to include an optional forced-transaction number, then uses it to delete rows from the forced-transactions DB table. Moderate risk because it affects finalization polling behavior and introduces additional data deletion during cleanup.
> 
> **Overview**
> **Finalization monitoring now consumes a new `FinalizedStateDataProvider` abstraction** instead of directly querying `finalizedL2BlockNumber`, and `FinalizationUpdate` gains an optional `forcedTransactionNumber` (null for pre-V8 contracts).
> 
> **Finalization-driven DB cleanup is expanded**: `RecordsCleanupFinalizationHandler` now deletes forced-transaction records up to the finalized forced-tx number (when present), and `L1DependentApp` wires the required `ForcedTransactionsDao` into this handler.
> 
> `Web3JLineaRollupSmartContractClientReadOnly` now implements `FinalizedStateDataProvider` (returning forced-tx data only on V8), and tests are updated to use a new `FakeFinalizedStateDataProvider` fixture plus an integration test that verifies forced-transaction cleanup behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1bf9b78d5d0034f3d069082f98bc5287f72b21dc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/L1DependentApp.kt
```diff
@@ -714,7 +714,7 @@ class L1DependentApp(
     fun setupL1FinalizationMonitorForShomeiFrontend(
       type2StateProofProviderConfig: Type2StateProofManagerConfig,
       httpJsonRpcClientFactory: VertxHttpJsonRpcClientFactory,
-      lineaRollupClient: LineaRollupSmartContractClientReadOnly,
+      lineaRollupClient: Web3JLineaRollupSmartContractClientReadOnly,
```
fluentcrafter: Why changing the type. That defeats the purpose of an interface and violates SOLID principles. 

We should revert this unless there is a strong reason otherwise..

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/finalization/FinalizationMonitor.kt
```diff
@@ -7,6 +7,7 @@ interface FinalizationMonitor {
   data class FinalizationUpdate(
     val blockNumber: ULong,
     val blockHash: Bytes32,
+    val forcedTransactionNumber: ULong? = null,
```
fluentcrafter: why nullable?

Also, I think `val blockHash: Bytes32` is not being used anywhere. If so, maybe we can just dele it and make the code simpler? Can you do this in an other PR please?

Filter94 (dismissed):

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/finalization/FinalizationMonitor.kt
jonesho: @Filter94 @fluentcrafter I checked that `blockHash` is needed when calling `rollupForkChoiceUpdated` to Shomei frontend:
https://github.com/Consensys/linea-monorepo/blob/70c085fc9378660da93b349f858ed5cff85f2759/coordinator/clients/shomei-client/src/main/kotlin/net/consensys/zkevm/coordinator/clients/ShomeiClient.kt#L57

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/L1DependentApp.kt
jonesho: Oh my bad, I was mistakenly thought `Web3JLineaRollupSmartContractClientReadOnly` was an interface, I will create a separate interface instead.

fluentcrafter: There is already `LineaRollupSmartContractClientReadOnly` whe do you need a new one. What other methods than those alreay provided by the interface do you need?

coordinator/ethereum/finalization-monitor/src/main/kotlin/net/consensys/zkevm/ethereum/finalization/FinalizationMonitorImpl.kt
```diff
@@ -67,15 +68,29 @@ class FinalizationMonitorImpl(
...
 
   private fun getFinalizationState(): SafeFuture<FinalizationMonitor.FinalizationUpdate> {
-    return contract
-      .finalizedL2BlockNumber(blockParameter = config.l1QueryBlockTag)
-      .thenCompose { lineaFinalizedBlockNumber ->
+    return contract.getVersion()
+      .thenCompose { version ->
+        when (version) {
+          LineaRollupContractVersion.V6,
+          LineaRollupContractVersion.V7,
```
fluentcrafter: IMO adding this version dependent code does not scale well. The interface should hide version breaking changes as much as possible... 

What other options have you considered?

jonesho: @fluentcrafter Another option is to add a periodic ftx db cleanup service in `ForcedTransactionsAppImpl` (similar to `ftxInvalidityProofService`) which polls latest finalized ftx number by periodically calling the `finalizedStateProvider.getLatestFinalizedState()`, in this way we don't even need to check the version, create a new interface, and modify other files like `FinalizationMonitorImpl.kt` and `FinalizationMonitor.kt‎`, wdyt? 

My initial intent is to put the ftx db clean up logic closer to the other db clean up rountines in `RecordsCleanupFinalizationHandler.kt`

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/L1DependentApp.kt
jonesho: I need the `getLatestFinalizedState` method from interface `LineaRollupSmartContractClientReadOnlyFinalizedStateProvider` for querying the finalized ftx number. The interface `LineaRollupSmartContractClientReadOnly` doesn't declare that method.

coordinator/ethereum/finalization-monitor/src/main/kotlin/net/consensys/zkevm/ethereum/finalization/FinalizationMonitorImpl.kt
fluentcrafter: > My initial intent is to put the ftx db clean up logic closer to the other db clean up

I think that is a good property to have. 

However I don't understand the relation of that answer with the changes done here in `FinalizationMonitorImpl` because it's doing the clean up here...

jonesho: @fluentcrafter the clean up is not done here, all the code changes here in `FinalizationMonitorImpl` are to fill in the `forcedTransactionNumber` in `FinalizationUpdate`, such that the listener (`RecordsCleanupFinalizationHandler`) would be able to use the `forcedTransactionNumber` to [clean up ftx db records](https://github.com/Consensys/linea-monorepo/pull/2658/changes/BASE..50ce0bf16b76e46e7318aa7ee29c6a4f4cb54c46#diff-7796f93ec8460010b793d9beb2f6a8f85b21155199a6bcb8fa5a016d35eb3237R50), just like the other db clean ups.

jonesho: As mentioned above, another way is to add a periodic ftx db cleanup service in `ForcedTransactionsAppImpl` (similar to `ftxInvalidityProofService`) which polls latest finalized ftx number by periodically calling the `finalizedStateProvider.getLatestFinalizedState()`, in this way we don't even need to do any version check, no need to create any new interface, and modify `FinalizationUpdate` data class. It will be less changes and all under `forced-transactions` folder. The cons is that the ftx db cleanup logic would not live close to other db cleanup logics, I'd created the [PR](https://github.com/Consensys/linea-monorepo/pull/2719) with this alternative approach

Filter94: For me the point of Pedro's comment is not about the cleanup logic, but about the fact that there's a switch here that handles different contract versions differently. Which IMO should be done somewhere in a contract client

jonesho: I see the point now, added changes to hide the version check inside the contract client, and add an provider interface instead

coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/finalization/FinalizationMonitor.kt
```diff
@@ -7,6 +7,7 @@ interface FinalizationMonitor {
   data class FinalizationUpdate(
     val blockNumber: ULong,
     val blockHash: Bytes32,
+    val forcedTransactionNumber: ULong? = null, // null means ftx number is not available as contract is still before V8
```
fluentcrafter: If the contract is before V8, we can surely default to `0` no?

jvm-libs/linea/clients/interfaces/src/main/kotlin/linea/contract/l1/FinalizedStateDataProvider.kt
```diff
@@ -0,0 +1,13 @@
+package linea.contract.l1
+
+import linea.domain.BlockParameter
+import tech.pegasys.teku.infrastructure.async.SafeFuture
+
+interface FinalizedStateDataProvider {
```
Filter94, line 6: Nit:
```suggestion
class FakeFinalizedStateProvider(
```

Filter94 (approved):

PR #2670, opened 2026-03-26
Feat/compressor instances

Tabaie:
Enables defining multiple compressor instances on the JVM side. Each `GoBackedBlobCompressor` object corresponds to a compressor instance.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes the JVM↔native (Go/JNA) compressor interface to a handle-based lifecycle with explicit `close`/`Free`, which can introduce leaks or crashes if misused. It also updates the shipped native library versions, so behavior differences/regressions are possible.
> 
> **Overview**
> Enables **multiple independent blob-compressor instances per JVM** by switching the native compressor API from a single global instance to a **handle-based** model (Go `Init` returns a handle; JVM wraps it and calls `Free` on `close`).
> 
> Introduces `BlobCompressorFactory` (replacing direct `GoBackedBlobCompressor.getInstance`) to select the legacy v2 API vs the new handle-based API, updates `BlobCompressor`/`BlobCompressorSelector` to be `Closeable`, and adds `GoBackedBlobCompressorV4` to manage per-handle state.
> 
> Updates native library downloads to `blob-libs-v4.0.0`, adjusts compressor version enums/mappings, removes older shnarf-calculator v1.2 references, and expands tests to cover multi-instance independence, proper cleanup, and updated integration paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit de9208fc12a3e1249908822db606b55eaebedcaa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/besu-plugins/state-recovery/appcore/logic/src/test/kotlin/linea/staterecovery/BlobDecompressorAndDeserializerTest.kt
+++ b/besu-plugins/state-recovery/appcore/logic/src/test/kotlin/linea/staterecovery/BlobDecompressorAndDeserializerTest.kt
@@ -24,7 +24,7 @@ import kotlin.jvm.optionals.getOrNull
 import kotlin.time.Instant
 
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
-class BlobDecompressorAndDeserializerV1Test {
+class BlobDecompressorAndDeserializerTest {
   private lateinit var compressor: BlobCompressor
   private val blockStaticFields = BlockHeaderStaticFields(
     coinbase = Address.ZERO.toArray(),
--- a/jvm-libs/linea/blob-compressor/src/test/kotlin/linea/blob/GoBackedBlobCompressorTest.kt
+++ b/jvm-libs/linea/blob-compressor/src/test/kotlin/linea/blob/GoBackedBlobCompressorTest.kt
@@ -172,7 +172,7 @@ class GoBackedBlobCompressorTest {
   @ParameterizedTest(name = "{0}")
   @MethodSource("compressorVersions")
   fun `close releases instance and a new one can be created`(version: BlobCompressorVersion) {
-    val c1 = newCompressor(version).use { c1 ->
+    newCompressor(version).use { c1 ->
       c1.appendBlock(TEST_DATA.first())
     }
 
--- a/jvm-libs/linea/blob-compressor/src/test/kotlin/net/consensys/linea/blob/GoNativeCompressorAndShnarfCalculatorIntTest.kt
+++ b/jvm-libs/linea/blob-compressor/src/test/kotlin/net/consensys/linea/blob/GoNativeCompressorAndShnarfCalculatorIntTest.kt
@@ -36,15 +36,12 @@ class GoNativeCompressorAndShnarfCalculatorIntTest {
   @ParameterizedTest()
   @ValueSource(booleans = [false, true])
   fun `should compress and shnarf with eip4844 enabled-disabled`(eip4844Enabled: Boolean) {
-    val compressorV2 = BlobCompressorFactory.getInstance(BlobCompressorVersion.V2, DATA_LIMIT)
     val compressorV3 = BlobCompressorFactory.getInstance(BlobCompressorVersion.V3, DATA_LIMIT)
     val compressorV4 = BlobCompressorFactory.getInstance(BlobCompressorVersion.V4, DATA_LIMIT)
 
-    val shnarfCalculatorV1 = GoNativeShnarfCalculatorFactory.getInstance(ShnarfCalculatorVersion.V1_2)
     val shnarfCalculatorV3 = GoNativeShnarfCalculatorFactory.getInstance(ShnarfCalculatorVersion.V3)
 
     val compressorAndShnarfCalculator = listOf(
-      Pair(compressorV2, shnarfCalculatorV1),
       Pair(compressorV3, shnarfCalculatorV3),
       Pair(compressorV4, shnarfCalculatorV3),
     )
--- a/jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/BlobCompressorSelector.kt
+++ b/jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/BlobCompressorSelector.kt
@@ -1,8 +1,9 @@
 package linea.blob
 
+import java.io.Closeable
 import kotlin.time.Instant
 
-interface BlobCompressorSelector<T> {
+interface BlobCompressorSelector<T> : Closeable {
   /**
    * Returns a BlobCompressor instance based on the provided selector.
    */
@@ -35,5 +36,9 @@ class BlobCompressorSelectorByTimestamp(
 
     return blobCompressors[blobCompressorVersion]
       ?: throw IllegalStateException("BlobCompressor for version $blobCompressorVersion not found")
+  }
+
+  override fun close() {
+    blobCompressors.values.forEach(BlobCompressor::close)
   }
 }
--- a/jvm-libs/linea/blob-shnarf-calculator/build.gradle
+++ b/jvm-libs/linea/blob-shnarf-calculator/build.gradle
@@ -28,7 +28,6 @@ def libsZipDownloadOutputDir = project.parent.layout.buildDirectory.asFile.get()
 
 task downloadNativeLibs {
   doLast {
-    fetchLibFromZip("https://github.com/Consensys/linea-monorepo/releases/download/blob-libs-v1.2.0/linea-blob-libs-v1.2.0.zip", "shnarf_calculator", libsZipDownloadOutputDir)
     fetchLibFromZip("https://github.com/Consensys/linea-monorepo/releases/download/blob-libs-v3.0.1/linea-blob-libs-v3.0.1.zip", "shnarf_calculator", libsZipDownloadOutputDir)
   }
 }
--- a/jvm-libs/linea/blob-shnarf-calculator/src/main/kotlin/linea/blob/GoNativeBlobShnarfCalculator.kt
+++ b/jvm-libs/linea/blob-shnarf-calculator/src/main/kotlin/linea/blob/GoNativeBlobShnarfCalculator.kt
@@ -58,7 +58,6 @@ interface GoNativeBlobShnarfCalculator {
 internal interface GoNativeBlobShnarfCalculatorJna : GoNativeBlobShnarfCalculator, Library
 
 enum class ShnarfCalculatorVersion(val version: String) {
-  V1_2("v1.2.0"),
   V3("v3.0.1"),
 }
 
--- a/jvm-libs/linea/blob-shnarf-calculator/src/test/kotlin/linea/blob/GoNativeBlobShnarfCalculatorTest.kt
+++ b/jvm-libs/linea/blob-shnarf-calculator/src/test/kotlin/linea/blob/GoNativeBlobShnarfCalculatorTest.kt
@@ -13,11 +13,9 @@ class GoNativeBlobShnarfCalculatorTest {
   companion object {
     @JvmStatic
     fun enumerateShnarfCalculators(): Stream<Arguments> {
-      val shnarfCalculatorV1 = GoNativeShnarfCalculatorFactory.getInstance(ShnarfCalculatorVersion.V1_2)
       val shnarfCalculatorV3 = GoNativeShnarfCalculatorFactory.getInstance(ShnarfCalculatorVersion.V3)
 
       return Stream.of(
-        Arguments.of(shnarfCalculatorV1),
         Arguments.of(shnarfCalculatorV3),
       )
     }


jvm-libs/linea/blob-compressor/build.gradle
```diff
@@ -38,6 +38,7 @@ task downloadNativeLibs {
     fetchLibFromZip("https://github.com/Consensys/linea-monorepo/releases/download/blob-libs-v1.2.0/linea-blob-libs-v1.2.0.zip", "blob_compressor", libsZipDownloadOutputDir)
```
fluentcrafter: I think we can decomission 1.2.0

jvm-libs/linea/blob-compressor/build.gradle
```diff
@@ -38,6 +38,7 @@ task downloadNativeLibs {
...
     fetchLibFromZip("https://github.com/Consensys/linea-monorepo/releases/download/blob-libs-v2.1.0-rc1/linea-blob-libs-v2.1.0-rc1.zip", "blob_compressor", libsZipDownloadOutputDir)
     fetchLibFromZip("https://github.com/Consensys/linea-monorepo/releases/download/blob-libs-v3.0.1/linea-blob-libs-v3.0.1.zip", "blob_compressor", libsZipDownloadOutputDir)
+    fetchLibFromZip("https://github.com/Consensys/linea-monorepo/releases/download/blob-libs-v4.0.0/linea-blob-libs-v4.0.0.zip", "blob_compressor", libsZipDownloadOutputDir)
```
fluentcrafter, line 39: AKIU, V4 it's actually V3 implementation with different API, right ? I mean, compression/decompression is done by the same prover?

If so, shall we:
- not import V3,
- when we see. V3 in coordinator's config entry fallback to V4 to no break compatibilitly?

jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/BlobCompressor.kt
```diff
@@ -40,68 +39,37 @@ interface BlobCompressor {
...
-
-      val goNativeBlobCompressor = GoNativeBlobCompressorFactory.getInstance(compressorVersion)
-      val initialized = goNativeBlobCompressor.Init(
-        dataLimit = dataLimit,
-        dictPath = GoNativeBlobCompressorFactory.dictionaryPath.toString(),
-      )
-      if (!initialized) {
-        throw InstantiationException(goNativeBlobCompressor.Error())
+object BlobCompressorFactory {
+  @JvmStatic
```
fluentcrafter, line 43: Make this Thread Safe? No sure the underlying go code can handle multiple threads during instantiation?

jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/GoBackedBlobCompressor.kt
```diff
@@ -0,0 +1,107 @@
...
+
+  override fun close() {}
+}
+
+class GoBackedBlobCompressorV4 internal constructor(
+  private val goNativeBlobCompressor: GoNativeBlobCompressorV4,
+  override val version: BlobCompressorVersion,
+  private val handle: Int,
+) : BlobCompressor {
+  private val log = LogManager.getLogger(GoBackedBlobCompressor::class.java)
```
fluentcrafter:
```suggestion
  private val log = LogManager.getLogger(GoBackedBlobCompressorV4::class.java)
```

jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/GoBackedBlobCompressor.kt
```diff
@@ -0,0 +1,107 @@
+package linea.blob
+
+import linea.kotlin.encodeHex
+import org.apache.logging.log4j.LogManager
+
+class GoBackedBlobCompressor internal constructor(
+  private val goNativeBlobCompressor: GoNativeBlobCompressor,
+  override val version: BlobCompressorVersion,
```
fluentcrafter, line 8: what is this used for?

jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/GoBackedBlobCompressor.kt
```diff
@@ -0,0 +1,107 @@
...
+  override fun compressedSize(data: ByteArray): Int {
+    return goNativeBlobCompressor.RawCompressedSize(data, data.size)
+  }
+
+  override fun close() {}
+}
+
+class GoBackedBlobCompressorV4 internal constructor(
+  private val goNativeBlobCompressor: GoNativeBlobCompressorV4,
+  override val version: BlobCompressorVersion,
```
fluentcrafter, line 58: what is this used for?

jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/GoBackedBlobCompressor.kt
```diff
@@ -0,0 +1,107 @@
...
+
+class GoBackedBlobCompressor internal constructor(
+  private val goNativeBlobCompressor: GoNativeBlobCompressor,
+  override val version: BlobCompressorVersion,
+) : BlobCompressor {
+  private val log = LogManager.getLogger(GoBackedBlobCompressor::class.java)
+
+  override fun canAppendBlock(blockRLPEncoded: ByteArray): Boolean {
+    return goNativeBlobCompressor.Write(blockRLPEncoded, blockRLPEncoded.size)
+  }
```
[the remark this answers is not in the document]
fluentcrafter: this seems legit. We should had a test for this case.

jvm-libs/linea/blob-compressor/src/test/kotlin/linea/blob/GoBackedBlobCompressorTest.kt
```diff
@@ -1,102 +1,198 @@
...
     private const val DATA_LIMIT = 24 * 1024
     private val TEST_DATA = CompressorTestData.blocksRlpEncoded
-    private val compressor = GoBackedBlobCompressor.getInstance(BlobCompressorVersion.V3, DATA_LIMIT)
   }
 
-  @BeforeEach
-  fun before() {
-    compressor.reset()
-  }
+  private fun newCompressor() = BlobCompressorFactory.getInstance(BlobCompressorVersion.V4, DATA_LIMIT)
```
fluentcrafter: how's V3 and V2 tested now?

fluentcrafter: Also we should add a test that canAppend does not mutate the compressor data.

jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/GoBackedBlobCompressor.kt
gauravahuja: Version was added for the blob compressor selector change

jvm-libs/linea/blob-compressor/build.gradle
gauravahuja: I think we can, but I would prefer doing it in a separate PR

jvm-libs/linea/blob-compressor/build.gradle
gauravahuja: ohk, I think we can do this. So I will remove both 1.2.0 and 3.0.1. For V3 I will load 4.0.0 but the compressor.version will still be V3

jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/GoBackedBlobCompressor.kt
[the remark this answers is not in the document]
gauravahuja: Fixed and added test

jvm-libs/linea/blob-compressor/build.gradle
gauravahuja: Removed compressor v1_2 in this PR

jvm-libs/linea/blob-compressor/build.gradle
gauravahuja: Done

jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/GoBackedBlobCompressor.kt
gauravahuja: Version was added for the blob compressor selector change

jvm-libs/linea/blob-compressor/src/test/kotlin/linea/blob/GoBackedBlobCompressorTest.kt
gauravahuja: V2 is test in the file GoBackedBlobCompressorLegacyTest

jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/BlobCompressor.kt
gauravahuja: BlobCompressorFactory does not need to be thread safe as it always creates a new instance of BlobCompressor. The underlying GoNativeBlobCompressorLibrary already uses locks and is thread safe while loading the `.so` library.

besu-plugins/state-recovery/appcore/logic/src/test/kotlin/linea/staterecovery/BlobDecompressorAndDeserializerTest.kt
```diff
@@ -37,8 +37,8 @@ class BlobDecompressorAndDeserializerV1Test {
   @BeforeEach
   fun setUp() {
     vertx = Vertx.vertx()
-    compressor = GoBackedBlobCompressor.getInstance(
-      compressorVersion = BlobCompressorVersion.V1_2,
+    compressor = BlobCompressorFactory.getInstance(
+      compressorVersion = BlobCompressorVersion.V2,
```
fluentcrafter, line 41: maybe we shall rename the class name now?

jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/GoBackedBlobCompressor.kt
fluentcrafter: where inside the `GoBackedBlobCompressor` or `GoBackedBlobCompressorV4` is version being used?

jvm-libs/linea/blob-compressor/src/test/kotlin/net/consensys/linea/blob/GoNativeCompressorAndShnarfCalculatorIntTest.kt
```diff
@@ -28,81 +31,84 @@ class GoNativeCompressorAndShnarfCalculatorIntTest {
...
+    val compressorV4 = BlobCompressorFactory.getInstance(BlobCompressorVersion.V4, DATA_LIMIT)
 
-  @ParameterizedTest
-  @MethodSource("enumerateValidBlobCompressorAndShnarfCalculatorPair")
-  fun `should compress and shnarf with eip4844 enabled`(
-    compressor: GoNativeBlobCompressor,
-    shnarfCalculator: GoNativeBlobShnarfCalculator,
-  ) {
-    testsCompressionAndsShnarfCalculationWithEip4844Enabled(compressor, shnarfCalculator)
+    val shnarfCalculatorV1 = GoNativeShnarfCalculatorFactory.getInstance(ShnarfCalculatorVersion.V1_2)
```
fluentcrafter: we shall decomission shnarf calculator V1_2 as well.

jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/GoBackedBlobCompressor.kt
gauravahuja: version is used in [CachingTransactionCompressor](https://github.com/Consensys/linea-monorepo/blob/main/besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java#L72)

jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/GoBackedBlobCompressor.kt
gauravahuja: version is used in [CachingTransactionCompressor](https://github.com/Consensys/linea-monorepo/blob/main/besu-plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/utils/CachingTransactionCompressor.java#L72)

jvm-libs/linea/blob-compressor/src/test/kotlin/net/consensys/linea/blob/GoNativeCompressorAndShnarfCalculatorIntTest.kt
gauravahuja: What about BlobDecompressorVersion V1_2_0

besu-plugins/state-recovery/appcore/logic/src/test/kotlin/linea/staterecovery/BlobDecompressorAndDeserializerTest.kt
gauravahuja: Rename to what? I already moved factory methods from `GoBackedBlobCompressor` to a new class `BlobCompressorFactory`

fluentcrafter: the class is called V1, however V1 does not exist anymore. Not very relevant though

jvm-libs/linea/blob-compressor/src/test/kotlin/net/consensys/linea/blob/GoNativeCompressorAndShnarfCalculatorIntTest.kt
fluentcrafter: It may be used for the state recovery? Otherwise, it can be removed as well.

jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/GoNativeBlobCompressor.kt
```diff
@@ -97,14 +99,41 @@ interface GoNativeBlobCompressor {
...
+interface GoNativeBlobCompressorLegacyJnaLib : GoNativeBlobCompressor, Library
+
+interface GoNativeBlobCompressorV4JnaLib : GoNativeBlobCompressorV4, Library
 
 enum class BlobCompressorVersion(val version: String) {
-  V1_2("v1.2.0"),
   V2("v2.1.0"),
-  V3("v3.0.1"),
+  V3("v4.0.0"),
+  V4("v4.0.0"),
```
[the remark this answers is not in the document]
gauravahuja: This is expected because we load the same library for v3 and v4.

jvm-libs/linea/blob-compressor/src/main/kotlin/linea/blob/GoBackedBlobCompressor.kt
```diff
@@ -0,0 +1,107 @@
...
+    goNativeBlobCompressor.Reset(handle)
+  }
+
+  override fun compressedSize(data: ByteArray): Int {
+    return goNativeBlobCompressor.RawCompressedSize(handle, data, data.size)
+  }
+
+  override fun close() {
+    goNativeBlobCompressor.Free(handle)
+  }
```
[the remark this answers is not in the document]
gauravahuja: This is expected.

jvm-libs/linea/blob-compressor/src/test/kotlin/net/consensys/linea/blob/GoNativeCompressorAndShnarfCalculatorIntTest.kt
```diff
@@ -28,81 +31,81 @@ class GoNativeCompressorAndShnarfCalculatorIntTest {
...
+      }
+    testCompressedSizeEstimationIsConsistentWithBlobMakerOutput(legacyCompressor)
+
+    BlobCompressorVersion.entries.filter { it != BlobCompressorVersion.V2 }.forEach { version ->
+      val compressor = GoNativeBlobCompressorFactory.getInstance(version)
+      val errOut = PointerByReference()
+      val handle = compressor.Init(DATA_LIMIT, GoNativeBlobCompressorFactory.dictionaryPath.toString(), errOut)
+      if (handle == -1) {
+        throw InstantiationException(
+          errOut.value?.getString(0) ?: "Failed to initialize compressor",
```
jonesho, line 80: @gauravahuja I suspect this would cause potential memory leak here as explicitly mentioned in the go code [here](https://github.com/Consensys/linea-monorepo/pull/2670/changes#diff-be858060b6f30c2dd28c835814865d0d04e51a485da8dcc26dd9047598f39010R42) and checked with Claude which seems positive:

```
 there's a memory leak here. Line 58 in BlobCompressor.kt defines val errOut = PointerByReference().                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                             
  When handle == -1, the Go side (libcompressor.go:50) allocates a C string via C.CString(err.Error()), which uses malloc. The Kotlin code reads it via errOut.value?.getString(0) but never frees the native memory.                                                                                                        
                                                                                                                                                                                                                                                                                                                             
  The fix:                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                             
  if (handle == -1) {                                                                                                                                                                                                                                                                                                        
    val errorMsg = errOut.value?.getString(0) ?: "Failed to initialize compressor"
    errOut.value?.let { com.sun.jna.Native.free(com.sun.jna.Pointer.nativeValue(it)) }
    throw InstantiationException(errorMsg)                                                                                                                                                                                                                                                                                   
  }
                                                                                                                                                                                                                                                                                                                             
  Native.free is JNA's equivalent of calling C.free on the pointer, which is the correct way to release memory allocated by C.CString on the Go side.  
```

gauravahuja: ok, I can fix this in a follow up. However this is not that severe as there is no way to recover if there is an exception.

Issue #2673, opened 2026-03-27
Restore Highest proven batch execution block number metric to previous behavior


gauravahuja:
After refactoring the proof flow we create the proof request file and add the proof index to a queue. This queue is unbounded. The poller checks if the first proof index from the queue is proven or not at every tick. If it is proven it process the proof and removes the proof index from the queue. This way the poller moves sequentially and the metric "Highest proven batch execution block number" reflects that.

Previously we would create a proof request and the future would only return when the response was created. So we would have multiple futures waiting and as soon as a proof response was availble it was processed. This proof response could be out of order too. But this way the metric "Highest proven batch execution block number" reflected the highest batch processed even though previous batches might not be proven.

We have another metric for consecutive proofs which along with this metric helps figure out gaps.

We need to restore "Highest proven batch execution block number" to previous behavior. This can be done by having the poller check all proof indexes in the queue or (some fixed number like 200) at every tick.

PR #2674, opened 2026-03-27
[Coordinator] Poll multiple execution proof response per tick

gauravahuja:
This PR implements issue(s) #2673

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> Changes the proof-status polling loop to issue many concurrent status checks per tick, which could increase load on the proof coordinator and surface concurrency/ordering issues when removing completed requests.
> 
> **Overview**
> Updates `ProofGeneratingConflationHandlerImpl` to poll **multiple** in-flight `ExecutionProofIndex` entries per polling tick (instead of only the head of the queue), running the checks concurrently and removing proven indices after `acceptNewBatch` completes.
> 
> Adds a new config knob, `executionProofPollsPerTick` (default `200`), to cap how many pending proofs are checked each tick.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 54fde7f0bea91f3b9137adc6408162c6514546b7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/conflation/ProofGeneratingConflationHandlerImpl.kt
+++ b/coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/conflation/ProofGeneratingConflationHandlerImpl.kt
@@ -44,6 +44,7 @@ class ProofGeneratingConflationHandlerImpl(
   data class Config(
     val conflationAndProofGenerationRetryBackoffDelay: Duration,
     val executionProofPollingInterval: Duration,
+    val executionProofPollsPerTick: Int = 200,
   )
 
   private val proofRequestsInProgress = ConcurrentLinkedDeque<ExecutionProofIndex>()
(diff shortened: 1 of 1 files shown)


coordinator/core/src/main/kotlin/net/consensys/zkevm/ethereum/coordination/conflation/ProofGeneratingConflationHandlerImpl.kt
```diff
@@ -57,28 +58,36 @@ class ProofGeneratingConflationHandlerImpl(
...
 
+  override fun action(): SafeFuture<*> {
+    if (proofRequestsInProgress.isEmpty()) {
+      return SafeFuture.completedFuture(Unit)
+    }
+    val iterator = proofRequestsInProgress.iterator()
+    val proofIndicesToPoll = mutableListOf<ExecutionProofIndex>()
+    while (iterator.hasNext() && proofIndicesToPoll.size < config.executionProofPollsPerTick) {
+      proofIndicesToPoll.add(iterator.next())
+    }
```
jonesho, line 86: This would not solve the issue if `proofRequestsInProgress` is more than `config.executionProofPollsPerTick`, can we iterate the entire list in multiple ticks (if the list size > `config.executionProofPollsPerTick`) but with a static number of elements in each tick? E.g. remember the last visited block number and iterate beginning with the next one in the subsequent tick, if the list size > `config.executionProofPollsPerTick`

gauravahuja: At every tick we poll config.executionProofPollsPerTick proof indices and if any of the have been proven they will be processed and removed from the queue. So next tick we will poll the proof indices that were not proven + new ones from the queue. The number of new ones will be equal to the number of proven proof indices from the previous tick. So even with a high static number we are progressing in the queue. The static number is 200 which is high enough that we won't have that many batch proofs pending at any time.

jonesho: I agree 200 should be high enough to cover most cases.

Filter94: I'm concerned that those 200 requests to the file system will be executed concurrently. I feel like we better set the parallelism factor separately from the reasonable lookahead window

Issue #2653, opened 2026-03-25
Coordinator metric for highest_consecutive_block_number was not updated correctly


jonesho:
🔐 🔐 `Should this be a security advisory instead?` 🔐 🔐

We noticed an [incident](https://consensys.slack.com/archives/C0AN7FHUF8W/p1774397868292359?thread_ts=1774395133.104719&cid=C0AN7FHUF8W) on mainnet that the coordinator metric was stuck at`Highest proven consecutive execution batch block number=29882788` but the next execution proof had been created. 

Also note that the coordinator atm was set to stop conflation after certain block (29887667), and restarted coordinator did bring us to correct metric value.

## Impact
__environment [Testnet/Mainnet]__

__an user is impacted [Y/N], if yes type of user: [External/Internal]__

### Steps to Reproduce
_A bug that we can't reproduce is hard to resolve, to ensure the bug is fixed in a timely manner, it is crucial to indicate how to reproduce it. If the bug can't be reproduced, give specific details on how it happened._

Steps to reproduce the bug:

### Actual result
Evidence of the bug including log extracts, screen shot, wallet addresses, Tx hash, ...

### Expected result

_Remember that an issue without proper labels and without assignee is unlikely to get prioritized -> contact the team or the product owner to get it included in the plan._

PR #2676, opened 2026-03-27
feat: create polling service to monitor highest consecutive proven batch block number

jonesho:
…ipped operations conditionally

This PR implements issue(s) #2653

We identified that root cause is that, when `force-stop-conflation-at-block-inclusive` is enabled and the target block has reached, the update of the metric `linea_batch_proven_highest_consecutive_block_number` would be stopped because the `BlockCreationMonitor` periodic action routine was completely stopped.

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> Touches core conflation/block polling flow by changing how `BlockCreationMonitor` gates fetching and by introducing a new periodic monitor; incorrect scheduling/caching could stall or overrun block processing or leave metrics inaccurate.
> 
> **Overview**
> Adds an independent `ProvenBlockNumberMonitor` periodic service in `ConflationApp` to keep the highest consecutive proven batch block number refreshed (and its metric updated) even when `BlockCreationMonitor` stops due to `forceStopConflationAtBlockInclusive`/timestamp limits.
> 
> Refactors `BlockCreationMonitor` (and backtesting + tests) to use a synchronous `LastProvenBlockNumberProviderSync.getLastKnownProvenBlockNumber()` for per-tick gap checks, with the async DB refresh handled by the new monitor; also slightly adjusts the reorg log message formatting.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 56b0afb2841e2af6f6177daf43c78b53891c6861. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflation/ConflationApp.kt
```diff
@@ -499,14 +501,28 @@ class ConflationApp(
     lastProvenConsecutiveBatchBlockNumberProvider
   }
 
+  // This object acts as an independent periodic polling service which is responsible
+  // for monitoring the highest consecutive proven block number in the batch db
+  private val provenBlockNumberMonitor = object : VertxPeriodicPollingService(
```
Filter94, line 506: I think we can replace this poller with a cache computing call in `lastProvenBlockNumberProvider`. And if you're concerned about DB queries triggered by the scraper, we can set a debouncer right inside the `BatchesRepoBasedLastProvenBlockNumberProvider`

jonesho: @Filter94 Yes, we can have a debouncer to control the rate of db operations; however, another point is that if we do what you suggested, we're pushing the responsibility of moving the cache `lastProvenBlockNumber` forward to the metric scraper (class from external), and if it was somehow not calling the `measurementSupplier` (e.g. being disabled) or the metric code get removed by another contributor, then not only the metric will be stuck, the other components that are depending on `lastProvenBlockNumber` will be impacted as well, e.g. the "block creation monitor" will run into stop conflation mode because the gap increases when `lastProvenBlockNumber` stuck.
imo letting the scraper to own the responsibility for advancing `lastProvenBlockNumber` doesn't sound plausible to me, having another long running service to solely monitor the `lastProvenBlockNumber` seems the most safest approach, maybe any suggestion?

jonesho: After a bit of research and it seems to me triggering db calls on scraper is not encouraged:
```
Claude: Do NOT trigger database calls on scrape. This is a high-risk anti-pattern, especially in Vert.x-based applications. Instead, use a push-based or cache-based approach.
```
In addition, the following change on this [line](https://github.com/Consensys/linea-monorepo/blob/23421dd596b74e1d09bdb15ec4e4988cff735be4/coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflation/ConflationApp.kt#L499)
```
measurementSupplier = { lastProvenConsecutiveBatchBlockNumberProvider.getLastProvenBlockNumber().get() }
```
will block the scrape thread which is the Vert.x event loop, given that get() will be called eventually, plus a potential deadlock:
```
- Potential deadlock: If batchesRepository uses a Vert.x reactive SQL client, its completion callback is dispatched back onto the same event loop — which is now blocked waiting for it → classic deadlock
```
Indeed, the current cache-based approach `measurementSupplier = { lastProvenConsecutiveBatchBlockNumberProvider.getLastKnownProvenBlockNumber() }` is deemed be the best practice for scraper supplier
```
  The safe approach:
  
  The class already has getLastKnownProvenBlockNumber() at line 38, which just reads the AtomicLong — zero blocking,
  zero DB access. Use that in the Micrometer gauge instead:

  // Safe: reads cached AtomicLong, no I/O
  Gauge.builder("last_proven_block") { provider.getLastKnownProvenBlockNumber().toDouble() }
      .register(registry)

  Let the coordinator's existing async polling cycle call findAndCacheLastProvenBlockNumberFromDb() on its own schedule to keep the cache fresh. The scrape path should only read the cached value, never drive DB I/O.
```
That being said, I think the current implementation in the PR should be the safest approach which aligns the suggestion and is simple without introducing any anti-pattern

fluentcrafter: Why not moving the Poller into BatchesRepoBasedLastProvenBlockNumberProvider that calls `findAndCacheLastProvenBlockNumberFromDb`  and change `getLastProvenBlockNumber` the cached value ?

PR #2696, opened 2026-03-31
[Feat] - deploy signing UI

thedarkjester:
This PR implements issue(s) #

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> Changes touch deployment signing and verification execution paths (including a new local HTTP bridge and Hardhat subtask hook), which could affect release/deploy reliability if misconfigured or if the UI session lifecycle has edge cases.
> 
> **Overview**
> Adds an **opt-in browser-wallet signing flow** for Hardhat deploys/selected operational tasks via `HARDHAT_SIGNER_UI=true`, backed by a local Next.js app + loopback HTTP bridge, with an optional `EXPECTED_SIGNER_ADDRESS` guard and workflow/status updates surfaced to the UI.
> 
> Hardhat config now enforces **exclusive signer modes** (UI vs `DEPLOYER_PRIVATE_KEY`), resolves deployer accounts via a new helper, and keeps a **single UI session for chained `--tags` runs** by hooking `deploy:runDeploy`; most deploy scripts are updated to use `withSignerUiSession`/`getUiSigner` and provide richer UI context while waiting for tx receipts.
> 
> Contract verification is made **best-effort**: it runs in a separate Hardhat child process with a timeout, reports progress to the UI, and no longer blocks deployments indefinitely on explorer verification. Docs/templates/linters are updated and Next build artifacts for `contracts/signer-ui` are ignored.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 18ad4884ed4bba6b6ab63db9485cab2b3f8554ee. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/contracts/common/helpers/verifyContract.ts
+++ b/contracts/common/helpers/verifyContract.ts
@@ -43,43 +43,59 @@ async function runVerifyTaskWithTimeout(
     );
 
     let settled = false;
+    let timedOut = false;
+    let killTimer: ReturnType<typeof setTimeout> | undefined;
+
     const finish = (result: "succeeded" | "failed" | "timed_out") => {
       if (settled) {
         return;
       }
       settled = true;
-      clearTimeout(timeout);
+      clearTimeout(timeoutHandle);
+      clearTimeout(killTimer);
+      child.off("exit", onExit);
+      child.off("error", onError);
       resolveRun(result);
     };
 
-    const timeout = setTimeout(() => {
-      try {
-        child.kill("SIGTERM");
-      } catch {
-        /* child may already be gone */
+    const onExit = (code: number | null, signal: NodeJS.Signals | null) => {
+      if (timedOut) {
+        finish("timed_out");
+        return;
       }
-      setTimeout(() => {
-        try {
-          child.kill("SIGKILL");
-        } catch {
-          /* child may already be gone */
-        }
-      }, 1000).unref();
-      finish("timed_out");
-    }, VERIFY_TIMEOUT_MS);
-
-    child.once("exit", (code, signal) => {
       if (code === 0) {
         finish("succeeded");
         return;
       }
       console.log(`Verification process exited with code ${code ?? "null"}${signal ? ` (signal: ${signal})` : ""}.`);
       finish("failed");
-    });
-    child.once("error", (error) => {
+    };
+
+    const onError = (error: Error) => {
       console.log(`Error happened during verification: ${error}`);
       finish("failed");
-    });
+    };
+
+    const timeoutHandle = setTimeout(() => {
+      timedOut = true;
+      try {
+        child.kill("SIGTERM");
+      } catch {
+        /* child may already be gone */
+      }
+      // No .unref(): SIGKILL must fire even if the parent event loop would otherwise exit.
+      // finish("timed_out") is deferred to the exit handler so the child is reaped first.
+      killTimer = setTimeout(() => {
+        try {
+          child.kill("SIGKILL");
+        } catch {
+          /* child may already be gone */
+        }
+      }, 1000);
+    }, VERIFY_TIMEOUT_MS);
+
+    child.on("exit", onExit);
+    child.on("error", onError);
   });
 }
 
--- a/contracts/common/helpers/environment.ts
+++ b/contracts/common/helpers/environment.ts
@@ -33,7 +33,7 @@ export function getEnvVarOrDefault(envVar: string, defaultValue: unknown) {
 
 export function getOptionalEnvVar(name: string): string | undefined {
   const envValue = process.env[name];
-  if (!envValue) {
+  if (envValue === undefined) {
     return undefined;
   }
   return envValue;
--- a/contracts/hardhat.config.ts
+++ b/contracts/hardhat.config.ts
@@ -122,8 +122,11 @@ const config: HardhatUserConfig = {
       url: blockchainNode,
       accounts: deployerAccounts(),
       timeout: BLOCKCHAIN_TIMEOUT,
-      // No fixed chainId: docker L1 is 31648428 (docker/config/l1-node/el/genesis.json);
-      // hosted devnet (e.g. rpc.devnet.linea.build) uses 59139. Hardhat HH101 if config ≠ RPC.
+      // docker L1 is 31648428 (docker/config/l1-node/el/genesis.json);
+      // hosted devnet (e.g. rpc.devnet.linea.build) uses 59139.
+      // Set ZKEVM_DEV_CHAIN_ID to enforce chain-ID validation and prevent wrong-chain deployments.
+      // Omitting it disables validation (Hardhat HH101) and preserves flexibility across environments.
+      ...(process.env.ZKEVM_DEV_CHAIN_ID ? { chainId: parseInt(process.env.ZKEVM_DEV_CHAIN_ID, 10) } : {}),
     },
     l2: {
       url: l2BlockchainNode ?? "",


contracts/scripts/hardhat/signer-ui-bridge.ts
```diff
@@ -0,0 +1,1809 @@
...
+
+      let normalizedAddress: string;
+      try {
+        normalizedAddress = getAddress(payload.address);
+      } catch {
+        writeJson(response, 400, { error: "Invalid wallet address." });
+        return;
+      }
+
+      this.walletState = {
```
alainncls: [P1] This session is still re-bindable to a different wallet mid-run. `/api/wallet` overwrites `walletState` on every sync from the UI, `getAddress()` always returns the latest `walletState.address`, and `/api/respond` validates against that same mutable value. If the operator switches accounts during a deploy batch or an OpenZeppelin proxy deployment, later transactions can be signed by a different EOA than the one that started the session. The safer behavior is to bind the session to the first connected address and reject subsequent wallet/account changes until the run ends.

contracts/scripts/hardhat/signer-ui-bridge.ts
```diff
@@ -0,0 +1,1809 @@
...
+    ) {
+      try {
+        process.kill(parsed.pid, "SIGTERM");
+      } catch {
+        /* process does not exist or cannot be signaled */
+      }
+    }
+
+    if (hasKnownPort) {
+      killTcpListenersOnPort(parsed.port);
```
alainncls: [P2] This cleanup can kill an unrelated local service when `.next/dev/lock` is stale and the recorded port has already been reused. At this point we have only shown that the lockfile PID is not the active listener anymore; we have not shown that the current listener on `parsed.port` belongs to this signer UI. Calling `killTcpListenersOnPort(parsed.port)` here is therefore unsafe. It should either be skipped entirely or restricted to a process we can prove we own.

contracts/scripts/hardhat/signer-ui-bridge.ts
```diff
@@ -0,0 +1,1809 @@
...
+        return;
+      }
+
+      this.walletState = {
+        address: normalizedAddress,
+        chainId: payload.chainId,
+        connectedAt: new Date().toISOString(),
+      };
+
+      const resolvers = this.walletResolvers.splice(0);
```
VGau: **Session is re-bindable to a different wallet mid-run.**

`/api/wallet` unconditionally overwrites `this.walletState` on every poll from the UI. If the operator switches MetaMask accounts during a deploy batch (e.g. between deploying a proxy and calling its initializer), later transactions get signed by a **different EOA** than the one that started the session. This can leave OpenZeppelin proxies owned by the wrong address.

`SignerUiSigner.getAddress()` always returns the latest `walletState.address`, and `/api/respond` validates against that same mutable value, so the check passes even after a switch.

Suggested fix: add a `boundAddress` field to `SignerUiSession`. On the first successful `/api/wallet` call, lock it. On subsequent calls, reject address changes with a 409:

```ts
if (this.boundAddress && this.boundAddress !== normalizedAddress) {
  writeJson(response, 409, {
    error: `Session is bound to ${this.boundAddress}. `
         + `Disconnect and reconnect the original account, or restart the deploy.`,
  });
  return;
}

if (!this.boundAddress) {
  this.boundAddress = normalizedAddress;
}
```

ChainId updates should still be allowed (wallet may need to switch chain), but the address must stay pinned for the session lifetime.

contracts/common/helpers/verifyContract.ts
```diff
@@ -1,38 +1,144 @@
+import { spawn } from "node:child_process";
+import { resolve } from "node:path";
+import { clearUiWorkflowStatus, setUiWorkflowStatus } from "../../scripts/hardhat/signer-ui-bridge";
```
VGau: **Circular import: `verifyContract.ts` ↔ `signer-ui-bridge.ts`**

This static import creates a cycle:
```
verifyContract.ts → signer-ui-bridge.ts → ../../common → common/helpers/index.ts → verifyContract.ts
```

Under Node CJS resolution, whichever module loads second receives a partially-initialized export object from the first. Depending on module evaluation order, `clearUiWorkflowStatus` or `setUiWorkflowStatus` could be `undefined` at call time, causing a runtime crash during contract verification.

The rest of `signer-ui-bridge.ts` already uses `createRequire` / lazy loading to avoid exactly this class of issue.

Suggested fix — use a lazy `await import()` inside `verifyBestEffort` instead of the static import:

```ts
// Remove line 3 static import, then inside verifyBestEffort:
const { clearUiWorkflowStatus, setUiWorkflowStatus } = await import("../../scripts/hardhat/signer-ui-bridge");
```

Alternatively, extract these two helpers into a small standalone module (like `signer-mode.ts` was extracted) that neither file imports the other through.

contracts/scripts/hardhat/signer-ui-bridge.ts
```diff
@@ -0,0 +1,1809 @@
...
+  const { ethers: hhEthers } = hh;
+  const deployer = await resolveNamedDeployerAddress(hh.getNamedAccounts);
+  if (deployer) {
+    return await hhEthers.getSigner(deployer);
+  }
+  return await hhEthers.provider.getSigner();
+}
+
+/**
+ * Attach constructor/initializer context for the next browser-signed transaction only.
```
VGau, line 2077: **`resolveUiRunner` (no-arg) and `getUiSigner` have subtly different fallback signer resolution.**

`getUiSigner` (used by deploy scripts) resolves via `hre.getNamedAccounts()` → named deployer, falling back to `signers[0]`.

`resolveUiRunner` with no argument (used by `deployUpgradableFromFactory` and `deployUpgradableFromFactoryWithConstructorArgs` in `utils.ts`) also tries named deployer, but its last-resort fallback is `hhEthers.provider.getSigner()` instead of `signers[0]`.

In practice these usually resolve to the same account (both try named deployer first), but the inconsistency is a maintenance hazard — future readers will expect identical behavior.

Suggested fix: have the no-argument fallback delegate to `getUiSigner` directly:

```ts
// Replace the no-argument fallback block (lines ~1676-1686) with:
warnIfUsingPrivateKeySigning();
const hh = loadHardhatRuntime();
return await getUiSigner(hh as unknown as HardhatRuntimeEnvironment);
```

This eliminates the divergence and ensures a single code path for signer resolution.

contracts/common/helpers/verifyContract.ts
thedarkjester:
```
• This report is not valid as written for this repo.

  Why:

  1. The claimed cycle path includes ../../common -> common/helpers/index.ts, but in reality scripts/hardhat/signer-ui-bridge.ts:20 imports ../../common, which resolves to common.ts, not common/helpers/index.ts.
  2. common.ts only exports node URL helpers/constants and does not import helpers/index.ts or verifyContract.ts.
  3. So the actual edge is one-way:
     verifyContract.ts -> signer-ui-bridge.ts -> common.ts
     No ... -> verifyContract.ts return edge from that path.

  So there isn’t a demonstrated runtime circular-import crash from this specific chain.
  ```

contracts/scripts/hardhat/signer-ui-bridge.ts
thedarkjester: fixing this now

VGau (approved):

Issue #2654, opened 2026-03-25
Set linea-besu as a gradle module in the linea-monorepo gradle


jonesho:
🔐 🔐 `Should this be a security advisory instead?` 🔐 🔐

## Description
To make linea-besu a gradle module such that the other modules would be able to depend on it. The linea-besu module would contain all the building and publishing logic we have nowadays, and it would allow gradle to build besu everytime we run `./gradle build` 

## Motivation
This section should indicate what value we are receiving from it.

## Tasks
- [ ] Medium-level task that needs to be implemented for this feature request to be considered complete.

## Acceptance criteria
- [ ] List tests that need to be implemented or checks to consider this task completed. 

## Risks
- [ ] These are risks that may prevent completion of the task as well as areas to pay special attention to

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #2720, opened 2026-04-02
feat: set linea-besu as a gradle module

jonesho:
This PR implements issue(s) #2654
- Added the `linea-besu` module
- Added the `linea-besu:clean` task to remove the tmp besu folder and cached version file and cleanup maven local
- Added the `linea-besu:build` task to replace the uses of `buildAndUpdateBesuVersionInLibsVersions` (no need to manually call it anymore gradle should take care of it) 
- Added cached besu commit and version file to avoid always spawning a bash process to resolve the besu version if the besuCommit doesn't change
- Removed `besu` field in `libs.versions.toml` and the relevant git pre-commit hook
- Removed the `linea-besu` plugin and move the logics into `linea-besu` module

Caveats:
- for unpublished besu commit (besu version that is not published to any remote/local maven), it will take time to build and publish besu to mavenlocal
- for published besu commit, the gradle build time will be much less and is around the same as before the PR (checked with "gradle --scan" command)
- Previously discussed approach about using runtimeOnly dependencies for besu libraries to skip the use of besu-plugin-library seems not a viable approach:                                   
- I tried moving Besu libraries out of compileClasspath (via runtimeOnly instead of api/implementation/compileOnly), but hitting some walls. Two blockers:
  - Compilation breaks. tracer:plugins and besu-plugins:linea-sequencer import Besu types (org.hyperledger.besu.plugin.*, internals, etc.). The compiler needs them on compileClasspath. runtimeOnly = not on compileClasspath = cannot find symbol.
  - `jvm-libs/linea/besu-libs` is a java-library whose entire job is re-exporting Besu types on api. By definition those have to be on consumers' compileClasspath.
- `besu-plugin-library` and `besu-plugin-distribution` are highly encouraged by besu team for plugin implementation as they help avoid dependency conflicts by enforcing versions with the Besu BOM

 

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes Gradle dependency resolution and CI release workflows around Besu, which can break builds/publishing if version resolution or caching behaves unexpectedly.
> 
> **Overview**
> **Besu is now managed as a first-class Gradle module.** This adds a new `linea-besu` project that lazily resolves the Besu version from `besuCommit`, caches it in `linea-besu/.besu-resolved`, and (when needed) builds/publishes Besu artifacts before downstream compilation; `jvm-libs:linea:besu-libs` now depends on this module.
> 
> **Versioning and CI flows are updated accordingly.** The `besu` version entry and its pre-commit protection are removed from `gradle/libs.versions.toml`, build scripts/plugins wiring are deleted (`LineaBesuPlugin`), and tracer/sequencer plugin builds are updated to reference `resolveBesuVersion()` instead of a pinned catalog version. GitHub workflows/actions are adjusted to only build/publish Besu on `besuCommit` changes, derive `besu_version` from the built distribution, and commit refreshed `.besu-resolved` to PR branches when applicable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5ef86fb2e63a4c5a4a1dff8ba42a93467b53e7ac. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/actions/parse-release-versions/action.yml
```diff
@@ -27,7 +27,7 @@ runs:
         ARTIFACT_VERSION=linea-${{ inputs.commit_tag }}
         TRACER_VERSION=$ARTIFACT_VERSION
         SEQUENCER_VERSION=$ARTIFACT_VERSION
-        BESU_VERSION=$(grep -E '^\s*besu\s*=' gradle/libs.versions.toml | sed -n 's/.*"\([^"]*\)".*/\1/p')
+        BESU_VERSION=$(ls tmp/besu-eth/build/distributions/besu-*.tar.gz 2>/dev/null | head -1 | sed 's/.*besu-\(.*\)\.tar\.gz/\1/' || true)
```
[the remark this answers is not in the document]
jonesho: It's intended, the downstream steps would just be skipped.

linea-besu/build.gradle
```diff
@@ -0,0 +1,353 @@
...
+}
+
+tasks.named('clean').configure { t ->
+  def cachedResolvedFilePath = cachedResolvedFile.absolutePath
+  def rootPath = rootDirFile.absolutePath
+  def mavenLocalBasePath = mavenLocalBase
+  doLast {
+    [
+      new File(rootPath, 'tmp/besu-eth'),
+      new File(mavenLocalBasePath, 'besu'),
```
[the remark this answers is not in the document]
jonesho: No, its the correct maven local path

linea-besu/build.gradle
```diff
@@ -0,0 +1,353 @@
...
+        skipDownloadBesuDist,
+        resolvedVersion,
+        cloudsmithUser,
+        cloudsmithApiKey,
+        mavenLocalBase
+        )
+  }
+
+  return resolvedVersion
+}
```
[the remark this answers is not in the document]
jonesho: It won't because projects that don't apply `besu-plugin-library` will rely on the following to resolve besu:
```
if (!cfg.name.toLowerCase().contains('spotless')
          && !cfg.name.toLowerCase().contains('ktlint')
          && details.requested.group in [
            'org.hyperledger.besu',
            'org.hyperledger.besu.internal'
          ] &&
          (!details.requested.version || details.requested.version.trim().isEmpty())) {
        details.useVersion(rootProject.ext.resolveBesuVersion())
        details.because('Versionless Besu dependencies are aligned to the lazily resolved Besu version') // e.g. jvm-libs:linea:besu-libs
      }
```

.github/actions/linea-besu-package/build-plugins-and-assemble/action.yml
```diff
@@ -69,8 +69,12 @@ runs:
   steps:
     - name: Setup Tracer Environment
       uses: ./.github/actions/setup-tracer-environment
-      with:
-        skip-download-besu-dist: false
+    
+    - name: Build besu from source
+      shell: bash
+      run: ./gradlew linea-besu:build -PskipDownloadBesuDist=false
```
Filter94, line 75: But since LineaBesuPlugin is deleted, `PskipDownloadBesuDist` shouldn't be a thing now, right?

linea-besu-package/Makefile
```diff
@@ -10,20 +10,25 @@ LOCAL_SEQUENCER_DIST_FOLDER ?= $(abspath ../besu-plugins/linea-sequencer/sequenc
...
+clean-besu-and-maven-local:
+	cd ..; \
+	echo "Clean up tmp besu folder and maven local" ; \
+	./gradlew linea-besu:clean
+
+build-besu:
 	cd ..; \
 	echo "Build besu-eth/besu locally at BESU_COMMIT=$(BESU_COMMIT)" ; \
-	./gradlew buildAndUpdateBesuVersionInLibsVersions
+	./gradlew linea-besu:build -PskipDownloadBesuDist=false
```
Filter94, line 21: I think `skipDownloadBesuDist` shouldn't be needed anymore

docs/linea-besu-package-releases.md
```diff
@@ -66,8 +66,8 @@ This builds tracer and sequencer from source (and besu if needed), assembles the
 ### When the desired besu version needs to be updated
 
 1. Create a PR to update the `besuCommit` field in [libs.versions.toml](../gradle/libs.versions.toml) which denotes the desired commit hash tag from the [besu-eth/besu](https://github.com/besu-eth/besu) repo
-- The triggered Github workflows would automatically build the besu at the desired commit hash tag and publish it to cloudsmith [maven](https://cloudsmith.io/~consensys/repos/linea-besu/packages/) AND it would automatically push a commit to the PR to update the `besu` field in [libs.versions.toml](../gradle/libs.versions.toml)
-- To get the corresponding `besu` version from the besu commit hash before running any Github workflow, one can run `./gradlew buildAndUpdateBesuVersionInLibsVersions` locally which shall see the update in [libs.versions.toml](../gradle/libs.versions.toml) OR run Makefile command i.e. `cd linea-besu-package && make build-besu-local-and-update-lib-version`
+- The triggered Github workflows will automatically build Besu at the desired commit hash and publish it to Cloudsmith [maven](https://cloudsmith.io/~consensys/repos/linea-besu/packages/)
```
Filter94, line 69: Do we really need it? We should be able to leverage gradle cache to reuse Besu module's build artifacts

gradle/libs.versions.toml
```diff
@@ -8,8 +8,8 @@ dependencyLicenseReport = { id = "com.github.jk1.dependency-license-report", ver
 lombok = { id = "io.freefair.lombok", version = "9.2.0" }
 gradleVersions = { id = "com.github.ben-manes.versions", version = "0.51.0" }
 dependencyManagement = { id = "io.spring.dependency-management", version = "1.1.5" }
-besuPluginLibrary = { id = "net.consensys.besu-plugin-library", version = "0.1.6" }
-besuPluginDistribution = { id = "net.consensys.besu-plugin-distribution", version = "0.1.6" }
+besuPluginLibrary = { id = "net.consensys.besu-plugin-library", version = "0.2.1" }
+besuPluginDistribution = { id = "net.consensys.besu-plugin-distribution", version = "0.2.1" }
```
Filter94, line 12: Do we still need the Besu Plugin and the library?

linea-besu/.besu-resolved
```diff
@@ -0,0 +1,6 @@
+# Auto-managed cache of the (besuCommit -> resolved besuVersion) mapping.
+# Written by Gradle (:linea-besu:resolveBesuVersion) and refreshed by CI
+# (.github/workflows/linea-besu-build-and-publish.yml) on besuCommit bumps.
+# Do NOT edit by hand and do NOT commit local changes — this file is gitignored.
+besuCommit=52a8db8db07441c6f346e2cdd768354b962eefb1
```
Filter94: Why is commit hash here? As a cache key?

linea-besu/src/main/java/net/consensys/linea/besu/prepare/LineaBesuPrepareMarker.java
```diff
@@ -0,0 +1,25 @@
...
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package net.consensys.linea.besu.prepare;
+
+/**
+ * Placeholder so this module produces a jar. Building {@code :linea-besu} triggers {@code
```
Filter94: But we have besu-libs module, which produces a jar without having any source code. Maybe you should enforce linea minimal conventions?

linea-besu/build.gradle
```diff
@@ -0,0 +1,353 @@
...
+
+description = 'Ensures Hyperledger Besu artifacts for besuCommit in gradle/libs.versions.toml are built or fetched before dependent modules compile'
+
+def besuCommit = rootProject.libs.versions.besuCommit.get()
+def rootDirFile = rootProject.layout.projectDirectory.asFile
+// Single-file cache of (besuCommit, resolved besuVersion). Managed automatically by
+// the :linea-besu:resolveBesuVersion task and by CI (linea-besu-build-and-publish
+// workflow). Gitignored — developers should not edit or commit this file; CI force-adds
+// updates on besuCommit bumps.
+def cachedResolvedFile = new File(rootDirFile, 'linea-besu/.besu-resolved')
```
Filter94, line 55: Can we move it inside gradle cache or to the build artifacts, which is practically the same?

linea-besu/build.gradle
```diff
@@ -0,0 +1,353 @@
...
+    spec.environment('VERSION_LABEL', '')
+    spec.commandLine('bash', rootDir.absolutePath + '/linea-besu/scripts/checkout-and-resolve-version.sh')
+  }.standardOutput.asText.get()
+
+  def ver = outputText.trim().readLines().last()
+  writeCachedBesuVersion(cachedResolvedFile, besuCommit, ver)
+  return ver
+}
+
+static void buildAndPublishBesuAction(
```
Filter94, line 234: To me all the logic around maven publishing is useless if we leverage Gradle cache for the same purpose. It should also simplify the module

linea-besu/build.gradle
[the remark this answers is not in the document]
jonesho: Good point, make a in-memory cache for the resolved besu verison

.github/actions/linea-besu-package/build-plugins-and-assemble/action.yml
jonesho: No, LineaBesuPlugin is kept, it has the following purposes which is worth keeping:
  1. Auto-injects the Besu-provided catalog into compileOnly. Without the plugin, we'd have to list each coordinates by hand across the tracer and sequencer subprojects.
  2. Filters the installDist CopySpec to exclude anything Besu itself provides at runtime. This is what keeps the plugin distribution slim — only plugin-specific jars ship.                                                       
  3. Version alignment via merged catalog + enforcedPlatform(BOM) — keeps transitive deps (tuweni, jackson, etc.) on the exact versions Besu uses at runtime, so the plugin doesn't accidentally load a different tuweni-bytes than what Besu ships.

linea-besu-package/Makefile
jonesho: It's needed here because when building the linea-besu-package image, we need the distribution tar.gz file which we'd download it from maven instead of building locally for published besu version. For unpublished besu version, we'd have no choice but to build the distribution locally

docs/linea-besu-package-releases.md
jonesho: Yes, we need to publish the besu at `besuCommit` to maven, otherwise, ppl or CI who checkout the PR would needs to build the besu locally, I don't see how we could leverage gradle cache to reuse Besu module's build artifacts here?

gradle/libs.versions.toml
jonesho: Yes, as mentioned above

linea-besu/.besu-resolved
jonesho: Yes, it acts as a cache key, such that if it matches the one in libs.versions.toml, no resolution of besu version is needed locally (i.e. checkout the besu repo and compute the besu version), which would save a lot of besu repo checkout operations from CI.

linea-besu/build.gradle
jonesho: No, the purpose of this cached file is the following:
When a PR bumps `besuCommit`, the `linea-besu-build-and-publish` workflow does `git add --force linea-besu/.besu-resolved && git push`. That commit lands on the PR branch, and every downstream consumer who pulls would get the refreshed `commit→version mapping` for free — no one else has to run the checkout script. If the file lives in build/ or the Gradle cache, there's no mechanism to distribute the mapping: each dev's first build after the bump would re-clone Besu upstream and rerun `checkout-and-resolve-version.sh` (~seconds-to-minutes).

linea-besu/build.gradle
jonesho: The main reason it's there is for the unpublished-commit path: 
- when a dev bumps besuCommit to a SHA that hasn't been published to maven yet, we need to build Besu from source and publishToMavenLocal so Gradle's resolve can find the jars othewise the build will fail. Gradle's cache doesn't help here since there's no remote to fetch from.
- for published-commit path, this function would not be called at all

docs/linea-besu-package-releases.md
Filter94: The idea is to treat Linea Besu the same way other modules are treated. To use the prover client locally you need to build it first. But if it hasn't changed, it won't be rebuilt, it will be cached by Gradle. This can be done in another PR if ever. Maven publish has its own advantages

linea-besu/build.gradle
Filter94: We discussed that if we manage to find the the way to find Besu's version by commit hash without checking out the repo OR if we drop the Besu version prefix (X.Y.Z-...) for the maven version, we could get rid of caching the version inside the repo. Did you try searching for a way?

Filter94 (approved):

linea-besu/build.gradle
jonesho: Yes, but I couldn't find a straight-forward solution without the need to use any github APIs (which is not more efficient than git repo checkout approach, given that multiple calls with loop search are needed), the problem is tag reachability requires walking the commit graph backwards from a commit. That graph lives in the .git object store. Without checking outt, we cannot do ancestry checks. So instead, we improve the tag retrieval speed by doing shallow clone on cleaned empty `tmp/besu-eth` folder: 

`git clone --filter=tree:0 --no-checkout`

The .git savings come from not downloading blobs for every commit in Besu's history — only blobs reachable from the specific checkout we request are fetched, on demand.

We can merge it as is, and add a follow up ticket to remove the use of cached `.besu-resolved` if it didn't take too long to retrieve the besu vesion from fresh.

linea-besu/scripts/checkout-and-resolve-version.sh
```diff
@@ -5,14 +5,22 @@ set -e
...
-  cd "$BESU_DIR" && git checkout "$BESU_COMMIT"
+  # Partial clone: skip trees + blobs on the initial fetch so we only download the
+  # commit graph + tag refs. That's all `git describe --tags --abbrev=0 <commit>`
+  # below needs, keeping the version-only resolve fast for published commits. If a
+  # later step (build-dist-and-publish.sh) actually needs source files, 
+  # a subsequent git checkout will fetch the needed blobs on demand.
+  git clone --filter=tree:0 --no-checkout https://github.com/besu-eth/besu.git "$BESU_DIR"
+  if [ "$SKIP_CHECKOUT_AFTER_CLONE" != "true" ]; then
+    (cd "$BESU_DIR" && git checkout "$BESU_COMMIT")
+  fi
```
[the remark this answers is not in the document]
jonesho: That full checkout is to prepare for the build afterward

PR #2743, opened 2026-04-06
spec(rollup): implements the draft specification of the rollup.

AlexandreBelling:
This PR implements a draft spec for the rollup-validity relation after migrating to R5. It comprises a MD document specifying how the system is structured in sub-proofs to aggregate and a python spec detailing what the proof checks (regardless of the sub-proof composition structure).

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it introduces a new rollup-spec Python model that touches proof inputs/outputs, KZG/blob authentication, forced-transaction handling, and bridge message commitments, but it is largely a draft/spec implementation with multiple `NotImplementedError` placeholders and no runtime integration changes.
> 
> **Overview**
> Adds a new `rollup_spec/` draft specification for a Type-1 RISC-V Linea rollup, including a detailed design doc (`Readme.md`) describing execution/compression/join proof composition, shnarf chaining, DA blob contents, and forced-transaction (FTX) public inputs/continuity.
> 
> Introduces Python reference code that models the validity relation: blob authentication via KZG (`RollupDataWitness`), truncated DA block parsing/consistency checks, forced-transaction validation and rolling-hash tracking, a rollup verifier entrypoint (`check_rollup_validity`) that replays Osaka state transitions and extracts bridge logs, plus a modified state transition helper that returns `BlockOutput`.
> 
> Also updates `.gitignore` to ignore Python virtualenvs (`**/.venv`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4e7987e35446280a59d00cf0018ae6a1de7fc665. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rollup_spec/blob.py
```diff
@@ -0,0 +1,223 @@
...
+    def is_authenticated_blob_bytes(self) -> Tuple[bool, Hash32] :
+        """
+        is_authenticated_blob_bytes checks the KZG proofs and returns true if
+        it successfully validates that the blob_hash relates to the blob_data.
+
+        The function also returns the blob_hash.
+        """
+        blob_hash = self.blob_hash()
+        blob_kzg_x = keccak256(self.blob_bytes + blob_hash)
+        blob_poly = parse_as_bls12_381_fr_vec(self.blob_bytes)
```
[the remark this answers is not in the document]
AlexandreBelling: Will implement later. I don't trust an LLM to implement this inline as it will just create a ton more confusion if there is an error. The implementation should be specifically done separately and tested vs a reference implementations for that.

rollup_spec/Readme.md
```diff
@@ -0,0 +1,416 @@
...
+
+The migration to a Type-1 RISC-V zkEVM, targeting the Fusaka/Glamsterdam Ethereum fork, is an opportunity to fundamentally simplify this stack. A general-purpose RISC-V virtual machine can execute standard software directly as a provable program, eliminating the need for most bespoke circuits.
+
+### 1.2 Design Principles
+
+The architecture is driven by four principles:
+
+1. **Move intelligence from circuits to software.** Logic that today lives in custom Gnark/Vortex circuits moves into standard RISC-V guest programs written in Rust or C. This trades circuit complexity for software simplicity.
+2. **Use industry-standard primitives.** SNARK-friendly workarounds (MiMC hashing, custom LZSS) are replaced with standard algorithms (Keccak256, LZ4/zstd, BLS12-381) that can be compiled directly into the guest.
```
Filter94, line 19: Are LZ4/zstd zk friendly? If yes, why exactly we couldn't use them before?

rollup_spec/Readme.md
```diff
@@ -0,0 +1,416 @@
...
+- **Compression proofs** — for each submitted blob, a prover generates a compression proof attesting to correct decompression and KZG polynomial evaluation. 
+    - Compared to before, the compression proof is also tasked to aggregate its related execution proofs. The compression proof thus serves as the smallest unit of aggregation. It chains the execution proofs, binds the last one to the compression proof, and produces the unified 16-field public-input tuple.
+
+**Stream 3 — Aggregation.** Once all execution proofs and compression proofs for a target finalization range are available, they are assembled into a proof tree:
+
+1. **Join proofs** — applied recursively, each combining two adjacent-range proofs (compression or prior join proofs) by verifying them inside the guest and asserting continuity in software. The aggregation topology is flexible: the coordinator may use a balanced tree, a sequential left-fold, or any other shape, since the only constraint is that A's end block equals B's start block.
```
Filter94, line 37: > the coordinator

I thought we agreed to change this terminology to avoid the collision with the existing Coordinator

rollup_spec/Readme.md
```diff
@@ -0,0 +1,416 @@
...
+
+### 1.3 System Overview
+
+The system is organized around three concurrent, independent streams that converge at finalization time.
+
+**Stream 1 — Data Availability.** The sequencer compresses batches of L2 blocks and submits the resulting blob to L1. For each blob, the L1 contract anchors a new shnarf — a cumulative hash that chains the blob's content to the preceding history. The shnarf chain is the canonical on-chain record of submitted DA data.
+
+**Stream 2 — Proving (Execution & Compression).** Two proof types are generated independently and in parallel:
+
+- **Execution proofs** — for each range of L2 blocks, a prover generates an execution proof attesting to the EVM state transition. Multiple execution proofs can be produced in parallel across different block ranges.
```
Filter94, line 31: > for each range of L2 blocks

What is the range of blocks? How is it defined?

rollup_spec/Readme.md
```diff
@@ -0,0 +1,416 @@
...
+- **Compression proofs** — for each submitted blob, a prover generates a compression proof attesting to correct decompression and KZG polynomial evaluation. 
+    - Compared to before, the compression proof is also tasked to aggregate its related execution proofs. The compression proof thus serves as the smallest unit of aggregation. It chains the execution proofs, binds the last one to the compression proof, and produces the unified 16-field public-input tuple.
+
+**Stream 3 — Aggregation.** Once all execution proofs and compression proofs for a target finalization range are available, they are assembled into a proof tree:
```
Filter94, line 35: We shouldn't need execution proofs anymore, since compression proof is the smallest unit or aggregation

rollup_spec/Readme.md
```diff
@@ -0,0 +1,416 @@
...
+| Field | Description |
+|---|---|
+| `blobContent` | The raw compressed blob bytes (4096 × 32-byte EIP-4844 payload) |
+| `blobHash` | The blobHash of the blob as submitted on L1 |
+| `KzgProof` | The KZG proof for the blob |
+| `KzgY` | The KZG proof evaluation claim |
+| `blockData₁ … blockDatam` | The block data for all related blocks |
+| `E₁ … Eₙ` | The executions proofs, ordered by block range |
+| `PI_E₁ … PI_Eₙ` | The public-input tuple for the execution proofs |
+| `L2L1MsgList₁ … L2L1MsgListₙ ` | The list of the L2L1 message hashes for each execution proof |
```
Filter94, line 120: The resulting ones?

rollup_spec/Readme.md
```diff
@@ -0,0 +1,416 @@
...
+
+`lastBlockHash` anchors the shnarf to the execution history; `blobHash` anchors it to the DA blob. Because the KZG polynomial evaluation is proven inside the zkVM, the evaluation point `X` and claim `Y` never appear on-chain — the L1 contract only checks `blobHash` against the transaction's `VERSIONED_HASH`.
+
+### 3.2 Blob Payload
+
+The DA blob must contain the exact inputs required to re-execute the L2 blocks from the previous finalized state. Because the zk-proof guarantees transition validity, any data that is a deterministic *output* of execution can be stripped.
+
+**What is included:**
+
+- **Block context variables** — fields required to resolve EVM opcodes: `timestamp` (for `TIMESTAMP`), `mixHash`/`prevrandao` (for `PREVRANDAO`). Note: `gasLimit` and fields that are deterministic outputs of execution (such as `withdrawalsRoot`) are not included. `coinbase` (for `COINBASE`), `baseFee` are static configuration fields of the rollup.
```
Filter94, line 246: I don't see how `gasLImit` is deterministic. It depends on the target gas limit

rollup_spec/Readme.md
```diff
@@ -0,0 +1,416 @@
...
+- **Target block hash** — `blockHash` is included so that users can infer the return value of the keccak opcode without having access to the entire block header since a Type-1 block hash requires a perfect RLP-encoding of the full header
+- **Signature-stripped transactions** — the ordered transaction list including `from` (sender address), `nonce`, gas parameters, `to`, `value`, `data`, and `accessLists`; ECDSA signatures `(v, r, s)` are omitted since the execution proof guarantees they were validly signed. `from` is stored explicitly so the sender can be recovered without signature verification during re-execution. However, the transaction must figure their corresponding blob-hashes and access-lists.
+
+**What is stripped:**
+
+- ECDSA signatures `(v, r, s)`
+- Intermediate state roots and receipt roots — these are deterministic outputs of execution, not inputs to it. Note: the current shnarf formula includes `newStateRootHash` as an explicit input; in the new design (§3.1) it is replaced by `lastBlockHash`, which is an execution input rather than an output, so no state root ever appears on-chain.
```
Filter94, line 253: > so no state root ever appears on-chain.

Isn't it an issue for interoperability?

rollup_spec/Readme.md
```diff
@@ -0,0 +1,416 @@
...
+
+**Stream 1 — Data Availability.** The sequencer compresses batches of L2 blocks and submits the resulting blob to L1. For each blob, the L1 contract anchors a new shnarf — a cumulative hash that chains the blob's content to the preceding history. The shnarf chain is the canonical on-chain record of submitted DA data.
+
+**Stream 2 — Proving (Execution & Compression).** Two proof types are generated independently and in parallel:
+
+- **Execution proofs** — for each range of L2 blocks, a prover generates an execution proof attesting to the EVM state transition. Multiple execution proofs can be produced in parallel across different block ranges.
+- **Compression proofs** — for each submitted blob, a prover generates a compression proof attesting to correct decompression and KZG polynomial evaluation. 
+    - Compared to before, the compression proof is also tasked to aggregate its related execution proofs. The compression proof thus serves as the smallest unit of aggregation. It chains the execution proofs, binds the last one to the compression proof, and produces the unified 16-field public-input tuple.
```
FlorianHuc, line 33: If I understand correctly, it means that the compression proof depends on the inluded execution proofs. Is there some hierarchical aggregation, or a max number of execution proof?

FlorianHuc, line 33: Do we tackle within this change the case of blocks spanning over multiple blbos? Or do we keep it for later?

rollup_spec/Readme.md
```diff
@@ -0,0 +1,416 @@
...
+Exec₄  ─┘
+```
+
+---
+
+## 2. Proving System
+
+### 2.1 Execution Proof
+
+**WE HAVEN'T CLARIFIED HOW THE GUEST PROGRAM CAN TELL WHICH TRANSACTIONS ARE TO BE REFUSED. PRESUMABLE? JUST ADDING THAT INFORMATION FOR EACH FORCED TRANSACTION IN THE PRIVATE INPUTS OF THE PROGRAM IS LIKELY ENOUGH**
```
FlorianHuc, line 55: this should be inputs passed by the coordinator, or some helper program.

FlorianHuc: to me, the only cases left are invalid nonce and balance=0, and filtered addresses.

rollup_spec/Readme.md
```diff
@@ -0,0 +1,416 @@
...
+| `newLastBlockNumber` | Block number at the end of this range; required for the L1 contract to update `currentL2BlockNumber` and support liveness checks |
+| `finalStateRootHash` | Explicitly exposed to avoid L1 calldata hashing overhead for withdrawal resolution; the guest enforces it belongs to the header of `newLastBlockHash` |
+| `L2L1MessagesHash` | keccak256 of the ordered list of L2→L1 withdrawal message hashes emitted in this range; the number of messages is bounded per execution proof |
+| `prevL1L2BridgeRollingHash` | Accumulated L1→L2 deposit rolling hash at the start of this range; enables chaining across execution proofs and L1 continuity verification |
+| `prevL1L2BridgeRollingHashMessageNumber` | Message number corresponding to `prevL1L2BridgeRollingHash` |
+| `L1L2BridgeRollingHash` | Accumulated L1→L2 deposit rolling hash at the end of this range |
+| `L1L2BridgeRollingHashMessageNumber` | Message number corresponding to `L1L2BridgeRollingHash` |
+| `dynamicChainConfig` | Hash of the L2 `chainID, baseFee, coinBase` and `L2MessageServiceContract` |
+| `prevFtxRollingHash` | Forced-transaction rolling hash at the start of this range |
```
FlorianHuc, line 73: is it rolling hash of all ftxs, or (more likely) of all that were taken into consideration so far?

rollup_spec/Readme.md
AlexandreBelling: LZSS was zk-friendly. We want to replace it with something more standard

rollup_spec/Readme.md
AlexandreBelling: Not specified there. But not needed for the guest program part

rollup_spec/Readme.md
AlexandreBelling: Depends on whether you think proving all of that in go is too much or not. I am not sure it is that trivial

rollup_spec/Readme.md
Filter94: I get that, but how can we replace it with something that's not ZK friendly?

rollup_spec/Readme.md
AlexandreBelling: It says you if you know the previous blocks gas limit and usage and your genesis block, then you know the gas limit of the next block

rollup_spec/Readme.md
AlexandreBelling: Why would that be?

rollup_spec/Readme.md
AlexandreBelling: ?

rollup_spec/Readme.md
Filter94: AFAIK, currently there are some interop solutions that rely on the L2 state merkle proof and having the state root hash on L1 allows verifying it

@thedarkjester does my point make sense to you?

rollup_spec/Readme.md
Filter94: You do only if your target is fixed. But strictly speaking the target is configurable by the client. Currently it's fixed, but it still worries me with the horizon of 10 years ahead

rollup_spec/Readme.md
Filter94: Right, I thought they're rolling hashes at first. My question doesn't make sense

rollup_spec/Readme.md
thedarkjester: @Filter94 there are multiple ways to prove the state (stateRoot, blockHash etc), however, in the spec of EIP-7888, it is recommended to use the `blockHash` which is the intent of this.

```
Block hashes are the recommended state commitment, but state roots or other commitments may be used, such as batch hashes (since some rollups don’t commit single blocks, but batches of blocks instead).
```

Filter94: I get that for the execution proof we don't need the state root hash, but I'm still unclear whether it's a good idea to remove the state root hash from the contract

thedarkjester: the only known major consumer of the `stateroothash` at the moment is the `CCIP gateway` with the `linea_getProof` verification. As I understand it, this will be changed to use the `blockhash` as part of the proof. I say known because with the recent change to Shomei V3 and the proof format/Poseidon2 change, no other parties have indicated issues.

In terms of other consumers, we only have the rollup itself.

I am finding it difficult to justify keeping it with no consumers other than us atm.

rollup_spec/Readme.md
AlexandreBelling: Cause it's gonna be in R5. It was zk-friendly most due to the fact that the circuit was "easy" to write. But now, we don't need to maintain the circuit.

rollup_spec/Readme.md
AlexandreBelling: Yep, this is a Claude artefact. I did not pay attention. I changed that to "- in any chosen topology -"

rollup_spec/Readme.md
AlexandreBelling: But that's configurable in the genesis, right? It's not just a ad-hoc Besu config, is it?

rollup_spec/Readme.md
AlexandreBelling: Oh OK, then with this spec, we expose the stateroot (and implictly, the blockhash via the shnarf). We could make the exposition of the blockhash explicit in the public-input.

Regardless, the right place to look for this thing are the Join proof public-inputs, not the data-availability (which describes what we put in the block).

AlexandreBelling: Here are the public input description:

https://github.com/Consensys/linea-monorepo/pull/2743/changes#diff-88010f6610e1743fed2b4ab34325e4fed74e94419314676929ddc08b587cdf6fR207-R222

On the removal of the stateroot. We can remove it, we could also keep it. From my POV, it makes it more straightforward to lookup into the state of the L2 thanks to it

rollup_spec/Readme.md
Filter94: No, it's an ad hoc config on all the clients. `--miner.gaslimit` for Geth, for example and `--target-gas-limit` for Besu

rollup_spec/Readme.md
Filter94: Makes sense. Can't get used to it yet

rollup_spec/Readme.md
Filter94: I think this needs to be clarified, because it will impact the implementation and maybe delivery timelines quite a bit

rollup_spec/Readme.md
AlexandreBelling: Ok, I just checked. Good call @Filter94 

We should add this to the dynamic chain config (cc. @thedarkjester). The alternative is to include it in the block header in the DA.

rollup_spec/Readme.md
AlexandreBelling: > If I understand correctly, it means that the compression proof depends on the inluded execution proofs.

Yes

> Is there some hierarchical aggregation, or a max number of execution proof?

The way, I see it. We could have a big for-loop aggregating a non-predetermined number of execution proof.

AlexandreBelling: I did not intend to, to limit the complexity of the migration. That being said, once the migration is done this will become a lot more trivial change than before.

rollup_spec/Readme.md
AlexandreBelling: Actually, this comment is a leftover I had been missing. The python spec includes a proposal where the coordinator tells the prover for each forced transaction, whether it was accepted, rejected-because-from or rejected-because-to. This seemed to be the least breaking change compared to the current solution.

rollup_spec/Readme.md
AlexandreBelling: prevFtxRollingHash = all that were executed before the current block
newFtxRollingHash = all that have been execution at the end of the current block

rollup_spec/Readme.md
```diff
@@ -0,0 +1,416 @@
...
+   assert_eq!(PI_A.L1L2BridgeRollingHash,                  PI_B.prevL1L2BridgeRollingHash)
+   assert_eq!(PI_A.L1L2BridgeRollingHashMessageNumber,     PI_B.prevL1L2BridgeRollingHashMessageNumber)
+   assert_eq!(PI_A.chainID,                                PI_B.chainID)
+   assert_eq!(PI_A.coinbase,                               PI_B.coinbase)
+   assert_eq!(PI_A.baseFee,                                PI_B.baseFee)
+   assert_eq!(PI_A.newFtxRollingHash,                      PI_B.prevFtxRollingHash)
+   ```
+   Block-hash continuity is implicit in the shnarf check: `PI_A.newShnarf` encodes A's last block hash, and asserting `PI_A.newShnarf == PI_B.prevShnarf` is sufficient.
+
+3. **Merge the L2→L1 root lists.** Receive the ordered root arrays `roots_A` and `roots_B` as private witnesses. Verify each against its committed hash:
```
AlexandreBelling, line 188: Note that this approach does not scale very well. In the EEZ-style rollup architecture they propose to add a PCD (proof-carrying data) structure where each rollup finalization proof recursively verifies the previous one. This structure allows verifying the entire rollup history in a single proof.

Our "Joint-Proof" architecture is compatible with this but the L2L1 message-passing mechanism is not compatible with this architecture because we would not be able to merge all the L2L1 messages appearing in the entire rollup history.

It's not something that we have pushed so far but we could revisit this idea I think.

AlexandreBelling: With our current approach it is not that much of an issue because we don't have that many withdrawals to compute at once.

rollup_spec/Readme.md
```diff
@@ -0,0 +1,416 @@
...
+
+* **Enforce sequencer consensus rules**: timestamp sequentiality, base-fee, coin-base and , enforces Ethereum consensus rules (fork-choice, timestamps), 
+
+* **Extract the canonical L2L1Message** hashes from the block receipt and compute their flat-hash using keccak256.
+
+* **Extract the L1L2RollingHash** by checking a Merkle proof on the old and the new state. Both for the `L1L2RollingHash` and the `L1L2RollingHashNumber`
+
+* **Inspect the forced transactions**: See the corresponding section.
+
+* **Output**: the public-inputs as computed in the above steps.
```
Filter94, line 99: It would be nice to explicitly enumerate the outputs from the execution guest program and the compression one

Filter94 (approved):

fluentcrafter (approved):

Issue #2648, opened 2026-03-24
Besu Sequencer Plugin - Block/Transaction Validation should require explicit opt-in for block importing


fluentcrafter:
## Impact
__environment [All]__

### Steps to Reproduce
Linea Besu package when it has Plugins that should only apply to 
- A) sequencer's block building or 
- B) transaction pool validation when accepting tx through RPC or P2P

Is also applying those rules to block validation in follower nodes when importing a block. This should not be default behaviour. The consensus of the canonical chain was already reached by the validators and follower nodes should import it. Maybe can be added as an explicit opt-in, but should not as default behaviour. 

We could observe the impact of this during release of 7702 when we enabled on Sequencer  and forgot to enable in Archive nodes by mistake. The archive nodes stopped synching at the first block with a 7702 transaction because was failing in the Transaction validator. 

Steps to reproduce the bug:
1. set sequencer with `--plugin-linea-delegate-code-tx-enabled=true`
2. set l2-besu-node with `--plugin-linea-delegate-code-tx-enabled=false`
3. send 7702 transaction to sequencer, get it included and l2-besu-node will fail to sync,

PR #2825, opened 2026-04-15
feat!: #2648, Validation plugin restructuring

Filter94:
This PR implements issue(s) #2648

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> Changes how transaction type (EIP-4844/EIP-7702) enforcement is wired by adding new validators and making certain plugins mutually exclusive at startup, which could break node configurations that previously loaded both.
> 
> **Overview**
> Refactors transaction type validation by introducing shared `TransactionTypeValidation` logic and wiring it into two distinct enforcement paths: a new pool-level `TransactionTypeValidator` inside `LineaTransactionPoolValidatorPlugin`, and a renamed protocol-level `LineaBlockTransactionValidatorPlugin` for block import/selection.
> 
> Adds a hard startup guard to prevent `LineaTransactionPoolValidatorPlugin` and `LineaBlockTransactionValidatorPlugin` from being enabled together, updates acceptance tests to separately cover pool rejection vs block-import rejection for blob and EIP-7702 transactions, and refreshes plugin documentation accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 71b598a34978034292ab5febdf831f74f203dfab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fluentcrafter:
Hi @Filter94 what's the migration/release plan of this? IIUC, it seems to be breaking change?
```
if (LineaBlockTransactionValidatorPlugin.registered.get()) {
      throw new IllegalStateException(
          "Both LineaBlockTransactionValidatorPlugin and LineaTransactionPoolValidatorPlugin are"
              + " enabled. Only one should be active at a time since their transaction type"
              + " validation functionality overlaps. Use LineaTransactionPoolValidatorPlugin for"
              + " RPC/P2P nodes or LineaBlockTransactionValidatorPlugin for validator nodes.");
    }
 ```
Can we have a backwards compatible during some time with just a WARNING? 

> The former LineaTransactionValidatorPlugin is replaced by LineaBlockTransactionValidatorPlugin
Also, this renaming will break because of config `plugins=[...]`, can we still have hold `LineaTransactionValidatorPlugin` as kind of alias point to `LineaBlockTransactionValidatorPlugin` ?

If not, I suggest to then:
- keep the previous classes;
- create the new ones, versioned `LineaBlockTransactionValidatorPluginV1` & `LineaTransactionValidatorPluginV1` that need to be explicitly enabled on the configs instead?

WDYT?

Filter94:
> Can we have a backwards compatible during some time with just a WARNING?

It can be just a warning, but rename of the plugin is a backwards incompatible change. And if we go by the book, we have to go through the deprecation process of `LineaTransactionValidatorPlugin`. And to me the rename isn't worth it, because the impact is only the linea rollup operators

fluentcrafter:
> And to me the rename isn't worth it, because the impact is only the linea rollup operators

We have few external node runners, e.g Infura, hence I would be more cautious. 
How much of an effort would you figure?

@julien-marchand , @Othryades what's your take on this, how much of impact you anticipate for external runners?

Filter94:
> > And to me the rename isn't worth it, because the impact is only the linea rollup operators
> 
> We have few external node runners, e.g Infura, hence I would be more cautious. How much of an effort would you figure?
> 
> @julien-marchand , @Othryades what's your take on this, how much of impact you anticipate for external runners?

The rename impacts the plugin that wasn't ever supposed to be enabled by the regular node runners. Note that Besu profiles serve as configuration reference and `LineaTransactionValidatorPlugin` wasn't enabled there

fluentcrafter (approved):

PR #2873, opened 2026-04-21
[Chore] - extend timeouts to match scenarios

thedarkjester:
The following adjustments have been made:

1. Shomei get proof - the sum of the two serial calls did not match the total of the test - this should allow for more time for success 

2. Bridging analysis of logs showed it would succeed, but timed out due to L2 activity and Coordinator restarts

---

> [!NOTE]
> **Low Risk**
> Changes are confined to E2E tests and test utilities (timeouts, retries, and event query ranges) and should mainly affect test stability rather than production behavior.
> 
> **Overview**
> Reduces flakiness in E2E suites by scoping event waits to `fromBlock` captured just before bridging/messaging actions and by increasing several per-test/per-wait timeouts (notably token bridging and Shomei `linea_getProof`).
> 
> Updates deny-list test helpers to reload **both** sequencer plugins (`TransactionPoolValidator` and new `TransactionSelector`), add short file-sync and fixed “effect” delays around deny-list mutations, and adjusts unit tests accordingly.
> 
> Improves robustness of infra helpers: `expectBlockedTransaction` now handles both immediate RPC rejection and “accepted-but-never-mined” pool discards, and public clients / account funding now wrap `estimateFeesPerGas` with a `BlockNotFoundError` retry extension.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5fba4d465d39871476389aacc9cd976805703dee. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


e2e/src/common/test-helpers/transaction.ts
```diff
@@ -1,5 +1,59 @@
...
+ * Asserts that a transaction was blocked by the sequencer's deny list.
+ *
+ * Handles two sequencer behaviours:
+ * 1. Synchronous RPC rejection — `sendTransaction` rejects with a "blocked" error.
+ * 2. Async pool discard — `sendTransaction` returns a hash that is never mined.
+ *    Waits for BLOCKS_TO_WAIT blocks to be produced before checking, so a tx that
+ *    is merely pending is not mistaken for a blocked one.
+ */
+export async function expectBlockedTransaction(
+  client: BlockAndReceiptClient,
```
VGau:
```suggestion
  client: PublicClient,
```

e2e/src/common/test-helpers/transaction.ts
```diff
@@ -1,5 +1,59 @@
...
+    throw new Error(`expectBlockedTransaction: transaction ${hash} was mined but should have been blocked`);
+  } catch (error) {
+    if (error instanceof Error && error.message.includes("expectBlockedTransaction")) {
+      throw error;
+    }
+    const name = (error as { name?: string }).name ?? "";
+    if (name !== "TransactionReceiptNotFoundError") {
+      throw new Error(
+        `expectBlockedTransaction: unexpected error checking receipt for ${hash}: ${name} — ${error instanceof Error ? error.message : String(error)}`,
+      );
```
VGau:
```suggestion
    const e = error as GetTransactionReceiptErrorType;

    if (e.name !== "TransactionReceiptNotFoundError") {
      throw error;
    }
```

e2e/src/bridge-tokens.spec.ts
```diff
@@ -40,7 +41,9 @@ describe("Bridge ERC20 Tokens L1 -> L2 and L2 -> L1", () => {
 
       logger.debug("Minting and approving tokens to L1 TokenBridge");
 
-      const { maxPriorityFeePerGas, maxFeePerGas } = await l1PublicClient.estimateFeesPerGas();
+      const { maxPriorityFeePerGas, maxFeePerGas } = await withRetryOnBlockNotFound(() =>
```
fluentcrafter, line 44: I see `withRetryOnBlockNotFound` being called in a few places, which feels to me like violating DRY...

What are your thoughts on making the retry logic part of the client, not the Raw client, but a RetryAble client that does the retry logic and delegates to the raw client.

something in the lines of:
`context.l1PublicClient(retries: true, retriesTimeout: 5seconds)`

fluentcrafter (dismissed):
Just a made a small suggestion that can done after you agree with it

e2e/src/bridge-tokens.spec.ts
VGau: The viem client has a default retry count = 3 :) https://viem.sh/docs/clients/transports/http#retrycount-optional

thedarkjester: According to the AI - @VGau is this accurate?
```
Short answer: viem's default retries run at the HTTP transport layer and only cover HTTP-level failures. BlockNotFoundError is thrown at the action layer, after a successful HTTP 200 response. The default retry machinery never sees it.
```

VGau: Yes it makes sense. The retry is done only when the underlying http call failed:
- Network error (fetch fails)
- Request timeout
- HTTP 429 (rate limited)
- HTTP 500 or 5XX
- etc

The block not found error in viem is a functional error and not a http error.

alainncls (dismissed):

alainncls (approved):

PR #2894, opened 2026-04-22
experiment: Go to Risc5

ivokub:
This PR implements issue(s) #

### Checklist

  PR.


verifier-risc5/third_party/tinygo-riscv64im_zicclsm.patch
```diff
@@ -0,0 +1,388 @@
+diff --git a/builder/library.go b/builder/library.go
```
AlexandreBelling, line 1: What is this patch? What is it patching?

verifier-risc5/internal/workload/workload.go
```diff
@@ -0,0 +1,31 @@
+package workload
```
AlexandreBelling, line 1: Is that the go guest program that you compiled ?

verifier-risc5/third_party/tinygo-riscv64im_zicclsm.patch
ivokub: It does several things to patch the Go runtime and the RISC5 target when compiling with TinyGo:
* correctly passes memory shape -mcmodel=midany
* short-circuits defer keyword to avoid creating new goroutine (we omit the scheduler so cannot have goroutines)
* adds dummy precompile which can be called (i.e. we can do ECALL which we then trap in our fake-VM)
* creates conrete compilation target where we are more aggressive in disabling Go runtime (no GC etc.)
* completely disables interrupts (we don't have OS)
* adds a way to handle guest program exit status (which we can trap in our fake-VM to assert completion of program)

verifier-risc5/internal/workload/workload.go
ivokub: Not directly, this is only the library which is being called. The relevant files I'm building with Tinygo:
* verifier-risc5/cmd/verifier/main_baremetal.go -- entry point
* verifier-risc5/cmd/verifier/input_baremetal.go -- loading of the input to the guest program which is stored in fixed location in memory (given by my fake-VM)
* verifier-risc5/cmd/verifier/compute_precompile_baremetal.go -- the wrapper for the dummy computation. It aligns the inputs to what we expect in the fake-VM and then does ECALL (through a Tinygo linked implementation for calling fake-VM accelerators)
* verifier-risc5/cmd/verifier/announce_qemu_virt.go -- writes the output to a known location in the memory (separated for fake-VM and QEMU, we have dedicated space in fake-VM where it reads the computation status)

QEMU is as is, but for fake-VM I implemented simple libriscv based VM which allows to trap ECALLs, stores inputs in known location and later reads the status from known location. See verifier-risc5/cmd/verifier/announce_qemu_virt.go (it also has the linked implementation of fake-VM accelerator)

PR #2911, opened 2026-04-23
fix(coordinator): ftx conflation and fix metrics

fluentcrafter:
This PR implements issue(s) #

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> Touches core conflation/aggregation triggering and forced-transaction handling; incorrect start/resume block selection or deadline trigger wiring could change blob/aggregation boundaries and liveness, but changes are largely refactoring to shared factories with added tests/updates.
> 
> **Overview**
> **Refactors conflation + aggregation trigger construction into a single `CalculatorsFactory`**, replacing the older `ConflationCalculatorFactory`/`AggregationCalculatorFactory` and removing `DynamicBlockNumberSet` in favor of an internal mutable target-end-blocks set shared between blob and aggregation cutoffs.
> 
> **Fixes resume/starting baselines and related metrics** by separately tracking last conflated vs last aggregated blocks (`LastProcessedBlocks`) and using the correct block numbers/timestamps to initialize monitors, calculators, DB cleanup, and gauges (including switching L1-dependent “highest accepted” trackers to start from `lastFinalizedBlock`).
> 
> **Moves safe-block plumbing into `linea.conflation`** (including `FixedLaggingHeadSafeBlockProvider`) and updates deadline runners to use the timer abstraction (`TimerFactory`) instead of raw Vertx timers.
> 
> **Simplifies forced-transactions integration** by dropping the separate FTX-driven aggregation calculator/queueing path and only feeding FTX events into conflation triggering (plus minor queue sizing and error-message tweaks).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 726ee1df8949bd6ff902db2a761022512a7788de. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
```diff
@@ -182,14 +183,22 @@ class ConflationBacktestingApp(
...
+    aggregationTargetEndBlockNumbers = backtestingCoordinatorConfig.conflation.proofAggregation.targetEndBlocks?.toSet()
+      ?: emptySet(),
+    timerFactory = VertxTimerFactory(vertx),
+    safeBlockProvider = FixedLaggingHeadSafeBlockProvider(
+      ethApiBlockClient = l2EthClient,
+      blocksToFinalization = 0UL,
+    ),
     metricsFacade = metricsFacade,
-    log = log,
+    clock = this.clock,
```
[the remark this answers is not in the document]
fluentcrafter: @gauravahuja do we want conflation deadline in BackTesting App?

That may remove predictability...

coordinator/core/src/main/kotlin/linea/conflation/calculators/CalculatorsFactory.kt
```diff
@@ -0,0 +1,266 @@
...
+    aggregationDeadline: Duration? = null,
+    aggregationDeadlineCheckInterval: Duration? = null,
+    aggregationDeadlineNoL2ActivityTimeout: Duration? = null,
+    extraSyncCalculators: List<ConflationTriggerCalculator> = emptyList(),
+    timerFactory: TimerFactory,
+    safeBlockProvider: SafeBlockProvider,
+    clock: Clock,
+    log: Logger = LogManager.getLogger(GlobalBlockConflationCalculator::class.java),
+  ): ConflationCalculators {
+    val aggregationTargetEndBlockNumbersSet = DynamicBlockNumberSet(aggregationTargetEndBlockNumbers)
```
gauravahuja: This can just be called dynamicTargetEndBlockNumberSet because it is also used to ConflationCalculatorByTargetBlockNumber.

coordinator/core/src/main/kotlin/linea/conflation/calculators/CalculatorsFactory.kt
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    return ConflationCalculators(
+      blockConflationCalculator = blockConflationCalculator,
+      aggregationCalculator = aggCalc.aggregationCalculator,
+      service = service,
+    )
+  }
+
+  fun createAggregationCalculator(
```
gauravahuja: One of these createAggregationCalculator needs to be private otherwise it is very confusing to find which one is used where.

coordinator/core/src/main/kotlin/linea/conflation/calculators/CalculatorsFactory.kt
```diff
@@ -0,0 +1,266 @@
...
+      aggregationBlobLimit = aggregationBlobLimit,
+      aggregationSizeMultipleOf = aggregationSizeMultipleOf,
+      aggregationDeadline = aggregationDeadline,
+      aggregationDeadlineCheckInterval = aggregationDeadlineCheckInterval,
+      aggregationDeadlineNoL2ActivityTimeout = aggregationDeadlineNoL2ActivityTimeout,
+      safeBlockProvider = safeBlockProvider,
+      timerFactory = timerFactory,
+      metricsFacade = metricsFacade,
+      clock = clock,
+    )
```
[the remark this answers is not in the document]
gauravahuja: This seems important. We have
```kotlin
private val lastProcessedBlockNumber = resumeConflationFrom(
    aggregationsRepository,
    lastFinalizedBlock,
  ).get()
  private val lastConsecutiveAggregatedBlockNumber = resumeAggregationFrom(
    aggregationsRepository,
    lastFinalizedBlock,
  ).get()
```
One is supposed to be used for conflation and other for aggregation. But after the refactor I can't keep track of which is where.

coordinator/core/src/main/kotlin/linea/conflation/calculators/GlobalBlobAwareConflationCalculator.kt
```diff
@@ -28,7 +28,18 @@ class GlobalBlobAwareConflationCalculator(
   private val blobCalculator: ConflationTriggerCalculatorByDataCompressed,
   private val batchesLimit: UInt,
   metricsFacade: MetricsFacade,
-  private val dynamicBlockNumberSet: DynamicBlockNumberSet,
+  private val aggregationTargetEndBlocksDynamicSet: DynamicBlockNumberSet,
```
gauravahuja: Also commented elsewhere about the naming fo this set.

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/ConflationBacktestingApp.kt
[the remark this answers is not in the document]
gauravahuja: Its not needed

coordinator/core/src/main/kotlin/linea/conflation/calculators/CalculatorsFactory.kt
fluentcrafter: I disagree. Variables/Classes shall be named after their **why & what** (intent and content), not where they are used. That abstract name made the original code actually harder for me to understand it.

coordinator/core/src/main/kotlin/linea/conflation/calculators/GlobalBlobAwareConflationCalculator.kt
fluentcrafter: Replied before.

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/L1DependentApp.kt
```diff
@@ -343,7 +337,7 @@ class L1DependentApp(
     )
   }
 
-  private val highestAcceptedBlobTracker = HighestULongTracker(lastProcessedBlockNumber).also {
+  private val highestAcceptedBlobTracker = HighestULongTracker(lastFinalizedBlock).also {
```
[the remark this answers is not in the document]
fluentcrafter: The previous implementation was a hidden bug.

coordinator/core/src/main/kotlin/linea/conflation/calculators/CalculatorsFactory.kt
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    return GlobalAggregationCalculator(
+      lastBlockNumber = lastAggregatedBlockNumber,
+      syncAggregationTrigger = syncAggregationTriggerCalculators,
+      deferredAggregationTrigger = listOfNotNull(aggregationDeadlineCalculator),
+      metricsFacade = metricsFacade,
+      aggregationSizeMultipleOf = aggregationSizeMultipleOf,
+    )
+  }
```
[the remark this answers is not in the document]
fluentcrafter: >But the AggregationTriggerCalculatorByTargetBlockNumbers never consumes/removes used target block numbers from the shared set, meaning stale entries accumulate and could incorrectly trigger future aggregations at already-processed block numbers.

That's not an issue because the conflation only move forward,  so it won't come evaluate same blockNumber twice. 

What can be a problem - in theory - is the list growing as we add to it because an Hardfork Timestamp was reached. However in practice this is completely negligible...

coordinator/core/src/main/kotlin/linea/conflation/calculators/CalculatorsFactory.kt
gauravahuja: > why & what (intent and content
So this is used for both conflation and aggregation target end block numbers.

fluentcrafter: > So this is used for both conflation and aggregation target end block numbers.

What is "conflation target end block numbers", how are they configured with the coordinator?

coordinator/core/src/main/kotlin/linea/conflation/calculators/CalculatorsFactory.kt
```diff
@@ -0,0 +1,267 @@
...
+      metricsFacade = metricsFacade,
+    )
+
+    return AggregationCalculators(
+      aggregationCalculator = aggregationCalculator,
+      service = aggregationDeadlineCalculator ?: DisabledService("Aggregation Deadline"),
+    )
+  }
+
+  fun createAggregationCalculator(
```
gauravahuja: I still think having two methods with same name `createAggregationCalculator` is confusing. This method that creates a `GlobalAggregationCalculator` can be removed and the `GlobalAggregationCalculator` can directly be created in the `private fun createAggregationCalculator`. The only other place `fun createAggregationCalculator` is used is in the  test class `GlobalAggregationCalculatorTest` which should directly use the constructor to instantiate just like other calculator tests.

gauravahuja (dismissed):

coordinator/core/src/main/kotlin/linea/conflation/calculators/CalculatorsFactory.kt
fluentcrafter: > The only other place fun createAggregationCalculator is used is in the test class GlobalAggregationCalculatorTest which should directly use the constructor to instantiate just like other calculator tests.

In theory yes... in practice that approach leads to: 
- 1) duplication, 
- 2) hidden errors - **like we had on FTX** - why? Because the "wiring" that is used for the production code it's not the same that is used in the tests, leaving testing gaps. 

The clients of `linea.conflation.calculators` package should instantiate the calculators through the Factory. So, in fact the tests should rely on the factory as much as possible as well. Maybe, not unit tests but some of them like the GlobalAggregationCalculatorTest.

I can move
```kotlin
fun createAggregationCalculator(
    lastAggregatedBlockNumber: ULong,
    lastAggregatedTimestamp: Instant,
    aggregationTargetEndBlockNumbers: Set<ULong> = emptySet(),
    timestampBasedHardForks: List<Instant> = emptyList(),
    aggregationProofsLimit: UInt,
    aggregationBlobLimit: UInt?,
    aggregationSizeMultipleOf: UInt,
    aggregationDeadlineCalculator: DeferredAggregationTriggerCalculator? = null,
    metricsFacade: MetricsFacade,
  ): GlobalAggregationCalculator
```
inside `GlobalAggregationCalculator.Companion` object as a static method if that's better.

gauravahuja: in that case we should set the constructor as private or internal, but that is not necessary in this PR.

gauravahuja (approved):

Issue #2914, opened 2026-04-24
Port the log-derivative compiler to prover-ray


AlexandreBelling:
## Summary

Port the log-derivative compiler from the existing prover into the `prover-ray` framework.

## Context

The compilation pipeline converts memory-bus arguments and lookup constraints into log-derivative sums (steps 1–3 in the Vortex compilation pipeline). This compiler is a core step before the global/local constraint compilation and must be ported to move forward with a complete pipeline.

## Scope

- Port the log-derivative reduction for lookups and memory-bus arguments
- Group log-derivative terms by module as in the current implementation
- Integrate with the new dynamic-sized module support

## Dependencies

- Dynamic-sized module support (#2913)

PR #2928, opened 2026-04-24
Prover-ray: port the range checks

### LINKED ISSUE #2919: Re-add support for RangeCheck in prover-ray
## Summary

Re-add support for `RangeCheck` constraints in the `prover-ray` framework so the arithmetization can express bounded-value assertions.

## Reference

The equivalent compiler pass is implemented in `prover/protocol/compiler/specialqueries/range.go`. The compilation method is quite simple, there is not strong reason to change. We still want the compilation to be "lazy" when creating the precomputed columns in the sense that we don't want N copies of the same precomputed columns.

## Scope

- Implement or port the `RangeCheck` compiler pass
- Integrate it into the compilation pipeline at the appropriate stage
- Add tests covering both valid witnesses and soundness (invalid witness must fail proof generation)

AlexandreBelling:
Here is a PR for the port of the range-checks in the new ray framework. The handling mechanism is the same as the previous one. It has been tested against dynamically-size modules as well.

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> Introduces new constraint/query semantics and a compiler rewrite that affect proof-system soundness if bounds or table construction are incorrect, though changes are well-covered by new unit tests.
> 
> **Overview**
> Adds a new `wiop.RangeCheck` query type to assert column values lie in `[0,B)` (with runtime `Check`, constructor validation, and module storage).
> 
> Introduces a `rangecheck` compiler pass that reduces each unreduced `RangeCheck` into an Inclusion `TableRelation` against a shared precomputed range column per bound (sized to `NextPowerOfTwo(B)`), and marks queries reduced for idempotent compilation.
> 
> Adds unit tests for both the query and compiler pass (structural/idempotency + basic soundness), and registers a new `NewRangeCheckScenario` in `wioptest`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 615da761815a5da662c2dc0edb8a83ce6eb219bb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/prover-ray/wiop/compilers/rangecheck/rangecheck.go
+++ b/prover-ray/wiop/compilers/rangecheck/rangecheck.go
@@ -1,0 +1,75 @@
+// Package rangecheck implements the RangeCheck compiler pass for the wiop
+// protocol framework.
+//
+// It reduces every [wiop.RangeCheck] query to an [wiop.TableRelation] inclusion:
+// the checked column must be a subset of a precomputed column that enumerates
+// [0, B). A single precomputed range column is shared across all RangeChecks
+// with the same bound B, keeping the number of precomputed columns minimal.
+package rangecheck
+
+import (
+	"github.com/consensys/linea-monorepo/prover-ray/maths/koalabear/field"
+	"github.com/consensys/linea-monorepo/prover-ray/utils"
+	"github.com/consensys/linea-monorepo/prover-ray/wiop"
+)
+
+// Compile reduces all [wiop.RangeCheck] queries in sys to [wiop.TableRelation]
+// inclusion constraints against precomputed range tables.
+//
+// For each unique bound B, one new module of size NextPowerOfTwo(B) is created
+// with a precomputed column containing [0, 1, …, B-1, 0, …, 0]. Each
+// RangeCheck is then replaced by an Inclusion asserting that the checked
+// column is contained in the range column. Already-reduced queries are skipped.
+func Compile(sys *wiop.System) {
+	rangeColumns := make(map[int]*wiop.Column)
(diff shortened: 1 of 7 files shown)


ThomasPiellard:
looks good, but there are few things that I don't understand:
* in `checkInclusion`, you sample `alpha := field.RandomElemExt()` randomly, normally alpha must be sent by the verifier (so derived via FS). Is it because it is a test that you emulate FS ?
* where is the multiplicity column created ? Where is the logup column ? Can we see those columns in the trace ?
* why only 'TableRelation' is used is the test on not 'LogDerivativeSums' ? If there is a range check logDeriv sums must appear somewhere no ?

AlexandreBelling:
> in checkInclusion, you sample alpha := field.RandomElemExt() randomly, normally alpha must be sent by the verifier (so derived via FS). Is it because it is a test that you emulate FS ?

Yes, it is an optimization to let the verifier collapse rows into a single field element that we can fit and lookup into a hashmap.

> where is the multiplicity column created ? Where is the logup column ? Can we see those columns in the trace ?
> why only 'TableRelation' is used is the test on not 'LogDerivativeSums' ? If there is a range check logDeriv sums must appear somewhere no ?

This is not in the scope of this PR. The multipliciplicity column arises when converting a lookup into a log-derivative-sum; the current PR only does the **range-check** -> **lookup**. 

These are part of the issue #2914 ; but the present pr is for #2919

AlexandreBelling:
Fixes #2919

ThomasPiellard (approved):

ThomasPiellard (approved):

PR #3011, opened 2026-05-04
[ZkTracer] Add test cases for `blake.rs` implementation, single-opcode tests and used linker script

### LINKED ISSUE #2991: Add to `blake.rs` test cases
I wouldn't mind adding 2 more test vectors with `r = 0` (i.e. 0 rounds) and `f = 0, 1`

_Originally posted by @OlivierBBB in https://github.com/Consensys/linea-monorepo/pull/2839#discussion_r3173048270_

lorenzogentile404:
This PR implements issue(s) #

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes how example ELFs are linked/laid out and how `main.go` serializes program bytes (now including `.rodata`/`.data`), which can affect execution correctness across tests and tooling.
> 
> **Overview**
> Adds a generated linker script–based build flow for RISC-V examples (assembly/Zig/Rust), including new `linker-script`/`verify-elf` targets, updated disassembly output, and support for tests in subfolders.
> 
> Extends the ELF→JSON converter (`main.go`) to serialize a contiguous program image from `.text` + `.rodata` + `.data` (with `.bss` implied zero) instead of only `.text`, aligning inputs with the new memory layout.
> 
> Introduces new test coverage: standalone Blake2b EIP-152 vector tests (embedded and input-memory driven), a stack-load regression test, and an “atomic” single-opcode test generator + runner scripts; also tweaks zkVM execution traces (extra `ECALL` prints, force `registers[0]=0`, and marks `ram` as `@bipartite`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 88a1b0522462be7f686b52786c4d4a537c240778. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


amkCha:
Great, do you think you can add a command line example in the README to launch the blake_with_in_bytes with the values of let's say TEST_VECTOR_4 ? Good job working on this !

arithmetization/src/test/examples/rust/src/blake_with_in_bytes.rs
```diff
@@ -0,0 +1,77 @@
...
+            in(reg) code,
+            options(noreturn)
+        );
+    }
+}
+
+// required by the compiler
+#[panic_handler]
+fn panic(_: &core::panic::PanicInfo) -> ! {
+    exit(3);
```
DavePearce: I would recommend using an exit code like e.g. `i32::MAX` or something so that its distinct from other exit codes you might use when debugging.

arithmetization/src/test/examples/rust/src/blake_with_in_bytes.rs
```diff
@@ -0,0 +1,77 @@
...
+        let input = core::str::from_utf8_unchecked(core::slice::from_raw_parts(base, INPUT_LEN));
+        let expected = core::str::from_utf8_unchecked(core::slice::from_raw_parts(
+            base.add(INPUT_LEN),
+            OUTPUT_LEN,
+        ));
+        (input, expected)
+    }
+}
+
+fn exit(code: i32) -> ! {
```
DavePearce: You repeat this function in both this file and also `blake_with_in_embedded.rs`.  I would suggest make a standalone file (e.g. `std.rs` or something) which contains the "system functions" available to us (currently only `exit()`, but we might expect more).  Then, you can reuse this file across different tests and, eventually, it can become our "standard library" for the arithmetization under Rust.

arithmetization/src/test/examples/rust/src/blake_with_in_bytes.rs
```diff
@@ -0,0 +1,77 @@
...
+}
+
+const INPUT_LEN: usize = 426; // 213 bytes as hex
+const OUTPUT_LEN: usize = 128; // 64 bytes as hex
+
+extern "C" {
+    static _input_start: u8;
+}
+
+fn get_test_vector() -> (&'static str, &'static str) {
```
DavePearce: I see what you are doing here.  I would suggest implementing the function `read_inputs(&mut [u8]) -> usize` and putting it in a standard library file (e.g. `std.rs` --- see below).  You can then use this function in `get_test_vector()` to do the initial read before splitting out the `input` / `expected` strings.

arithmetization/src/test/examples/rust/src/blake_with_in_bytes.rs
DavePearce: This would then include `read_inputs()` and `write_output()`.

DavePearce (dismissed):
LGTM --- I left some comments though which you should consider, perhaps in a separate PR.

arithmetization/src/test/examples/rust/src/blake_with_in_bytes.rs
lorenzogentile404: Good idea.

arithmetization/src/main/riscv/interpreter.zkc
```diff
@@ -57,6 +57,50 @@ fn interpreter<ram, registers>(instruction:u32, pc:Address) -> (new_pc:Address)
...
+    registers[0] = 0
+
+    // Useful for debugging
+    // print_registers()
+
     printf "-----------\n"
+
+}
+
+fn print_registers<registers>() -> () {
```
DavePearce, line 70: I was wondering when someone was going to write this function!

arithmetization/src/test/examples/rust/src/blake_with_in_bytes.rs
```diff
@@ -0,0 +1,47 @@
...
+    exit(code);
+}
+
+fn get_test_vector() -> (&'static str, &'static str) {
+    const INPUT_LEN: usize = 426; // 213 bytes as hex
+    const OUTPUT_LEN: usize = 128; // 64 bytes as hex
+    const TOTAL_LEN: usize = INPUT_LEN + OUTPUT_LEN;
+    static mut BUF: [u8; TOTAL_LEN] = [0u8; TOTAL_LEN];
+    unsafe {
+        read_memory((&raw mut BUF) as *mut u8, &raw const _input_start as usize, TOTAL_LEN);
```
DavePearce: This is pretty messy.  I would rewrite to make it cleaner like so:

```
fn get_test_vector() -> (&'static str, &'static str) {
    ...
    static mut BUF: [u8; TOTAL_LEN] = [0u8; TOTAL_LEN];
    
    unsafe {
        read_memory(&mut BUF, TOTAL_LEN);
         ...
    }
}
```

and update `read_memory` to something like this:

```
pub fn read_memory(buf: &mut [u8], len: usize) {
    unsafe {
        let base = &raw const _input_start;
        for i in 0..len {
            buf[i] = *base.add(i);
        }
    }
}
```

Its must cleaner.  All the yucky `unsafe` stuff related to the inputs is hidden away inside `custom_std.rs`.  Otherwise, you are forcing the person using `read_memory()` to understand what actually is going on!

DavePearce (changes requested):
Hey,

Ok, a couple of minor changes requested.  Also (and maybe this is for another PR) but you can optimise the `memory` somewhat I think:

```
pub input riscV_program(address:Address) -> (bytes:u8)

pub input riscV_program_length(rv:u1) -> (length:Address)
pub input riscV_program_offset(i:u1) -> (address:Address)
pub input entry_point(i:u1) -> (address:Address)
```

The last three could be collapsed into a single memory like this:

```
pub input riscV_program_lengths(rv:u3) -> (length:Address)
```

Then `riscV_program_lengths[0]` could be `riscV_program_length`, and `riscV_program_lengths[1]` could be `riscV_program_offset`, etc.

Just means few modules are created, and with less constraints overall.

arithmetization/Makefile
```diff
@@ -11,7 +11,7 @@ REPO_ROOT := $(abspath $(MAKEFILE_DIR)/..)
 GO_CORSET ?= $(REPO_ROOT)/go-corset
 GO_CORSET_REPO ?= https://github.com/consensys/go-corset.git
 # Optional: pin a commit or branch (e.g. 763f30878e5fc260f0d88863c9883c904f8fc056). Leave empty for default branch.
-GO_CORSET_REF_FOR_ZKC ?=
+GO_CORSET_REF_FOR_ZKC ?= v1.2.14
```
amkCha, line 15: Why are you fixing a version for zkc ? For now I think it's better we use the regular updates on main

.gitignore
```diff
@@ -175,6 +175,7 @@ contracts/.claude/settings.local.json
 e2e/e2e-runtime-report.html
 e2e/e2e-runtime-report-summary.json
 .zig-cache/
+arithmetization/src/test/examples/linker_script.ld
```
amkCha, line 178: Why do we ignore the ld script ?

arithmetization/src/main/riscv/interpreter.zkc
```diff
@@ -57,6 +57,50 @@ fn interpreter<ram, registers>(instruction:u32, pc:Address) -> (new_pc:Address)
         }
     }
 
+    // Forcefully set registers[0] = zero back to 0
+    registers[0] = 0
+
+    // Useful for debugging
+    // print_registers()
```
amkCha, line 64: Nice

arithmetization/src/test/examples/asm/scripts/gen_op_tests.py
```diff
@@ -0,0 +1,348 @@
+#!/usr/bin/env python3
+"""Generate atomic single-opcode RV64IM tests for the zkc test harness.
```
amkCha, line 2: Where does this python come from ? Is it the link we found to test opcodes ?

lorenzogentile404: this is the python script that generated the assembly tests. It was generated by Claude.

amkCha: so you are not using this test base ? https://github.com/riscv/riscv-arch-test

arithmetization/Makefile
lorenzogentile404: The default one was not aligned with v1.2.14, leading to compilation issues on this branch, so this is a workaround until is fixed in go-corset.

.gitignore
lorenzogentile404: because it is re-regenerated every time for now, based on the parameters passed to the `Makefile`.

arithmetization/src/test/examples/asm/scripts/gen_op_tests.py
lorenzogentile404: not yet. This is the step I am working on.

lorenzogentile404: Claude tests were just a preliminary step to have something good enough in a quick way. This repository you mentioned requires a lot of configuration, so I should be able to run them this week (likely by end of tomorrow)

lorenzogentile404: Once they are running, we may even delete those artigianal ones.

amkCha (dismissed):

arithmetization/src/test/examples/asm/scripts/gen_op_tests.py
```diff
@@ -0,0 +1,348 @@
...
+
+emit("add",     rr_test("add",   5, 7, 12))
+emit("sub",     rr_test("sub",   12, 7, 5))
+emit("sll",     rr_test("sll",   1, 4, 16))
+emit("slt_neg", rr_test("slt", U64_NEG(-1), 0, 1))
+emit("slt_pos", rr_test("slt", 0, 1, 1))
+emit("sltu",    rr_test("sltu", U64_NEG(-1), 0, 0))
+emit("xor",     rr_test("xor", 0xff, 0x0f, 0xf0))
+emit("srl",     rr_test("srl", 16, 4, 1))
+emit("sra",     rr_test("sra", U64_NEG(-16), 1, U64_NEG(-8)))
```
OlivierBBB, line 117: What's the pattern here to determine which arguments you made positive, negative, zero etc ... ? Since we can make this programmatic I would suggest testing all combinations of signs (>0, =0, <0) for all input arguments of all inputs. This has a nontrivial impact on several instructions. Also, some instructions operate on `Words` rather than `DoubleWords` etc and having tests that involve what happens under truncation is necessary.

arithmetization/src/test/examples/asm/scripts/gen_op_tests.py
```diff
@@ -0,0 +1,348 @@
...
+
+emit("sh", STORE_PROLOGUE + """\
+    li t1, 0xabcd
+    sh t1, 0(t0)
+    lhu a2, 0(t0)
+    li a3, 0xabcd
+    sub a0, a2, a3
+""" + EPILOGUE)
+
+emit("sw", STORE_PROLOGUE + """\
```
OlivierBBB, line 329: The tests for `SW`, `SH`, `SB` don't account for truncations: inputs already have the correct size IIUC. Do these instructions perform sign extension on the truncated inputs ? Then this would have to be accounted for, too.

arithmetization/src/test/examples/asm/src/op/op_add.s
OlivierBBB, line 1: All of these files are generated, correct ? If so maybe we want to add everything under `op/.*\.s` to `.gitignore` and regenerate them with your script.

arithmetization/src/test/examples/Makefile
```diff
@@ -2,54 +2,94 @@
 # Note that Zicclsm extension does not affect the generated ELF so it is omitted.
 # Moreover, ABI being LP64 (soft-float) is relevant only for float numbers, which we do not use, so it can be omitted as well.
 
+# Declare all targets as phony to prevent conflicts with files of the same name
+.PHONY: exec debug compile zkc-exec zkc-debug clean clean-all linker-script verify-elf
```
OlivierBBB, line 6: What is this ?

OlivierBBB (changes requested):
LGTM but would request more systematic opcode tests in particular wrt input signs, input truncation and possible output sign extension.

arithmetization/src/test/examples/asm/src/op/op_add.s
lorenzogentile404: Yes.

arithmetization/src/test/examples/asm/scripts/gen_op_tests.py
lorenzogentile404: Those test can be thrown away potentially, once the [act4](https://github.com/riscv/riscv-arch-test) tests are completely set, so I do not think it is worth testing corner cases ourself.

arithmetization/src/test/examples/Makefile
lorenzogentile404: It is probably not necessary, but basically if the folder of the Makefile there is a file that has the same name of one of the targets, then the body of the target is not executed as it is intepreted as "the file already exists so do nothing". This tells to `make`  to never interpret targets as files to look for.

arithmetization/src/test/examples/asm/src/op/op_add.s
lorenzogentile404: Done.

arithmetization/src/test/examples/asm/scripts/gen_op_tests.py
lorenzogentile404: Same as above.

DavePearce (approved):
LGTM

arithmetization/Makefile
```diff
@@ -11,7 +11,7 @@ REPO_ROOT := $(abspath $(MAKEFILE_DIR)/..)
...
```
letypequividelespoubelles, line 14: Why do we have 2 corset version specified ?

lorenzogentile404: Temporary workaround that is going to be reverted in the next PR.

arithmetization/src/test/examples/asm/scripts/gen_op_tests.py
lorenzogentile404: As agreed, we keep both, but verify act4 cover corner cases we are interested.

OlivierBBB (approved):

PR #3020, opened 2026-05-05
[RISCV] fix: endianness inconsistency between `XXX_disassembled.elf` and `XXX.json`

### LINKED ISSUE #3019: [RISCV] endianness inconsistency
# Main issue

There is an inconsistency in the endianness conventions between
- `src/test/examples/rust/target/riscv64im-unknown-none-elf/release/XXX_disassembled.elf` and
- `src/test/examples/rust/target/riscv64im-unknown-none-elf/release/XXX.json` file

produced when running `zkcexec path/to/XXX.rs`. The same issue might plague the zig files, too, though I didn't check. In one example I find for the json

```
// the json
"riscV_program": "0x370180081301f1ff..."

// the disassembled elf
0000000000011318 <_start>:
_start():
   11318:	08800137          	lui	sp,0x8800
   1131c:	fff10113          	addi	sp,sp,-1 # 87fffff <memset+0x87edceb>
```

This is annoying since it makes going back and forth between both files hard.

# Fix
Claude has a suggestion for how to fix it:

```
The suggestion was to use llvm-objdump instead of riscv64-unknown-elf-objdump, which shows bytes in memory order (little-endian) by default, matching the JSON. It's available via the Rust toolchain as either:

llvm-objdump -d --line-numbers -S $(BIN) > $(BIN)_disassembled.elf

or:

rust-objdump -d --line-numbers -S $(BIN) > $(BIN)_disassembled.elf
```
This solution should work with all extensions:
```
❯ Will this work too when building a zig program or when building from an assembly file ? As the Makefile allows one to do ?

⏺ Yes, it would work for all three. Line 71 is unconditional — the same command runs on $(BIN) regardless of whether it came from .s, .zig, or .rs. Since all three produce a RISC-V ELF binary, llvm-objdump (which operates at the binary level, not the source level) handles them identically.

The only caveat: rust-objdump requires the Rust toolchain on PATH, so it's less suitable in a general environment. llvm-objdump is the safer choice since it's a standalone LLVM tool.
```

OlivierBBB:
We process the output of `objdump` with `sed` to reverse the 4 byte instructions to comply with endianness of the risc5 binary. This aligns it with the endianness found in the json.

Ex:
```
   11158: 08800137         lui    sp, 0x8800
```
becomes
```
   11158: 08800137 (37018008)         lui    sp, 0x8800
```

The first `08800137` is the original mnemonic produced by objdump. It can be fed into an online tool that decomposes risc5 instructions and be successfully decoded. (`37018008`) on the other hand is obtained by reversing the bytes of the previous string. One can find this 4 byte string in the json file.

---

> [!NOTE]
> **Low Risk**
> Low risk: changes are confined to the test/examples build tooling and documentation, affecting only how disassembly output is generated.
> 
> **Overview**
> Fixes an endianness mismatch between `<name>_disassembled.elf` and the generated `<name>.json` by post-processing `objdump` output to also display instruction bytes in memory order (including compressed 16-bit instructions).
> 
> The examples `Makefile` now auto-detects a suitable disassembler via a new `OBJDUMP` variable (preferring `llvm-objdump`/`rust-objdump`, with macOS Homebrew and GNU fallbacks) and documents the new requirement/option in `README.md`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 03c9e1ae35f564e8610846f0caf9451b60119e64. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


DavePearce (approved):
LGTM.  The `json` looks correct btw.  If you use `xxd -e -g1` you can see that the order of bytes is `37` first then `01`, etc.

arithmetization/src/test/examples/README.md
```diff
@@ -86,17 +87,18 @@ zkc-test <name>.zig ZIG_STRIP=false
...
+| `STRIP`           | `false`                                                                                 | Strip debug symbols from the ELF after compilation                                                                   |
+| `ZIG_STRIP`       | `true`                                                                                  | Strip when compiling Zig (reduces binary size), ignored for `.s` and `.rs`                                           |
+| `IN_BYTES`        | `""`                                                                                    | Input bytes written to memory at `IN_BYTES_OFFSET` before execution                                                  |
+| `PROGRAM_OFFSET`  | `0`                                                                                     | Memory offset where the program is loaded (up to 128 MB)                                                             |
+| `IN_BYTES_OFFSET` | `0x8000000`                                                                             | Memory offset where input bytes are written (up to 1 GB)                                                             |
```
lorenzogentile404, line 100: To make clear we did not forget a 0, we may write `0x08000000`.

arithmetization/src/test/examples/Makefile
```diff
@@ -27,6 +27,11 @@ STRIP ?= false
 # Ignored for .s and .rs files
 ZIG_STRIP ?= true
 
+# Disassembler: prefer llvm-objdump (shows bytes in memory order, matching the JSON format).
+# Falls back to the Homebrew keg-only path on macOS, then to riscv64-unknown-elf-objdump.
+# Override with: make OBJDUMP=/path/to/objdump
+OBJDUMP ?= $(or $(shell command -v llvm-objdump 2>/dev/null),$(shell command -v rust-objdump 2>/dev/null),$(shell test -x /opt/homebrew/opt/llvm/bin/llvm-objdump && echo /opt/homebrew/opt/llvm/bin/llvm-objdump),$(shell command -v riscv64-unknown-elf-objdump 2>/dev/null))
```
lorenzogentile404, line 33: `OBJDUMP :=` can be used if this never changes.

arithmetization/src/test/examples/Makefile
```diff
@@ -68,7 +73,20 @@ compile: $(SRC)
...
+  # the following regular expression maps
+  #
+  # <hex-pc>: <byte_1><byte_2><byte_3><byte_4><non hex><etc>
+  #
+  # to
+  #
+  # <hex-pc>: <byte_1><byte_2><byte_3><byte_4> (<byte_4><byte_3><byte_2><byte_1>)<non hex><etc>
+  #
+  # Claude also included the same type of operation for COMPRESSED instructions ... the 2nd -e expression
+	@$(OBJDUMP) -d --line-numbers -S $(BIN) \
```
lorenzogentile404, line 85: so here the 3 commands above are tried until one of them is succeed? Also, I would add a comment saying that the generated file is modified with `sed` command for debugging purposes or so, so it is very clear it is not the standard one.

lorenzogentile404 (changes requested):

arithmetization/src/test/examples/Makefile
lorenzogentile404: Also, I do not understand the comment about Claude. We do not expect COMPRESSED instructions to appear.

lorenzogentile404: Actually the description of this PR may be a commented in the `Makefile`, as it helps understanding what happens.

lorenzogentile404: `-d --line-numbers -S` those flags I believe were specific for `riscv64-unknown-elf-objdump`, now it seems they are applied in all cases. Does it work with assembly, rust and zig?

lorenzogentile404: Could `riscv64-unknown-elf-objdump` be used in all cases, with some adjustement using `sed` to make them consistent otherwise?

lorenzogentile404 (changes requested):
I believe this adds complexity to the script for a little advantage in terms of debugging. Unless really needed, I would not merge this one.

PR #3032, opened 2026-05-07
fix(coordinator): require all three proof flows to reach target before marking conflation backtesting complete

gauravahuja:
fix: track per-flow progress to determine conflation backtesting completion

Previously ConflationBacktestingApp.onConflationProgress was wired only to the
aggregation request handler and decided completion solely from aggregation's
end block. Because the execution, compression, and aggregation flows submit
proof requests asynchronously and independently (in particular, execution
proof request creation can lag arbitrarily on traces conflation API retries
while compression and aggregation proceed off the in-memory blob store), this
could mark backtesting complete while execution proof requests for the final
batches were still pending.

Replace the single completion flag with BacktestingProgressTracker, which
records the highest end block reached by each flow with atomic max-updates
and reports complete only when all three reach the configured target. The
tracker also serves as the LastProvenBlockNumberProviderSync for
BlockCreationMonitor, so its baseline is initialised to startBlockNumber - 1
to preserve the previous fetch-gap behaviour at startup.

This PR implements issue(s) #

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> Changes backtesting job completion and block-fetch gating to rely on new per-flow progress tracking; incorrect tracking could cause premature job shutdown or stalled block fetching.
> 
> **Overview**
> Fixes conflation backtesting completion logic by introducing `BacktestingProgressTracker`, which atomically tracks the highest end-block reached by **execution**, **compression**, and **aggregation** proof-request flows and reports completion only when *all three* reach the configured target.
> 
> `ConflationBacktestingApp` is rewired to record progress from each flow’s proof-request handler and to use the tracker’s execution baseline as `BlockCreationMonitor`’s last-proven provider, preserving startup fetch-gap behavior. Adds unit tests covering completion conditions, out-of-order callbacks, and the startup baseline regression guard.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6b7d5cb44e513be147e541b8f813d1032f99f664. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/BacktestingProgressTracker.kt
+++ b/coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/BacktestingProgressTracker.kt
@@ -1,0 +1,76 @@
+package net.consensys.zkevm.coordinator.app.conflationbacktesting
+
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.Logger
+import kotlin.concurrent.atomics.AtomicBoolean
+import kotlin.concurrent.atomics.AtomicLong
+import kotlin.concurrent.atomics.ExperimentalAtomicApi
+import kotlin.concurrent.atomics.fetchAndUpdate
+
+/**
+ * Tracks per-flow progress (execution, compression, aggregation) for a conflation backtesting job.
+ *
+ * Each flow submits its proof request asynchronously and independently:
+ *  - Execution proof request creation depends on a remote traces conflation API call with retries.
+ *  - Compression proof request creation depends only on blob creation in the conflation calculator.
+ *  - Aggregation proof request creation depends on compression proof requests being submitted.
+ *
+ * Because these flows run independently, an aggregation request for the target end block can be
+ * created while earlier execution proof requests are still retrying. Backtesting is complete only
+ * when every flow has reached [targetEndBlockNumber].
+ */
+@OptIn(ExperimentalAtomicApi::class)
+class BacktestingProgressTracker(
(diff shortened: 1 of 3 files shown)


coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/BacktestingProgressTracker.kt
```diff
@@ -0,0 +1,76 @@
...
+ *  - Execution proof request creation depends on a remote traces conflation API call with retries.
+ *  - Compression proof request creation depends only on blob creation in the conflation calculator.
+ *  - Aggregation proof request creation depends on compression proof requests being submitted.
+ *
+ * Because these flows run independently, an aggregation request for the target end block can be
+ * created while earlier execution proof requests are still retrying. Backtesting is complete only
+ * when every flow has reached [targetEndBlockNumber].
+ */
+@OptIn(ExperimentalAtomicApi::class)
+class BacktestingProgressTracker(
```
Filter94, line 23: Can you track the same with metrics?

Filter94 (approved):

coordinator/app/src/main/kotlin/net/consensys/zkevm/coordinator/app/conflationbacktesting/BacktestingProgressTracker.kt
gauravahuja: Metrics would mostly be for us to know the progress but this tracking also helps to stop the job and cleanup resources.

PR #3048, opened 2026-05-11
fix(contracts): remove token from url and value from logs

thedarkjester:
> [!NOTE]
> **Medium Risk**
> Changes how the local Hardhat signer UI authenticates to the loopback bridge (new bootstrap route and state tracking) and tweaks RPC transaction polling, so failures could break deploy signing flows even though scope is limited to local tooling.
> 
> **Overview**
> **Removes secret leakage from logs and URLs.** `getRequiredEnvVar` no longer prints environment variable values, and the Hardhat signer UI URL no longer includes the bearer `sessionToken` query param.
> 
> **Adds loopback session bootstrapping for the signer UI.** The bridge now exposes `POST /api/bootstrap` (origin-restricted) to return the session token once per session, and the Next.js UI calls `bootstrapSession()` to fetch/store it instead of relying on URL/sessionStorage secrets.
> 
> **Hardens bridge transaction lookup and error handling.** RPC `getTransaction` polling is wrapped with per-attempt timeouts, sleep intervals respect remaining time, and `/api/respond` now detects cancelled pending requests and produces clearer failure messaging when a tx never appears on the configured RPC.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aea1d4016cbc5765a5de85359ea094cccf82d33f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/contracts/scripts/hardhat/signer-ui-bridge.ts
+++ b/contracts/scripts/hardhat/signer-ui-bridge.ts
@@ -768,12 +768,18 @@ async function getTransactionWithRetry(
   intervalMs: number,
 ): Promise<TransactionResponse | null> {
   const startedAt = Date.now();
-  while (Date.now() - startedAt < timeoutMs) {
-    const tx = await getTransactionWithTimeout(provider, hash, Math.min(TX_LOOKUP_RPC_ATTEMPT_TIMEOUT_MS, timeoutMs));
+  let remainingMs = timeoutMs;
+  while (remainingMs > 0) {
+    const tx = await getTransactionWithTimeout(provider, hash, Math.min(TX_LOOKUP_RPC_ATTEMPT_TIMEOUT_MS, remainingMs));
     if (tx) {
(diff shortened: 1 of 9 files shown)


contracts/scripts/hardhat/signer-ui-bridge.ts
```diff
@@ -1646,6 +1681,22 @@ class SignerUiSession {
     writeJson(response, 404, { error: "Not found" });
   }
 
+  private handleBootstrapRoute(request: IncomingMessage, response: ServerResponse): void {
```
alainncls, line 1690: Do not rely on `Origin` alone before returning `sessionSecret`: local processes can spoof this header. Require a one-time bootstrap nonce or another unspoofable local challenge.

alainncls (approved):

PR #3092, opened 2026-05-13
perf(prover-ray): Switch U_alpha from evaluation form to monomial coefficient form 

YaoJGalteland:
Switch the Vortex opening-proof linear combination U_alpha from evaluation form (T × blowup elements) to monomial coefficient form (T elements), shrinking the linear-combination payload by a factor of `blowup` (e.g. 16x at blowup=16). This mirrors the optimization previously landed in prover/... under #2631 / 6b4141d7d, but lives entirely in the prover-ray crypto layer. 

Perfomance: 
By integrating  #2631 / 6b4141d7d to the mainnet, it reduces the prover's internal wizard proof by ~3×, which shows up as an 8% smaller outer circuit (86M-> 79M constraints) and 27% faster setup (1h27m->1h3m on r7i.48xlarge), 65.6% smaller Outer Plonk witness size (3.7M->1.3M), runtime-neutral end-to-end (within 1%)

Changes:

1. reedsolomon.go adds two helpers:
   - LCEvalsToCoefficients: T-evals (small domain) -> T monomial coeffs via IFFT_T + bit-reverse. Used by the prover after computing the linear combination over the T-length source polynomials.
   - ExtCoefficientsToAllEvaluations: T coeffs -> N RS-domain evaluations via DIT FFT_N. Used by the verifier to look up U_alpha at the opened column indices.

2. vortex/prover.go's Prove signature now takes (*Params, polys) so it can compute the linear combination over T-length source polys (cheaper than over N-length encoded matrices) and immediately convert to coefficients via LCEvalsToCoefficients.

3. vortex/verifier_common.go switches to coefficient-mode checks:
   - CheckStatement evaluates U_alpha at x via Horner on coefficients (replacing the Lagrange interpolation on T evaluations).
   - CheckLinComb expands the T coefficients to all N evaluations once via ExtCoefficientsToAllEvaluations, then looks up the relevant U_alpha(omega^selectedColID) per opened column.
   - CheckIsCodeWord is dropped: any T coefficients describe a unique degree-<T polynomial, so the codeword property is implicit and no separate RS check is needed.

4. vortex/verifier.go drops the CheckIsCodeWord call from VerifyCommon and threads params.RsParams into the new CheckLinComb signature.

5. verifier_test.go updates the Prove call to the new signature.

TestVerifier and the reedsolomon tests pass.

This PR implements issue(s) #

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> Changes the prover/verifier protocol representation for the opening proof’s linear combination and rewires verification checks, so subtle mismatches in FFT/IFFT conventions or domain sizing could cause invalid proofs to verify/fail.
> 
> **Overview**
> **Switches Vortex opening-proof linear combination `U_alpha` from evaluation form to monomial-coefficient form** to shrink proof payload and avoid operating on blowup-sized vectors.
> 
> The prover now computes the linear combination from the *T-length source polynomials* and converts evaluations→coefficients via a new `polynomials.LCEvalsToCoefficients` (IFFT+bit-reverse); `Prove`’s signature is updated accordingly. The verifier is updated to validate in coefficient mode: it evaluates `U_alpha(x)` via `EvalCanonical`, expands coefficients to RS-domain evaluations via new `RsParams.EncodeFromMonomials`, removes the explicit Reed–Solomon codeword check, and adds stricter input/shape validation and better error propagation for Merkle/SIS inclusion checks. Tests are extended to cover the new FFT-based encoding path and multiple inverse rates (including `16`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e38b4398f444f5aa55b5598c32a693e4caa510ff. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


prover-ray/crypto/koalabear/reedsolomon/reedsolomon.go
```diff
@@ -195,3 +195,40 @@ func (r *RsParams) IsCodewordExt(v []field.Ext) error {
 
 	return nil
 }
+
+// LCEvalsToCoefficients takes a length-T vector of evaluations of a polynomial
+// over the small domain (T = NbColumns()) and returns its T monomial
+// coefficients. This is used to send the Vortex linear combination U_alpha in
+// coefficient form (T elements) rather than evaluation form (T × blowup
+// elements), shrinking the proof by a factor of `blowup`.
+func (r *RsParams) LCEvalsToCoefficients(v []field.Ext) []field.Ext {
```
AlexandreBelling: The function, (the package) and the docs are leaky; they should not talk about Vortex (the package is for Reed-Solomon which exists independently from Vortex. Beside, IIUC the function does not have any relation with RS directly; it's just Lagrange to Monomial conversion.

I'd suggest,
* Moving the function in the maths directory close to either FFT or polynomials
* Renaming it with a name that does not echo Vortex

prover-ray/crypto/koalabear/vortex/verifier_common.go
```diff
@@ -33,12 +32,14 @@ type VerifierInput struct {
...
+// CheckStatement verifies that the linear combination U_alpha — given as T
+// monomial coefficients — is consistent with the claimed evaluations ys at
+// point x. The check is U_alpha(x) == ∑ᵢ αⁱ · yJoinedᵢ where U_alpha is
+// evaluated by Horner on its coefficients.
+func CheckStatement(coefficients []field.Ext, ys [][]field.Ext, x, alpha field.Ext) error {
 
-	// Check the consistency of Ys and proof.Linear combination
 	yJoined := slices.Concat(ys...)
-	alphaY := polynomials.EvalLagrange(field.VecFromExt(linComb), field.ElemFromExt(x))
+	alphaY := vortex.EvalFextPolyHorner(coefficients, x)
```
AlexandreBelling: That also should not live in the Vortex scope. It's a helper function that belongs in the maths package. Otherwise, we will end up with 50 different implementation of the same function

prover-ray/crypto/koalabear/reedsolomon/reedsolomon.go
```diff
@@ -195,3 +195,40 @@ func (r *RsParams) IsCodewordExt(v []field.Ext) error {
...
+	return coeffs
+}
+
+// ExtCoefficientsToAllEvaluations evaluates the degree-<T polynomial given by T
+// monomial coefficients at all N = NbEncodedColumns() points of the RS
+// domain (ω_N^0, ω_N^1, ..., ω_N^{N-1}). The returned slice has length N. This
+// is the verifier-side counterpart of LCEvalsToCoefficients: it expands the
+// T coefficients back to all N evaluations so the verifier can look up
+// U_alpha(ω_N^selectedColID).
+func (r *RsParams) ExtCoefficientsToAllEvaluations(coefficients []field.Ext) []field.Ext {
```
AlexandreBelling: Rename it, EncodeFromMonomials

AlexandreBelling (dismissed):

prover-ray/crypto/koalabear/reedsolomon/reedsolomon.go
```diff
@@ -195,3 +195,21 @@ func (r *RsParams) IsCodewordExt(v []field.Ext) error {
...
+	t := r.NbColumns()
+	n := r.NbEncodedColumns()
+	if len(coefficients) != t {
+		panic(fmt.Sprintf("EncodeFromMonomials: expected %d coeffs got %d", t, len(coefficients)))
+	}
+	buf := make([]field.Ext, n)
+	copy(buf, coefficients)
+	// DIT FFT expects bit-reversed input; natural-order evaluations come out.
+	utils.BitReverse(buf)
+	r.Domains[1].FFTExt(buf, fft.DIT, fft.WithNbTasks(1))
```
ivokub: Why `fft.WithNbTasks(1)`? When I look at the callgraph it doesn't seem that this method is called in parallel context so that we would want to force NbTasks=1 here.

prover-ray/maths/koalabear/polynomials/canonical.go
```diff
@@ -29,6 +33,22 @@ func EvalCanonical(poly field.Vec, z field.Gen) field.Gen {
...
+// ω a primitive n-th root of unity), via p[j] = (1/n)·Σᵢ v[i]·ω⁻ⁱʲ, implemented as
+// a single IFFT_n + bit-reverse over the extension field.
+func LCEvalsToCoefficients(d *fft.Domain, v []field.Ext) []field.Ext {
+	n := int(d.Cardinality)
+	if len(v) != n {
+		panic(fmt.Sprintf("LCEvalsToCoefficients: expected %d evals got %d", n, len(v)))
+	}
+	coeffs := make([]field.Ext, n)
+	copy(coeffs, v)
+	d.FFTInverseExt(coeffs, fft.DIF, fft.WithNbTasks(1))
```
ivokub: Similarly why we force `fft.WithNbTasks(1)` here?

prover-ray/crypto/koalabear/vortex/prover.go
```diff
@@ -3,22 +3,31 @@ package vortex
...
+	polys [][][]field.Element,
 	encodedMatrices []EncodedMatrix,
 	trees []*smt.Tree, alpha field.Ext) (*OpeningProof, [][]smt.Proof) {
 
 	proof := &OpeningProof{}
 
-	_encodedMatrices := make([][]field.Element, 0, len(encodedMatrices))
-	for _, m := range encodedMatrices {
-		_encodedMatrices = append(_encodedMatrices, m...)
+	allPolys := make([][]field.Element, 0)
```
ivokub: We could loop over polys before to get the full length for allocation and then just `copy`.

prover-ray/crypto/koalabear/vortex/prover.go
```diff
@@ -3,22 +3,31 @@ package vortex
...
+	for _, pl := range polys {
+		allPolys = append(allPolys, pl...)
 	}
-	LinearCombination(proof, _encodedMatrices, alpha)
+	LinearCombination(proof, allPolys, alpha)
+	proof.LinearCombination = polynomials.LCEvalsToCoefficients(params.RsParams.Domains[0], proof.LinearCombination)
```
ivokub: Why not do `LCEvalsToCoefficients` directly in LinearCombination? Right now imo it is a bit fragile as at some point we hold `proof.LinearCombination` as eval form and then later in coefficient form. Right now it works, but I think later if anyone does some changes or reorders something we could start having mismatch.

ivokub (commented):
Looks good, though I have a few questions about fft.WithNbTasks and a future-proofing implementation doing the EvalToCoefficient directly inside LinearCombination.

The new method are not covered by tests as well, I think it would be a good to have for test coverage for different rates.

---

I think we're right now also quite genereous with allocations as well. For example in prover-ray/crypto/koalabear/vortex/prover_common.go we allocate `linCombs` before the parallel context and then inside every goroutine again `localLinComb`. But we could essentially just use `localLinComb = linComb[start:stop]`.

And not in this PR, but I think we should have an overhaul later to also provide destinations in methods which could possibly allocate. For example right now we have:
```go
func SomeMethod(inputs []field.Ext) []field.Ext {
    output := make([]field.Ext, len(inputs))
    // .. do something
   return output
}
```
this means that we always allocate memory. But we could instead do:
```go
func SomeMethod(inputs []field.Ext, output []field.Ext) []field.Ext {
    if cap(output) < len(inputs) {
        output = make([]field.Ext, 0, len(inputs))
    }
    output = output[0: len(inputs)]
    // do something
    return output // we still need to return it in case we allocated new if the dst was small
}
```
now we could use
```go
scratch := make([]field.Ext, 10)
dst = SomeMethod(inputs, scratch)
// use it, but we don't need to store dst anymore so can reuse it
dst = SomeMethod(inputs, dst) // no allocations here anymore!
```
and if we want to create something new
```go
dst = SomeMethod(inputs, nil) // allocates new
```

But lets keep it into another perf-optimized PR for now.

---

In this PR I think we could have more defensive programming as well:

verifier.go (line 54), inside Verify, before VerifyCommon:
* len(proof.Columns) == len(commitments)
* len(merkleProofs) == len(commitments)
* if WithSis != nil, len(WithSis) == len(commitments)
* if defaulting WithSis, size it from len(commitments), not len(merkleProofs)

verifier.go (line 78), inside VerifyCommon, after the linear-combination length check:
* reject len(vi.EntryList) == 0
* check every selectedColID is 0 <= selectedColID < params.RsParams.NbEncodedColumns()
* check len(vi.Ys) == len(proof.Columns)
* for each commitment i, check len(proof.Columns[i]) == len(vi.EntryList)
* for each opened column proof.Columns[i][j], check len(proof.Columns[i][j]) == len(vi.Ys[i])




verifier.go (line 20), inside CheckColumnInclusion:
* defensively verify len(columns) == len(commitments) == len(merkleProofs) == len(WithSis)
* for each i, verify len(columns[i]) == len(merkleProofs[i])
* replace the panic(err) at verifier.go (line 27) with return fmt.Errorf(...)

prover-ray/crypto/koalabear/vortex/verifier.go
```diff
@@ -49,43 +64,66 @@ func CheckColumnInclusion(sis *ringsis.Key, columns [][][]field.Element,
...
+			if got, want := len(proof.Columns[i][j]), len(vi.Ys[i]); got != want {
+				return fmt.Errorf("proof.Columns[%d][%d] length %d does not match vi.Ys[%d] length %d", i, j, got, i, want)
+			}
+		}
 	}
 
-	err = CheckStatement(proof.LinearCombination, vi.Ys, vi.X, vi.Alpha)
-	if err != nil {
+	if err := CheckLinComb(params.RsParams, proof.LinearCombination, vi.EntryList, vi.Alpha, proof.Columns); err != nil {
 		return err
```
ivokub, line 125: For the future, if there are multiple methods returning errors and you return them directly, then we cannot really distinguish where the error came from. Better to wrap:
```go
return fmt.Errorf("CheckLinComb failed: %w", err)
```

Then if there is an error we get a more descriptive
```
"CheckLinComb failed: [inner reason]"
"CheckStatement failed: [inner reason]"
```

ivokub (approved):
Thanks for the fixes!

prover-ray/crypto/koalabear/vortex/verifier.go
ivokub: But for now lets keep it as is, no need to change. The errors returned in the functions are distinct anyway.

AlexandreBelling (approved):

PR #3097, opened 2026-05-13
feat(prover-ray): support dynamic module sizes in global compiler

### LINKED ISSUE #2922: Port the global constraint compiler with dynamic-sized column support
## Summary

Port the global constraint compiler to `prover-ray`, replacing the current vibe-coded version with a clean implementation that properly supports dynamic-sized columns.

## Context

The current global constraint compiler computes one quotient per module, which is the right structure, but was written without dynamic-sized column support in mind. This port should be a deliberate, well-tested reimplementation rather than a copy.

## Scope

- Compute one quotient polynomial per module
- Correctly handle dynamic-sized columns (variable row counts per module)
- Shift evaluation points to save time as described in the compilation pipeline
- Comprehensive tests including invalid-witness (soundness) cases

## Dependencies

- Dynamic-sized module support (#2913)

Soleimani193:
This PR implements issue(s) #2922

Closes #2922

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 7ed2a5b0daa2ce1c88daa0a04eafc4dd3ec5b412. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/prover-ray/wiop/compilers/global/global.go
+++ b/prover-ray/wiop/compilers/global/global.go
@@ -838,20 +838,19 @@ func collectRootColumns(expr wiop.Expression) []*wiop.Column {
 // d + c = degreeFactor * (n-1) + c. Dividing by the annihilator (x^n - 1)
 // gives a quotient of degree at most:
 //
-//	quotientDeg = degreeFactor * (n-1) + c - n
-//	            = (degreeFactor - 1) * n + (c - degreeFactor)
-//	            = (degreeFactor - 1) * (n - 1) + (c - 1)
+//	quotientDeg = degreeFactor * (n-1) + c - n +1
+//	            = (degreeFactor - 1) * n + (c - degreeFactor + 1)
 //
 // For this to fit in ratio shares of size n (i.e., degree < ratio * n), we need:
 //
 //	ratio * n > quotientDeg
-//	ratio > (degreeFactor - 1) + (c - degreeFactor) / n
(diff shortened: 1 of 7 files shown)


prover-ray/wiop/compilers/global/global.go
```diff
@@ -746,15 +828,32 @@ func collectRootColumns(expr wiop.Expression) []*wiop.Column {
...
+// gives a quotient of degree at most:
+//
+//	quotientDeg = degreeFactor * (n-1) + c - n +1
+//	            = (degreeFactor - 1) * n + (c - degreeFactor + 1)
+//
+// For this to fit in ratio shares of size n (i.e., degree < ratio * n), we need:
+//
+//	ratio * n > quotientDeg
+//	ratio > (degreeFactor - 1) + (c - degreeFactor +1) / n
+func computeRatio(v *wiop.Vanishing) int {
```
AlexandreBelling, line 848: Running the lower bound I get `ratio <= 2 * degreeBound + len(cancel) - 2` corresponding to n=1.

If degreeBound = 4 and len(cancel) = 2, then we get ratio=8. But we would like to bound the ratio to 4 so that we can have columns of size 2**22 without breaking the 2 adicity of koalabear.

Perhaps, we need a system of upper-bound in the module where modules with a larger exprDegree may not have too many rows.

AlexandreBelling (approved):

PR #3154, opened 2026-05-19
feat: interpreter optim for i-family for funct6 and funct7 destructur…

amkCha:
Running the benchmark on 100 keccaks, this improves the speed by 8%.

Indeed the impact of this change in the i-family could be explained on keccak by the compiled program decoded being composed of ADDI at 22%.

5 local runs, alternate baseline and optim version

---

> [!NOTE]
> **Medium Risk**
> Touches core RISC-V I-type instruction execution (LOAD/OP-IMM/OP-IMM-32/JALR), so any mistake could change program semantics, but the change is mostly a localized refactor/optimization reusing existing computations.
> 
> **Overview**
> **Optimizes I-type decoding/execution** by computing `rs1 + sign_extend(imm12)` once (`rs1_plus_signext_imm12`) and reusing it across `LOAD`, `ADDI/ADDIW`, and `JALR`, instead of recomputing and casting to an `Address` early.
> 
> **Reduces work in non-shift paths** by moving `funct6`/`uimm6` and `funct7`/`uimm5` extraction into the `OP_IMM` and `OP_IMM_32` cases respectively, and removes the `funct7` debug print from the default I-type error branch.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b50be18726328214482c33d440cce9ab462ca082. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


OlivierBBB (dismissed):
So the optimization is to
- do the `funct6` / `uimm6` stuff only for `OP_IMM`
- do the `funct7` / `uimm5` stuff only for `OP_IMM_32`
- on the contrary to compute `rs1_plus_imm` in all cases

LGTM.

I'm wondering: the casting
```
var mem_address:Address = rs1_plus_imm as Address
```
from `DoubleWord`, which is an alias for `u64`, to `Address`, which is also an alias for `u64`: will that be free ? It should be a noop but from what I know of go-corset it will create a new register for `mem_address` when parsing the function body.

amkCha:
> So the optimization is to
> 
>     * do the `funct6` / `uimm6` stuff only for `OP_IMM`
> 
>     * do the `funct7` / `uimm5` stuff only for `OP_IMM_32`
> 
>     * on the contrary to compute `rs1_plus_imm` in all cases
> 

Yes for the first 2, and for the last one, the computation was already done in all cases but then it wasn't reused in the subcases, so I just removed the extra computation and used the variable. It's used in all cases except SYSTEM

amkCha:
> I'm wondering: the casting
> 
> ```
> var mem_address:Address = rs1_plus_imm as Addressfrom `DoubleWord
> ```
> 
> from `DoubleWord`, which is an alias for `u64`, to `Address`, which is also an alias for `u64`: will that be free ? It should be a noop but from what I know of go-corset it will create a new register for `mem_address` when parsing the function body.

Agreed that I'm unsure about keeping the casting, as I don't think it's free. I was trying to balance things out, as we used alias to make it easier to read vs u64 can be the input / output of many things. I can also just add a comment

letypequividelespoubelles (approved):
I think we'll have to see the tradeoff between compute only in subcase as it raises the constraint's  degree but make the tracing faster.

Anyway, LGTM for me

PR #3160, opened 2026-05-19
[ZkTracer] feat: run act4 test suite for arith, add install-zkc dependency on te…

amkCha:
…st runs

This PR implements issue(s) #

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> CI now installs additional toolchains (Rust nightly, `mise`, xPack RISC-V GCC) and runs a new ACT4 suite, which increases build complexity and the chance of flaky or failing pipelines if tool downloads/versions change.
> 
> **Overview**
> Adds a new GitHub Actions workflow to build and run the RISC-V ACT4 reference test suite for `arithmetization`.
> 
> Updates the shared `setup-arithmetization-riscv` composite action to install Rust nightly (with `rust-src`), `mise`, and an xPack-provided `riscv64-unknown-elf-gcc` (plus binutils), replacing the previous single apt-based assembler install.
> 
> Tightens `zkc` installation behavior by making it a dependency of example/test targets (`zkc-exec`, `zkc-debug`, `blake-all`, `run-act4`) and by skipping reinstall when `zkc` is already on `PATH` unless a pinned `GO_CORSET_REF_FOR_ZKC` is provided; also tweaks existing workflows/job names to use the consolidated install/run steps.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 76a2bf633b402aa18e77a5a3a29222cbafadb84e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


arithmetization/src/test/examples/Makefile
```diff
@@ -143,10 +144,13 @@ compile: $(SRC) linker-script
...
 
-zkc-exec:
+install-zkc:
+	$(MAKE) -C $(ARITHMETIZATION_DIR) install-zkc
+
+zkc-exec: install-zkc
 	zkc exec $(JSON) $(ZKC_MAIN)
 
-zkc-debug:
+zkc-debug: install-zkc
```
lorenzogentile404, line 153: `zkc` is among the prerequisites to use this `Makefile`, so for consistency I would avoid adding that (as we do not do it for other dependencies).

amkCha: I see your point, but I think the custom way to see it is more : pre-requisites are more well-know stacks, with different ways to install etc, that is up to the user to decide. And then for instance zkc, it's our recipe, only one way to install, and for the test run, you need a special version we (will) pin. so in that case, it becomes a burden to outsource it. That's why previously go-corset was handled by gradle directly. You want it to run out-of-the-box as long as you have the major ones, go, rust etc

lorenzogentile404: Sounds good.

lorenzogentile404 (changes requested):
Also the alias in the README needs to be updated with `install-zkc` target:

```
riscv-test() {
    local makefile="path/to/linea-monorepo/arithmetization/src/test/examples/Makefile"
    case "$1" in
        clean-all|linker-script|blake-all|build-act4|run-act4|install-zkc)
            # targets that do NOT require TEST argument
            make -f "$makefile" "$1" "${@:2}"
            ;;
        exec|debug|compile|zkc-exec|zkc-debug|clean|verify-elf)
            # targets that require TEST argument
            make -f "$makefile" "$1" TEST="$2" "${@:3}"
            ;;
        *)
            # default target (riscv-test foo.<ext> is the same as riscv-test exec TEST=foo.<ext>)
            make -f "$makefile" TEST="$1" "${@:2}"
            ;;
    esac
}
```

arithmetization/Makefile
```diff
@@ -43,7 +43,15 @@ checkout-go-corset:
 	fi
 
 # Install the zkc CLI from the go-corset repository.
-install-zkc: checkout-go-corset
+# - When GO_CORSET_REF_FOR_ZKC is set: always checkout the pinned ref and reinstall.
+# - When GO_CORSET_REF_FOR_ZKC is empty: skip if zkc is already on PATH.
+install-zkc:
+	@set -eu; \
+	if [ -z "$(GO_CORSET_REF_FOR_ZKC)" ] && command -v zkc >/dev/null 2>&1; then \
```
lorenzogentile404, line 50: do you actually want `set -eu` here? you could use `|| exit $$?;` to make the recipe stop right after a failure.

lorenzogentile404 (approved):
LGTM, small comment left.

PR #3166, opened 2026-05-20
feat(verifier-ray): implement multi round coin generation in verifier-ray

### LINKED ISSUE #3121: [RV recursion] - implement per-round random coin generation
Per-round Fiat–Shamir / random-coin derivation in the verifier.

 In the IOP protocol the verifier and prover trade messages in rounds; at the start of each round the verifier needs a
  fresh random "coin" (challenge) that the prover couldn't have predicted. #3121 is the verifier-side API for producing those coins via Fiat-Shamir — the named operation, not the underlying transcript machinery.

  For each protocol round, expose a function that:
  - [ ] Asserts the verifier is in the expected round state (no skipping, no replaying).
  - [ ] Squeezes the right number of bytes from the transcript.
  - [ ] Maps those bytes into the right output type — often a Koalabear-extension element for soundness, sometimes a list of column indices for Vortex queries, sometimes a small modulus reduction for selection from a finite set.
  - [ ] Advances the round counter so the next call gets the next coin.

  - [ ] testing: a scripted protocol trace: prover does these absorbs, verifier squeezes these coins, expected coin values are [v1, v2, …]. Replay the script with one prover-side absorb byte flipped, expect the first downstream coin to diverge.
  This is what locks in agreement between the Go prover and Zig verifier transcripts.

arijitdutta67:
This PR implements issue(s) #3121 

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> Changes Fiat-Shamir transcript absorption and round/coin derivation that must stay bit-for-bit aligned with prover-ray; mistakes would break proof verification, though golden traces and round-guard tests mitigate this.
> 
> **Overview**
> Adds **verifier-side multi-round Fiat-Shamir coin derivation** so the Zig runtime can mirror `prover-ray`’s `AdvanceRound` without pulling in the full Go WIOP runtime.
> 
> **Runtime API:** Replaces the placeholder `advanceRound()` with `advanceRoundWithMessage`, which absorbs caller-supplied oracle/public column vectors and public cells, bumps `current_round`, and squeezes the requested number of Koalabear E6 extension coins into caller-owned storage. It tracks `total_rounds`, validates round order (no skip/replay), blocks advancing the last round, and errors on undersized output buffers. `Visibility` tags match prover-ray’s oracle/public encoding; internal columns stay out of the message.
> 
> **Transcript / types:** Fiat-Shamir gains `absorbVector` / `absorbScalar` over new `field/value` `Vector` and `Scalar` unions (replacing the old `vec` export in `lib.zig`). Generated verifier stubs no longer fake `rt.advanceRound()`.
> 
> **Tests & fixtures:** Go `testdata/generate` builds a scripted multi-round WIOP trace via `prover-ray` and commits `runtime_trace_cases` in `vectors.zig`. Golden tests replay rounds and assert coin equality; a tamper test and unit tests cover round guards and absorption. README documents runtime behavior and the regenerate workflow.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a4fc04a31f49cdb4c15729b6057d224c6b929d47. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


verifier-ray/src/runtime.zig
```diff
@@ -1,17 +1,117 @@
...
+
+    pub fn initWithRoundCount(round_count: usize) Runtime {
         return .{
             .transcript = fiat_shamir.Transcript.init(),
             .current_round = 0,
+            .total_rounds = round_count,
         };
     }
 
     pub fn advanceRound(self: *Runtime) void {
```
YaoJGalteland: is advanceRound redundant? There are no transcript absorption, no coin derivation, no validation of assignments, no round-ordering enforcement.

verifier-ray/src/runtime.zig
```diff
@@ -1,17 +1,117 @@
...
+        self: *Runtime,
+        expected_round: usize,
+        message: RoundMessage,
+        out_coins: []ext.Ext,
+    ) Error![]const ext.Ext {
+        if (self.total_rounds == 0) return Error.NoRounds;
+        if (self.current_round != expected_round or expected_round >= self.total_rounds) {
+            return Error.UnexpectedRound;
+        }
+        if (self.current_round + 1 >= self.total_rounds) return Error.LastRound;
```
YaoJGalteland, line 76: I think the last round should never call `advanceRoundWithMessage`. That method exists to close out a round’s transcript and derive challenges for the next round, so calling it when there’s no next round doesn’t make sense. 

The fix should be at the call site: the Zig codegen shouldn’t call `advanceRoundWithMessage` for the final round. 

Same issue in prover-ray: the codegen currently calls `rt.AdvanceRound`() after every round, including the last one, which will panic at runtime. It could be guarded with if i < len(sys.Rounds)-1.

verifier-ray/src/runtime.zig
```diff
@@ -1,17 +1,117 @@
...
+            .ext => |value| transcript.updateExt(&.{value}),
+        }
+    }
+};
+
+pub const ColumnAssignment = struct {
+    visibility: Visibility,
+    /// Null means the column has not been assigned. Oracle and public columns
+    /// must be assigned before their round can advance.
+    assignment: ?Vector,
```
ivokub: But for the verifier, all columns are either oracle and public columns? Why would the assignment be optional?

verifier-ray/src/runtime.zig
```diff
@@ -1,17 +1,117 @@
...
+};
+
+pub const Visibility = enum(u8) {
+    internal = 0,
+    oracle = 1,
+    public = 2,
+};
+
+pub const Vector = union(enum) {
+    base: []const field.Element,
```
ivokub: Perhaps use the Vector types from `verifier-ray/src/field/vec.zig`? They are called ...Slice there, I think we could potentially rename to `ElementVector` and `ExtVector` there?

verifier-ray/src/runtime.zig
```diff
@@ -1,17 +1,117 @@
...
+    NoRounds,
+    UnexpectedRound,
+    LastRound,
+    MissingColumnAssignment,
+    MissingCellAssignment,
+    OutputTooSmall,
+};
+
+pub const Visibility = enum(u8) {
+    internal = 0,
```
ivokub: Would we need internal visibility for the verifier?

YaoJGalteland (commented):
The coin generation flow is very clear and easy to follow. Left a few comments below

verifier-ray/src/runtime.zig
ivokub: And perhaps we could move the Vector and Scalar definitions to somewhere it is better reusable?

verifier-ray/src/runtime.zig
```diff
@@ -1,17 +1,117 @@
...
+    oracle = 1,
+    public = 2,
+};
+
+pub const Vector = union(enum) {
+    base: []const field.Element,
+    ext: []const ext.Ext,
+
+    fn absorb(self: Vector, transcript: *fiat_shamir.Transcript) void {
```
ivokub: Imo it would be better to have the absorb methods in fiat-shamir package. It is just a wrapper around `update....` methods on union type.

verifier-ray/test/golden_test.zig
```diff
@@ -217,3 +250,70 @@ fn expectDigest(actual: poseidon2.Digest, expected: [8]u32) !void {
         try expectElem(actual_limb, expected_limb);
     }
 }
+
+const max_trace_columns = 8;
+const max_trace_values = 8;
+const max_trace_cells = 8;
+const max_trace_coins = 8;
```
ivokub: These values are not generated, but rather some heuristically chosen values? Why couldn't we derive them from test case?

verifier-ray/test/golden_test.zig
```diff
@@ -217,3 +250,70 @@ fn expectDigest(actual: poseidon2.Digest, expected: [8]u32) !void {
...
+
+const TraceRoundBacking = struct {
+    columns: [max_trace_columns]runtime.ColumnAssignment = undefined,
+    cells: [max_trace_cells]?runtime.Scalar = undefined,
+    base_values: [max_trace_columns][max_trace_values]field.Element = undefined,
+    ext_values: [max_trace_columns][max_trace_values]ext.Ext = undefined,
+
+    fn fill(self: *TraceRoundBacking, round_case: anytype, tamper_first_absorb: bool) !runtime.RoundMessage {
```
ivokub: Why use anytype for round_case?

verifier-ray/test/golden_test.zig
```diff
@@ -217,3 +250,70 @@ fn expectDigest(actual: poseidon2.Digest, expected: [8]u32) !void {
...
+                    try std.testing.expect(column_case.ext_values.len <= max_trace_values);
+                    fillExts(&self.ext_values[i], column_case.ext_values);
+                    assignment = .{ .ext = self.ext_values[i][0..column_case.ext_values.len] };
+                } else {
+                    try std.testing.expect(column_case.base_values.len <= max_trace_values);
+                    fillElems(&self.base_values[i], column_case.base_values);
+                    if (tamper_first_absorb and !tampered and column_case.base_values.len != 0) {
+                        // the lsb of the first value is bit flipped to simulate a tampered absorb, which should cause
+                        //downstream coins to diverge
+                        self.base_values[i][0] = field.Element.init(self.base_values[i][0].value ^ 1);
```
ivokub: for consistency, we could use `elem(...)` instead of `field.Element.init(...)` here

verifier-ray/src/runtime.zig
```diff
@@ -1,17 +1,117 @@
...
         };
     }
 
     pub fn advanceRound(self: *Runtime) void {
         self.current_round += 1;
     }
+
+    /// Absorb one round's verifier-visible messages, advance the round counter,
+    /// and derive the next round's extension-field coins.
+    pub fn advanceRoundWithMessage(
```
ivokub, line 66: I think we should have a bit more comments -- out_coins should be backing array where we store the coins etc.

ivokub (changes requested):
Thanks! The test generation and verification look great. I think everything is well covered as well. There are only a few comments on refactoring etc. Have a look what makes sense and what not.

verifier-ray/src/runtime.zig
arijitdutta67: Good catch. Indeed I forgot to remove it. Also removed it from stub.zig. We can later use `advanceRoundWithMessage` here.

verifier-ray/src/runtime.zig
arijitdutta67: Agreed. We can follow this principle when we are going to use `advanceRoundWithMessage` in verifier-ray

verifier-ray/src/runtime.zig
arijitdutta67: True, we don't need it in verifier. Removing it.

verifier-ray/src/runtime.zig
```diff
@@ -1,17 +1,113 @@
...
 
-    pub fn advanceRound(self: *Runtime) void {
+    /// Absorb one round's verifier-visible messages, advance the round counter,
+    /// and derive the next round's extension-field coins.
+    pub fn advanceRoundWithMessage(
+        self: *Runtime,
+        expected_round: usize,
+        message: RoundMessage,
+        out_coins: []ext.Ext,
+    ) Error![]const ext.Ext {
```
YaoJGalteland: Can we have a Coin type? For example, `pub const Coin = ext.Ext`. It makes the data’s purpose much clearer and improves readability.

It also ensures coins are always represented as ext.Ext, preventing accidental use of buggy base-field values. Additionally, it avoids mixing the meaning of these values with other ext.Ext usages, such as polynomial evaluations.

verifier-ray/src/runtime.zig
```diff
@@ -1,17 +1,92 @@
...
+    NoRounds,
+    UnexpectedRound,
+    LastRound,
+    OutputTooSmall,
+};
+
+/// For the verifier, there are only two meaningful visibilty, the internal visisbilty
+/// of prover-ray is not relevant. The `oracle` visibility is for columns that are only visible to the prover, and the `public` visibility is for columns that are visible to both the prover and verifier.
+pub const Visibility = enum(u8) {
+    oracle = 0,
```
ivokub: Even though we don't have the private visibility, then I think it would be good to keep same mapping with prover i.e. oracle = 1 and public = 2. This also allows to remove the mapping in golden_test.zig
```zig
/// Convert generated visibility tags into verifier runtime visibility values.
fn visibility(value: u8) !runtime.Visibility {
    return switch (value) {
        1 => .oracle,
        2 => .public,
        else => error.InvalidVisibility,
    };
}
```

verifier-ray/codegen/internal/generator/generator_test.go
```diff
@@ -24,8 +24,14 @@ func TestGenerate_DefaultEntryPoint(t *testing.T) {
 	if !strings.Contains(src, "pub fn verifyGenerated") {
 		t.Fatalf("generated source is missing default entry point:\n%s", src)
 	}
-	if got := strings.Count(src, "rt.advanceRound();"); got != 2 {
-		t.Fatalf("expected one advance per round, got %d:\n%s", got, src)
+	if !strings.Contains(src, "_ = rt;") {
+		t.Fatalf("generated source should mark runtime unused until real round messages exist:\n%s", src)
+	}
+	if strings.Contains(src, "rt.advanceRound") {
```
ivokub, line 27: Imo we don't need this test -- we have removed it from source code and it would be compile time failure. This test would carry only some historical assumptions around and we couldn't in the future figure out why it is like this.

ivokub (commented):
Yup, I think it looks good now. Perhaps also have a look at the copilot comments to keep README up-to-date. And also see a comment about visibility, imo it is good to keep aligned with prover side just in case.

YaoJGalteland (approved):
Thank you for the changes, looks good!

ivokub (approved):
Thanks! Looks good!

PR #3168, opened 2026-05-20
feat(verifier-ray): migrate verifier-ray extension field from E4 to E6

### LINKED ISSUE #3148: [RV recursion] - migrate to E6 from E4 in the stub Zig verifier
https://github.com/Consensys/linea-monorepo/pull/3111 updated Ray verifier to use extension 6 from extension 4. Migrate Zig verifier to E6 as well for compatibility

Implementation:
* update the extension definition and operations

Testing:
- [ ] update the golden unit tests to accomodate the change

YaoJGalteland:
This PR implements issue(s) #3148 

Migrates the Zig verifier (verifier-ray) extension field from degree-4 (E4) to degree-6 (E6), matching the prover-ray migration done in PR #3111.

Closes #3148 

---

> [!NOTE]
> **High Risk**
> High risk because it changes core finite-field arithmetic (`mul`, `inverse`, serialization) and Fiat–Shamir challenge derivation, which can invalidate proofs or cause subtle consensus/verification mismatches if incorrect.
> 
> **Overview**
> Migrates `verifier-ray`’s KoalaBear extension field from a 4-limb representation to a 6-limb *Fp6 over Fp2* construction, introducing a dedicated `E2` type and rewriting `Ext` arithmetic (notably `mul`, `inverse`, `toBytes/fromBytesCanonical`, and `pow`).
> 
> Updates the Fiat–Shamir transcript to absorb and sample the new 6-limb extension elements, refreshes golden vectors + generator to the new `prover-ray` version, and adjusts tests accordingly (including a new `a^(p^6-1)=1` check).
> 
> Separately, adds `verifier-ray` as an allowed Conventional Commit scope in `.husky/commit-msg`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 823285ef9d032b995ab558bc8c5d9e24a809c25f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


verifier-ray/src/crypto/fiat_shamir.zig
YaoJGalteland, line 26: The name `randomField` is slightly misleading. I recall the old implementation returned a base field element, which is no longer useful. The current `randomField` in prover-ray returns an Octuplet. I think `randomDigest` or `randomOctuplet` would be much clearer. @ivokub Do you have a better alternative? It would be great to clean up this naming convention everywhere.

ivokub: Good catch -- perhaps randomDigest is clearer? Imo `randomOctuplet` kind of forces the output type to be Octuplet which imo is too opinionated API.

arijitdutta67 (dismissed):
LGTM

verifier-ray/src/crypto/fiat_shamir.zig
YaoJGalteland: I will create a separate PR for the renaming so we don't mess up this current PR. The dominant part of that change will be inside prover-ray.

verifier-ray/testdata/generate/go.sum
```diff
@@ -8,6 +8,8 @@ github.com/consensys/gnark-crypto v0.20.2-0.20260514182922-df0578435b08 h1:EdljQ
 github.com/consensys/gnark-crypto v0.20.2-0.20260514182922-df0578435b08/go.mod h1:NzeBHSZ49bIM7RtrNTYYR2kymTqwvI/A4eTgQlyQc+Q=
 github.com/consensys/linea-monorepo/prover-ray v0.0.0-20260518154448-1f6880839cd2 h1:Ggp+FPN2SmxdPC1k77w9YCNef4RUOL9C4algaGIUuWk=
 github.com/consensys/linea-monorepo/prover-ray v0.0.0-20260518154448-1f6880839cd2/go.mod h1:HZncWpppP5LS0qB3moZR1ESi/vIbCEnkG1w65TsAojU=
+github.com/consensys/linea-monorepo/prover-ray v0.0.0-20260519010204-24a53941da53 h1:DpSMTgN3kJYo7vbBGnSXGJXQx4yVVR35WUE/NFqYdWY=
```
ivokub, line 9: In the future, please also run `go mod tidy` when updating dependencies - it will remove the stale checksums

verifier-ray/src/field/koalabear_ext.zig
```diff
@@ -1,11 +1,12 @@
...
-pub const degree = 4;
+pub const degree = 6;
 pub const bytes = degree * base.bytes;
 
 pub const Ext = struct {
     /// Limb order matches prover-ray and gnark-crypto:
-    /// B0.A0, B0.A1, B1.A0, B1.A1 for B0 + B1*v with v^2 = u and u^2 = 3.
+    /// B0.A0, B0.A1, B1.A0, B1.A1, B2.A0, B2.A1
+    /// for the tower F_{p^6} = F_{p^2}[v]/(v^3 - (u+1)) with F_{p^2} = F_p[u]/(u^2 - 3).
     limbs: [degree]base.Element,
```
ivokub: Perhaps we should follow the definition in gnark-crypto/ray i.e.
```
pub const Ext = struct {
    B0: E2,
    B1: E2,
    B2: E2,
}
```

Memory-wise it doesn't make a difference, but corresponds more clearly to the gnark-crypto definition. And imo this makes the implementation also a bit clearer in mul where we right now manually construct E2 values.

ivokub (commented):
Have a look if makes sense to use the concrete tower definition for E6 as E6{E2, E2, E2} instead of flattened.

verifier-ray/src/field/koalabear_ext.zig
YaoJGalteland: Good idea! That's more structured and highly readable.

verifier-ray/src/field/koalabear_ext.zig
```diff
@@ -1,174 +1,144 @@
...
+        const d_inv = d.inverse();
+
+        return .{ .B0 = cap_a.mul(d_inv), .B1 = cap_b.mul(d_inv), .B2 = cap_c.mul(d_inv) };
     }
 
     pub fn div(self: Ext, rhs: Ext) Ext {
         return self.mul(rhs.inverse());
     }
 
+    pub fn pow(self: Ext, exponent: u256) Ext {
```
ivokub, line 99: I think we can keep it for now, but perhaps we should reconsider in the future if exponent should be `u256` - it is correct mathematically that this bound makes sense, but in the application point of view (verifying proofs) the exponent is usually the domain size, i.e. not bigger than 2^24. So u32 should be fine.

But lets keep as is and fix in the future when it makes sense.

ivokub (approved):
LGTM

arijitdutta67 (approved):

PR #3178, opened 2026-05-20
refactor(prover-ray): rename misleading field names to digest/fext

YaoJGalteland:
Rename misleading `Field`/`Element` names to `Digest` or `Fext` in prover-ray and verifier-ray

The previous names mixed up the concept of a raw field element with the specific semantic types returned, either a Poseidon2 digest (an Octuplet) or an extension field element (Ext/Fext). Three renames were applied consistently across both packages:

`SumElement` → `SumDigest` (prover-ray)

- `MDHasher.SumElement()` returns a Poseidon2 digest (an `Octuplet`), not a generic field element. Renamed across poseidon2.go, fiatshamir, vortex, smt, and all test files.

`RandomField` → `RandomDigest` (prover-ray)

- FiatShamir.RandomField() samples a random Poseidon2 digest from the transcript, not an arbitrary field element. Renamed in fiatshamir/poseidon2.go, fiat_shamir.zig, and the test vector generator.

`RandomFieldFromSeed` → `RandomFextFromSeed` (prover-ray)

- FiatShamir.RandomFieldFromSeed() returns a field.Ext (extension field element), not a base field element. Renamed to reflect the actual return type.

---

> [!NOTE]
> **Low Risk**
> Low risk: this is a mechanical rename of APIs and call sites to better reflect return types, with no algorithmic changes. Main risk is missed external call sites causing compile/runtime integration issues outside the updated packages.
> 
> **Overview**
> Renames Poseidon2 MD hasher output APIs to reflect semantics: `MDHasher.SumElement()` is now `SumDigest()`, and all internal callers (SMT hashing, Vortex commitments/verifier, and Poseidon2 tests) are updated accordingly.
> 
> Renames Fiat-Shamir transcript sampling to match what is produced: `RandomField()` -> `RandomDigest()` (octuplet digest) and `RandomFieldFromSeed()` -> `RandomFextFromSeed()` (extension field element), updating usages like `RandomFext()` and integer sampling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e9f775cdbd28d49a544053dc0d06f0b43386bd30. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/prover-ray/crypto/koalabear/fiatshamir/poseidon2.go
+++ b/prover-ray/crypto/koalabear/fiatshamir/poseidon2.go
@@ -89,17 +89,17 @@ func (fs *FiatShamir) UpdateSV(sv field.Vec) {
 	}
 }
 
-// RandomField samples a random octuplet from the current transcript state.
-func (fs *FiatShamir) RandomField() field.Octuplet {
+// RandomDigest samples a random octuplet from the current transcript state.
+func (fs *FiatShamir) RandomDigest() field.Octuplet {
 	defer fs.safeguardUpdate()
-	return fs.h.SumElement()
+	return fs.h.SumDigest()
 }
 
 // RandomFext samples a random extension field element from the transcript.
 // Uses the first ExtensionDegree (=6) of the 8 octuplet coordinates; the
 // remaining two are discarded.
 func (fs *FiatShamir) RandomFext() field.Ext {
-	s := fs.RandomField() // already calls safeguardUpdate()
+	s := fs.RandomDigest() // already calls safeguardUpdate()
 	var res field.Ext
 	res.B0.A0 = s[0]
 	res.B0.A1 = s[1]
@@ -111,8 +111,8 @@ func (fs *FiatShamir) RandomFext() field.Ext {
 	return res
 }
 
-// RandomFieldFromSeed derives a deterministic extension field element from a seed and name.
-func (fs *FiatShamir) RandomFieldFromSeed(seed field.Octuplet, name string) field.Ext {
+// RandomFextFromSeed derives a deterministic extension field element from a seed and name.
+func (fs *FiatShamir) RandomFextFromSeed(seed field.Octuplet, name string) field.Ext {
 
 	// The first step encodes the 'name' into a single field element. The
 	// field element is obtained by hashing and taking the modulo of the
@@ -149,7 +149,7 @@ func (fs *FiatShamir) RandomManyIntegers(num, upperBound int) []int {
 
 	for i < num {
 		// take the remainder mod n of each limb
-		c := fs.RandomField() // already calls safeguardUpdate()
+		c := fs.RandomDigest() // already calls safeguardUpdate()
 		for j := 0; j < 8; j++ {
 			b := c[j].Bits()
 			res[i] = int(b[0]) % upperBound
--- a/prover-ray/crypto/koalabear/poseidon2/poseidon2.go
+++ b/prover-ray/crypto/koalabear/poseidon2/poseidon2.go
@@ -86,7 +86,7 @@ func (d *MDHasher) GetStateOctuplet() field.Octuplet {
 	oldBuffer := slices.Clone(d.buffer)
 	oldBufferPosition := d.bufferPosition
 
-	_ = d.SumElement() // this flushes the hasher
+	_ = d.SumDigest() // this flushes the hasher
 	res := slices.Clone(d.state[:])
 
 	d.state = oldState
@@ -101,8 +101,8 @@ func (d *MDHasher) WriteElements(elmts ...field.Element) {
 	d.buffer = append(d.buffer, elmts[:]...)
 }
 
-// SumElement finalizes the hash and returns the digest as an octuplet.
-func (d *MDHasher) SumElement() field.Octuplet {
+// SumDigest finalizes the hash and returns the digest as an octuplet.
+func (d *MDHasher) SumDigest() field.Octuplet {
 	for len(d.buffer) != 0 {
 		var buf [BlockSize]field.Element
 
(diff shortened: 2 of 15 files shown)


verifier-ray/testdata/generate/go.mod
```diff
@@ -4,6 +4,8 @@ go 1.25.7
 
 require github.com/consensys/linea-monorepo/prover-ray v0.0.0-20260519010204-24a53941da53
 
+replace github.com/consensys/linea-monorepo/prover-ray => ../../../prover-ray
```
ivokub: Imo it is not good to add it here. We just removed the replace directive https://github.com/Consensys/linea-monorepo/commit/a52ce40dd481b9d203aa63d4c622f09fd75a102b because with it any changes in prover-ray API need to be mirrored in verifier-ray, otherwise we get failing CI. And for now we want to decouple prover and verifier implementation so that prover side could implement things without needing to update the verifier implementation yet.

If we pin prover-ray in this package to concrete version, then we can define on the verifier side what prover version we concretely depend on, without breaking anything.

ivokub: Now, if we remove the replace directive, then this means that we have to split the prover and verifier changes into two PRs -- first we do the prover and later in a separate PR we do the verifier.

So I would remove the verifier side changes from this PR for now and push as a separate PR when this PR is merged with updating go.mod to concrete pinned prover-ray version.

Perhaps later we will remove the separate versioning of the prover and verifier when things are settled, otherwise I think it becomes difficult to make more bigger refactoring (as this PR) without touching a lot of codebase at once.

ivokub (changes requested):
The prover side looks good, but lets split the prover and verifier changes into two PRs.

prover-ray/crypto/koalabear/fiatshamir/poseidon2.go
```diff
@@ -89,17 +89,17 @@ func (fs *FiatShamir) UpdateSV(sv field.Vec) {
 	}
 }
 
-// RandomField samples a random octuplet from the current transcript state.
-func (fs *FiatShamir) RandomField() field.Octuplet {
+// RandomDigest samples a random octuplet from the current transcript state.
+func (fs *FiatShamir) RandomDigest() field.Octuplet {
```
AlexandreBelling, line 93: RandomOctuplet?

prover-ray/crypto/koalabear/poseidon2/poseidon2.go
```diff
@@ -101,8 +101,8 @@ func (d *MDHasher) WriteElements(elmts ...field.Element) {
 	d.buffer = append(d.buffer, elmts[:]...)
 }
 
-// SumElement finalizes the hash and returns the digest as an octuplet.
-func (d *MDHasher) SumElement() field.Octuplet {
+// SumDigest finalizes the hash and returns the digest as an octuplet.
+func (d *MDHasher) SumDigest() field.Octuplet {
```
AlexandreBelling, line 105: SumOctuplet

prover-ray/crypto/koalabear/poseidon2/poseidon2_native_test.go
```diff
@@ -78,23 +78,23 @@ func TestResetWithoutPriorSum(t *testing.T) {
...
-	r2 := h2.SumElement()
+	r2 := h2.SumDigest()
 
-	require.Equal(t, r1, r2, "Reset() must clear the buffer even when SumElement was not called first")
+	require.Equal(t, r1, r2, "Reset() must clear the buffer even when SumDigest was not called first")
 }
 
-// TestSumElementIdempotent checks that calling SumElement again on an already-flushed hasher is a no-op.
-func TestSumElementIdempotent(t *testing.T) {
+// TestSumDigestIdempotent checks that calling SumDigest again on an already-flushed hasher is a no-op.
```
AlexandreBelling, line 90: TestSumOctupletIdempotent

AlexandreBelling (commented):
Left some comments as suggestion

prover-ray/crypto/koalabear/fiatshamir/poseidon2.go
YaoJGalteland: Discussed with @ivokub, we agreed that Digest is the better name here. Digest describes what the value is (a Poseidon2 hash output / Fiat-Shamir challenge), while Octuplet describes how it's stored (8 KoalaBear field elements). If the hash function ever changed its output width, randomDigest stays correct but randomOctuplet becomes a lie.

ivokub (approved):
Thanks for the changes. Now looks good.

My preference would indeed be "SumDigest" for outputting values. Perhaps it would make sense to have `type Digest = field.Octuplet` and have this as return value. It should be non-breaking change as it is type alias, but perhaps a bit cleaner? Or if we don't use Octuplets outside of Poseidon2 context at all, then we could just define `type Digest [8]field.Element` in poseidon2.go and leave the Octuplet definition out from `field` package.

But I don't have strong preference, I'm good with anything imo it is clear enough both ways (except for "Field" when we output Octuplets :) )

YaoJGalteland:
I found keccak package also calls it `RandomField` (I won't change that code name right now). That makes sense because they use BLS12-377, so the hash output is a single field element. It would be good to introduce a generic name that covers all hashes :)

AlexandreBelling:
Keccak cannot return on a BLS12-377 as it outputs on 256 bits and the field is slightly smaller than that. Do we forsee a situation where we won't use octuplet as a digest type in the prover/verifier? I can't think of one. 

Maybe when tackling the outer-proof?

AlexandreBelling (approved):

PR #3187, opened 2026-05-21
chore(misc): New rollup spec, python part update

Filter94:
This PR implements issue(s) #

### Checklist

  PR.

---

> [!NOTE]
> **High Risk**
> Changes the proving protocol surface (public inputs, DA/KZG binding, bridge and FTX semantics) and reference guest logic that L1 finalization and production zkVM implementations must match.
> 
> **Overview**
> This PR **restructures the Type-1 RISC-V rollup spec and Python reference** around three guests—**l2-execution**, **rollup**, and **rollup-aggregation**—plus separate **`l1_rollup.py`** for contract checks, replacing the old monolithic `blob.py` / execution+blob naming.
> 
> **Spec (`Readme.md`)** renames proof layers, documents guest entry points, and tightens semantics: rollup guests **recompute** DA from witnessed **`blockRlps`** (truncate → RLP → LZ4 → zero-pad) and verify **KZG** in-guest (no witnessed `blobContent` / commitment); **final L1 public inputs** drop **`endStateRootHash`** and add **`endBlockTimestamp`**; L1→L2 bridge boundaries are read from **L2MessageService storage** (no `l1L2Messages` witness); **forced transactions** use five **`acceptance`** outcomes with L1 sanction checks on refused addresses; rollup adds **full parent-hash chain** binding for intermediate blocks; **§2.5** maps Python `raise` to zkVM failed termination.
> 
> **Python reference** implements **`run_l2_execution_guest`**, **`run_rollup_guest`** (ckzg + lz4, shnarf, recursive l2-exec checks, L2→L1 Merkle roots), and **`run_rollup_aggregation_guest`**, with **`block.py`** FTX/chain-config cleanup and **`requirements.txt`** adding **`ckzg`** and **`lz4`**. **Prover JSON** is renamed (`getZkL2ExecutionProof`, `getZkRollupProof`, `getZkRollupAggregationProof`) and aligned to the new witnesses and 14/15-field PIs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8422f8425748a14e891db4c5d2b8a689b4b9df8d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rollup_spec/blob.py
```diff
@@ -1,243 +1,670 @@
...
 
-    def blob_hash(self) -> bytes:
+    def blob_hash(self) -> Hash32:
         """
         blob_hash returns the versioned hash of the blob KZG commitment.
         """
-        return kzg_commitment_to_versioned_hash(self.blob_kzg_commitment)
-    
-    def is_authenticated_blob_bytes(self) -> Tuple[bool, Hash32] :
+        return Hash32(kzg_commitment_to_versioned_hash(self.blob_kzg_commitment))
```
AlexandreBelling: I'd advise not having a field storing it in the body of the "self", the reason is that the guest program already has everything it needs to recompute it. The risk of adding it is adding an invariant in the code and an assertion enforcing the invariant.

AlexandreBelling: Talking about self.blob_kzg_commitment

rollup_spec/blob.py
```diff
@@ -1,243 +1,670 @@
...
+                padded,
+                self.blob_kzg_commitment,
+                self.blob_kzg_proof,
+                _trusted_setup(),
             )
+        except Exception as exc:
+            # Malformed witness bytes (invalid G1 encoding, blob length
+            # mismatch, field-element overflow, …) raise from ckzg; surface
+            # them as a proof-verification failure with the original cause.
+            raise Exception("invalid KZG proof") from exc
```
AlexandreBelling: I'd be against a code that "throws" because this is non-trivial to transcribe in R5. Or document what we want to do exactly if this throws. IMO, I think we want to program to early exit and returning some exit code in its output when that happen. That we can prove that the program crash. Another interpretation is that a throws corresponds to un-provable execution flow

rollup_spec/blob.py
AlexandreBelling: And by extension to any intermediate value

rollup_spec/l2_execution.py
```diff
@@ -0,0 +1,472 @@
...
+
+    for ftx in block.forced_transactions:
+        # FTXs are processed in ascending L1-assigned number.
+        if ftx.number != last_processed_ftx_number + 1:
+            raise Exception("forced transactions must be processed in ascending sequence")
+
+        # Deadline constraint (§6.5): the FTX must be handled in a block whose
+        # number does not exceed its declared deadline.
+        if ftx.deadline < block_header.number:
+            raise Exception("deadline exceeded")
```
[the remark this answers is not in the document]
AlexandreBelling: Likely a typo in the readme

rollup_spec/l2_execution.py
```diff
@@ -0,0 +1,470 @@
...
+        if resolved_ftx.acceptance == ForcedTransactionAcceptance.INCLUDED:
+            if not tx_in_block:
+                raise Exception("INCLUDED FTX was not found in the block's transaction list")
+            # The EVM state transition proves the FTX executed validly as part
+            # of the block; nothing more to check at this layer.
+            continue
+
+        if resolved_ftx.acceptance not in (
+            ForcedTransactionAcceptance.BAD_NONCE,
+            ForcedTransactionAcceptance.BAD_BALANCE,
```
AlexandreBelling, line 182: @thedarkjester brought up more forced transaction cases that we were missing : intrinsic-gas-too-low

rollup_spec/blob.py
Filter94: I was looking at the Rust implementations of the EVM where it's common to panic. It probably means that the program just exits with an exitcode != 0. Of course there is a difference between what exactly happens in Python vs Zig on the machine code level, but I think for the spec it's sufficient that throw = panic

rollup_spec/l2_execution.py
Filter94: I thought it will be handled by the contract itself, no?

rollup_spec/rollup.py
```diff
@@ -1,243 +1,176 @@
...
+
+from .blob import AggregatedPublicInput, L2_L1_TREE_DEPTH, ShnarfWitness
+from .execution import hash_address_list, hash_hash_list
 
-BRIDGE_L2L1_MESSAGE_SENT_TOPIC_0 = Bytes32(bytes.fromhex("e856c2b8bd4eb0027ce32eeaf595c21b0b6b4644b326e5b7bd80a1cf8db72e6c"))
-BRIDGE_L1L2_ROLLING_HASH_UPDATED_TOPIC_0 = Bytes32(bytes.fromhex("99b65a4301b38c09fb6a5f27052d73e8372bbe8f6779d678bfe8a41b66cce7ac"))
 
 @dataclass
-class RollupPrivateInput:
+class PlonkVerifier:
```
AlexandreBelling: I am happy with us having a mock model for the contract, but I am not sure we want this to live in the rollup spec, as we would implicitly have to maintain consistency with the contract. 

I think this also lives at a different level compared to the contract, so I may be missing the point.

Filter94: Good point about the consistency, but it makes the whole protocol complete. 

@thedarkjester What's your opinion?

rollup_spec/blob.py
AlexandreBelling: Hum, but are these panics part of the spec of the EVM? Honest question: I don't know. We need to also differentiate between the defensive assertion of the implementation and the "throws" that are part of the EVM spec. Also, I don't know how the zkVM process panics internally. On a normal CPU this is done via interruption,  but I don't know if R5 has interruptions.

rollup_spec/l2_execution.py
AlexandreBelling: From discussing with @thedarkjester , it is not that trivial to do on the contract side.

rollup_spec/blob.py
Filter94: Google says yes

Filter94: I found it in the standards https://github.com/eth-act/zkvm-standards/blob/main/standards/standard-termination-semantics/README.md

rollup_spec/l2_execution.py
Filter94: I caught up with that. I'll wait for the consensus on each question related to the intrinsic gas and will update the spec after that

AlexandreBelling (approved):

PR #3194, opened 2026-05-21
fix(prover): Incorporate `isAllowedCircuitID` into aggregation FPI

Tabaie:
This PR adds `IsAllowedCircuitID` to the list of hashed aggregation public inputs to prevent a dishonest prover from running dummy proofs in production.

---

> [!NOTE]
> **High Risk**
> Changes proof public-input hashing and L1 verifier constructor configuration; a prover/deploy mismatch would reject valid aggregations or weaken circuit allowlisting guarantees.
> 
> **Overview**
> **`IsAllowedCircuitID`** is now part of the aggregation public input and the MiMC chain-configuration hash (prover, gnark circuit, and L1 verifier deploy path), so the allowed-circuit bitmask is cryptographically bound to proofs and must match between prover and on-chain verifier.
> 
> PlonkVerifier deployment requires **`VERIFIER_IS_ALLOWED_CIRCUIT_ID`** (constructor arg `isAllowedCircuitId`); deployment docs and Hardhat test helpers pass it (tests use `0`). Hardhat aggregated-proof and forced-transaction fixtures were regenerated with updated **`aggregatedProofPublicInput`**, proofs, and **`isAllowedCircuitID`** fields. Prover tests assert the mask changes the public input and refresh chain-config golden vectors for the fifth hashed field.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dba64b97f1f384480380ae8fa5a428c7be4e0d10. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AlexandreBelling (approved):

AlexandreBelling:
Approved for the prover side.

Why the changes on the contract side though?

gusiri:
I feel like IsAllowedCircuitID shouldn't be part of the chain config. Chain config parameters are designed to define 'a specific chain', which IsAllowedCircuitID does not do. Introducing it here is unrelated to that purpose and compromises the original design. Furthermore, chain config parameters belong in the [layer2] section of the TOML file, and IsAllowedCircuitID clearly doesn't fit there (we currently have it in [aggregation]). This mismatch confirms that it should not be part of the chain config. It's okay to add them here to minimize modifications though... 

Anyway, this is not a trivial change. This needs to be discussed thoroughly with the contract team and carefully tested across both teams before proceeding.

AlexandreBelling (approved):

AlexandreBelling (changes requested):
Let's rediscuss this before merging

Soleimani193 (changes requested):
The PR adds allowed circuits to the public inputs, but it does not bind them to the circuit. I mean, as PI, it is not bound to the circuit. A malicious prover can just generate the right PI but use the wrong circuits. What prevents this attack?

Tabaie:
> The PR adds allowed circuits to the public inputs, but it does not bind them to the circuit. I mean, as PI, it is not bound to the circuit. A malicious prover can just generate the right PI but use the wrong circuits. What prevents this attack?

`IsAllowedCircuitID` appears twice in the PI Interconnection circuit. As a subfield of the chain configuration, and as a top-level public input. The two are asserted to be equal. The former is incorporated into the configuration PI and hence into the aggregation PI. The latter is extracted and used in the aggregation circuit as the bit mask. So the bit mask _is_ bound to the top-level public input. i.e. if the prover uses the wrong circuits it will be visible in the PI.

Soleimani193:
> > The PR adds allowed circuits to the public inputs, but it does not bind them to the circuit. I mean, as PI, it is not bound to the circuit. A malicious prover can just generate the right PI but use the wrong circuits. What prevents this attack?
> 
> `IsAllowedCircuitID` appears twice in the PI Interconnection circuit. As a subfield of the chain configuration, and as a top-level public input. The two are asserted to be equal. The former is incorporated into the configuration PI and hence into the aggregation PI. The latter is extracted and used in the aggregation circuit as the bit mask. So the bit mask _is_ bound to the top-level public input. i.e. if the prover uses the wrong circuits it will be visible in the PI.

I just checked the changes from the PR, but indeed some part of binding is already in place just hanging was missing.

Soleimani193 (dismissed):

gusiri (approved):

gusiri (dismissed):

AlexandreBelling (dismissed):

contracts/deploy/01_deploy_PlonkVerifier.ts
```diff
@@ -23,6 +23,7 @@ const func: DeployFunction = withSignerUiSession(
     const baseFee = getRequiredEnvVar("VERIFIER_BASE_FEE");
     const coinbase = getRequiredEnvVar("VERIFIER_COINBASE");
     const l2MessageServiceAddress = getRequiredEnvVar("L2_MESSAGE_SERVICE_ADDRESS");
+    const isAllowedCircuitId = getRequiredEnvVar("VERIFIER_IS_ALLOWED_CIRCUIT_ID");
```
thedarkjester, line 33: Please ask the LLM to complete missing documentation.

thedarkjester (changes requested):
we are missing the testcase generation to make the smart contract tests pass.

Also, we should consider integrating this into the new deploys and static artifacts to be prover compatible

gusiri (approved):

contracts/test/hardhat/rollup/helpers/deploy.ts
```diff
@@ -209,6 +209,10 @@ async function deployTestPlonkVerifierForDataAggregation(): Promise<string> {
       value: toBeHex("0x508Ca82Df566dCD1B0DE8296e70a96332cD644ec", 32),
       name: "l2MessageServiceAddress",
     },
+    {
+      value: toBeHex(0n, 32),
```
thedarkjester, line 213: this gives the impression 0 is a valid circuitId ? is this accurate?

gusiri: var GlobalCircuitIDMapping = map[string]uint{
	// Dummy circuits (bits 0-1) - for testing environments
	"execution-dummy":         0,
	"data-availability-dummy": 1,

yes, 0 is a valid circuit ID (execution dev mode).

AlexandreBelling (approved):

thedarkjester (approved):

Issue #3120, opened 2026-05-15
[RV recursion] - implement queries supported in Ray


ivokub:
Verifier-side action for each query type the prover supports.

  Queries are all handled via this pattern:

  fn verify_<query>(claim: <QueryClaim>, transcript: *Transcript, openings: *VortexOpener) !void {
      // 1. Read challenge points from the transcript (Fiat-Shamir, via #3107).
      // 2. Request openings of the relevant columns via the Vortex verifier (#3108).
      // 3. Evaluate the claimed algebraic relation at the opened values.
      // 4. Return error.Rejected if the relation fails; success means the query is consistent.
   }

Implement it for the currently identified queries:
- [ ] log-derivative sum.
- [ ] Lookup.
- [ ] Local constraint.
- [ ] Local opening.
- [ ] Range check.

 Consistency tests vs. `prover-ray`:
 - [ ] take a small honest proof from the Go side (one with at least one example of each query type), run it through the Zig verifier, expect accept.
 - [ ] Mutate the proof bytes systematically (flip a bit in
   each opening, in each claimed value) and expect reject for each mutation.
 - [ ] The fixture set has to cover every query type

ivokub:
Changing in favor of #3192

PR #3202, opened 2026-05-25
feat(verifier-ray): global compiler for verifier-ray

### LINKED ISSUE #3192: [RV recursion] - implement global constraint verification

arijitdutta67:
This PR implements issue(s) #3120 

### Checklist

  PR.


verifier-ray/src/compiler/global.zig
```diff
@@ -0,0 +1,451 @@
...
+    buckets: []CompiledBucket,
+
+    fn checkLagrangeClaims(self: CompiledModule, input: CheckInput, eval_coin: ext.Ext) Error!void {
+        for (self.witness_views, 0..) |view, i| {
+            if (view.column >= input.initial_round.columns.len) return Error.InvalidColumnCount;
+            const column = input.initial_round.columns[view.column];
+            const point = try shiftedPoint(self.spec.size, eval_coin, view.shift);
+            const got = try evaluateVectorAtExt(column.assignment, point);
+            const want = input.witness_claims[self.witness_claim_offset + i];
+            if (!got.eql(want)) return Error.WitnessClaimMismatch;
```
YaoJGalteland: Why do we need to prove witness_claims[i] == Lagrange(column_i, r)? Won't this be proven by delegating authentication to the PCS layer?

A real verifier never receives the raw column vector. In the real protocol:
- prover sends:  commitment(of column_i)  +  witness_claims[i]  +  opening_proof_i
- verifier runs: PCS.verify(commitment, r, witness_claims[i], opening_proof_i)
               → this proves that witness_claims[i] == poly_i(r)

I think the current global query only checks the algebraic identity P_agg(r) = (rⁿ−1)·Q(r) (and trust their evaluation.)

verifier-ray/src/compiler/global.zig
```diff
@@ -0,0 +1,451 @@
...
+        }
+
+        for (self.buckets) |bucket| {
+            for (0..bucket.ratio) |k| {
+                const column_index = bucket.quotient_column_offset + k;
+                if (column_index >= input.quotient_round.columns.len) return Error.InvalidColumnCount;
+                const got = try evaluateVectorAtExt(input.quotient_round.columns[column_index].assignment, eval_coin);
+                const want = input.quotient_claims[bucket.quotient_claim_offset + k];
+                if (!got.eql(want)) return Error.QuotientClaimMismatch;
```
YaoJGalteland: the same as above, why do we need to prove quotient_claims[i] == Lagrange(column'_i, r)?

YaoJGalteland (commented):
Thank you for the implementation! It looks great and works perfectly on my end. I think adding a more step-by-step description would make it clearer what the verifier is doing, such as: 

What the Verifier Checks

**Inputs (trusted after PCS):**
```
witness_claims[i]  = C_i(r)   // each unique column, evaluated at r
quotient_claims[k] = Q_k(r)   // each quotient share, evaluated at r
merge_coin,  r                 // Fiat-Shamir coins
n                              // module size
```

**Step 1 — evaluate each vanishing expression at r:**
```
V_t(r) = eval_expression(V_t, witness_claims)

// e.g. V = C_i· C_j     →  V(r) = witness_claims[i] * witness_claims[j]
// e.g. V = C_k·(C_k−1) →  V(r) = witness_claims[k] * (witness_claims[k] − 1)
```

**Step 2 — apply cancellation:**
```
P_j(r) = V_j(r) · ∏_{k ∈ cancelled_j} (r − ω^k)
```

**Step 3 — aggregate with merge coin:**
```
P_agg(r) = Σ_j  merge_coin^j · P_j(r)
```

**Step 4 — reconstruct Q(r) from shares:**
```
Q(r) = Q_0(r) + r^n·Q_1(r) + r^{2n}·Q_2(r) + …
```

**Step 5 — identity check:**
```
assert  P_agg(r)  ==  (r^n − 1) · Q(r)
```

verifier-ray/src/compiler/global.zig
arijitdutta67: You are absolutely correct. In reality, the global verifier checks only the quotient identity. I’d frame the current code as an interim/golden-vector verifier: useful for proving that prover-ray-generated full vectors and claims match the Zig algebra, but not the final protocol boundary. Having said that it indeed makes sense to remove this checks and simplify the verifier.

YaoJGalteland:
This work keeps the circuit knowledge lives in Zig as runtime data (expression trees interpreted at verify time). I'm working on another approach (suggested by @ivokub) that pushes it entirely into Go at codegen time, leaving the Zig runtime with no circuit knowledge — just arithmetic on already-evaluated values. This aligns with the design goal of a small fixed runtime suitable for a zkVM precompile. The latter approach seems to work as well, I'm cleaning up the code now, and trying to understand the diff.

prover-ray/wiop/compilers/global/global.go
```diff
@@ -548,6 +548,45 @@ type Verifier struct {
 	buckets       []verifierBucket
 }
 
+// VerifierExport is the minimal public view of a compiled global verifier
```
ivokub, line 551: I think we can omit the ..Export structs and just make the fields public.

verifier-ray/src/lib.zig
```diff
@@ -18,6 +18,10 @@ pub const pcs = struct {
     pub const polynomial = @import("pcs/polynomial.zig");
 };
 
+pub const compiler = struct {
+    pub const global = @import("compiler/global.zig");
```
ivokub, line 22: Why compiler package?

verifier-ray/src/compiler/global.zig
```diff
@@ -0,0 +1,451 @@
...
+};
+
+pub const Compiled = struct {
+    allocator: std.mem.Allocator,
+    modules: []CompiledModule,
+    total_witness_claims: usize,
+    total_quotient_claims: usize,
+    total_quotient_columns: usize,
+
+    pub fn deinit(self: *Compiled) void {
```
ivokub, line 82: In principle we can just drop it and leak memory - in zkVM the deallocation doesn't have a point, whatever we have in the lookupbus is already there. This should also mean we could remove the allocator imo from `Compiled`

verifier-ray/src/compiler/global.zig
```diff
@@ -0,0 +1,451 @@
...
+    pub fn Check(self: *const Compiled, allocator: std.mem.Allocator, input: CheckInput) Error!void {
+        if (input.initial_round.next_round_coin_count != self.modules.len) return Error.InvalidCoinCount;
+        if (input.quotient_round.next_round_coin_count != self.modules.len) return Error.InvalidCoinCount;
+        if (input.witness_claims.len != self.total_witness_claims) return Error.InvalidClaimCount;
+        if (input.quotient_claims.len != self.total_quotient_claims) return Error.InvalidClaimCount;
+        if (input.quotient_round.columns.len != self.total_quotient_columns) return Error.InvalidColumnCount;
+
+        var rt = runtime.Runtime.initWithRoundCount(3);
+
+        const merge_coin_buf = try allocator.alloc(runtime.Coin, self.modules.len);
```
ivokub, line 102: Here we currently have runtime allocations -- I think in principle we should be able to know the shapes (i.e. number of modules) already, so we wouldn't do any dynamic allocations, which we would need to account for in the zkvm context where we don't have OS. In principle this could work, but means we would have to implement a custom allocator using the zkc ram.

verifier-ray/src/compiler/global.zig
```diff
@@ -0,0 +1,451 @@
...
+    };
+}
+
+fn shiftedPoint(cardinality: usize, point: ext.Ext, shift: i32) Error!ext.Ext {
+    if (shift == 0) return point;
+    const omega_shift = try rootPower(cardinality, shift);
+    return point.mul(ext.Ext.lift(omega_shift));
+}
+
+fn rootPower(cardinality: usize, shift: i32) Error!field.Element {
```
ivokub: Again cardinality won't be bigger than 2^22 I think so it should be sufficient to use u32

verifier-ray/src/compiler/global.zig
```diff
@@ -0,0 +1,451 @@
...
+
+        const eval_coin_buf = try allocator.alloc(runtime.Coin, self.modules.len);
+        defer allocator.free(eval_coin_buf);
+        const eval_coins = try rt.advanceRoundWithMessage(1, input.quotient_round, eval_coin_buf);
+        if (eval_coins.len != self.modules.len) return Error.InvalidCoinCount;
+
+        for (self.modules, 0..) |module, module_index| {
+            const merge_coin = merge_coins[module_index];
+            const eval_coin = eval_coins[module_index];
+            try module.checkLagrangeClaims(input, eval_coin);
```
ivokub: I don't also understand the Lagrange claim right now.

ivokub (commented):
Looks quite good. It took a bit time to understand, but I think it is really straightforward implementation.

Imo what is missing are a few comments (a la mathematical formulas what we're computing in the global check).

Right now the runtime is dynamic, but perhaps when we code generate the system into Zig (i.e. the representations of vanishing polynomials), then we could use comptime directly and Zig already evaluates at compile time.

verifier-ray/src/compiler/global.zig
arijitdutta67: Sorry @YaoJGalteland already mentioned about it and I forgot to remove it. Removed now.

PR #3220, opened 2026-05-27
feat(arithmetization): benchmark tool for interpreter and native zkc + keccak in rust

amkCha:
Benchmark workflows for 
- zkc interpreter against 2 branches running blake and keccakf guest programs in rust
- native zkc keccakf against 2 versions of zkc

Results in summary look like (disregard actual figures, not testing anything in particular)

For zkc interpreter 

image

For zkc native

image

---

> [!NOTE]
> **Low Risk**
> Changes are confined to arithmetization test examples and CI benchmarking; no production runtime or security-sensitive paths are modified.
> 
> **Overview**
> Adds **GitHub Actions** workflows to compare **base vs optim** on large runners: one runs **`zkc execute`** on Rust **Keccak** and **Blake** guests (git worktrees, warmup, alternating timed iterations, **`cache_thrasher`**, **`/usr/bin/time -v`** logs, **`render_bench_summary`** → step summary + artifacts); another benchmarks **native `zkc exec`** on **`keccakf_batched.zkc`** with two **`go-corset`** refs. The RISC-V setup action now installs **`time`**.
> 
> **Harness changes:** Makefile **`keccak-rust-{build,json,exec}`** and **`blake-rust-{build,json,exec}`**; **`KECCAK_N_VECTORS`** at compile time; new **`keccak`** RISC-V guest; Go helpers for accepts → inputs/JSON and synthetic Blake rounds; **`main.go`** **`@hexFile`** inputs with distinct endian handling for Blake vs Keccak. **`.gitignore`** covers script outputs and stray host binaries. ACT4/Blake runs use quieter **`zkc exec -q`** on success.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d111972941b757b07b245fb6a96f33b2847dbf6a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


arithmetization/src/test/examples/Makefile
```diff
@@ -133,14 +136,25 @@ compile: require-src linker-script
...
+			cargo +nightly rustc \
+				--manifest-path $(MAKEFILE_DIR)rust/Cargo.toml \
+				--target $(MAKEFILE_DIR)rust/riscv64im-unknown-none-elf.json \
+				-Z build-std=core \
+				-Z json-target-spec \
+				--bin $(NAME) \
+				--release \
+				-- -C link-arg=-T$(LINKER_SCRIPT); \
+		fi; \
 	else \
```
lorenzogentile404, line 158: I believe this has an overlap with something I am doing, that is adding a target to run a vector of test cases against a generic test (written in rust, assembly or zig). Also, it feels odd that the compile target handles special cases. I would rather extract this logic in a similar way to what is done for define `define ELF_TO_JSON`.

lorenzogentile404 (dismissed):

arithmetization/src/test/examples/scripts/blake_rounds_to_accepts/main.go
```diff
@@ -0,0 +1,158 @@
+// Generates a synthetic Blake-2f `IN_BYTES` test vector with a chosen number of
```
lorenzogentile404, line 1: This script is going to be affected by https://github.com/Consensys/linea-monorepo/pull/3227, as `IN_BYTES` is expected to be big-endian.

arithmetization/src/test/examples/scripts/blake_rounds_to_accepts/main.go
```diff
@@ -0,0 +1,158 @@
...
+// Output: one `0x<hex>` line on stdout (277 bytes = 554 hex chars), matching
+// the .all format consumed by the blake_accepts_to_in_bytes helper / the
+// blake-rust-json Makefile target.
+//
+// Usage:
+//
+//	blake_rounds_to_in_bytes <rounds>
+//
+// Reference: RFC 7693 / EIP-152. Cross-checked against the rounds=12 test
+// vector documented in blake_with_in_bytes.rs (canonical Blake2b-512("abc")).
```
lorenzogentile404, line 35: This script is to generate new test cases, right?

arithmetization/src/test/examples/scripts/blake_accepts_to_in_bytes/main.go
```diff
@@ -0,0 +1,103 @@
...
+// bytes = 277 bytes = 554 hex chars + the `0x` prefix). Emits the first N
+// validated lines to stdout, one IN_BYTES per line — the natural shape for
+// Blake's per-vector loop (unlike Keccak's single batched blob).
+//
+// Usage:
+//
+//	blake_accepts_to_in_bytes <accepts-file> <n-vectors>
+//
+// `n-vectors == 0` means "all lines". Exits non-zero if any line fails
+// validation or fewer than N vectors are available.
```
lorenzogentile404, line 15: I am not sure I get the purpose of this script. Does it extract lines from a `blake.all` file?

arithmetization/src/test/examples/scripts/keccakf_accepts_to_json/main.go
```diff
@@ -0,0 +1,153 @@
+// Converts a native-zkc keccakf `.accepts` fixture into the input JSON that
```
lorenzogentile404, line 1: For `blake` we did the conversion from the JSON format to the one-test-case-per-line format once, saved the result, and then did not store the script for the conversion. Is it useful to keep this?

arithmetization/src/test/examples/main.go
```diff
@@ -22,11 +22,32 @@ type memoryBlob struct {
 	name   string
 }
 
+// parseInBytes accepts either an inline hex literal / raw string, or @path to a
+// file containing a 0x-prefixed hex blob (used for large batched Keccak inputs
+// that would exceed the shell's per-argument size limit).
+func parseInBytes(arg string) ([]byte, error) {
```
lorenzogentile404, line 42: Useful feature!

lorenzogentile404:
New rust tests have been added, but `arithmetization/src/test/examples/rust/Cargo.toml` has not been updated it seems, so they probably will not run.

arithmetization/src/test/examples/Makefile
lorenzogentile404: Here is the PR I am talking about https://github.com/Consensys/linea-monorepo/pull/3232. Basically, `blake-rust-exec` target was running a vector containing 10 test cases against a specific rust test implementing indeed blake. The PR adds a target called `vector-exec` that generalise this, by taking a vector file and a test (in rust, assembly or zig). Then `blake-rust-exec `just uses the more generic target `vector-exec`.

arithmetization/src/test/examples/scripts/blake_rounds_to_accepts/main.go
amkCha: Yes good question. For benchmark, it made more sense if the user inputs how many rounds they want, so we have to compute on the fly the output and put it into the correct format as I can't store the results for 1 to 100,000 rounds for instance, so that is what the script is doing, going from a number of rounds to the output h so you can store it in a blake.accepts file

arithmetization/src/test/examples/scripts/blake_rounds_to_accepts/main.go
amkCha: Noted, I'm updating it now

arithmetization/src/test/examples/scripts/blake_rounds_to_accepts/main.go
amkCha: I fixed the name of the script to `blake_rounds_to_accepts`

arithmetization/src/test/examples/scripts/blake_accepts_to_in_bytes/main.go
amkCha: Now this script does the bit where you go from .accepts (or .all) file, to IN_BYTES. I split the original blake test in 3 steps ( blake-rust-build, blake-rust-json and blake-rust-exec) so I could reuse from blake-rust-json in the benchmark steps once I've created the custom blake.all file with `blake_rounds_to_accepts`. The benchmark and the test now run the same, it's just that they start with a different .accepts file

amkCha:
> New rust tests have been added, but `arithmetization/src/test/examples/rust/Cargo.toml` has not been updated it seems, so they probably will not run.

It has :) else indeed my benchmark would have never run
image

arithmetization/src/test/examples/scripts/keccakf_accepts_to_json/main.go
amkCha: I thought about how to design things and the scripts for benchmarking are more needed because we ask for a user input, we don't run against the same data all the time, and potentially we might be adding more. Here for this script, I take only n test vectors that I extract from the single line of the .accepts file (the vectors are concatenated together for this version of keccak) and format it in a json

.github/workflows/arithmetization-benchmark-zkc-interpreter.yml
```diff
@@ -0,0 +1,194 @@
...
+          # Helper binaries used by the timing loop and the summary step.
+          go build -o /tmp/bench/cache_thrasher       ./scripts/cache_thrasher/main.go
+          go build -o /tmp/bench/render_bench_summary ./scripts/render_bench_summary/main.go
+
+      - name: Run benchmark (warmup + alternating timed iterations with per-iter cache reset)
+        shell: bash
+        run: |
+          set -uo pipefail
+          mkdir -p /tmp/bench/logs
+          ITERS=${{ inputs.iterations || 0 }}
```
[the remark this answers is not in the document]
amkCha: Done on purpose as it's the default value used when PR is running

arithmetization/src/test/examples/scripts/blake_accepts_to_in_bytes/main.go
DavePearce: FYI --- if I understood it correctly, it looks very similar to unix `tail` command.  However, `tail -n 0` does not return all lines.

arithmetization/src/test/examples/scripts/keccakf_accepts_to_json/main.go
DavePearce: One option here would be to change the format of the `accepts` files to be valid JSON.  The only thing about that is that I use comments with `;;` in the files.  But otherwise, it would be straightforward.

DavePearce (approved):
LGTM.  I think it is adding a reasonable amount of code (which is fine), but just have to keep in mind that we have to maintain it.  Claude obviously helps with this!!  Also, if there are changes we can make upstream (e.g. in `go-corset`) which would simplify the pipelines, then its worth considering them at some point.

arithmetization/src/test/examples/scripts/blake_rounds_to_accepts/main.go
lorenzogentile404: Ok, cool! For those kind of scripts I would something such as `generator` in the name.

arithmetization/src/test/examples/scripts/blake_accepts_to_in_bytes/main.go
lorenzogentile404: Sounds good @amkCha. So in the next PR I could generalise these 3 steps: `vector-build`, `vector-json` and `vector-exec` where the original code is split in 3 generic targets.

arithmetization/src/test/examples/scripts/keccakf_accepts_to_json/main.go
lorenzogentile404: I see, thanks for the clarification.

PR #3270, opened 2026-06-04
fix(verifier-ray): absorb commitments into transcript when present

ivokub:
This PR implements issue(s) #

### Checklist

  PR.

---

> [!NOTE]
> **High Risk**
> Changes how oracle data is hashed into the transcript, which directly affects challenge coins and proof verification; mitigated by prover-ray–sourced golden traces and updated tests.
> 
> **Overview**
> Aligns **verifier-ray** Fiat–Shamir round absorption with what the verifier actually sees: **oracle columns go into the transcript as 8-limb commitments** (`updateElements`), while **public columns and cells still use raw values**.
> 
> `RoundMessage` columns are now a **`ColumnMessage` union** (`oracle_commitment` vs `public_column`) instead of visibility-tagged full assignments. Golden trace generation in Go was reworked to export **Poseidon commitments** for oracle rounds (plus union-shaped column/cell fixtures), and Zig golden/transcript tests were updated to match—including tampering a commitment limb to assert coin divergence.
> 
> The **Makefile** zkc path adds **`prepare-zkc-input`** (reverse-order hex encoding for the ELF-to-JSON helper) and points **`ELF_TO_JSON`** at the script under `arithmetization`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 14f7fd78bf8b7e7779a7aea924be2a0af0739e46. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


verifier-ray/src/runtime.zig
```diff
@@ -20,16 +21,18 @@ pub const Visibility = enum(u8) {
 pub const Vector = value.Vector;
 pub const Scalar = value.Scalar;
 pub const Coin = ext.Ext;
+pub const Commitment = proof_mod.Commitment;
 
-pub const ColumnAssignment = struct {
-    visibility: Visibility,
-    assignment: Vector,
+pub const ColumnMessage = union(enum) {
```
arijitdutta67, line 26: Just a thought, should we change `ColumnMessage` -> `TranscriptMessage`? Saying because commitment is not a column. But maybe we cannot extract it out of `ColumnMessage`, because there is a chance of order mismatch between prover and the verifier in doing so. We need to change the `RoundMessage` struct also if we go with this.

ivokub: Makes sense -- but I would leave it for the later PR where we add a separate TranscriptReplay/TranscriptSpec structure for reusable transcripts passing along verifiers etc. In the PR train right now I'm trying to do minimal PRs by function and leave refactoring/cleanup for later so that it would be better to review individually.

But I do like the idea, it is way cleaner -- I also think we should separate the transcript stuff into a separate package so the separation would be very clean.

arijitdutta67 (dismissed):
LGTM

verifier-ray/Makefile
```diff
@@ -4,14 +4,15 @@ SHELL := /bin/bash
...
 ZKC_MAIN := ../arithmetization/src/main/riscv/main.zkc
 
 PASSING_INPUT_FILE := testdata/inputs/passing.bin
 FAILING_INPUT_FILE := testdata/inputs/failing.bin
 INPUT_FILE ?= $(PASSING_INPUT_FILE)
-# Encode the binary fixture for the ELF-to-JSON helper; the verifier sees decoded bytes.
-IN_BYTES ?= 0x$(shell xxd -p -c 256 $(INPUT_FILE))
+# The ELF-to-JSON helper treats 0x input as a big-endian integer and reverses it
+# before writing guest RAM, so encode binary fixtures in reverse transport order.
+IN_BYTES ?= 0x$(shell xxd -p -c1 $(INPUT_FILE) | tac | tr -d '\n')
```
YaoJGalteland: `tac` is common on Linux, but not installed by default on macOS. the `awk` version is more portable, especially on systems like macOS where tac is not installed by default. But `awk` version stores all lines in memory before printing them in reverse. For small binary inputs, that is fine. For very large files, tac may be more efficient. Do you know the input file size?

verifier-ray/test/golden_test.zig
```diff
@@ -259,6 +272,22 @@ fn expectExt(actual: ext.Ext, expected: ext.Ext) !void {
     try std.testing.expect(actual.eql(expected));
 }
 
+fn extToUints(value: ext.Ext) [6]u32 {
+    return .{
+        value.B0.a0.value,
```
YaoJGalteland, line 277: can we document the ordering for flattening `ext6` elements, and also how to form an `ext6` element from 6 base elements?

The order is also embedded in `fiatshamir.UpdateExt`, but I think it would be good to make it explicit.

verifier-ray/test/golden_test.zig
```diff
@@ -313,39 +362,46 @@ const TraceRoundBacking = struct {
...
+                    if (tamper_first_absorb and !tampered and values.len != 0) {
+                        self.base_values[i][0] = elem(self.base_values[i][0].value ^ 1);
+                        tampered = true;
+                    }
+                    self.columns[column_count] = .{ .public_column = .{ .base = self.base_values[i][0..values.len] } };
+                    column_count += 1;
+                },
+                .public_ext => |values| {
+                    try std.testing.expect(values.len <= max_trace_values);
+                    fillExts(&self.ext_values[i], values);
```
YaoJGalteland, line 394: The oracle and public_base arms both apply the bit-flip when tamper_first_absorb && !tampered. The public_ext arm does not:


```
if (tamper_first_absorb and !tampered and values.len != 0) {
                        self.ext_values[i][0].B0.a0 = elem(self.ext_values[i][0].B0.a0.value ^ 1);
                        tampered = true;
                    }
```

currently, the tamper is applied before public_ext is reached, so `!tampered = false`, no flip is needed.

if a future fixture revision puts public_ext first (e.g., `!tampered = true`, to test a round with no oracle columns), tamper_first_absorb=true would silently produce no modification, the coins would match the untampered run, and the assertion !got[0].eql(expected[0]) would panic with a misleading failure rather than clearly flagging missing coverage.

verifier-ray/test/golden_test.zig
```diff
@@ -313,39 +362,46 @@ const TraceRoundBacking = struct {
...
-                break :assignment .{ .base = self.base_values[i][0..column_case.base_values.len] };
-            };
-
-            self.columns[column_count] = .{
-                .visibility = @enumFromInt(column_case.visibility),
-                .assignment = assignment,
-            };
-            column_count += 1;
+            switch (column_case) {
+                .oracle => |commitments| {
```
YaoJGalteland, line 369: `public_base` and `public_ext` both have this bounds check:

```
try std.testing.expect(values.len <= max_trace_values);
```

but the oracle arm doesn’t seem to check the bounds before incrementing `oracle_commitment_count` and `column_count`.

verifier-ray/Makefile
ivokub: Currently 108 bytes:
```
ls -l testdata/inputs/
total 8
-rw-rw-r-- 1 ivo ivo 108 Jun  4 13:01 failing.bin
-rw-rw-r-- 1 ivo ivo 108 Jun  4 13:01 passing.bin
```

I'm checking that awk is not binary-based and can fail on odd inputs (i.e. containing null bytes). Another recommendation is using python3, I'll try if this works in CI

ivokub: Oh, actually we don't need to handle binary file -- the elf-to-json tool expexts as input hex encoded, so awk should be fine actually.

ivokub: Fixed -- added separate target and removed the option to provide the input as environment variable.

verifier-ray/test/golden_test.zig
ivokub: Thanks! Added tampering for public_ext as well as you recommended. Good catch!

verifier-ray/test/golden_test.zig
ivokub: Thanks! Good finding, also added check for commitment length as well.

verifier-ray/test/golden_test.zig
ivokub: Good point -- I think the extToUints should go into koalabear_ext.zig actually to have the flattening in a single place.

ivokub: What do you think if I instead make the change in the cleanup PR to unblock this PR and the global vanishing PR merging?

YaoJGalteland: yes, please go ahead. I think this fits well as cleanup/doc-related work.

ivokub: Thanks!

YaoJGalteland (approved):
Thank you for the changes, they look good!

PR #3271, opened 2026-06-04
feat(verifier-ray): vanishing query check using codegen

ivokub:
This PR adds code generation to extract the data necessary for generating the Zig-side verifier structure. It also implements the vanishing query check, tests, fixture generation.

Right now the transcript advancing is embedded inside the vanishing query verification -- in an upcoming PR we will clearly separate Transcript into separate structure so it would be better reusable and we have direct separation of steps.

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> New proof-checking logic on quotient identities and Fiat-Shamir coin derivation must stay aligned with prover-ray; witness/quotient claims and dynamic sizes are still trusted fixture inputs until PCS is wired in.
> 
> **Overview**
> Adds **Go codegen** that walks prover-ray `global.Verifier` actions and emits comptime **`vanishing.System`** Zig data, plus a new **`query/vanishing`** checker that replays the transcript, evaluates vanishing expressions, and asserts the PLONK quotient identity **P_agg(r) = (r^n − 1) · Q(r)**.
> 
> The Zig path uses **comptime** system metadata and `inline for` over modules/buckets/expressions so static module sizes specialize `r^n` and cancellation roots; dynamic modules read sizes from proof input. **`powComptime`** on extension fields supports that static path.
> 
> **Fixture generation** extends `testdata/generate` to run full compiler pipelines on `wioptest` scenarios, write **`testdata/generated/vanishing.zig`**, and wire **`test_vanishing`** into the build. **Go tests** cover extraction and Zig rendering; **Zig tests** check honest proofs, invalid quotient failures, and dynamic size validation. Docs describe the global constraint, codegen flow, and PCS integration assumptions (trusted claims/sizes for now; transcript still embedded in vanishing verify).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a8777194f9bf8441569ccf801d26a6f5e7525666. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


verifier-ray/codegen/vanishing.go
```diff
@@ -0,0 +1,261 @@
...
+
+type NamedVanishingSystem struct {
+	Name   string
+	System VanishingSystem
+}
+
+type VanishingSystem struct {
+	SourceName          string
+	Modules             []VanishingModule
+	RoundCoinCounts     []int
```
arijitdutta67, line 35: I think we don't use this slice `RoundCoinCounts` anywhere? Can we remove it? Or I am missing something here?

ivokub: Hmm - lets me trace through. We use it in template generation
```
const system_{{$case.Index}}_round_coin_counts = [_]usize{{intArray $case.System.RoundCoinCounts}};
```
which embeds it into system:
```zig
pub const System = struct {
    modules: []const Module,
    round_coin_counts: []const usize = &.{}, <-----
    round_coin_offsets: []const usize = &.{},
    max_round_coins: usize = 0,
    total_round_coins: usize = 0,
    dynamic_module_count: usize = 0,
    total_witness_claims: usize = 0,
    total_quotient_claims: usize = 0,
};
```
which we use it for runtime/transcript initialization (in vanishing.zig)
```zig
    var rt = runtime.Runtime.initWithRoundCount(system.round_coin_counts.len); <----
    var all_coins: [system.total_round_coins]runtime.Coin = undefined;
    var round_coins: [system.max_round_coins]runtime.Coin = undefined;
    var merge_coins: [system.modules.len]runtime.Coin = undefined;
    var eval_coins: [system.modules.len]runtime.Coin = undefined;

    const merge_advance_index = system.round_coin_counts.len - 3; <----
    const eval_advance_index = system.round_coin_counts.len - 2; <----
```

But you're completely right we won't need to store it -- we already know the offsets and total coin count, so everything is deducible. Here it is only used for validity checking, which imo is kind of pointless indeed.

I agree it makes sense to remove.

ivokub: Actually, double checking we do:
```zig
const next_coin_count = system.round_coin_counts[round_index + 1];
        const message = runtime.RoundMessage{
            .columns = round.columns,
            .cells = round.cells,
            .next_round_coin_count = next_coin_count,
        };
```
to estimate how many coins we need to sample. In principle we could take `coin_offsets[round_index+2]-coin_offsets[round1_index+1]`, but then we would have to start also handling the edge case for the last round where we should instead use `total_coins-coin_offsets[round_index+1]`. It works, but imo ugly and possibly error prone, I would rather keep explicit coin counts as well, imo it is also good description for the users how many coins we need per round.

YaoJGalteland: the coin count per round is determined by the prover-ray protocol definition, specifically by how many `NewCoinField` calls are registered on each round object during protocol **compilation**.

Both prover and verifier need to agree on this count exactly, because it controls how many times `randomExt()` is called before the next round starts. This is shared protocol structure, not a runtime decision.

To me, the verifier reading order should be:

1. load the compiled system, where the round structure is already known: columns, cells, and coin counts
2. read the proof, which constructs the `RoundMessage` for each round
3. for each round:

   * absorb the `RoundMessage` into the transcript
   * squeeze `round_coin_counts[i]` coins, where the count comes from step 1, not step 2
   * check the proof claims against the derived coins

Putting `next_round_coin_count` inside `RoundMessage` seems to conflate these two sources. It makes it look like the squeeze count can vary proof-to-proof, or that the verifier needs to construct it dynamically, while it is actually fixed at compile time.

I think the codegen could register the counts as a compile-time constant and pass them to Runtime.init:

YaoJGalteland (dismissed):
The core implementation is correct and the architecture is solid. we can work on the coin separation in a separate PR.

verifier-ray/codegen/vanishing.go
ivokub: Yup - actually we could make the Runtime type-parametric on the System spec, so we have type safety on that as well. It should ensure that all the sampled coins count are consistent how the system is defined.

verifier-ray/src/query/vanishing.zig
```diff
@@ -0,0 +1,280 @@
...
+
+    var rt = runtime.Runtime.initWithRoundCount(system.round_coin_counts.len);
+    var all_coins: [system.total_round_coins]runtime.Coin = undefined;
+    var round_coins: [system.max_round_coins]runtime.Coin = undefined;
+    var merge_coins: [system.modules.len]runtime.Coin = undefined;
+    var eval_coins: [system.modules.len]runtime.Coin = undefined;
+
+    const merge_advance_index = system.round_coin_counts.len - 3;
+    const eval_advance_index = system.round_coin_counts.len - 2;
+
```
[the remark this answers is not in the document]
ivokub: Good point - but we will separate the transcript reply in a later PR and potentially use comptime information about the number of coins. So will postpone it for now.

ivokub:
I rebased on top of main now (as this PR depended on the absorb commitment PR).

arijitdutta67 (approved):
LGTM

YaoJGalteland (approved):

PR #3275, opened 2026-06-05
feat(prover-ray): Add LagrangeSelectors in the wizard framework

AlexandreBelling:
### Summary

Introduces `LagrangeSelector` as a first-class expression leaf — the public indicator polynomial that is `1` at one row and `0` elsewhere — and rewires the local-constraint compiler to lift via a selector instead of a **precomputed, statically-sized** Lagrange column. This removes a latent limitation: local constraints on **dynamic-size modules** previously could not compile at all.

### Motivation / the bug

The `localvanishing` pass lifts a row-pinned (scalar) predicate into a domain-wide identity by multiplying with a Lagrange indicator. It did this by creating a **precomputed column** of length `m.Size()` with a single `1`. Precomputed columns require a statically-known size, so this approach is fundamentally incompatible with dynamic modules.

This never surfaced because **no test ever combined a local constraint with a dynamic module** — every `NewLocalConstraint` call site targets a `NewSizedModule`, and the dynamic-module tests only use *global* (multi-valued) `NewVanishing`. Run against a dynamic module, the pass panics (`m.Size() == 0` → precomputed column rejected). It was a coverage gap, not a silent miscompilation.

The fix is the `LagrangeSelector`: it carries no assignment and is evaluated from `RuntimeSize`, so it works for both static and dynamic modules.

### What changed

**`LagrangeSelector` (`wiop`)**
- A `VectorPromise` leaf evaluating to `(0,…,1,…,0)`; not committed.
- `EvaluateOutOfDomain(rt, x)` evaluates its low-degree extension `L_p(x) = ω^p·(xⁿ−1)/(n·(x−ωᵖ))`, with an explicit panic if `x` is the in-domain point `ωᵖ` (rather than silently returning `0` via the field's `1/0 = 0`).
- `NewLagrangeSelector` validates the position (non-negative; bounded by size when statically sized).

**Global-quotient compiler (`compilers/global`)**
- Treats `LagrangeSelector` as an expression leaf. Prover: `computeLagrangeSelectorCoset` evaluates `L_p` analytically over the quotient coset (batch-inverted, same parametrization as the cancellation polynomial). Verifier: evaluates it at the random point via `EvaluateOutOfDomain` — **no commitment and no `LagrangeEval` claim** (a net reduction vs. the old committed Lagrange column).

**Local-vanishing compiler (`compilers/localvanishing`)**
- Lifts via `NewLagrangeSelector(m, anchor)` instead of a precomputed column; `newLagrangeColumn` removed. Now size-agnostic.

**`Size()` made fail-fast (`wiop`)**
- `ColumnView.Size()` and `Constant.Size()` now panic on dynamic/unsized modules instead of silently returning `0`, matching `Degree()` and the scalar promises. (`IsSized()` remains the gate callers check first.)

### Tests
- **Unit:** `EvaluateOutOfDomain` (vs. barycentric reference, base + extension points), selector inside an expression, selector inside a `Vanishing`.
- **Global layer in isolation:** hand-written global constraint containing a selector leaf, linear (`ratio == 1`) and quadratic (`ratio > 1`, large-coset path), completeness + soundness.
- **Pipeline:** three new dynamic-module `LocalVanishingScenarios` (first-row, shifted, quadratic) running the full `localvanishing → global` pipeline, completeness + soundness.

### Notes
- No behavior change for static-module local constraints; the lifted polynomial is identical, just computed analytically rather than via FFT of a committed column.

---

> [!NOTE]
> **Medium Risk**
> Changes core quotient prover/verifier expression evaluation and local-constraint lowering; behavior on static modules is intended to be equivalent but the proof path is security-sensitive.
> 
> **Overview**
> Introduces **`LagrangeSelector`** as an uncommitted expression leaf (1 at one domain row, 0 elsewhere) with in-domain **`EvaluateVector`** and out-of-domain **`EvaluateOutOfDomain`** via the Lagrange basis closed form.
> 
> **Local constraints** no longer allocate static precomputed Lagrange columns; **`localvanishing`** lifts row-pinned predicates with **`NewLagrangeSelector`**, which fixes **local constraints on dynamic-size modules** (size comes from **`RuntimeSize`**).
> 
> The **global quotient compiler** treats selectors as expression leaves: the prover evaluates them on the large quotient coset via **`computeLagrangeSelectorCoset`** (no FFT/commitment); the verifier uses **`EvaluateOutOfDomain`** at the random point instead of witness **`LagrangeEval`** claims for a committed indicator column.
> 
> **`ColumnView`**, **`Constant`**, and related **`Size()`** paths now **panic** on dynamic/unsized modules instead of returning 0. Tests cover selector math, hand-written global vanishings (linear and large-coset), and **dynamic-module** local→global pipeline scenarios.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 94e066250a770ef153cf6b1f3e111b21f5224cb8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/prover-ray/wiop/compilers/global/global.go
+++ b/prover-ray/wiop/compilers/global/global.go
@@ -353,6 +353,98 @@ func computeCancellationCoset(cancelled []int, n, N int) []field.Element {
 	return cVals
 }
 
+// computeLagrangeSelectorCoset returns the base-field evaluation of the
+// Lagrange selector polynomial
+//
+//	L_p(X) = ω^p · (X^n − 1) / (n · (X − ω^p))
+//
+// at all N = n·ratio coset points {g · ω_N^j : j = 0…N-1}, where ω =
+// RootOfUnityBy(n), ω_N = RootOfUnityBy(N), g = MultiplicativeGen, and p is
+// `position` normalised into [0, n). This is the polynomial that
+// [wiop.LagrangeSelector] represents (1 at row p, 0 elsewhere on the domain);
+// the same closed form is used pointwise by
+// [wiop.LagrangeSelector.EvaluateOutOfDomain].
+//
+// The coset parametrisation mirrors [computeCancellationCoset] exactly so the
+// result lines up with the FFT-coset evaluations of the witness columns. The
+// denominator never vanishes: a coset point g·ω_N^j is never a pure n-th root
+// of unity, so X − ω^p ≠ 0.
+func computeLagrangeSelectorCoset(position, n, N int) []field.Element {
+	// Normalise the anchor into [0, n).
+	p := ((position % n) + n) % n
+
+	omega := field.RootOfUnityBy(n)
+	var omegaP field.Element
+	field.ExpToInt(&omegaP, omega, p)
+
+	// Constant numerator coefficient ω^p / n.
+	var nInv field.Element
+	nInv.SetUint64(uint64(n))
+	nInv.Inverse(&nInv)
+	var numCoef field.Element
+	numCoef.Mul(&omegaP, &nInv)
+
+	omegaN := field.RootOfUnityBy(N)
+	var g field.Element
+	g.SetUint64(field.MultiplicativeGen)
+
+	// x_j^n = (g·ω_N^j)^n = g^n · (ω_N^n)^j cycles with period ratio = N/n,
+	// advanced incrementally to avoid a per-point exponentiation.
+	var gPowN, omegaNPowN field.Element
+	field.ExpToInt(&gPowN, g, n)           // g^n
+	field.ExpToInt(&omegaNPowN, omegaN, n) // ω_N^n
+	var one field.Element
+	one.SetOne()
+
+	denom := make([]field.Element, N) // x_j − ω^p
+	num := make([]field.Element, N)   // x_j^n − 1
+	x := g
+	xPowN := gPowN
+	for j := 0; j < N; j++ {
+		denom[j].Sub(&x, &omegaP)
+		num[j].Sub(&xPowN, &one)
+		x.Mul(&x, &omegaN)
+		xPowN.Mul(&xPowN, &omegaNPowN)
+	}
+
+	invDenom := make([]field.Element, N)
+	field.VecBatchInvBase(invDenom, denom)
+
+	res := make([]field.Element, N)
+	for j := 0; j < N; j++ {
+		res[j].Mul(&numCoef, &num[j])
+		res[j].Mul(&res[j], &invDenom[j])
+	}
+	return res
+}
+
+// collectLagrangeSelectorPositions records the Position of every
+// [wiop.LagrangeSelector] leaf reachable from expr into out.
+func collectLagrangeSelectorPositions(expr wiop.Expression, out map[int]struct{}) {
+	switch e := expr.(type) {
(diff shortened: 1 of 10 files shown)


prover-ray/wiop/compilers/global/global.go
```diff
@@ -743,12 +847,13 @@ func evalCancellationAtPoint(cancelled []int, n int, r field.Gen) field.Gen {
 	return result
 }
 
-// evalExprAtPoint evaluates a symbolic expression at a pre-computed scalar
-// point using the witness column evaluation map (from LagrangeEval claim cells).
-// Coins and cells are looked up directly from the runtime.
+// evalExprAtPoint evaluates a symbolic expression at the point r using the
+// witness column evaluation map (from LagrangeEval claim cells). Coins and
+// cells are looked up directly from the runtime.
 func evalExprAtPoint(
```
AlexandreBelling, line 853: @ivokub @YaoJGalteland @arijitdutta67 
The changes here are going to impact the verifier-ray? Want me to code the corresponding changes in this PR?

YaoJGalteland: The verifier-ray's appendExpr (in this [PR](https://github.com/Consensys/linea-monorepo/pull/3271/changes#diff-e3fa3cd96a2a12ce13e335876eb1e26b7058f730e64b30777ba11ac477edadea)) switches on the concrete expression type. It has no case for *wiop.LagrangeSelector and hits the default branch:


```
default:
    return 0, &UnsupportedExpressionError{Type: fmt.Sprintf("%T", expr)}
```

I think it’s probably best to update the verifier-ray changes after both #3275 and #3271 are merged.

ivokub: We don't need to do this in this PR -- we pin to prover-ray at particular commit, so the change shouldn't break verifier-ray right now. Verifier side changes can be done later separately after this PR is merged so that we have commit on main we can pin against.

YaoJGalteland:
the flat wiop/ root has around 40 files all in one package. I think it could be split into a few subpackages to make the dependency structure explicit, maybe query, expr.

Tabaie:
Great! I was wondering why the Lagrange basis was being committed to, when it could be easily computed verifier-side. Ideally we do want to integrate that on our side if we want a realistic estimate of verifier costs no? @ivokub @YaoJGalteland

ivokub:
> Great! I was wondering why the Lagrange basis was being committed to, when it could be easily computed verifier-side. Ideally we do want to integrate that on our side if we want a realistic estimate of verifier costs no? @ivokub @YaoJGalteland

Yes, but it doesn't need to be part of this PR. We pin used prover-ray version in verifier by git tag, so can develop prover and verifier independently.

AlexandreBelling:
> the flat wiop/ root has around 40 files all in one package. I think it could be split into a few subpackages to make the dependency structure explicit, maybe query, expr.

That's fine I think. I am more leaning toward that than splitting in more folders because otherwise we had dependency problems in the previous version as a result of splitting.

We can still use thematic naming to help grouping related files in the list but we the sheer number of files per package does make a good argument for splitting I think.

bogdanbear (approved):

PR #3311, opened 2026-06-08
feat(arithmetization): added Zig wrapper for keccak zkc implementation

### LINKED ISSUE #3302: [RV] zig wrapper for `keccak`
# Main point
Write a zkc-keccak-impl wrapper in Zig that implements the [standardized keccak interface](https://github.com/eth-act/zkvm-standards/blob/282cd356c3a0498416bb0619f9c8a347ce9933fb/standards/c-interface-accelerators/zkvm_accelerators.h#L158-L166) using https://github.com/LFDT-Lineth/lineth-monorepo/pull/3040 as a reference (even if written in rust rather than zkc).

## Interface

```C
/**
 * Compute Keccak-256 hash
 *
 * @param data Pointer to input data
 * @param len Length of input data in bytes
 * @param[out] output Pointer to output hash
 * @return ZKVM_EOK on success, ZKVM_EFAIL on failure
 */
zkvm_status zkvm_keccak256(const uint8_t* data, size_t len, zkvm_keccak256_hash* output);
```

## Zig example

With actual zig program being the equivalent in zig of the following in-lined assembly written in Rust:

```rust
fn in_line_assembly_keccak_call(input_offset: usize, input_size: usize, output_offset: usize) {
  unsafe {
    core::arch::asm!(
      ".insn r 0x0b, 0b111, 0b1111111, {2}, {0}, {1}",
      in(reg) input_offset,
      in(reg) input_size,
      in(reg) output_offset,
    );
  }
}
```

lorenzogentile404:


arithmetization/src/test/examples/zig/src/keccak_wrapper.zig
```diff
@@ -0,0 +1,30 @@
...
+    // no OS to return to, signal halt via ecall
+    asm volatile (
+        \\li a0, 0   # exit code 0
+        \\li a7, 93  # syscall number for exit
+        \\ecall
+    );
+    unreachable;
+}
+
+fn in_line_assembly_keccak_call(
```
OlivierBBB: Shouldn't we use the same name as in the standardized interface ?

OlivierBBB: Adapting what Claude writes:

```
implementing zkvm_keccak256 against that C ABI.

  C → Zig type mapping

  ┌───────────────────────────────┬──────────────────────────────────────────┬────────────────────────────────────┐
  │               C               │                   Zig                    │               reason               │
  ├───────────────────────────────┼──────────────────────────────────────────┼────────────────────────────────────┤
  │ zkvm_status (enum, EOK=0,     │ enum(c_int)                              │ the -1 forces the C enum's         │
  │ EFAIL=-1)                     │                                          │ underlying type to signed int      │
  ├───────────────────────────────┼──────────────────────────────────────────┼────────────────────────────────────┤
  │ zkvm_bytes_32 (_Alignas(8)    │ extern struct { data: [32]u8 align(8) }  │ extern for C layout; align(8)      │
  │ uint8_t[32])                  │                                          │ replicates ALIGN8                  │
  ├───────────────────────────────┼──────────────────────────────────────────┼────────────────────────────────────┤
  │ const uint8_t* data           │ [*]const u8 (or ?[*]const u8 to honour   │ unknown-length buffer, paired with │
  │                               │ the panic-on-NULL note)                  │  len                               │
  ├───────────────────────────────┼──────────────────────────────────────────┼────────────────────────────────────┤
  │ size_t len                    │ usize                                    │                                    │
  ├───────────────────────────────┼──────────────────────────────────────────┼────────────────────────────────────┤
  │ zkvm_keccak256_hash* output   │ *zkvm_keccak256_hash (or optional)       │                                    │
  └───────────────────────────────┴──────────────────────────────────────────┴────────────────────────────────────┘

  Interface + implementation

  const std = @import("std");
  const Keccak256 = std.crypto.hash.sha3.Keccak256;

  /// Matches the C enum; underlying type is c_int because of ZKVM_EFAIL = -1.
  pub const zkvm_status = enum(c_int) {
      ZKVM_EOK = 0,
      ZKVM_EFAIL = -1,
  };

  /// zkvm_bytes_32 == zkvm_keccak256_hash: 32 bytes, 8-byte aligned.
  pub const zkvm_bytes_32 = extern struct {
      data: [32]u8 align(8),
  };
  pub const zkvm_keccak256_hash = zkvm_bytes_32;

  /// zkvm_status zkvm_keccak256(const uint8_t* data, size_t len,
  ///                           zkvm_keccak256_hash* output);
  export fn zkvm_keccak256(
      data: ?[*]const u8,
      len: usize,
      output: ?*zkvm_keccak256_hash,
  ) zkvm_status {
      // Spec: "If a function is called with a NULL pointer, the function SHOULD panic."
      const out = output orelse @panic("zkvm_keccak256: output is NULL");
      const buf: []const u8 = if (data) |p|
          p[0..len]
      else if (len == 0)
          &.{} // empty input with NULL data is well-defined
      else
          @panic("zkvm_keccak256: data is NULL");

      // Keccak256.hash(buf, &out.data, .{});
      // do the custom instruction instead

      return .ZKVM_EOK;
  }

  Notes

  - export fn already gives the C calling convention and exports the symbol as zkvm_keccak256 — no callconv needed.
  - I made the pointers optional (?*/?[*]) purely to implement the "SHOULD panic on NULL" clause. If you'd rather
  trust the caller (the header says the caller MUST ensure validity), drop the ? and the orelse branches; non-optional
  Zig pointers can't be null at the type level.
  - std.crypto.hash.sha3.Keccak256 is the legacy/Ethereum Keccak (0x01 padding), not NIST SHA3-256
  (std.crypto.hash.sha3.Sha3_256) — correct for zkvm_keccak256.
  - The function never returns .ZKVM_EFAIL here since Keccak256.hash is infallible; that path only exists for
  accelerators that can actually fail (bad curve points, etc.).
  - To generate the matching C header automatically, build with -femit-h, or for a freestanding zkVM guest target
  compile as a static lib (e.g. addStaticLibrary with .target = riscv...-freestanding) and ship the provided .h.
```

OlivierBBB (approved):
LGTM

PR #3319, opened 2026-06-09
feat(arithmetization): integrate the keccak precompile into the interpreter

### LINKED ISSUE #3301: [RV] integrate `keccak` in the interpreter
This requires updating @amkCha's keccak zkc-implementation to work with RV-ram inputs and RV-ram outputs.
- [ ] rather than a zkc `input` / `output`, it should consume the
  - RV's RAM
  - input pointer
  - input length
  - output pointer
- [ ] porting it to the monorepo
- [ ] integrating it in the interpreter, one can take inspiration from [Poseidon1](https://github.com/LFDT-Lineth/lineth-monorepo/pull/3040)

```rust
// zkc keccak
fn keccak<ram>(input :Address, size :u32, output :Address) {
  ...
}
```

OlivierBBB:


arithmetization/src/main/riscv/main.zkc
```diff
@@ -13,7 +13,7 @@ include "utils/register_utils.zkc"
 include "memory.zkc"
 
 // TODO @Ghost
-fn main<ram, registers>() {
+fn main<registers, ram, state_row0, state_row1, state_row2, state_row3, state_row4, b0, b1, b2, b3, b4, c, d, padded_input_block>() {
     var opcode:Opcode
```
lorenzogentile404, line 17: @DavePearce it seems at the moment we need to propagate all memory variables specific to keccak from main down to keccak implementation. In so, this is not ideal in the long term. We raise an issue for that.

arithmetization/src/main/riscv/memory.zkc
```diff
@@ -18,3 +19,18 @@ pub input blobs_offset_and_size(i:u64) -> (blob_offset:Address, blob_size:Length
...
+memory state_row3(address:u3) -> (state:lane)
+memory state_row4(address:u3) -> (state:lane)
+memory b0(address:u3) -> (state:lane)
+memory b1(address:u3) -> (state:lane)
+memory b2(address:u3) -> (state:lane)
+memory b3(address:u3) -> (state:lane)
+memory b4(address:u3) -> (state:lane)
+memory c(address:u3) -> (state:lane)
+memory d(address:u3) -> (state:lane)
+memory padded_input_block(address:u5) -> (state:lane)
```
lorenzogentile404, line 36: this needs to be moved once we support memory variables being declared at any spot.

lorenzogentile404 (dismissed):
LGTM

arithmetization/src/main/riscv/memory.zkc
OlivierBBB: We can declare memory variables whereever we like. We just have to add the import path.

lorenzogentile404 (approved):
LGTM

ivokub (approved):
LGTM

Issue #3286, opened 2026-06-08
feat: separate transcript reply logic from verifier actions


ivokub:
Right now the transcript reply is very deeply integrated with the verifier action. But it should be a separate step for cleaner separation.

We know the structure of the system, and all the coins etc already at verifier compile time, so we can preallocate all space we need and have compile time checks for ranges etc.

PR #3322, opened 2026-06-09
feat(verifier-ray): split transcript replay from sub-verifier vanishing

YaoJGalteland:
This PR implements issue(s) #3286 

Extracts Fiat-Shamir transcript ownership out of vanishing.verify and into a new shared verifier.verify entry point, so sub-verifiers is only: **given pre-derived coins and cell openings, check the mathematical identity. Nothing else.**

| Concern | Before | After |
|---|---|---|
| Transcript ownership | inside `vanishing.verify` | `protocol.replay`, called once |
| Coin counts | `next_round_coin_count` in `RoundMessage` | comptime `protocol.Spec` |
| Sub-verifier input | `rounds + claims` | `ctx + claims` (math only) |


verifier-ray/codegen/internal/generator/generator.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// Round describes one verifier round in generated output.
+type Round struct {
+	ID              int
+	VerifierActions []string
+}
+
+// Generate emits compilable Zig verifier constants.
+func Generate(system System, dst io.Writer) error {
```
ivokub: I think we shouldn't have several methods for code generation. Here we use the line-based writing and in verifier-ray/codegen/vanishing_zig.go we use template-based.

verifier-ray/codegen/internal/generator/generator.go
```diff
@@ -0,0 +1,117 @@
...
+// System is the codegen input for one compiled verifier. Rounds carry
+// per-round metadata; Vanishing, when non-nil, supplies the compiled
+// vanishing query system and drives coin-routing values in the emitted Spec.
+type System struct {
+	Rounds []Round
+	// Vanishing carries the compiled vanishing system. When non-nil, Generate
+	// uses its RoundCoinCounts for the protocol Spec and emits full module
+	// metadata into the output file. When nil, the spec has zero coins and the
+	// vanishing system is emitted as an empty placeholder.
+	Vanishing *codegen.VanishingSystem
```
ivokub: We embed the VanishingSystem here, but the coin counts for vanishing etc are still embedded in `*codegen.VanishingSystem`. Wouldn't it make sense to extract it into a separate field?

verifier-ray/codegen/internal/generator/generator.go
```diff
@@ -0,0 +1,117 @@
...
+	"strings"
+
+	codegen "github.com/consensys/linea-monorepo/verifier-ray/codegen"
+)
+
+// System is the codegen input for one compiled verifier. Rounds carry
+// per-round metadata; Vanishing, when non-nil, supplies the compiled
+// vanishing query system and drives coin-routing values in the emitted Spec.
+type System struct {
+	Rounds []Round
```
ivokub: Why would we need rounds and inside every round we have `VerifierAction`?

The point of this PR is that we split the transcript replying and different query verifications and the rounds imo only matter for the transcript reply, i.e. in which order we sample the coins. So, would we need the slice Rounds which has internally slice of VerifierActions? Imo it just mimics the structure of the System on the prover side, not what the verifier actually does.

verifier-ray/src/generated/stub.zig
```diff
@@ -0,0 +1,20 @@
+// Code generated by verifier-ray/codegen. DO NOT EDIT.
```
ivokub: I would omit the stub generation for now. Lets just consider the tests and test cases and lets figure out later how we code-generate the IOP system code which will get built.

The point being that we shouldn't have a "dummy" system included in the source code, otherwise it may happen that we accidentally build a verifier which does nothing. Better not to have source code at all so that we would have concrete Zig compile time failure when input is missing.

This was already removed in the scaffolding cleanup PR.

verifier-ray/test/all.zig
```diff
@@ -3,4 +3,5 @@ comptime {
     _ = @import("golden_test.zig");
     _ = @import("transcript_test.zig");
     _ = @import("vanishing_test.zig");
+    _ = @import("generated_stub");
```
ivokub: Can remove

verifier-ray/src/protocol/root.zig
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    comptime if (coin_offsets.len != advance_counts.len + 1)
+        @compileError("coin_offsets must have length advance_counts.len + 1");
+    comptime for (0..advance_counts.len) |i| {
+        if (coin_offsets[i + 1] + advance_counts[i] > total_coins)
+            @compileError("coin_offsets/advance_counts inconsistent with total_coins");
+    };
+
+    const all_coins = try allocator.alloc(Coin, total_coins);
```
ivokub: We don't need to use runtime allocator for memory. In zkvm every time we allocate everything it stays around forever. So we can just predefine the whole memory what we need as `total_coins` is known.

verifier-ray/src/protocol/sampler.zig
```diff
@@ -0,0 +1,59 @@
...
+        ///
+        /// `round_index` is comptime-known; an out-of-bounds index is a
+        /// compile error. Calling with a round index that does not equal the
+        /// current round returns `UnexpectedRound`.
+        pub fn advanceRoundWithMessage(
+            self: *Self,
+            comptime round_index: usize,
+            message: RoundMessage,
+        ) Error![advance_counts[round_index]]Coin {
+            if (self.current_round != round_index) return Error.UnexpectedRound;
```
ivokub: I think in principle this check could be done at comptime as well -- from specification we already know how many times `advanceRoundWithMessage` will be called, so all the updates to `self.current_round` are already known as well.

The problem is that we cannot provide `self` as a comptime argument as it containts the runtime initialized Fiat-Shamir. When we would provide the runtime FS instanse and comptime Sampler separately, then we could have strong compile-time checks that all rounds are correctly defined and in order. This also means that we can reduce the error catching with `try` expressions.

verifier-ray/src/protocol/root.zig
```diff
@@ -0,0 +1,87 @@
...
+    errdefer allocator.free(all_coins);
+
+    inline for (0..advance_counts.len) |advance_index| {
+        const round_coins = try sampler.advanceRoundWithMessage(advance_index, rounds[advance_index]);
```
ivokub: Here we can also just give reference of all_coins to `advanceRoundWithMessage` instead of letting it allocate and we copy here it to our slice.

verifier-ray/src/verifier.zig
```diff
@@ -0,0 +1,107 @@
+const std = @import("std");
+const protocol = @import("protocol/root.zig");
+const vanishing = @import("query/vanishing.zig");
+const ext = @import("field/koalabear_ext.zig");
+// TODO(new-sub-verifier): add import here — step 1 above.
+
+// ── Adding a new sub-verifier (e.g. logderiv, rangecheck) ────────────────────
```
ivokub, line 6: Nice! Very clean separation!

verifier-ray/build.zig
```diff
@@ -41,6 +41,14 @@ pub fn build(b: *std.Build) void {
             .{ .name = "verifier_ray", .module = verifier_mod },
         },
     });
+    const generated_stub_mod = b.addModule("generated_stub", .{
```
ivokub: I also guess we can remove it for now

verifier-ray/test/vanishing_test.zig
```diff
@@ -4,18 +4,19 @@ const fixtures = @import("test_vanishing");
...
 test "vanishing quotient honest scenarios match prover-ray" {
     try std.testing.expect(fixtures.scenarios.len > 0);
     inline for (fixtures.scenarios) |case| {
         const system = case.system;
         var arena = std.heap.ArenaAllocator.init(std.testing.allocator);
         defer arena.deinit();
-        const input = try proofInput(arena.allocator(), case.honest);
-        try vanishing.verify(system, input);
+        const proof = try buildProofData(arena.allocator(), case.honest);
+        try verifier.verify(arena.allocator(), specOf(system), .{ .vanishing = system }, proof);
```
ivokub: I would keep these tests local to vanishing query tests -- i.e. we call vanishing.verify with already prepared coins. Right now we're calling the full `verifier.verify` method which, when we add more queries later, may start failing on paths unrelevant to vanishing queries. Similarly also below for other tests.

verifier-ray/test/vanishing_test.zig
```diff
@@ -44,52 +48,119 @@ test "dynamic vanishing module sizes are required and validated" {
...
+                .all_coins = &[_]protocol.Coin{},
+                .rounds = &[_]protocol.RoundMessage{},
+            },
+            .witness_claims = &[_]ext.Ext{},
+            .quotient_claims = &[_]ext.Ext{},
+        }),
+    );
+}
+
+test "verifier frees replay coins after verification" {
```
ivokub: When we remove the allocator for sampled coin array allocation, then this test becomes irrelevant.

verifier-ray/test/vanishing_test.zig
```diff
@@ -44,52 +48,119 @@ test "dynamic vanishing module sizes are required and validated" {
...
+    // mismatch. Assert at least one case does to confirm the check is live.
     try std.testing.expect(wrong_size_failures > 0);
 }
 
-fn proofInput(allocator: std.mem.Allocator, proof: fixtures.VanishingProofView) !vanishing.CheckInput {
-    const witness_claims = try allocator.alloc(ext.Ext, proof.witness_claims.len);
-    for (proof.witness_claims, 0..) |claim, i| {
-        witness_claims[i] = ext.Ext.fromUints(claim);
-    }
+test "verifier rejects malformed round counts" {
```
ivokub: This test is more relevant for `verifier.verify`, not for vanishing. I'd move to another file.

ivokub (changes requested):
It looks really nice! I think the separation is very clear and makes a lot of sense. Thanks!

Though I don't understand completely what is happening in the new `codegen/internal/generator` package or why it is needed. I see that we need to amend the code generation, but imo right now it seems very ad-hoc (different generation style).

Particularly - we would need to add some tests and testdata generation as well imo. Right now the `verifier.verify` tests utilize the `vanishing_test.zig` file, but I think it should be separate. And I think we could prepare testdata generation as well for the full spec/system into `verifier_test.zig` perhaps.

Finally - for coin sampling we can omit using the allocator as all the sizes of slices are already known at compile time. In zkvm having an allocator doesn't make a sense as we never de-allocate anyway. And in this case having everything known upfront just makes it a bit easier to read and avoid some odd possible errors in the future.

And also perhaps we should separate the "Sampler" and the underlying "Fiat-SHamir", so that we could have comptime round advancing checks separate from runtime Fiat-Shamir instance.

And finally, I don't have a strong preference, but why choose Sampler vs Transcript? You can keep Sampler though, it is also clear enough.

verifier-ray/src/verifier.zig
```diff
@@ -0,0 +1,107 @@
+const std = @import("std");
+const protocol = @import("protocol/root.zig");
+const vanishing = @import("query/vanishing.zig");
+const ext = @import("field/koalabear_ext.zig");
+// TODO(new-sub-verifier): add import here — step 1 above.
```
arijitdutta67: above -> below

verifier-ray/docs/verifier-design.md
```diff
@@ -0,0 +1,155 @@
...
+│  3. dispatch — call each sub-verifier        │
+└────────────────┬─────────────────────────────┘
+                 │ protocol.Context
+                 │  all_coins  []Coin
+                 │  rounds     []RoundMessage
+                 ▼
+        ┌────────────────┬──────────────────┐
+        ▼                ▼                  ▼
+  vanishing.verify  logderiv.verify   ...future...
+  (math only)       (math only)
```
arijitdutta67: May be: `math` -> `polynomial identity`

arijitdutta67 (dismissed):
Looks good to me. Thanks for doing this

verifier-ray/codegen/internal/generator/generator.go
YaoJGalteland: cool. The template-based approach seems like a good tradeoff here, easy to read, and the generated output is mostly static.

verifier-ray/docs/verifier-design.md
YaoJGalteland: good point. I think constraint identity is probably more specific to verifier checks.

verifier-ray/src/protocol/sampler.zig
YaoJGalteland: This check couldn’t really be done at comptime because current_round was a runtime field.

I changed it as you suggested and split the transcript out of Sampler. Sampler has no runtime state anymore, so this check is no longer needed.

The old current_round field existed to detect if someone called advanceRoundWithMessage(2, ...) before advanceRoundWithMessage(1, ...). With inline for, that misuse is structurally impossible — the compiler generates exactly the right sequence of calls, in order, and there is no variable an outside caller could tamper with.

verifier-ray/src/protocol/root.zig
```diff
@@ -0,0 +1,82 @@
...
+            @compileError("coin_offsets/advance_counts inconsistent with total_coins");
+    };
+
+    var transcript = fiat_shamir.Transcript.init();
+    var all_coins: [total_coins]Coin = undefined;
+
+    inline for (0..advance_counts.len) |advance_index| {
+        const offset = coin_offsets[advance_index + 1];
+        const count = advance_counts[advance_index];
+        Sampler(advance_counts).advanceRoundWithMessage(advance_index, &transcript, rounds[advance_index], all_coins[offset..][0..count]);
```
ivokub: Now when Sampler doesn't hold the transcript instance anymore, then do we really need the separate Sampler? I think we can essentially inline the `advanceRoundWithMessage` here now?

verifier-ray/src/protocol/root.zig
```diff
@@ -0,0 +1,82 @@
...
+///
+/// Parameters expected by callers:
+///   `advance_counts`  — `round_coin_counts[1..]`: squeeze counts starting from
+///                       round 1 (round 0 always produces 0 coins).
+///   `coin_offsets`    — full `round_coin_offsets` (unsliced); the loop uses
+///                       `coin_offsets[advance_index + 1]`, so the slice must
+///                       have length `advance_counts.len + 1`.
+///   `total_coins`     — `system.total_round_coins`; length of the returned slice.
+pub fn replay(
+    comptime advance_counts: []const usize,
```
ivokub: could take as input comptime spec: Spec instead. All the fields we provide are part of Spec already.

verifier-ray/codegen/spec_zig.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	return tmpl.Execute(w, specZigData{Options: opts, Routing: routing})
+}
+
+type specZigData struct {
+	Options SpecZigOptions
+	Routing CoinRouting
+}
+
+const specZigTemplate = `{{if .Options.EmitHeader}}// Code generated by verifier-ray/testdata/generate; DO NOT EDIT.
```
ivokub, line 45: This imo is way cleaner and more succinct way to code generate the spec.

verifier-ray/codegen/vanishing.go
```diff
@@ -115,16 +115,16 @@ type viewKey struct {
...
-	for i, round := range sys.Rounds {
-		out.RoundCoinOffsets[i] = out.TotalRoundCoins
-		out.RoundCoinCounts[i] = len(round.Coins)
-		out.TotalRoundCoins += len(round.Coins)
-		if len(round.Coins) > out.MaxRoundCoins {
-			out.MaxRoundCoins = len(round.Coins)
+func BuildVanishingSystem(sys *wiop.System, routing CoinRouting) (VanishingSystem, error) {
+	out := VanishingSystem{
+		SourceName:       sys.Context.Path(),
+		RoundCoinCounts:  routing.RoundCoinCounts,
```
ivokub: I think when we don't add the `round_coin_counts`, `round_coin_offsets` and `total_round_coins` into vanishing.System anymore then we don't need to store them in this structure we use for code generation as well.

Though we need access to CoinRouting to be able to write down the indices of the merge and eval coins in the flattened coin list.

verifier-ray/src/protocol/root.zig
YaoJGalteland: good point! Since `advanceRoundWithMessage` is only called once from `replay`, the absorb-and-squeeze logic could probably be implemented directly in `replay`'s `inline for` loop instead.

verifier-ray/src/protocol/root.zig
YaoJGalteland: Indeed! we can drop those counts, offsets and total_coins, just take Spec instead

verifier-ray/codegen/vanishing.go
```diff
@@ -183,6 +185,26 @@ func BuildVanishingSystem(sys *wiop.System) (VanishingSystem, error) {
 	return out, nil
 }
 
+// flatCoinIndex returns the absolute index of coin in the flat all_coins slice
+// described by routing. It reads the round index and within-round position
+// directly from coin.Context.ID, so it is correct regardless of which rounds
+// the vanishing compiler chose for merge and eval coins.
+func flatCoinIndex(routing CoinRouting, coin *wiop.CoinField) (int, error) {
```
ivokub, line 192: We could also in principle use this function for assining the indices below in appendExpr function as well.

Or we can just inline the implementation of this function into BuildVanishingSystem without error checking. We just essentially need `idx := routing.RoundCoinOffsets[roundIdx] + posInRound`

ivokub (dismissed):
Thanks, looks good! A small comment in code generation if you'd like to implement. But it is stylistic only.

ivokub (approved):

arijitdutta67 (approved):

PR #3329, opened 2026-06-09
feat(arithmetization): memory from u8 to u32 for interpreter efficiency

amkCha:
This PR implements issue(s) #

### Checklist

  PR.

### DIFF
--- a/arithmetization/src/main/riscv/ram/read.zkc
+++ b/arithmetization/src/main/riscv/ram/read.zkc
@@ -12,14 +12,14 @@ include "../utils/constants.zkc"
 // 5. Return the byte(s) value(s) via a mask operation
 
 fn read_8<ram>(address:Address) -> (value:u8) {
-    var word_idx:Address = address >> 2
-    var offset:Address = address & 3
+    var word_idx:Address = address / 4
+    var offset:Address = address % 4
     value = ((ram[word_idx] >> (offset * 8)) & 0xFF) as u8
 }
 
 fn read_16<ram>(address:Address) -> (value:u16) {
-    var word_idx:Address = address >> 2
-    var offset:Address = address & 3
+    var word_idx:Address = address / 4
+    var offset:Address = address % 4
     switch offset {
         case 0: {
             value = (ram[word_idx] & 0xFFFF) as u16
@@ -41,8 +41,8 @@ fn read_16<ram>(address:Address) -> (value:u16) {
 }
 
 fn read_32<ram>(address:Address) -> (value:u32) {
-    var word_idx:Address = address >> 2
-    var offset:Address = address & 3
+    var word_idx:Address = address / 4
+    var offset:Address = address % 4
     switch offset {
         case 0: {
             value = ram[word_idx]
@@ -64,8 +64,8 @@ fn read_32<ram>(address:Address) -> (value:u32) {
 }
 
 fn read_64<ram>(address:Address) -> (value:u64) {
-    var word_idx:Address = address >> 2
-    var offset:Address = address & 3
+    var word_idx:Address = address / 4
+    var offset:Address = address % 4
     if offset == 0 {
         value = ram[word_idx + 1]::ram[word_idx]
     } else {
--- a/arithmetization/src/main/riscv/ram/read.zkc
+++ b/arithmetization/src/main/riscv/ram/read.zkc
@@ -14,7 +14,8 @@ include "../utils/constants.zkc"
 fn read_8<ram>(address:Address) -> (value:u8) {
     var word_idx:Address = address / 4
     var offset:Address = address % 4
-    value = ((ram[word_idx] >> (offset * 8)) & 0xFF) as u8
+    var trash: u24
+    trash::value = ram[word_idx] >> (offset * 8)
 }
 
 fn read_16<ram>(address:Address) -> (value:u16) {
@@ -22,15 +23,18 @@ fn read_16<ram>(address:Address) -> (value:u16) {
     var offset:Address = address % 4
     switch offset {
         case 0: {
-            value = (ram[word_idx] & 0xFFFF) as u16
+            var trash: u16
+            trash::value = ram[word_idx]
             return
         }
         case 1: {
-            value = ((ram[word_idx] >> 8) & 0xFFFF) as u16
+            var trash: u16
+            trash::value = ram[word_idx] >> 8
             return
         }
         case 2: {
-            value = (ram[word_idx] >> 16) as u16
+            var trash: u16
+            trash::value = ram[word_idx] >> 16
             return
         }
         default: {
--- a/arithmetization/src/main/riscv/ram/read.zkc
+++ b/arithmetization/src/main/riscv/ram/read.zkc
@@ -22,23 +22,15 @@ fn read_16<ram>(address:Address) -> (value:u16) {
     var word_idx:Address = address / 4
     var offset:Address = address % 4
     switch offset {
-        case 0: {
-            var trash: u16
-            trash::value = ram[word_idx]
-            return
-        }
-        case 1: {
-            var trash: u16
-            trash::value = ram[word_idx] >> 8
-            return
-        }
-        case 2: {
-            var trash: u16
-            trash::value = ram[word_idx] >> 16
+        case 3: {
+            var trash: u48
+            // Concatenate highest byte of (word_idx) and lowest byte of (word_idx + 1)
+            trash::value = (ram[word_idx + 1]::ram[word_idx]) >> 24
             return
         }
         default: {
-            value = ((ram[word_idx] >> 24) | ((ram[word_idx + 1] & 0xFF) << 8)) as u16
+            var trash: u16
+            trash::value = ram[word_idx] >> (offset * 8)
             return
         }
     }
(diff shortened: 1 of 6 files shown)


arithmetization/src/main/riscv/ram/read.zkc
```diff
@@ -1,14 +1,84 @@
...
-    value = ram[address + 3]::ram[address + 2]::ram[address + 1]::ram[address + 0]
+fn read_8<ram>(address:Address) -> (value:u8) {
+    var word_idx:Address = address >> 2
+    var offset:Address = address & 3
+    if offset == 0 {
+        value = (ram[word_idx] & 0xFF) as u8
+        return
+    }
+    if offset == 1 {
+        value = ((ram[word_idx] >> 8) & 0xFF) as u8
```
DavePearce: instead of have three cases for misaligned reads, you could have just one:

```
value = ((ram[word_idx] >> (offset*8)) & 0xFF) as u8
```

Its just less stuff, and performance is not important since misaligned does not occur.

arithmetization/src/main/riscv/ram/read.zkc
```diff
@@ -1,14 +1,84 @@
...
 
 fn read_16<ram>(address:Address) -> (value:u16) {
-    value = ram[address + 1]::ram[address + 0]
+    var word_idx:Address = address >> 2
+    var offset:Address = address & 3
+    if offset == 0 {
+        value = (ram[word_idx] & 0xFFFF) as u16
+        return
+    }
+    if offset == 1 {
```
DavePearce: You could use a `switch` and collapse cases for `offset==1` and `offset==2` together.  Not sure if its worth it though, tbh.  Maybe at least using a `switch` instead of the `if` chain is an improvement.

arithmetization/src/main/riscv/ram/read.zkc
```diff
@@ -1,14 +1,84 @@
...
 }
 
+fn read_32<ram>(address:Address) -> (value:u32) {
+    var word_idx:Address = address >> 2
+    var offset:Address = address & 3
+    if offset == 0 {
+        value = ram[word_idx]
+        return
+    }
+    if offset == 1 {
```
DavePearce: Again, a `switch` at least would be better.  You can collapse misaligned cases down to make it a little tidier, but as before not a big deal.

arithmetization/src/main/riscv/ram/write.zkc
```diff
@@ -1,52 +1,86 @@
...
-
-    v1::v0 = value
-
-    ram[address + 0] = v0
-    ram[address + 1] = v1
+    var word_idx:Address
+    var offset:Address = address & 3
+    if offset == 0 {
+        word_idx = address >> 2
+        ram[word_idx] = (ram[word_idx] & 0xFFFF0000) | (value as u32)
```
DavePearce: Worth noting that you can do this another way.  Something like this I guess:

```
hi::lo = ram[word_idx]
ram[word_idx] = hi::value
```

Not sure which is better in the long run.

DavePearce (dismissed):
Overall, it looks good to me.  I have nothing that I think must be changed, only some minor comments really.

arithmetization/src/main/riscv/ram/read.zkc
amkCha: I replaced with a switch and kept the `offset==1` and `offset==2` separate as there was a small optimization where in the latter, you don't need a mask :ok_hand:

arithmetization/src/main/riscv/ram/read.zkc
amkCha: done !

amkCha: (even if offset == 3  didn't need the mask, it makes it much clearer/cleaner)

arithmetization/src/main/riscv/ram/write.zkc
amkCha: Applied !  and after checking we do gain a thousand steps with destructuring on 100 keccaks

arithmetization/src/main/riscv/ram/read.zkc
```diff
@@ -1,14 +1,76 @@
...
+// 3. Load the word from RAM at the calculated word index word_idx
+// 4. Shift the word right by the offset * 8 bits to get the byte as the least significant byte
+// 5. Return the byte(s) value(s) via a mask operation
 
-fn read_32<ram>(address:Address) -> (value:u32) {
-    value = ram[address + 3]::ram[address + 2]::ram[address + 1]::ram[address + 0]
+fn read_8<ram>(address:Address) -> (value:u8) {
+    var word_idx:Address = address >> 2
+    var offset:Address = address & 3
+    value = ((ram[word_idx] >> (offset * 8)) & 0xFF) as u8
```
letypequividelespoubelles: I think a destruct here is better (ie:
trash: u24
value = trash::((ram[word_idx] >> (offset * 8))

letypequividelespoubelles: same for other places btw

arithmetization/src/main/riscv/ram/read.zkc
```diff
@@ -1,14 +1,76 @@
...
 }
 
 fn read_16<ram>(address:Address) -> (value:u16) {
-    value = ram[address + 1]::ram[address + 0]
+    var word_idx:Address = address >> 2
```
letypequividelespoubelles: stupid question, hy not using zkc native / and % ?

arithmetization/src/main/riscv/ram/read.zkc
```diff
@@ -1,14 +1,76 @@
...
+        }
+        case 1: {
+            value = ((ram[word_idx] >> 8) & 0xFFFF) as u16
+            return
+        }
+        case 2: {
+            value = (ram[word_idx] >> 16) as u16
+            return
+        }
+        default: {
```
letypequividelespoubelles, line 31: default case is only 3 right ? 

I guess it generates less instruction than case 3 + default fail, but at least add a comment to explain why you wrote this.

arithmetization/src/main/riscv/ram/read.zkc
```diff
@@ -1,14 +1,76 @@
...
+            value = ((ram[word_idx] >> 24) | ((ram[word_idx + 1] & 0xFF) << 8)) as u16
+            return
+        }
+    }
 }
 
+fn read_32<ram>(address:Address) -> (value:u32) {
```
letypequividelespoubelles, line 39: all of this PR is only a win if the huge majority of case is aligned ?

letypequividelespoubelles (dismissed):
left some perf comment (use destruct instead of 0xFF/0xFFFF/ etc masks, use zkc native / and % when available).

For benchmarking, be sure to use lower-zknative flag to have an accurate estimate of the nb of instruction dealt by corset

arithmetization/src/main/riscv/ram/read.zkc
amkCha: Yes, we put out this PR after checking on blake and keccak, there are 0 misaligned addresses. We handle the case to be sure, and as we still have the guest program to inspect :ok_hand:

amkCha:
> For benchmarking, be sure to use lower-zknative flag to have an accurate estimate of the nb of instruction dealt by corset

For now, as we are targeting to improve the execution time, and lower-native takes 50x more time to run, I'm not sure it's that important. Once we reach a more reasonable time and we refine the step counting, we should definitely do the switch

arithmetization/src/main/riscv/ram/read.zkc
amkCha: because I didn't know we had these operators :) ! I often check the README to know what I can use `zkc/docs/ZKC_LANGUAGE.md` and it's not updated yet I think

letypequividelespoubelles:
> For now, as we are targeting to improve the execution time, and lower-native takes 50x more time to run, I'm not sure it's t

well, imho it is, for exemple trash::a and a a&0xFF will create the same number of instruction without `lower-zknative`, but for trace generation, the flag `lower-zknative` must be on, and then a&0xFF is at least more expensive. 
So if the aim is to benchmark the preflight (aka fast execution) we don't care, but imho we want to benchmark the trace generation (ie with lower-zknative on)

amkCha:
> well, imho it is, for exemple trash::a and a a&0xFF will create the same number of instruction without `lower-zknative`, but for trace generation, the flag `lower-zknative` must be on, and then a&0xFF is at least more expensive. So if the aim is to benchmark the preflight (aka fast execution) we don't care, but imho we want to benchmark the trace generation (ie with lower-zknative on)

yes I understand - to continue the discussion, the number of instructions are already smaller when I do destructuring vs the mask, I already saw an improvement (I answered it in one of David's comment as well). Do you see another example I can check ?

arithmetization/src/main/riscv/ram/read.zkc
amkCha: Sry, only did destructuring for writes, should be good for reads now as well !

arithmetization/src/main/riscv/ram/read.zkc
amkCha: done

arithmetization/src/main/riscv/ram/read.zkc
letypequividelespoubelles: noted :)

letypequividelespoubelles:
> > well, imho it is, for exemple trash::a and a a&0xFF will create the same number of instruction without `lower-zknative`, but for trace generation, the flag `lower-zknative` must be on, and then a&0xFF is at least more expensive. So if the aim is to benchmark the preflight (aka fast execution) we don't care, but imho we want to benchmark the trace generation (ie with lower-zknative on)
> 
> yes I understand - to continue the discussion, the number of instructions are already smaller when I do destructuring vs the mask, I already saw an improvement (I answered it in one of David's comment as well). Do you see another example I can check ?

nice. There was the one aboit / and %, I ay have mised other, but should be ok for now

letypequividelespoubelles (dismissed):
Lgtm

letypequividelespoubelles (approved):

PR #3333, opened 2026-06-10
feat(verifier-ray): benchmark for global constraint, transcript generation, poseidon2 compress

### LINKED ISSUE #3122: [RV verifier] - measure verifier operations
Count what the M1 verifier actually does on a real proof.

1. Instrument the native verifier to count invocations of each operation type.
2. Run the instrumented build against representative proofs — the fibonacci / pythagorean_triplet scenarios used in #3102, plus at least one larger Ray proof if available.
3. Record counts per operation type so #3123 has data to compare against.

**Acceptance items**
Instrumentation hooks in the native verifier emit counts for: 
- [ ] Poseidon compressions
- [ ] field operations
- [ ] extension operations
- [ ] polynomial evaluation (Lagrange, canonical form)
- [ ] Counts collected for the M1 reference proof and at least one larger Ray proof.
- [ ] Results recorded somewhere referenceable for #3123 — table in this doc, in the roadmap notes, or as a markdown report in the repo.

arijitdutta67:
## Description
Resolves issue #3122 .

### Checklist

  PR.


YaoJGalteland (commented):
Nice, this is a well-structured benchmark harness, and the comparison report is genuinely useful.

A few thoughts:

1. `bench/vanishing-benchmark-comparison.md` currently reports one `cycle_count` per scenario. That covers all of `vanishing.verify` plus the small entry/exit path, but there is no breakdown. This lets us rank scenarios, but it doesn’t really show where the cycles go, so it is hard to use the report to drive optimization.

   I think it would be useful to add coarse phase counters, at least:

   * Fiat-Shamir transcript
   * quotient identity verify

   Ideally we could also split expression evaluation vs final batch.

   The motivating case is `RangeCheckCompiler/LargeBound`: `7,236,804` cycles, around 44x the floor and `+4288%` vs the fastest case. Right now we can’t tell whether that is transcript hashing, expression evaluation, or the quotient check. A coarse phase split would make that immediately visible, and help tell whether this is expected or a bug.

2. We could also analyze perf by drawing a cycles-vs-size curve. Right now the size axis is a few hand-picked points tied to different constraints, like Fibonacci size 8, LargeFibonacci size 16, and SizeThirtyTwoCubic size 32.

   We can already see size matters: the marginal cost above the fixed floor jumps from ~97k at size 8 to ~320k at size 16, which is 3.3x for a 2x size bump.

   A proper power-of-two sweep on one fixed constraint, e.g. 4/8/16/32/64/128, would give a clearer cycles-vs-size curve. That would help confirm whether the behavior is `n log n` or worse, and explain the large-size outliers.

3. Only 3 of 83 cases exercise a dynamic module: `DynamicFibonacci`, `LeftPadDynamic`, and `DynamicQuadratic`. All three are single-module with `dynamic_module_count = 1`.

   Both padding directions are covered individually, which seems fine:

   * right padding: `DynamicFibonacci` / `DynamicQuadratic`
   * left padding: `LeftPadDynamic`

   What is not covered is:

   * multiple dynamic modules
   * dynamic + static mix in one system
   * mixed padding directions within one system
   * size edge cases like size 1 or non-power-of-two

   Maybe these are not needed for perf evaluation? but they could be useful for correctness coverage.

verifier-ray/docs/vanishing-benchmark.md
```diff
@@ -0,0 +1,66 @@
+# Vanishing Benchmarking
+
+This document explains how verifier-ray measures the RISC-V cost of the vanishing verifier. The generated single-case result lives in `bench/vanishing-benchmark.md`, and the generated comparison report lives in `bench/vanishing-benchmark-comparison.md`.
+
+## Technique
+
+The benchmark builds a tiny R5 guest from `bench/vanishing_main.zig`. That guest imports one generated vanishing fixture, calls `vanishing.verify(selected_case.system, selected_case.input)`, and exits with code `0` on success or `1` on verifier rejection.
```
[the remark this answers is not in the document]
arijitdutta67: That is a valid point. Either we should create a sub issue for the original issue or include benchmark for (Poseidon compressions, field/ext ops, poly evals, etc.). Maybe we can try the latter in a separate PR? Let me know what you think @ivokub @YaoJGalteland @Tabaie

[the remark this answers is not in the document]
YaoJGalteland: I think a separate PR would be clearer.

The current one can focus on comparing different scenarios, and the follow-up can dig into the breakdown for a few specific scenarios to see how much each crypto operation costs.

verifier-ray/Makefile
```diff
@@ -142,5 +171,63 @@ zkc-verify-failing-expected:
...
+	zig build --release=$(VANISHING_BENCH_RELEASE) -Dstrip=true -Dr5=true -Dvanishing-bench=true -Dvanishing-case=$(VANISHING_BENCH_CASE) -Dvanishing-invalid=$(VANISHING_INVALID)
+
+bench-vanishing-json: bench-vanishing-build
+	go run $(ELF_TO_JSON) $(VANISHING_BENCH_BIN) "" $(IN_BYTES_OFFSET) > $(VANISHING_BENCH_JSON)
+
+bench-vanishing-zkc: bench-vanishing-json
+	mkdir -p $(dir $(VANISHING_BENCH_LOG))
+	set -o pipefail; zkc exec -q $(VANISHING_BENCH_JSON) $(VANISHING_BENCH_ZKC_MAIN) | tee $(VANISHING_BENCH_LOG)
+
+bench-vanishing-zkc-failing:
```
ivokub: I think we can simplify the bench -- we wouldn't need to handle the failing cases.

verifier-ray/Makefile
```diff
@@ -142,5 +171,63 @@ zkc-verify-failing-expected:
 		echo "zkc-verify-failing failed as expected"; \
 	fi
 
+bench-vanishing-build: generate-testdata
+	zig build --release=$(VANISHING_BENCH_RELEASE) -Dstrip=true -Dr5=true -Dvanishing-bench=true -Dvanishing-case=$(VANISHING_BENCH_CASE) -Dvanishing-invalid=$(VANISHING_INVALID)
```
ivokub: No need for the VANISHING_BENCH_RELEASE env variable -- we're only intersted in small release. The non-stripped and non-small case is good for local debugging, but for R5 target we always want to have smallest build. What we could probably try later as well is to have "fast" build as well which would use different optimization tags for fast target.

verifier-ray/bench/vanishing_main.zig
```diff
@@ -0,0 +1,43 @@
+const builtin = @import("builtin");
```
ivokub, line 1: Could we instead of having a separate main target use `verifier-ray/src/main.zig` and update it to call `verifier.verify` instead?

verifier-ray/bench/riscv_main_bench.zkc
```diff
@@ -0,0 +1,56 @@
+// Benchmark RISC-V runner for verifier-ray guests.
```
ivokub, line 1: Or we can also run `zkc exec` without `-q` and it prints out the clock cycles as well. This would also give us access to the instructions as well which we could use to have better statistics (i.e. what instruction is most called). Then we don't need a custom interpreter.

verifier-ray/testdata/generate/main.go
```diff
@@ -59,6 +59,9 @@ func main() {
 	if err := writeVanishingFixtures(vanishingCases, vanishingSystems); err != nil {
 		panic(err)
 	}
+	if err := writeVanishingBenchFixtures(vanishingCases, vanishingSystems); err != nil {
```
ivokub: Currently the global repo .gitignore ignores the generated testdata. We should add explicit exception there (as we have for vanishing)

verifier-ray/testdata/generate/main.go
```diff
@@ -825,6 +841,280 @@ func writeVanishingScenarioList(out *bytes.Buffer, count int) {
...
+
+	fmt.Fprintf(out, "const %s_input = vanishing.CheckInput{\n", prefix)
+	fmt.Fprintf(out, "    .ctx = %s_ctx,\n", prefix)
+	fmt.Fprintf(out, "    .witness_claims = &%s_witness_claims,\n", prefix)
+	fmt.Fprintf(out, "    .quotient_claims = &%s_quotient_claims,\n", prefix)
+	fmt.Fprintf(out, "    .module_sizes = &%s_module_sizes,\n", prefix)
+	fmt.Fprintln(out, "};")
+	fmt.Fprintln(out)
+
+	fmt.Fprintf(out, "fn %s_keepRuntime() void {\n", prefix)
```
ivokub: I don't we need the doNotOptimizeAway part at all

verifier-ray/testdata/generate/main.go
```diff
@@ -825,6 +841,280 @@ func writeVanishingScenarioList(out *bytes.Buffer, count int) {
 	fmt.Fprintln(out, "};")
 }
 
+func writeVanishingBenchFixtures(cases []vanishingFixtureCase, systems []codegen.NamedVanishingSystem) error {
```
ivokub: I think we're duplicating quite a lot here.

Right now we have fixtures generated which target the vanishing tests, but eventually we would like to have fixtures for all scenarios stored in `testdata/generated/verify.zig` perhaps. And for benchmarking we could reuse it, instead of writing the fixtures in `bench/generated/...`.

For also running the coin sampling we need to write down the Spec which defines the rounds etc.

verifier-ray/build.zig
```diff
@@ -2,6 +2,9 @@ const std = @import("std");
 
 pub fn build(b: *std.Build) void {
     const r5 = b.option(bool, "r5", "Build for the Linea R5 zkVM target") orelse false;
+    const vanishing_bench = b.option(bool, "vanishing-bench", "Build the R5 vanishing benchmark guest") orelse false;
```
ivokub, line 5: If we update the `src/main.zig` to run `verifier.verify` then I don't think we need separate build target for benchmark anymore and we can just use the r5 target.

ivokub (changes requested):
Looks already quite good.

Though mainly what I would do is to run the whole `verifier.verify` instead of only vanishing to get actual numbers which also has hashing cost included. But this means that we also need to know the Spec for the transcript reply.

Second thing imo which we would like to try is to load the input data at runtime. Right now we have some static fixture, but I think we should try to figure out how to write them in file. I don't have a good idea yet, perhaps we can write some simple serializer.

arijitdutta67:
> Looks already quite good.
> 
> Though mainly what I would do is to run the whole `verifier.verify` instead of only vanishing to get actual numbers which also has hashing cost included. But this means that we also need to know the Spec for the transcript reply.
> 
> Second thing imo which we would like to try is to load the input data at runtime. Right now we have some static fixture, but I think we should try to figure out how to write them in file. I don't have a good idea yet, perhaps we can write some simple serializer.

Agree. I think this is the right direction, but I propose to split it into a follow-up PR because runtime input is more than moving the fixture to a file: we need a pointer-free proof serializer plus a guest-side decoder. For this PR, the static typed fixture gives us stable verifier-cost numbers without parsing cost. In the follow-up, we can reuse the existing _input_start / elf_to_json_gen @file path and add a separate decode phase so parsing cost is visible rather than mixed into transcript/verify cost.

verifier-ray/build.zig
arijitdutta67: Agree with the long-term direction. Once verifier-ray has a real runtime proof input format, src/main.zig should load that input and call verifier.verify, and then the benchmark can use the normal R5 target.

For this PR I’d keep the separate benchmark guest because src/main.zig is still a temporary smoke-test entrypoint, while the benchmark needs generated case selection, phase markers, and Poseidon compression counters. The current benchmark is semantically close to verifier.verify today: it runs protocol replay and then vanishing.verify, but keeps the split so we can report transcript vs vanishing cost.

I propose to handle the src/main.zig unification together with the runtime input serializer follow-up.

verifier-ray/testdata/generate/main.go
arijitdutta67: addressed with `testdata/generated/verify.zig`

verifier-ray/bench/vanishing_main.zig
arijitdutta67: Yes, that is the end state. Now that verify.zig exists, src/main.zig can eventually load/select a verifier fixture and call verifier.verify. This PR keeps the benchmark guest separate because we still need phase markers and replay stats for the vanishing-specific benchmark report.

verifier-ray/bench/riscv_main_bench.zkc
arijitdutta67: The custom interpreter is a helper runner of the original interpreter with phase markers. For this PR, I think it is okay to have it. Please let me know if you have a  better idea here.

verifier-ray/test/golden_test.zig
```diff
@@ -170,6 +170,9 @@ test "poseidon2 compression and merkle-damgard match prover-ray golden cases" {
         fillElems(&message, case.message);
         h.writeElements(message[0..case.message.len]);
         try expectDigest(h.sumDigest(), case.expected);
+
```
ivokub, line 173: This overloads this test. I'd rather move the expected profiling test into a separate test.

verifier-ray/src/crypto/poseidon2.zig
```diff
@@ -45,12 +45,14 @@ pub const MDHasher = struct {
     state: Digest,
     buffer: [block_size]field.Element,
     buffer_len: usize,
+    compression_count: usize,
```
ivokub, line 48: Right now we unconditionally always count the compressions. I'd rather move it out and only store the compression counts when we have some build-time constant set.

verifier-ray/src/crypto/fiat_shamir.zig
```diff
@@ -52,4 +52,8 @@ pub const Transcript = struct {
             .B2 = .{ .a0 = challenge[4], .a1 = challenge[5] },
         };
     }
+
+    pub fn compressionCount(self: *const Transcript) usize {
```
ivokub, line 56: I think we could remove this as well.

verifier-ray/src/protocol/root.zig
```diff
@@ -50,6 +57,17 @@ pub fn replay(
...
     rounds: []const RoundMessage,
 ) Error![spec.total_round_coins]Coin {
+    return (try replayWithStats(spec, rounds)).coins;
+}
+
+/// Replays the transcript and returns benchmark-friendly Poseidon2 statistics.
+///
+/// `replay` intentionally keeps the original API and delegates here, so coin
+/// generation remains implemented in one place.
+pub fn replayWithStats(
```
ivokub, line 67: I would remove the separate method and wrapping as well.

verifier-ray/test/transcript_test.zig
```diff
@@ -45,10 +59,17 @@ test "replay absorbs commitments, public columns, and cells, then squeezes coins
...
     };
-    const coins = try protocol.replay(spec, &rounds);
+    const replay_stats = try protocol.replayWithStats(spec, &rounds);
+    const coins = replay_stats.coins;
 
     // Consecutive coins must be distinct: randomDigest() absorbs a zero
     // separator between squeezes, so identical back-to-back outputs indicate
     // a broken separator mechanism.
     try std.testing.expect(!coins[0].eql(coins[1]));
+    try std.testing.expectEqual(@as(usize, 3), replay_stats.compression_count);
```
ivokub, line 69: I'd also remove it here.

verifier-ray/Makefile
```diff
@@ -142,5 +166,50 @@ zkc-verify-failing-expected:
 		echo "zkc-verify-failing failed as expected"; \
 	fi
 
+bench-vanishing-build: generate-testdata
+	zig build --release=small -Dstrip=true -Dr5=true -Dvanishing-bench=true -Dvanishing-case=$(VANISHING_BENCH_CASE)
+
+bench-vanishing-json: bench-vanishing-build
+	go run $(ELF_TO_JSON) $(VANISHING_BENCH_BIN) "" $(IN_ORIGIN) > $(VANISHING_BENCH_JSON)
```
ivokub, line 173: We could do `$(MAKE) zkc-build R5_BIN=$(VANISHING_BENCH_BIN) R5_JSON=$(VANISHING_BENCH_JSON)`, so we have single entry point for elf-to-json.

verifier-ray/Makefile
```diff
@@ -142,5 +166,50 @@ zkc-verify-failing-expected:
...
+
+bench-vanishing-zkc: bench-vanishing-json
+	mkdir -p $(dir $(VANISHING_BENCH_LOG))
+	set -o pipefail; zkc exec -q $(VANISHING_BENCH_JSON) $(VANISHING_BENCH_ZKC_MAIN) | tee $(VANISHING_BENCH_LOG)
+
+bench-vanishing-doc: bench-vanishing-zkc
```
ivokub, line 179: I would remove the individual-case benchmarks and the selectors for being able to choose which benchmark to run. We should always run all of them.

It is nice to have that we can run individually, but I think it starts overloading makefile -- its easy for the agents to parse, but we should target humans to read and understand.

verifier-ray/bench/riscv_main_bench.zkc
ivokub: Perhaps for this PR we can keep it, but I would still remove it as soon as possible, it will not be maintainable and we will have discrepancy with the actual interpreter. And when we have discrepancy then the benchmark data is also meaningless as it doesn't correspond to the actual executor.

For the phase markers for example we could discuss with the arithmetization team to add a dummy instruction in the ECALL parsing (in r-type) do not fail on some specific input, but rather continue interpreting (perhaps even in debug mode). I.e. in `r_type.zkc` instead of:
```
case CUSTOM_1: {
    if (funct3 == FUNCT3_KECCAK) && (funct7 == FUNCT7_KECCAK) {
        printf "KECCAK precompile "
        keccak(v1 as Address, v2 as u32, registers[rd] as Address)
    } else {
        printf "[ERROR] Unsupported R-type custom-1 instruction with\n"
        printf "\tfunct3 ≡ %x ≡ %b\n", funct3, funct3
        printf "\tfunct7 ≡ %x ≡ %b\n", funct7, funct7
        fail
    }
}
```
we could have
```
case CUSTOM_1: {
    if (funct3 == FUNCT3_KECCAK) && (funct7 == FUNCT7_KECCAK) {
        printf "KECCAK precompile "
        keccak(v1 as Address, v2 as u32, registers[rd] as Address)
    } else if (funct3 == MARKER) {
        printf "\tMARKER register[10] = %d\n" registers[10]
    } else {
        printf "[ERROR] Unsupported R-type custom-1 instruction with\n"
        printf "\tfunct3 ≡ %x ≡ %b\n", funct3, funct3
        printf "\tfunct7 ≡ %x ≡ %b\n", funct7, funct7
        fail
    }
}
```
etc. There is also some debug execution facility now for zkc, perhaps it could be defined inside that even so that in actual tracing it wouldn't lead to any additional constraints/cycles.

verifier-ray/bench/vanishing-benchmark-comparison.md
```diff
@@ -0,0 +1,116 @@
...
+| Slowest | `80 RangeCheckCompiler/LargeBound` (7,237,154 cycles) |
+| Git commit | `b15e97a84` |
+| Zig | `0.16.0` |
+| zkc | `mod	github.com/consensys/go-corset	v1.2.18-0.20260521022840-414e7b73971e` |
+| zkc main | `bench/riscv_main_bench.zkc` |
+| Logs | `zig-out/vanishing-bench/case-<index>.log` |
+
+## Results
+
+| Case | Scenario | RISC-V cycles | Transcript | Poseidon2 comps | Transcript / comp | Vanishing | Other | vs fastest | Modules | Dynamic | Rounds | Expressions | Buckets | Vanishings | Witness | Quotient |
```
ivokub, line 25: Imo vs fastest column is meaningless, these are different system descriptions and it doesn't tell anything useful.

verifier-ray/bench/vanishing-benchmark-comparison.md
```diff
@@ -0,0 +1,116 @@
+# Vanishing RISC-V Benchmark Comparison
```
ivokub, line 1: I would simplify this output. Lets just output it into csv. We don't imo need the Summary section at all and the descriptions.

verifier-ray/bench/vanishing-benchmark.md
```diff
@@ -0,0 +1,47 @@
+# Vanishing RISC-V Benchmark
```
ivokub, line 1: I think we can omit outputting this individual bench and just output the CSV of all scenarios.

verifier-ray/Makefile
```diff
@@ -49,12 +61,16 @@ help:
...
 		'zkc variables:' \
-		'  IN_ORIGIN            Input bytes address, default 0x08800000'
+		'  IN_ORIGIN            Input bytes address, default 0x08800000' \
+		'' \
+		'benchmark variables:' \
+		'  VANISHING_BENCH_CASE     Generated vanishing case index, default 0' \
+		'  VANISHING_BENCH_CASES    Case selector for comparison, default 0-5; examples: all, 0-10, 0,2,5-8'
 
-fmt: fmt-zig fmt-codegen fmt-testdata-generate
+fmt: generate-testdata fmt-zig fmt-codegen fmt-testdata-generate fmt-bench
```
YaoJGalteland, line 70: `fmt` and `generate-testdata` are different kinds of operations. One is a read-only check, while the other is a write that produces artifacts.

With the current setup, running `make fmt` will regenerate `verify.zig` and `vanishing.zig` as a side effect. That may surprise developers who only want to check formatting.

We can have ci run `generate-testdata` before `fmt`, so the generated files already exist when `fmt-zig` checks them. Then `fmt` can stay format-check-only, without regenerating files as a side effect.

YaoJGalteland (commented):
The benchmarking architecture is sound and well documented. Thank you for the updates.

I left one minor comment.

verifier-ray/bench/vanishing_main.zig
ivokub: Imo we could still try to merge the two. Imo there are two issues right now we have:
* input serialization
* conditional marker setting

For the second I think we can use compile-time options, see https://ziglang.org/learn/build-system/#conditional-compilation. I.e. when we have at build time set some option "profile" etc, then we set the markers. And we can access it as comptime and conditionally set the markers if we need to.

For the input deserialization indeed lets postpone this a bit. So what I would recommend right now for short term:
* define build-time option if profiling is set or not. Based on that we perform the marking of cycle start/end in R5 interpreter. And we should also have a separate profiling package which is called in Poseidon2 etc if the profiling is set or not.
* for input parsing we have another set option where we get the given spec/system/input from build time.

Later on after we have input serialization, we only keep the build-time option for defining the spec/system and input comes from runtime.

ivokub (commented):
I left a few comments. Dunno if they make sense, I think I could try instead implementing some ideas. But one is that the profiling should be separated by a build-time option and not called at all when not in profiling mode.

I.e. we could have a profiling package `profile.zig` which does a la

inline fn markR5(comptime marker: u64, comptime value: u64) {
    if (comptime profile_option.is_profiling) {
        asm volatile (
        \\mv a0, %[phase]
        \\mv a1, %[value]
        \\li a7, 4242
        \\ecall
        :
        : [phase] "r" (marker),
          [value] "r" (value),
        : .{ .a0 = true, .a1 = true, .a7 = true, .memory = true });
    }
}

var compressions: u64;

inline fn countCompression() {
    if (comptime profile_option.is_profiling) {
        compressions += 1;
    }
}
```
and then inside poseidon2.zig compress function we call countCompression. It should reduce to no-op when the profiling option is not set and otherwise update it.

ivokub:
Superseded by #3394

PR #3334, opened 2026-06-10
feat(prover-ray): proper Prove / Verify API with an explicit Proof object

AlexandreBelling:
# Motivation

Testing a WIOP protocol used a single shared wiop.Runtime for both the prover and verifier jobs: wioptest.RunAndVerify(rt) ran every round's ProverActions and then every round's VerifierActions against the same runtime that held all witness data. There was no Proof object and no real prover/verifier separation — the verifier simply read back whatever the prover left in the runtime, including oracle and internal columns.

This PR introduces a clean, real-proof-shaped API:

```
proof := sys.Prove(assign)   // runs the prover, returns a self-contained Proof
err   := sys.Verify(proof)   // re-checks the proof, returns an error on failure
```

### 1. New Proof / Prove / Verify in package wiop (wiop/prove_verify.go)

Proof carries only verifier-visible data: VisibilityPublic and VisibilityOracle columns, cells (always public), coins (always public), and dynamic-module sizes. Internal columns are deliberately excluded.
(*System).Prove(assign func(*Runtime)) Proof runs the prover loop (the same loop RunAndVerify used) and extracts the public transcript. assign is the witness hook — e.g. driver.AssignWithPreRead(rt, inputs) or a scenario's AssignHonest.
(*System).Verify(Proof) error rebuilds a fresh runtime from the public data alone (no oracle columns, no Fiat-Shamir replay) and runs every verifier action. Being in-package, it loads the unexported runtime maps directly.
Kept as methods on *System — no new package, no closures, no package split.

### 2. logderivativesum refactor so the proof can stay strictly public (wiop/compilers/logderivativesum/logderivativesum.go)

Previously the verifier read the oracle witness columns at row 0 to check the initial condition Z[0]·zDen[0] = zNum[0]. That boundary is now pinned in-circuit via Module.NewLocalConstraint(..., 0), discharged through the standard localvanishing → global pipeline (strictly more sound than the old verifier-only check, and it also covers the single-row n == 1 case).
The verifier action now performs only the final-sum check (Σ Z[n-1] == Result), reading only cells. Removed the now-unused evaluateRowZero helper and the zNum/zDen/zInit fields from zEntry.
Net effect: all three verifier actions (global, lookuptologderivsum, logderivativesum) read only public columns, cells, and coins.

### 3. Propagated the new API to existing tests

Migrated the scenario-suite compilation tests across pipeline_test.go, global, localvanishing, rangecheck, lookuptologderivsum, and logderivativesum to Prove/Verify (rangecheck's panic test now asserts Prove panics). localvanishing gained a small proveVerify(sys, assign) wrapper replacing its local runAndVerify copy.
Migrated zkcdriver/example_test.go's runTestCase to the new API.
Left the mutator tests (those that inject wrong values by mutating intermediate runtime state mid-flight, e.g. tampering Z/Result/M before the prover runs) on the manual runRound/checkAllVerifierActions loop — they don't fit the honest-prover-then-verify model.

## Notes

Coins live in the Proof (they're VisibilityPublic by definition); the verifier cannot re-derive them, since Fiat-Shamir also absorbs the oracle columns it does not hold.
A higher-level Compile wrapper that orchestrates the compiler passes was intentionally left out — it would need its own package to avoid an import cycle (the passes import wiop). Callers still run the passes before Prove, as before.

## Testing

`go build ./... && go test ./wiop/... ./zkcdriver/...` — all green, including soundness cases (Verify rejects invalid witnesses) across the vanishing, local-vanishing, lookup, range-check, and log-derivative pipelines.


prover-ray/wiop/prove_verify.go
```diff
@@ -0,0 +1,120 @@
...
+//     reconstruct module domains without re-assigning columns.
+//
+// A Proof is self-contained: [System.Verify] rebuilds a fresh [Runtime] from it
+// and runs every verifier action against that data alone. The fields are
+// unexported for now; they can be promoted once a serialized proof format is
+// needed.
+type Proof struct {
+	columns      map[ObjectID]*ConcreteVector
+	cells        map[ObjectID]field.Gen
+	coins        map[ObjectID]field.Gen
```
ivokub: We don't need to pass the coins in the proof -- they should be computed at the verifier runtime from messages (i.e. columns and cells).

prover-ray/wiop/prove_verify.go
```diff
@@ -0,0 +1,120 @@
...
+		rt.cells[id] = v
+	}
+	for id, v := range proof.coins {
+		rt.coins[id] = v
+	}
+	for k, v := range proof.dynamicSizes {
+		rt.dynamicSizes[k] = v
+	}
+
+	for _, r := range sys.Rounds {
```
ivokub: No coin computation, they come trusted from the proof.

ivokub (changes requested):
I would refactor the `compiler/logderivatesum` changes into a separate PR, right now this PR does two very different things imo.

And for the Proof and verifying based on proof, it is very unsound right now, there is no FS challenge computation and we plainly trust the coins given in the proof. It is good for development, but imo more clearly indicated in documentation etc.

AlexandreBelling:
The logderivativesum thing is a direct consequence of that. We have a verifier action that touches oracle columns and that is impossible. So I'd prefer to keep it in the PR.

prover-ray/wiop/wiop_prove_verify.go
```diff
@@ -0,0 +1,141 @@
...
+// every Fiat-Shamir challenge itself by replaying the transcript, so a prover
+// cannot influence the challenges by supplying forged coin values.
+//
+// Including the oracle columns is a testing convenience: it lets the verifier
+// reconstruct the exact Fiat-Shamir state without a commitment scheme (the raw
+// oracle values stand in for their commitments). In a production proof an oracle
+// column is sent only as its commitment, never in full; the pipeline should
+// ultimately guarantee that no VisibilityOracle column survives compilation.
+//
+// The fields are unexported for now; they can be promoted once a serialized
```
ivokub: Stale comment. Now fields are public.

prover-ray/wiop/wiop_prove_verify.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	// assignRound loads the proof's committed columns and cells for r into the
+	// runtime. AssignColumn / AssignCell require r to be the current round, so
+	// this is always called on rt.CurrentRound().
+	assignRound := func(r *Round) {
+		for _, col := range r.Columns {
+			if v, ok := proof.Columns[col.Context.ID]; ok {
+				rt.AssignColumn(col, v)
+			}
```
ivokub, line 137: It currently fails silently when the proof doesn't have a column/cells what the verifier/system expects.

prover-ray/wiop/wiop_prove_verify.go
```diff
@@ -0,0 +1,141 @@
...
+// oracle values stand in for their commitments). In a production proof an oracle
+// column is sent only as its commitment, never in full; the pipeline should
+// ultimately guarantee that no VisibilityOracle column survives compilation.
+//
+// The fields are unexported for now; they can be promoted once a serialized
+// proof format is needed.
+type Proof struct {
+	Columns      map[ObjectID]*ConcreteVector
+	Cells        map[ObjectID]field.Gen
+	DynamicSizes map[int]int
```
ivokub, line 33: What is the key in DynamicSizes map? The module id? Perhaps we have a dedicated type for it? Or minimally we should document it here as well.

prover-ray/wiop/wiop_prove_verify.go
```diff
@@ -0,0 +1,141 @@
...
+// challenge. Verifier actions read the re-derived coins, the cells, and the
+// public columns.
+func (sys *System) Verify(proof Proof) error {
+	rt := NewRuntime(sys) // currentRound = r0, preloads precomputed columns
+
+	// Restore dynamic-module sizes so Module.RuntimeSize resolves during the
+	// replay and the subsequent verifier checks.
+	for k, v := range proof.DynamicSizes {
+		rt.dynamicSizes[k] = v
+	}
```
ivokub: Afaik here we don't do any validation that all dynamic sizes what are needed to set are set. Do we have a list in System which indicates which modules have dynamic size?

ivokub (changes requested):

AlexandreBelling:
I added extra checks to verify all the fields of the proof. I think this should be tight enough.

Namely:
* dynamicSize must be non-zero power of two
* dynamicSize only specified for dynamic-sized modules
* the size must be consistent with column size if the module contains public columns
* the column must exists and have the right visibility [public/oracle]
* the cells must exist and they must all be provided

I added explicit sanity-check flagging missing cells and or columns, but note that this would be caught in the verifier actions anyway as they would fail to fetch the column.

Also, I implemented it so that we can pass Oracle column. But in production, we will not attempt to use them. This is mostly for convenience. Additionally, we are currently thinking of removing public column as we think they will not be needed anywhere for us. So we may strip out the Column field from the proof in the near future.

ivokub (dismissed):
Looks good now, though the comment about Proof fields being unexported is still stale.

YaoJGalteland (dismissed):
LGTM.

One additional thing, but fine to defer:

When the verifier-ray (Zig) codegen ports the final-sum check, it will need a declarative handle on them, because it can't introspect the Go Check body. Specifically it needs the structural skeleton of the scalar relation, expressed in transcript coordinates: which opened cells participate as addends, what the target is (a cell, or zero), and each cell's (round, index) position so the Zig side can index ctx.rounds[round].cells[index] consistently with the Fiat-Shamir replay. The global action needs no such handle — it's already declarative via its Vanishing queries — but the two logderiv scalar checks have no other declarative form.

I don’t think the read-only accessor needs to land in this PR. It’s only consumed by codegen, so I’ll add it in the follow-up codegen PR together with its only caller, instead of leaving it unused here.

arijitdutta67 (dismissed):
LGTM. However a small concern in the note of the PR description:
>Coins live in the Proof (they're VisibilityPublic by definition); the verifier cannot re-derive them, since Fiat-Shamir also absorbs the oracle columns it does not hold.

How can verifier-ray trust the coins that is there in the proof? I think it should regenerate them and match. To regenerate, the FS should absorb the commitments to the Oracle columns instead of the whole columns themself. The proof should contain those commitments so that verifier-ray can re-derive the coins. At least this is how I see. Please correct me if I am wrong @YaoJGalteland @ivokub

YaoJGalteland:
@arijitdutta67 this PR note is a bit misleading. verifier-ray can re-derive coins themself. And yes, the FS should absorb the **commitments** to the Oracle columns instead of the whole columns themself. Those commitments are part of the proof and are available to the verifier. 

That means the transcript over {oracle commitments, public columns, cells} is fully reconstructible by the verifier, and the coins are derived directly from that transcript. Nothing is missing.

The prover and verifier absorb the same commitments, so the transcripts match, and therefore the derived coins match as well.

YaoJGalteland (approved):

bogdanbear (approved):

Issue #3287, opened 2026-06-08
feat(verifier-ray): add Lagrange selector as supported op


ivokub:
The verifier-ray's appendExpr (in this [PR](https://github.com/Consensys/linea-monorepo/pull/3271/changes#diff-e3fa3cd96a2a12ce13e335876eb1e26b7058f730e64b30777ba11ac477edadea)) switches on the concrete expression type. It has no case for *wiop.LagrangeSelector and hits the default branch:

```
default:
    return 0, &UnsupportedExpressionError{Type: fmt.Sprintf("%T", expr)}
```

I think it’s probably best to update the verifier-ray changes after both #3275 and #3271 are merged.

_Originally posted by @YaoJGalteland in https://github.com/LFDT-Lineth/lineth-monorepo/pull/3275#discussion_r3361511696_

PR #3351, opened 2026-06-11
feat(verifier-ray): add Lagrange selector as supported op

YaoJGalteland:
This PR implements issue(s) #3287 

Mirrors #3275 (Add LagrangeSelectors in the wizard framework) in verifier-ray:
prover-ray gained `LagrangeSelector` as a wizard expression leaf, and this PR
teaches the verifier-ray codegen and Zig evaluator to handle it.

## What

Adds support for `LagrangeSelector` leaves to the verifier-ray vanishing
verifier — both the Go codegen that lowers a prover-ray system into the Zig
`System`, and the Zig evaluator that checks the quotient identity.

A Lagrange selector `L_p` is the row-`p` indicator column (1 at position `p`,
0 elsewhere), used to lower local constraints into global ones. It is evaluated
out-of-domain by the verifier rather than committed.

## Changes

**Codegen (Go)**
- New `ExprLagrangeSelector` expression kind with a `SelectorPosition` field
  (`codegen/vanishing.go`).
- Renders to `.{ .lagrange_selector = N }` in the generated Zig
  (`codegen/vanishing_zig.go`).

**Evaluator (Zig)**
- New `lagrange_selector` `ExprNode` and `evalLagrangeSelector`, implementing
      L_p(r) = ω^p · (rⁿ − 1) / (n · (r − ω^p))
  matching prover-ray's `LagrangeSelector.EvaluateOutOfDomain`.
- Explicit in-domain guard: rejects `r = ω^p` with `error.LagrangeSelectorInDomain`
  instead of silently returning 0 (the field defines 1/0 = 0).
- Comptime/runtime split: static modules fold `ω^p` and `n` to comptime
  constants; dynamic modules derive `ω` from the runtime size.

**Comptime parameter separation**
- Introduces `EvalCtx` to bundle the per-module evaluation context
  (`coin`, `annihilator`, `dynamic_n`) threaded through `evalExpr`/`evalOp`,
  replacing the previous loose scalar parameters.

## Testing

- Go codegen unit tests: `TestBuildVanishingSystemSupportsLagrangeSelector`,
  `TestExprNodeLiteralRendersLagrangeSelector`.
- Golden fixtures replayed against prover-ray proof data:
  - `Vanishing/LagrangeSelectorBoundary` — static module, exercises the
    comptime-folded selector path.
  - `Vanishing/DynamicLagrangeSelectorBoundary` — dynamic module, exercises the
    runtime-`ω` path that production dynamic modules hit.
- Zig unit test for the in-domain guard, with an out-of-domain control
  confirming the error is specific to the guard and not structural.


verifier-ray/src/query/vanishing.zig
```diff
@@ -166,25 +169,46 @@ fn verifyBucket(
     inline for (bucket.vanishings) |v| {
         // Aggregate the vanished numerators with the merge coin alpha:
         // P_agg(r) = sum_i alpha^i * P_i(r) * C_i(r).
-        const value = evalExpr(module, v.expression, input);
-        const cancellation = try cancellationAtPoint(v.cancelled_positions, static_n, dynamic_n, eval_coin);
+        const value = try evalExpr(module, v.expression, static_n, ctx, input);
+        const cancellation = try cancellationAtPoint(v.cancelled_positions, static_n, dynamic_n, ctx.coin);
```
ivokub: We could modify cancellatonAtPoint to also accept EvalCtx perhaps?

verifier-ray/src/query/vanishing.zig
```diff
@@ -195,20 +219,67 @@ fn scalarToExt(value: protocol.Scalar) ext.Ext {
...
+// (static_n == 0) derive the n-th root of unity from ctx.dynamic_n and take the
+// runtime pow. Everything else (the annihilator, the r - omega^position
+// denominator, the division) depends on the runtime eval coin r and stays
+// runtime in both cases.
+fn evalLagrangeSelector(comptime position: usize, comptime static_n: usize, ctx: EvalCtx) Error!ext.Ext {
+    const omega_pos = if (static_n != 0)
+        comptime staticRootPower(static_n, position)
+    else blk: {
+        const omega = field.rootOfUnityBy(ctx.dynamic_n) catch return error.InvalidModuleSize;
+        break :blk omega.pow(@as(u64, @intCast(position)));
```
ivokub: For extension we have defined `powComptime` which computes a power with compile-time known exponent. I think we could implement it similarly for base elements. And when we do, then we can also use it in staticRootPower as well.

Another thing which I think we should also address is that currently our pow methods for base and extension elements take runtime exponent with u256 or u64 inputs. But in practice it is always u32, so the loops would be slightly smaller.

verifier-ray/src/query/vanishing.zig
```diff
@@ -195,20 +219,67 @@ fn scalarToExt(value: protocol.Scalar) ext.Ext {
...
+    // modules (a comptime constant that folds here) else the runtime dynamic_n.
+    // The field defines 1/0 = 0, so an in-domain eval coin (r == omega^position)
+    // would silently yield 0; reject it explicitly to match the Go evaluator's
+    // out-of-domain contract.
+    const r_minus_omega = ctx.coin.sub(ext.Ext.lift(omega_pos));
+    if (r_minus_omega.isZero()) return error.LagrangeSelectorInDomain;
+    const n = if (static_n != 0) static_n else ctx.dynamic_n;
+    const denominator = r_minus_omega.mulByBase(field.Element.init(@as(u64, n)));
+
+    return numerator.div(denominator);
```
ivokub, line 279: Imo we should observe and keep in mind for later that we do an inverse here. This leads to computing base field inverses when computing the extension divisions.

Right now we have implemented a naive inverse by exp with `modulus-2`, but we should later probably have a dedicated zkVM precompile as well.

verifier-ray/testdata/generate/main.go
```diff
@@ -627,6 +627,33 @@ func buildVanishingFixtureCases() ([]vanishingFixtureCase, []codegen.NamedVanish
 			return nil, nil, err
 		}
 	}
+
```
ivokub, line 654: Perhaps we can move the new test cases to the end of fixture generation? Then there are less code changes in the stored fixtures :)

ivokub (dismissed):
Very clean implementation. It is good to merge on my side, but I left a few comments.

ivokub (commented):
And I think this is also something for the prover side, but it seems that the Lagrange Selectors are not deduplicated. This means that the verifier also recomputes it, which may be expensive when we have dynamic-sized modules. See for example:
```zig
const system_55_module_0_expressions = [_]vanishing.ExprNode{
    .{ .column_claim = 0 }, // col: "c0"
    .{ .constant = field.Element.init(1) },
    .{ .op = .{ .operator = .mul, .operands = &.{ 0, 1 } } },
    .{ .constant = field.Element.init(1) },
    .{ .op = .{ .operator = .mul, .operands = &.{ 2, 3 } } },
    .{ .column_claim = 1 }, // col: "c1"
    .{ .constant = field.Element.init(1) },
    .{ .op = .{ .operator = .mul, .operands = &.{ 5, 6 } } },
    .{ .constant = field.Element.init(1) },
    .{ .op = .{ .operator = .mul, .operands = &.{ 7, 8 } } },
    .{ .op = .{ .operator = .add, .operands = &.{ 4, 9 } } },
    .{ .column_claim = 2 }, // col: "c2"
    .{ .constant = field.Element.init(1) },
    .{ .op = .{ .operator = .mul, .operands = &.{ 11, 12 } } },
    .{ .constant = field.Element.init(1) },
    .{ .op = .{ .operator = .mul, .operands = &.{ 13, 14 } } },
    .{ .op = .{ .operator = .add, .operands = &.{ 10, 15 } } },
    .{ .column_claim = 3 }, // col: "z-b0-k0"
    .{ .column_claim = 4 }, // col: "z-b0-k0"
    .{ .op = .{ .operator = .sub, .operands = &.{ 17, 18 } } },
    .{ .constant = field.Element.init(1) },
    .{ .constant = field.Element.init(1) },
    .{ .op = .{ .operator = .mul, .operands = &.{ 20, 21 } } },
    .{ .constant = field.Element.init(1) },
    .{ .op = .{ .operator = .mul, .operands = &.{ 22, 23 } } },
    .{ .op = .{ .operator = .mul, .operands = &.{ 19, 24 } } },
    .{ .op = .{ .operator = .sub, .operands = &.{ 16, 25 } } },
    .{ .column_claim = 5 }, // col: "c3"
    .{ .column_claim = 6 }, // col: "z-b0-k1"
    .{ .column_claim = 7 }, // col: "z-b0-k1"
    .{ .op = .{ .operator = .sub, .operands = &.{ 28, 29 } } },
    .{ .constant = field.Element.init(1) },
    .{ .op = .{ .operator = .mul, .operands = &.{ 30, 31 } } },
    .{ .op = .{ .operator = .sub, .operands = &.{ 27, 32 } } },
    .{ .cell_value = .{ .round = 1, .index = 1 } }, // cell: "result"
    .{ .column_claim = 3 }, // col: "z-b0-k0"
    .{ .op = .{ .operator = .sub, .operands = &.{ 34, 35 } } },
    .{ .lagrange_selector = 0 },
    .{ .op = .{ .operator = .mul, .operands = &.{ 36, 37 } } },
    .{ .cell_value = .{ .round = 1, .index = 2 } }, // cell: "result"
    .{ .column_claim = 3 }, // col: "z-b0-k0"
    .{ .op = .{ .operator = .sub, .operands = &.{ 39, 40 } } },
    .{ .lagrange_selector = 3 },
    .{ .op = .{ .operator = .mul, .operands = &.{ 41, 42 } } },
    .{ .cell_value = .{ .round = 1, .index = 3 } }, // cell: "result"
    .{ .column_claim = 6 }, // col: "z-b0-k1"
    .{ .op = .{ .operator = .sub, .operands = &.{ 44, 45 } } },
    .{ .lagrange_selector = 0 },
    .{ .op = .{ .operator = .mul, .operands = &.{ 46, 47 } } },
    .{ .cell_value = .{ .round = 1, .index = 4 } }, // cell: "result"
    .{ .column_claim = 6 }, // col: "z-b0-k1"
    .{ .op = .{ .operator = .sub, .operands = &.{ 49, 50 } } },
    .{ .lagrange_selector = 3 },
    .{ .op = .{ .operator = .mul, .operands = &.{ 51, 52 } } },
};
```

I don't think we would have to deduplicate right now on the verifier side during codegen, but keep in mind for the future. And when the prover doesn't start deduplicating, then we should do it ourselves. But I think this could be part of bigger optimization pipeline (perhaps there is something else we could optimize out, but only if the number show it is worth it. When most of cycles are inside hash computation, then doesn't make a big difference I guess).

verifier-ray/src/query/vanishing.zig
YaoJGalteland: yes, we can. Since the function needs more than a bare coin, using EvalCtx could also remove dynamic_n from the input.

YaoJGalteland:
> And I think this is also something for the prover side, but it seems that the Lagrange Selectors are not deduplicated. This means that the verifier also recomputes it, which may be expensive when we have dynamic-sized modules. See for example:
> 
> ```zig
> const system_55_module_0_expressions = [_]vanishing.ExprNode{
>     .{ .column_claim = 0 }, // col: "c0"
>     .{ .constant = field.Element.init(1) },
>     .{ .op = .{ .operator = .mul, .operands = &.{ 0, 1 } } },
>     .{ .constant = field.Element.init(1) },
>     .{ .op = .{ .operator = .mul, .operands = &.{ 2, 3 } } },
>     .{ .column_claim = 1 }, // col: "c1"
>     .{ .constant = field.Element.init(1) },
>     .{ .op = .{ .operator = .mul, .operands = &.{ 5, 6 } } },
>     .{ .constant = field.Element.init(1) },
>     .{ .op = .{ .operator = .mul, .operands = &.{ 7, 8 } } },
>     .{ .op = .{ .operator = .add, .operands = &.{ 4, 9 } } },
>     .{ .column_claim = 2 }, // col: "c2"
>     .{ .constant = field.Element.init(1) },
>     .{ .op = .{ .operator = .mul, .operands = &.{ 11, 12 } } },
>     .{ .constant = field.Element.init(1) },
>     .{ .op = .{ .operator = .mul, .operands = &.{ 13, 14 } } },
>     .{ .op = .{ .operator = .add, .operands = &.{ 10, 15 } } },
>     .{ .column_claim = 3 }, // col: "z-b0-k0"
>     .{ .column_claim = 4 }, // col: "z-b0-k0"
>     .{ .op = .{ .operator = .sub, .operands = &.{ 17, 18 } } },
>     .{ .constant = field.Element.init(1) },
>     .{ .constant = field.Element.init(1) },
>     .{ .op = .{ .operator = .mul, .operands = &.{ 20, 21 } } },
>     .{ .constant = field.Element.init(1) },
>     .{ .op = .{ .operator = .mul, .operands = &.{ 22, 23 } } },
>     .{ .op = .{ .operator = .mul, .operands = &.{ 19, 24 } } },
>     .{ .op = .{ .operator = .sub, .operands = &.{ 16, 25 } } },
>     .{ .column_claim = 5 }, // col: "c3"
>     .{ .column_claim = 6 }, // col: "z-b0-k1"
>     .{ .column_claim = 7 }, // col: "z-b0-k1"
>     .{ .op = .{ .operator = .sub, .operands = &.{ 28, 29 } } },
>     .{ .constant = field.Element.init(1) },
>     .{ .op = .{ .operator = .mul, .operands = &.{ 30, 31 } } },
>     .{ .op = .{ .operator = .sub, .operands = &.{ 27, 32 } } },
>     .{ .cell_value = .{ .round = 1, .index = 1 } }, // cell: "result"
>     .{ .column_claim = 3 }, // col: "z-b0-k0"
>     .{ .op = .{ .operator = .sub, .operands = &.{ 34, 35 } } },
>     .{ .lagrange_selector = 0 },
>     .{ .op = .{ .operator = .mul, .operands = &.{ 36, 37 } } },
>     .{ .cell_value = .{ .round = 1, .index = 2 } }, // cell: "result"
>     .{ .column_claim = 3 }, // col: "z-b0-k0"
>     .{ .op = .{ .operator = .sub, .operands = &.{ 39, 40 } } },
>     .{ .lagrange_selector = 3 },
>     .{ .op = .{ .operator = .mul, .operands = &.{ 41, 42 } } },
>     .{ .cell_value = .{ .round = 1, .index = 3 } }, // cell: "result"
>     .{ .column_claim = 6 }, // col: "z-b0-k1"
>     .{ .op = .{ .operator = .sub, .operands = &.{ 44, 45 } } },
>     .{ .lagrange_selector = 0 },
>     .{ .op = .{ .operator = .mul, .operands = &.{ 46, 47 } } },
>     .{ .cell_value = .{ .round = 1, .index = 4 } }, // cell: "result"
>     .{ .column_claim = 6 }, // col: "z-b0-k1"
>     .{ .op = .{ .operator = .sub, .operands = &.{ 49, 50 } } },
>     .{ .lagrange_selector = 3 },
>     .{ .op = .{ .operator = .mul, .operands = &.{ 51, 52 } } },
> };
> ```
> 
> I don't think we would have to deduplicate right now on the verifier side during codegen, but keep in mind for the future. And when the prover doesn't start deduplicating, then we should do it ourselves. But I think this could be part of bigger optimization pipeline (perhaps there is something else we could optimize out, but only if the number show it is worth it. When most of cycles are inside hash computation, then doesn't make a big difference I guess).

Interesting observation! I agree as well, probably not worth addressing until profiling shows it actually matters.

verifier-ray/testdata/generated/vanishing.zig
```diff
@@ -2572,7 +2572,7 @@ pub const system_29_spec = protocol.Spec{
 // expression: "global"
 const system_29_module_0_expressions = [_]vanishing.ExprNode{
     .{ .column_claim = 0 }, // col: "col"
-    .{ .column_claim = 1 }, // col: "lagrange-row0"
+    .{ .lagrange_selector = 0 },
```
ivokub, line 2575: Now we have the LagrangeSelector node in the expression, we're left with only a single column claim. But looking at the corresponding Scenario below:
```zig
 .honest = .{
        .rounds = &.{
            .{
                .columns = &.{
                    .{ .public_base = &.{ 0, 9, 9, 9 } },
                },
                .cells = &.{},
            },
            .{
                .columns = &.{
                    .{ .public_ext = &.{ .{ 799014909, 0, 0, 0, 0, 0 }, .{ 551477246, 0, 0, 0, 0, 0 }, .{ 266338303, 0, 0, 0, 0, 0 }, .{ 2111905793, 0, 0, 0, 0, 0 } } },
                },
                .cells = &.{},
            },
        },
        .witness_claims = &.{.{ 1753098428, 1553015963, 1571627753, 1945326155, 133823156, 1601688298 }},
        .quotient_claims = &.{.{ 1106198120, 1204206469, 472755402, 934873305, 1294184935, 781850424 }},
        .module_sizes = &.{},
    },
```
then we still have the column claim in the extracted scenario. I think we should be able to remove the
```zig
    .{ .public_ext = &.{ .{ 799014909, 0, 0, 0, 0, 0 }, .{ 551477246, 0, 0, 0, 0, 0 }, .{ 266338303, 0, 0, 0, 0, 0 }, .{ 2111905793, 0, 0, 0, 0, 0 } } },
```
from the input? Now I'm not sure how this affecs the coin computations though.

ivokub (approved):
Looks good, thanks for the changes. I left one new comment about the column which previously was provided by the prover, but now computed by the verifier, but still is present in the generated fixture input.

verifier-ray/testdata/generated/vanishing.zig
YaoJGalteland: no, we should keep `public_ext`.

Currently, with `LagrangeSelector`, we have:

```text
round 0: col (`public_base`)  → merge coin α
round 1: Q₀ (`public_ext`)    → eval coin r
claims: col(r), Q₀(r)
```

`col(r) =  witness_claims[0]`, `witness_claims[1]` used to be `sel(r)`, but that has been dropped.

The verifier checks:

```text
col(r) · L₀(r) == Z_H(r) · Q₀(r)
```

If we drop `public_ext`, the prover can wait for `r` and then compute:

```text
Q₀(r) = col(r) · L₀(r) / Z_H(r)
```

which would trivially satisfy the identity.

Additionally, the commitment to `Q₀` is also needed and should be committed before r, so we can later check opening consistency by FRI, i.e. that `Q₀(r)` really matches the committed `Q₀`.

PR #3352, opened 2026-06-11
feat(arithmetization): predecoding perf branch

amkCha:
This PR implements issue(s) #

### Checklist

  PR.


arithmetization/src/test/scripts/elf_to_json_gen/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+## What pre-decoding is and why
+
+Without pre-decoding, the ZkC interpreter decodes each instruction *at runtime,
+every step*: fetch the 32-bit word from RAM, peel off the `opcode`, derive the
+instruction `type`, then split out the per-format fields (`rd`, `rs1`, `imm`, …).
+That work repeats for every executed instruction and adds to the machine's cost.
+
+Because the program text is static, this tool decodes **every instruction word
+once, ahead of time**, and ships the results as extra input tables in the JSON.
```
OlivierBBB: Quick question: why is the pre-decoding included in the JSON rather than being derived at runtime from the input program ?

OlivierBBB: I ask in particular (and I believe that was @DavePearce 's point today) given that the one-time decoding could be done in constraints. Including it in the JSON may still make sense, but it seems redundant if we were to do provable one time decoding (which makes sense and connects to the discussions we had about justifying the one-time-decoding).

arithmetization/src/main/riscv/instruction_processing/u_type.zkc
```diff
@@ -1,62 +1,66 @@
...
+    var imm:DoubleWord
     var rd:Register
-    // var uimm5  : UImm5
-    // var uimm6  : UImm6
-    // var imm12  : Imm12
-    var imm20:Imm20
-    // var opcode : Opcode
-
-    var tmp:DoubleWord
+    compute_op, imm, rd = decoded_utype[index]
```
OlivierBBB: So AFAICT [confirmed by inspection of `arithmetization/src/main/riscv/memory.zkc`]  there are `input` tables `decoded_xtype`, for `x = b, i, j, r, s, u`, and `decoded_core` table rather than a single `decoded` table. Given that the index is derived as `(pc - base) / 4` and that in `input` tables the `address` is required to be consecutive / gap free starting at address 0 up to the max address of that input table, see https://github.com/LFDT-Lineth/zkc/pull/1834, this means, AFAICT, that we will have 6/7 tables of size roughly `max_pc` rather than 1/2 such table(s).

What is the obstruction to having a single table ? The lookups from the `process_X_type_instruction` could target different columns in the one `decoded` table, so you could re-purpose columns to prevent the width of the module being too large.

arithmetization/src/main/riscv/interpreter.zkc
```diff
@@ -3,54 +3,54 @@ include "utils/register_utils.zkc"
...
+    // (see decoded_core in memory.zkc) and looked up by instruction index.
     var opcode:Opcode
-    instruction_parameters::opcode = instruction
-    instruction_type = instruction_type_from_opcode(opcode)
-
-    // TODO: handle the case where rd = 0: we will handle this in RAM itself, likely by adding an extra read-write pair
-    // to reset the value in registers[rd] to 0
+    var instruction_type:Type
+    var instruction_parameters:u25
+    opcode, instruction_type, instruction_parameters = decoded_core[index]
```
OlivierBBB: It seems to me that `instruction_parameters` isn't used (which aligns with my expectations) so should be removed.

arithmetization/src/main/riscv/interpreter.zkc
```diff
@@ -3,54 +3,54 @@ include "utils/register_utils.zkc"
...
 
     switch instruction_type {
```
OlivierBBB: My intuition was that decoding would completely remove this switch statement. Maybe that's overly simplistic. But I was imagining that, either here in `interpreter.zkc`, and maybe initially in the `main.zkc` module (when populating memory with the RV instructions corresponding to the compiled guest program's code), you would do a case free lookup to the one time decoding module. And actually do a full pass over the binary.

One reason for doing the one-time-decoding upfront in main is that this enforces non-malleability: you can't have different decoded programs (where differences would reside in sections of the program that aren't touched during execution) for a given RV binary. Otherwise you can tinker with the inputs by modifying the decoding of instructions that don't get touched during execution (e.g. if the binary contains unreachable, dead code).

This would be one attack direction.

One reason not to do it is not paying for sections of code that don't get executed. In particular for tiny blocks this could have an impact.

OlivierBBB: Maybe Claude has an opinion on this ...

OlivierBBB: **Note:** none of the above is an issue if the one-time decoding is baked into the prover's setup and we provide both the underlying RV binary and its supposed decoding. People can then just check that the one time decoding is valid offline and see the respective commitments being used in the proof.

arithmetization/src/main/riscv/interpreter.zkc
amkCha: yes, :100: - it's a leftover from continuous refactoring

arithmetization/src/main/riscv/instruction_processing/u_type.zkc
amkCha: As discussed, to reduce the number of lookups, decode_*_type tables have been collapsed into one

arithmetization/src/main/riscv/interpreter.zkc
amkCha: Yes, after our discussion, moved to a "giant" switch design

arithmetization/Makefile
```diff
@@ -11,7 +11,7 @@ REPO_ROOT := $(abspath $(MAKEFILE_DIR)/..)
 ZKC ?= $(REPO_ROOT)/zkc
 ZKC_REPO ?= https://github.com/LFDT-Lineth/zkc.git
 # Optional: pin a commit or branch (e.g. 763f30878e5fc260f0d88863c9883c904f8fc056). Leave empty for default branch.
-ZKC_REF ?=v1.2.20
+ZKC_REF ?=main
```
amkCha, line 15: pending the new version, to support wide range switches

arithmetization/src/test/scripts/elf_to_json_gen/README.md
```diff
@@ -0,0 +1,230 @@
+# elf_to_json_gen
+
+Generates the ZkC JSON input file for a given RISC-V ELF binary. It does two
+jobs:
+
+1. Converts the ELF's loadable sections (and an optional `IN_BYTES` input) into
+   the RAM blobs the interpreter loads before execution.
+2. **Statically pre-decodes** every instruction in the executable region into
+   lookup tables, so the interpreter can skip per-step instruction decoding.
+
```
lorenzogentile404, line 10: As the script is extremely big now, it seems it could make sense splitting it in two. Does the second part of the current script works on basically the JSON we had before? 
In that case we may have two scripts, one that generates the JSON and one that applies the predecoding to it.

arithmetization/src/test/scripts/elf_to_json_gen/README.md
```diff
@@ -0,0 +1,230 @@
...
+| ----------------------------- | ---------------------------------------------------- | ------------------------------------------------------------------ |
+| `entry_point_and_blobs_count` | `entry_point:Address, blobs_count:u64`               | ELF entry point + number of sparse RAM blobs                       |
+| `blobs_offset_and_size`       | `blob_offset:Address, blob_size:Length`              | RAM offset and byte length of each blob                            |
+| `blobs_executable`            | `executable:u1`                                      | `1` if blob `i` holds executable (`SHF_EXECINSTR`) bytes, else `0` |
+| `blobs_data`                  | `byte:u8`                                             | concatenated bytes of all blobs, in blob order                    |
+| `instruction_base`            | `base:Address`                                       | lowest executable address (4-aligned); maps `pc` → table index    |
+| `decoded`                     | `compute_op, imm, rs1, rs2, rd`                      | unified pre-decoded instruction table (one record per 4-byte word)|
+
```
lorenzogentile404, line 29: Could you add a visual representation of the predecoded JSON similar to this:

```json
{
  "entry_point_and_blobs_count": "0x<entry_point:u64>_<blobs_count:u64>",
  "blobs_offset_and_size": "0x<blob0_offset:u64>_<blob0_size:u64>____<blob1_offset:u64>_<blob1_size:u64>____...",
  "blobs_data": "0x<blob0_bytes>____<blob1_bytes>____..."
}
```

This one comes from `arithmetization/src/test/README.md`, which could be updated.

arithmetization/src/main/common/inputs.zkc
```diff
@@ -0,0 +1,27 @@
...
+// Whether blob i holds executable (SHF_EXECINSTR) section bytes (1) or data (0).
+pub input blobs_executable(i:u64) -> (executable:u1)
+
+// Concatenated bytes for all sparse memory blobs, in blob order
+pub input blobs_data(i:Address) -> (byte:u8)
+
+// Lowest executable instruction address (4-aligned). Used to convert a pc into
+// an index into the pre-decoded instruction tables below: index = (pc - base) >> 2
+pub input instruction_base(i:u1) -> (base:Address)
+
```
lorenzogentile404, line 25: It seems this could be concatenated to `blobs_offset_and_size`, turning it into  `blobs_offset_and_size_and_instruction_base`.

OlivierBBB (dismissed):
LGTM

letypequividelespoubelles (approved):
LGTM

PR #3353, opened 2026-06-12
feat(prover-ray): cleanup behind FRI in crypto

AlexandreBelling:
# PR: Consolidate and finish the KoalaBear FRI stack

## Context

The `crypto/koalabear` tree had accumulated several parallel, half-finished
attempts at the pieces FRI needs — separate `commitment`, `merkle`, `reedsolomon`,
`hash`, `smt`, `poly`, `poseidon2refactor`, and `fiatshamirrefactor` packages,
some duplicating each other, none wired together into a working prover/verifier.

This PR consolidates everything FRI depends on into a single `crypto/koalabear/fri`
package, finishes the protocol end to end over the extension field, and backs it
with a thorough bottom-up test suite. Net result is a large simplification
(~5.2k lines removed) and a FRI prover/verifier that builds, runs, and is tested.

## Summary

- **Consolidation.** Delete the redundant/legacy packages and move the live code
  into `fri`: Reed–Solomon encoder, Merkle tree, multi-size commitment, the FRI
  prover/verifier. Polynomial helpers move to `maths/koalabear/polynomials`.
- **Extension-field-only FRI.** Drop the base/ext "rail" split; the running
  polynomial and all levels are `field.Ext`. One prover (`proveExt`-equivalent),
  one verifier.
- **Prover as a coin-fed state machine** (`ProverState`): one folding challenge
  per transition, instead of consuming all coins in a single loop.
- **Bit-reversed codewords.** The RS encoder emits codewords in N-bit-reversed
  order so FRI conjugate pairs `(x, -x)` are adjacent; folding and the Merkle
  layout are aligned to this (mirrors Plonky3).
- **Verifier hardening + tests.** Structural validation that rejects malformed
  proofs early, plus a bottom-up test suite including reflection-based mutation
  testing.

## Package layout

Deleted (legacy / superseded / merged): `crypto/koalabear/{commitment, merkle,
reedsolomon, hash, smt, poly, poseidon2refactor, fiatshamirrefactor}`.

New/rewritten in `crypto/koalabear/fri/`:

- [reedsolomon.go](crypto/koalabear/fri/reedsolomon.go) — `RSEncoder`, `Encode`/`EncodeExt` (bit-reversed output).
- [tree.go](crypto/koalabear/fri/tree.go) — `Tree`, `Branch`, `NewTree` (3-ary multi-size), `OpenBranch`, `RecoverRoot`, `hashNode`.
- [multi_size_table.go](crypto/koalabear/fri/multi_size_table.go) — `MultiSizeTable`, `SizedTable`, well-formedness.
- [commitment.go](crypto/koalabear/fri/commitment.go) — `Commit`, `Encode`, `Merkleize` (crafts the values FRI verifies).
- [fri.go](crypto/koalabear/fri/fri.go) — `Params`, `Prove`, `Verify`, `foldLayerInternally`, `buildTreeExt`, query open/check.
- [fri_prover_state.go](crypto/koalabear/fri/fri_prover_state.go) — `ProverState` (`NewProverState`/`HasNext`/`Fold`/`Open`).

## Key design decisions

- **Bit-reversed folding.** `foldLayerInternally` folds adjacent pairs `(2j, 2j+1)`
  with twiddles built in natural order then bit-reversed (`gutils.BitReverse`),
  matching the encoder's output and Plonky3's `fold_matrix`. Output stays
  bit-reversed for the next round.
- **α²-batched multi-size levels.** A lower-degree level entering at round `jl`
  is mixed into the fold output as the auxiliary half-codeword scaled by `α²`
  (replacing a separate `γ` challenge). `ProverState.Fold` reads it from
  `levelAtRound[j+1]`; the verifier replays the same batching.
- **Two tree builders.** `NewTree` is the 3-ary multi-size builder (used by
  `Merkleize`); `newCompleteBinaryTree` is the plain power-of-two builder used by
  `buildTreeExt` for each FRI layer (no auxiliary leaves).
- **Leaf encoding.** Each `field.Ext` is padded into the low six coordinates of a
  `field.Octuplet` (`mapExtToOctuplet` / `octupletToExt`); conjugate pairs are
  adjacent sibling leaves, so one branch opens both halves.

## Bugs fixed (all surfaced by the new tests)

- `buildTreeExt` built no tree — it routed a single layer through `NewTree`, which
  cannot build a complete binary tree from one bottom layer. Now uses
  `newCompleteBinaryTree`.
- `openQueryExt` had a bogus layer-size assertion (`1<<(numRounds-1-j)`) that
  panicked for any rate > 1; replaced with "layers halve each round".
- `NewTree` level start position was `1<<i-1` instead of `n-1`, so it hashed the
  wrong nodes, never computed the root, and indexed `aux` out of range.
- `Merkleize` omitted the `log2(blowup)` empty top levels, so `NewTree` rejected
  every rate > 1 commitment; now pads to the full tree height.
- Latent: the prover folded with a zeroed local `alphas` slice rather than the
  supplied challenges (fixed by the state-machine refactor); `Open` opened extra
  levels at `s>>l` instead of `s>>jl`.

## Verifier security pass

`Verify` now performs structural validation before any hashing, so a malformed
proof is rejected cleanly instead of panicking or being silently accepted:
`len(openedPositions) ≥ NumQueries` and each position in `[0,N)`;
`len(alphas) ≥ numRounds`; `len(FinalPolyExt) == N>>numRounds` (exact); each query
has `numRounds` layers; every branch has `log2(layerLeaves)` siblings + matching
aux siblings (`checkBranchShape`). `RecoverRoot`'s empty-proof `panic` is now a
returned error.

## Testing

Bottom-up, table-driven, in `crypto/koalabear/fri/*_test.go`:

- [tree_test.go](crypto/koalabear/fri/tree_test.go) — octuplet↔ext round-trip; tree open/recover with sibling-layout and tamper checks.
- [fri_test.go](crypto/koalabear/fri/fri_test.go) — `foldLayerInternally` vs. independent re-evaluation of `Q = P_e + αP_o` (+ α²·aux); end-to-end `Prove`/`Verify` over single / one-extra / two-extra-level configs with tamper rejection.
- [commitment_test.go](crypto/koalabear/fri/commitment_test.go) — well-formedness, encoder validation, `Encode`, and full `Commit` round-trip (blowups 2/4, empty-middle level).
- [fri_mutation_test.go](crypto/koalabear/fri/fri_mutation_test.go) — reflection mutation testing: ~147 single-field/length mutations of the `Proof`, each must be rejected by `Verify` without panicking.
- [reedsolomon_test.go](crypto/koalabear/fri/reedsolomon_test.go) — encode evaluation correctness.

Run:

```
cd prover-ray && go test ./crypto/koalabear/fri/... && go vet ./crypto/koalabear/fri/...
```

All tests pass; `vet` is clean.

## Follow-ups (not in this PR)

- Commitment→FRI **integration** test (the commitment crafts the FRI-verified
  values, but the two are currently tested in isolation).
- A larger realistic-size config (e.g. `N=1024, D=64`) to stress depth/bit-reversal.
- Minor: `provePlan.numLevels` is redundant with `len(levels)`; a few low-priority
  TODOs remain (`Level.Tree` stores leaves twice; `NewParams` field-export decision).


AlexandreBelling:
Added @gbotrel @ivokub @arijitdutta67 @YaoJGalteland @Tabaie as reviewer as the PR might have ripple on their work.

prover-ray/crypto/koalabear/fri/commitment.go
```diff
@@ -0,0 +1,137 @@
...
+					ext.B1.A0, ext.B1.A1,
+					ext.B2.A0, ext.B2.A1)
+			}
+
+			leaves[i][j] = hasher.SumDigest()
+		}
+	}
+
+	// NewTree expects the levels from largest to smallest, one per tree level.
+	slices.Reverse(leaves)
```
YaoJGalteland: why don’t you build `MultiSizeTable` in descending order directly? Then you wouldn’t need the reverse computation here.

prover-ray/crypto/koalabear/fri/fri.go
```diff
@@ -1,49 +1,33 @@
...
-	D          int // 2^m: degree of the purported polynomial
+	N          int // 2^n: the dimension of the code
+	D          int // 2^m: the size of the plaintext polynomial
 	NumQueries int // number of independent queries (controls soundness error ≈ (1-δ)^Q)
-	LeafHasher commitment.LeafHasher
-	NodeHasher commitment.NodeHasher
 
 	numRounds    int // numRounds = m
-	invTwo       koalabear.Element
+	invTwo       field.Element
```
Tabaie: It's better to halve elements by `x -> (x&1==0) ? (x>>1) : (x+p >> 1)`

prover-ray/crypto/koalabear/fri/fri.go
```diff
@@ -1,49 +1,33 @@
...
 	domainsLight []domainLight // domainLight stores only the cardinality and the domain generator
-	grinding     int           // grinding bits for PoW, on the alpha
 }
 
 type Config struct {
 	WoFullDomainAllocation bool
-	// Grinding is the number of grinding bits for PoW. More grinding bits => more cpu work for the
-	// prover, but security goes from log_blowup * num_queries to
-	// log_blowup * num_queries + query_proof_of_work_bits.
-	Grinding int
```
Tabaie, line 46: Is this a definite decision that we won't do grinding?

prover-ray/crypto/koalabear/fri/fri.go
```diff
@@ -266,70 +146,50 @@ func (p Params) BuildLevelTreeExt(layer []ext.E6) (*merkle.Tree, error) {
 // with the query positions. levels[0].D must equal p.D and every Level must
 // contain one evaluation vector on exactly one rail. levels is sorted in-place
 // in decreasing order of D.
-// ts must already have been initialised with any prior-round context.
-func Prove(p Params, levels []Level, ts *fiatshamirrefactor.Transcript) (Proof, []int, error) {
-	sort.Slice(levels, func(i, j int) bool { return levels[i].D > levels[j].D })
+func Prove(p Params, levels []Level, alphas []field.Ext, openedPositions []int) Proof {
```
Tabaie: How can it be sound for the prover to take all the verifier challenges as a-priori input?

prover-ray/crypto/koalabear/fri/fri.go
```diff
@@ -338,40 +198,35 @@ func buildProvePlan(p Params, levels []Level) (provePlan, error) {
...
+		jl := utils.Log2Ceil(p.D / levels[l].D)
+		if jl < 1 || jl >= p.numRounds {
+			return plan, fmt.Errorf(
+				"fri: Prove: levels[%d].D=%d gives intro round %d, must be in 1..%d",
+				l, levels[l].D, jl, p.numRounds-1)
 		}
-		if levelRail != rail {
-			return plan, fmt.Errorf("fri: Prove: levels[%d] is %s, running rail is %s", l, levelRail, rail)
+		if _, dup := plan.levelAtRound[jl]; dup {
+			return plan, fmt.Errorf("fri: Prove: two levels share intro round %d", jl)
```
Tabaie, line 202: What if there are two tables of the same height in different rounds of the ambient protocol?

prover-ray/crypto/koalabear/fri/tree.go
```diff
@@ -0,0 +1,268 @@
...
+	Siblings []field.Octuplet
+	// AuxSiblings are the auxiliary siblings. We have
+	// `len(Siblings) == len(AuxSiblings)``
+	AuxSiblings []*field.Octuplet
+}
+
+// NewTree builds a new Tree from the given leaves. The leaves must be  provided
+// in the following order.
+//
+//	for all 0 <= i < len(leaves): leaves[i] = 2**(N-i-1) or 0
```
Tabaie: ~`leaves[i]`~`len(leaves[i])`

prover-ray/crypto/koalabear/fri/tree.go
```diff
@@ -0,0 +1,268 @@
...
+		branch.Siblings[level] = t.Nodes[sibling]
+		idxRemBits >>= 1
+		current = parent
+	}
+
+	return branch
+}
+
+// RecoverRoot recovers the root of the tree from a branch and a position. The
+// function errors if
```
Tabaie: Incomplete sentence

prover-ray/maths/koalabear/polynomials/lagrange.go
```diff
@@ -91,36 +93,104 @@ func EvalLagrangeBatch(p field.Vec, zs []field.Gen) []field.Gen {
...
+	omega, err := fft.Generator(uint64(n))
+	if err != nil {
+		panic(err) // there is no actual error case, this is a void error
+	}
+
+	for j := 0; j < n; j++ {
+		var num, den field.Ext
+		num.Sub(&v, &p[j])
+		field.SubByBase(&den, &z, &omegaJ)
+		den.Inverse(&den)
```
Tabaie, line 190: Not urgent but probably better to batch invert here

Tabaie (dismissed):
I only see one major issue: `fri.Prove` expects the coins to be known a-priori, which is unsound.

prover-ray/crypto/koalabear/fri/fri.go
Tabaie: Ah, the DEEP layer which will make the commitment scheme will already have folded all polynomials of the same size into one. So do we expect each leaf of each individual tree to be of length 1?

Tabaie: If that is so we can skip pre-hashing the leaves and save a lot of work.

prover-ray/crypto/koalabear/fri/tree.go
```diff
@@ -0,0 +1,268 @@
...
+	if currPos > 0 {
+		return field.Octuplet{}, errors.New("all bits of currPos should have been bitshifted beyond LSb")
+	}
+
+	return ancestor, nil
+}
+
+// hashNode hashes two field.Octuplets and an optional field.Octuplet.
+func hashNode(left, right field.Octuplet, aux *field.Octuplet) field.Octuplet {
+	hasher := poseidon2.NewMDHasher()
```
Tabaie: I think since the prover and verifier already agree on the tree shape, we can use the compression function directly and save one hash permutation.
i.e. `C(left, right)` or `C(aux, C(left, right))` if there is an aux,
rather than `C(C(0,left), right)` or `C(C(C(0,left), right), aux)`.

AlexandreBelling:
> I only see one major issue: `fri.Prove` expects the coins to be known a-priori, which is unsound.

That's OK. The Prove function is for testing. The real function is the state-machine implementation. It was implemented like that in the earlier version. The reason is that wiop has already a way to track the FS state and we don't want to duplicate that part. The FRI implementation is agnostic to Fiat-Shamir but this is expected.

prover-ray/crypto/koalabear/fri/fri.go
AlexandreBelling: It's for testing. Lemme rename that

prover-ray/crypto/koalabear/fri/fri.go
AlexandreBelling: >What if there are two tables of the same height in different rounds of the ambient protocol?

That's expected but they would still get merged into the same quotient. My thinking that that we should update the protocol so that we can explicitly support passing the complete list of quotient to fold directly. This would remove one step.

prover-ray/crypto/koalabear/fri/fri.go
AlexandreBelling: Not definitive, but a contrario no decision to use it were made. So I prefer to not maintain the implementation to make thing simple.

prover-ray/maths/koalabear/polynomials/lagrange.go
AlexandreBelling: Can leave a comment for that

prover-ray/crypto/koalabear/fri/fri.go
```diff
@@ -847,523 +365,270 @@ func verifyBase(
...
-
-			diff.Sub(&p, &q)
-			diff.Mul(&diff, &invTwo)
-			diff.Mul(&diff, &xInv)
-			diff.Mul(&diff, &alpha)
-
-			next[i].Add(&sum, &diff)
-			xInv.Mul(&xInv, &domain.GeneratorInv)
+		// if there is an aux, add it
+		if len(aux) > 0 {
```
YaoJGalteland, line 468: `len(aux) > 0` has the same value for every `j` in the loop. `aux` either has `half` elements or is nil/empty, and it doesn’t change inside the loop.

So I think we can check it once before the `for` loop instead of checking it on every iteration. `half` can be as large as `2^20`, so avoiding the repeated check seems worthwhile.

prover-ray/crypto/koalabear/fri/fri.go
```diff
@@ -847,523 +365,270 @@ func verifyBase(
...
+			auxTerm := new(field.Ext).Mul(&aux[j], alpha2)
```
YaoJGalteland: `auxTerm` could be declared once outside the loop as a stack var:

```go
var auxTerm field.Ext
```

Then we can reuse it with `auxTerm.Mul(...)` on each iteration.

Right now `new(field.Ext)` allocates a fresh heap object every iteration, which could happen ~ `2^20` times in a fold round.

prover-ray/crypto/koalabear/fri/fri.go
```diff
@@ -847,523 +365,270 @@ func verifyBase(
...
-	return res
-}
+// octupletToExt converts an octuplet into a field extension. It expects its
+// coordinates 6 and 7 to be zero.
+func octupletToExt(o field.Octuplet) (field.Ext, error) {
 
-func queryIndex(challenge hash.Digest, modulus int) int {
-	if modulus <= 0 {
-		return 0
+	if !o[6].IsZero() && !o[7].IsZero() {
```
YaoJGalteland: && -> ||

If o6 != 0 and o7 == 0, then:

!o[6].IsZero() && !o[7].IsZero() == false, so the panic is skipped

prover-ray/crypto/koalabear/fri/fri.go
```diff
@@ -847,523 +365,270 @@ func verifyBase(
...
+			}
+			if lvRoot != levelRoots[li-1] {
+				return fmt.Errorf("level %d: Merkle proof invalid", li)
 			}
+			aux, err := octupletToExt(lv.Leaf)
+			if err != nil {
+				return fmt.Errorf("level %d: decode leaf: %w", li, err)
+			}
+			var alpha2, term field.Ext
+			alpha2.Square(&alphas[j])
```
YaoJGalteland: alpha2 is recomputed per query, best to put it before the query loop: 

```
// precompute alpha2
alpha2= ...

for k := 0; k < p.NumQueries; k++ {
    ...
    checkQueryExt(..., alpha2s, ...)
}
```

prover-ray/crypto/koalabear/fri/commitment.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	if k != 1 {
+		panic("k must be one")
+	}
+
+	encoded := witness.Encode(encoder)
+	tree := encoded.Merkleize()
```
YaoJGalteland, line 32: the commitment tree's full lifepath seems to be: built → root sent to verifier via transcript → abandoned. no consistency check in Verify below

prover-ray/crypto/koalabear/fri/fri.go
```diff
@@ -700,7 +229,9 @@ func proveExt(p Params, levels []Level, plan provePlan, ts *fiatshamirrefactor.T
 // called (i.e. decreasing D).
 //
 // ts must be in the same state as when Prove was called.
-func Verify(p Params, levelRoots []hash.Digest, levelDs []int, prf Proof, ts *fiatshamirrefactor.Transcript) error {
+func Verify(p Params, levelRoots []field.Octuplet, levelDs []int, prf Proof,
```
YaoJGalteland, line 232: I didn’t find any consistency check for the 3-ary commitment tree, i.e. the Merkle authentication that binds the auxiliary third-child commitments into the main commitment tree.

I also don’t see preimage binding for the aux leaves.

I think we need to prove that the leaves are hashes of `table[i]`, and that the tree root is built from those leaves.

Am I missing where that check happens?

prover-ray/crypto/koalabear/fri/fri.go
```diff
@@ -847,523 +365,270 @@ func verifyBase(
 			}
 		}
 
-		if err := checkQuery(s, prf.FRIQueries[k], levelQueriesForQuery, levelRootsExtra,
-			introductions, gammas, roots, prf.FinalPolyBase, alphas, p); err != nil {
+		if err := checkQueryExt(s, prf.FRIQueries[k], levelQueriesForQuery, levelRootsExtra,
```
YaoJGalteland, line 368: do we need some consistency check for `levelRoots`, for example linking them with the 3-ary commitment tree?

I feel a bit confused now. A clear diagram showing which values are committed, and how they are related, would make this much easier to follow. It would also help ensure we are not copying or regenerating some data without properly linking it back to the committed values.

YaoJGalteland (commented):
I have a few questions about the consistency checks. not sure all of the required checks are included.

prover-ray/crypto/koalabear/fri/commitment.go
YaoJGalteland: I found the two trees confused me a bit.

`levelRoots` are the roots of the quotient polynomial trees, while the 3-ary multi-layer tree is built over all the initial polynomials grouped by degree. This seems to belong more to the DEEP-FRI process.

I’m not sure if it is clear to keep all of this in the `fri` folder, or whether part of it should live in a PCS layer.

To me, the split could be:

**FRI layer** — what is currently in the `fri` folder:
pure polynomial commitment to a single codeword: RS encode, Merkleize, fold, query. It doesn’t need to know about DEEP quotients or what the committed polynomial means.

**PCS layer** — not sure if this exists yet:
the DEEP-FRI protocol that ties everything together:

* commit to `P_i` via the 3-ary initial commitment tree
* absorb `P_i(z)` into the transcript, then draw `z` and batching challenges
* compute `Q = Σ λ_i · (P_i - P_i(z)) / (· - z)`
* run FRI on `Q`, calling into the `fri` folder
* at query positions, open the initial commitment tree to get raw `P_i(x_s)`, verify against the 3-ary root, then check the DEEP relation against `Q(x_s)`

Does this split sound right to you?

prover-ray/crypto/koalabear/fri/fri.go
AlexandreBelling: Agreed. But as the same, time I was take this as a potential future optimization that we can do. I'd prefer merging the PR as is and leaving this for future work.

prover-ray/crypto/koalabear/fri/fri.go
AlexandreBelling: Likewise, this is an optimization that we can leave in further PR. I don't think this should block the merging even though the insight is right.

prover-ray/crypto/koalabear/fri/commitment.go
AlexandreBelling: My thinking is that everything should sit in the same package. But the PCS should be a separate structure than FRI (but may contain FRI as part of itself). To me, they cannot be completely separated.

prover-ray/crypto/koalabear/fri/fri.go
AlexandreBelling: It's in recover root, L555

```
		// Authenticate the opened leaf against the round-j root. The hashing now
		// lives in the tree itself (Branch.RecoverRoot replays hashNode up to the
		// root), so there is no separate leaf/node hasher to call.
		root, err := branch.RecoverRoot(base)
		if err != nil {
			return fmt.Errorf("round %d: recover root: %w", j, err)
		}
		if root != roots[j] {
			return fmt.Errorf("round %d: Merkle proof invalid (base=%d)", j, base)
		}
```

prover-ray/crypto/koalabear/fri/fri.go
AlexandreBelling: Resolving as this folds into the previous comment.

prover-ray/crypto/koalabear/fri/fri.go
AlexandreBelling: I ended up doing it anyway alongside the other changes

prover-ray/crypto/koalabear/fri/fri.go
AlexandreBelling: Yeah, I checked again and optimizing the if-in-for-loop sounds premature to me.

prover-ray/crypto/koalabear/fri/fri.go
```diff
@@ -593,6 +597,9 @@ func checkQueryExt(s int, fq Query,
 		diff.Mul(&diff, &alphas[j])
 		expected.Add(&sum, &diff)
 
+		var alpha2, term field.Ext
+		alpha2.Square(&alphas[j])
```
YaoJGalteland, line 601: I think this should be moved outside of `checkQueryExt` to avoid recomputing it.

In `verifyExt`, before the query loop:

```go
alpha2s := make([]field.Ext, len(alphas))
for j := range alphas {
    alpha2s[j].Square(&alphas[j])
}

for k := 0; k < p.NumQueries; k++ {
    checkQueryExt(alpha2s, ...)
}
```

Then `checkQueryExt` can reuse the precomputed `alpha2s`.

prover-ray/crypto/koalabear/fri/fri.go
YaoJGalteland: Thank you for pointing the check. 

My understanding is that `RecoverRoot` only proves that a leaf is included in the commitment tree, i.e. the Merkle path inside (1).

Let me try to map the full picture more precisely.

```text
witness columns (raw evaluations)
       │
       ├────────────────────────────── (1) Commitment tree
       │   leaf[j]    = MDHash(col0[j], col1[j], ...)
       │   auxLeaf[j] = MDHash(col0'[j], col1'[j], ...)
       │   tree       = NewTree(leaves)
       │   root       → transcript
       │
       └────────────────────────────── (2) FRI levels
           DQ[j] = Σᵢ γⁱ · (colᵢ[j] - colᵢ(r)) / (ωʲ - r)  (The deep-quotient polynomial doesn't seem to implemented yet, the levels are random elements in test, no relation with the witness evaluations)
           Level{Evals: DQ, Tree: buildTreeExt(DQ)}
```



It does not prove:

* `leaf[j] = MDHash(col0[j], col1[j], ...)`, so the leaf is actually the hash of the opened column values
* `DQ[j]` was correctly computed from those same witness columns
* the bridge between (1) and (2), meaning the columns used to build the commitment tree are the same columns used to build `DQ`

So I think the missing part is the consistency check linking the opened initial polynomial values to both the commitment tree and the DEEP quotient computation.

prover-ray/crypto/koalabear/fri/fri.go
```diff
@@ -847,523 +359,270 @@ func verifyBase(
...
 
-		pair := []commitment.PairBase{{layer.LeafPBase, layer.LeafQBase}}
-		leaf := p.LeafHasher.HashLeaf(pair, nil)
-		if !merkle.Verify(roots[j], layer.Path, leaf, p.NodeHasher) {
-			return fmt.Errorf("round %d: Merkle proof invalid (base=%d)", j, base)
-		}
+		// x is the domain point of the opened leaf. The codeword is bit-reversed,
+		// so the natural-order index of position base is bitReverse(base) and
+		// x = gⱼ^{bitReverse(base)} with gⱼ the size-Nⱼ generator.
+		xExp := int(bits.Reverse64(uint64(base)) >> (64 - kj))
```
Tabaie, line 585: Wouldn't it be better for the loop to iterate "in bit-reversed order" so that the exp and inversion at each iteration are replaced with two multiplications?

Tabaie: I suppose cache misses will negate some of the gains but the difference in the amount of arithmetic done should overwhelm that.

prover-ray/crypto/koalabear/fri/fri.go
AlexandreBelling: You can do that. But at this point, this becomes a 0.000001% saving. I prefer to let my code as is as it is simpler.

prover-ray/crypto/koalabear/fri/fri.go
AlexandreBelling: We could try that. But I'd support doing it on a separate PR as this is optimization.

arijitdutta67 (approved):

gusiri (approved):

PR #3354, opened 2026-06-12
refactor(prover-ray): export logderivativesum internals for verifier-ray codegen

YaoJGalteland:
Exported `zEntry` → `ZEntry`, `ZFinal`, `verifierAction` → `VerifierAction` with `Ld` and `Entries`, and `resultIsZeroVerifierAction` → `ResultIsZeroVerifierAction` with `Ld`, in both `logderivativesum` and `lookuptologderivsum`.

No behavioral change — this is purely a visibility refactor. All exported fields carry the same values as before, and the rename follows Go’s capitalization convention for exported identifiers.

### DIFF
--- a/prover-ray/wiop/codegen/codegen_test.go
+++ b/prover-ray/wiop/codegen/codegen_test.go
@@ -16,6 +16,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const testPackageName = "generated"
+
 // compiledFibSystem builds the Fibonacci scenario, runs the global compiler,
 // and returns the compiled system together with collected ActionVars.
 func compiledFibSystem(t *testing.T) (*wiop.System, codegen.ActionVars) {
@@ -37,7 +39,7 @@ func generate(t *testing.T, sys *wiop.System, vars codegen.ActionVars, opts code
 // TestGenerate_ValidGoSyntax verifies that the output parses as valid Go source.
 func TestGenerate_ValidGoSyntax(t *testing.T) {
 	sys, vars := compiledFibSystem(t)
-	src := generate(t, sys, vars, codegen.Options{Package: "generated"})
+	src := generate(t, sys, vars, codegen.Options{Package: testPackageName})
 
 	fset := token.NewFileSet()
 	_, err := parser.ParseFile(fset, "prove_gen.go", src, parser.AllErrors)
@@ -54,7 +56,7 @@ func TestGenerate_FuncName(t *testing.T) {
 // TestGenerate_DefaultFuncName verifies that the default function name is "Prove".
 func TestGenerate_DefaultFuncName(t *testing.T) {
 	sys, vars := compiledFibSystem(t)
-	src := generate(t, sys, vars, codegen.Options{Package: "generated"})
+	src := generate(t, sys, vars, codegen.Options{Package: testPackageName})
 	assert.Contains(t, src, "func Prove(")
 }
 
@@ -64,7 +66,7 @@ func TestGenerate_DefaultFuncName(t *testing.T) {
 //   - there is exactly one .Check(rt) call per registered VerifierAction
 func TestGenerate_RoundStructure(t *testing.T) {
 	sys, vars := compiledFibSystem(t)
-	src := generate(t, sys, vars, codegen.Options{Package: "generated"})
+	src := generate(t, sys, vars, codegen.Options{Package: testPackageName})
 
 	var wantProver, wantVerifier int
 	for _, r := range sys.Rounds {
@@ -83,7 +85,7 @@ func TestGenerate_RoundStructure(t *testing.T) {
 // TestGenerate_RoundComments verifies that each round has a section comment.
 func TestGenerate_RoundComments(t *testing.T) {
 	sys, vars := compiledFibSystem(t)
-	src := generate(t, sys, vars, codegen.Options{Package: "generated"})
+	src := generate(t, sys, vars, codegen.Options{Package: testPackageName})
 
 	for i := range sys.Rounds {
 		assert.Contains(t, src, "=== Round", "round %d must have a section comment", i)
@@ -165,7 +167,7 @@ func normaliseSource(t *testing.T, src string) string {
 // imports, var block (with concrete action types), and Prove function body.
 func TestGenerate_GoldenFibonacci(t *testing.T) {
 	sys, vars := compiledFibSystem(t)
-	actual := generate(t, sys, vars, codegen.Options{Package: "generated"})
+	actual := generate(t, sys, vars, codegen.Options{Package: testPackageName})
 
 	assert.Equal(t,
 		normaliseSource(t, fibExpectedSource),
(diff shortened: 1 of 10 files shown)


prover/backend/invalidity/prove_test.go
```diff
@@ -169,6 +169,7 @@ func TestProveBadPrecompile(t *testing.T) {
 	builder := invalidity.NewBuilder(
 		invalidity.Config{
 			ZkEvmComp: comp,
+			MaxL2Logs: 16,
```
ivokub: Pinging @Soleimani193 for checking the change.

prover/circuits/invalidity/setup_test.go
```diff
@@ -59,6 +59,7 @@ func TestSetupRoundTripPrecompileLogs(t *testing.T) {
 	builder := invalidity.NewBuilder(
 		invalidity.Config{
 			ZkEvmComp: comp,
+			MaxL2Logs: 16,
```
ivokub: @Soleimani193 - can you confirm?

prover-ray/wiop/codegen/codegen_test.go
```diff
@@ -16,6 +16,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const testPackageName = "generated"
```
ivokub, line 19: It is not consistently applied, see for example method `TestGenerate_FuncName`

prover-ray/wiop/compilers/logderivativesum/logderivativesum.go
```diff
@@ -374,32 +377,38 @@ func genToExt(v field.Gen) field.Ext {
...
-type verifierAction struct {
-	ld      *wiop.LogDerivativeSum
-	entries []zEntry
+//
+// Exported (with exported fields) so out-of-package consumers — notably the
+// verifier-ray codegen — can read the endpoint openings and the Result cell.
+type VerifierAction struct {
+	// Ld is exported so codegen can read Result (cell coordinate) and use the
+	// pointer as a map key to correlate with ResultIsZeroVerifierAction.
+	Ld *wiop.LogDerivativeSum
```
ivokub: When it becomes exported, then I would use some more descriptive name for the field i.e. `LogDerivativeSum`. We can embed i.e.
```go
type VerifierAction struct {
    *wiop.LogDerivativeSum
    Entries []ZEntry
}
```
What we have to keep in mind when embedding is that `*wiop.LogDerivativeSum` also defines `Check(wiop.Runtime)` method, but we overwrite it here.

prover-ray/wiop/compilers/logderivativesum/logderivativesum.go
```diff
@@ -374,32 +377,38 @@ func genToExt(v field.Gen) field.Ext {
...
 // row-0 local constraint registered in buildZ, so this action reads only
 // local openings (cells) — never the oracle witness columns.
-type verifierAction struct {
-	ld      *wiop.LogDerivativeSum
-	entries []zEntry
+//
+// Exported (with exported fields) so out-of-package consumers — notably the
+// verifier-ray codegen — can read the endpoint openings and the Result cell.
+type VerifierAction struct {
+	// Ld is exported so codegen can read Result (cell coordinate) and use the
```
ivokub: I think we can drop the comment. It is redundant with type-level comment already. Also same for `[]Entries` below.

prover-ray/wiop/compilers/logderivativesum/logderivativesum.go
```diff
@@ -127,13 +127,16 @@ func fractionModule(f wiop.Fraction) *wiop.Module {
 	return f.Denominator.Module()
 }
 
-// zEntry collects the per-Z artefacts shared by the prover and verifier
+// ZEntry collects the per-Z artefacts shared by the prover and verifier
 // actions: the Z column, the raw fractions (filter, num, den) used by the
 // prover for filter-aware row skipping, and the column endpoint opening.
-type zEntry struct {
+// ZFinal is exported so out-of-package consumers (e.g. the verifier-ray
```
ivokub: We can drop the comment change.

prover-ray/wiop/compilers/lookuptologderivsum/lookuptologderivsum.go
```diff
@@ -411,22 +411,28 @@ func viewExprs(cols []*wiop.ColumnView) []wiop.Expression {
...
 // T-side fractions exactly when every selected A row is in the union of
 // selected B rows.
-type resultIsZeroVerifierAction struct {
-	ld *wiop.LogDerivativeSum
+//
+// Exported (with an exported field) so out-of-package consumers — notably the
+// verifier-ray codegen — can recognise a lookup-reduced LogDerivativeSum and
+// flag that its Result must be zero.
+type ResultIsZeroVerifierAction struct {
+	// Ld is exported so codegen can use the pointer as a map key to flag the
```
ivokub: We can drop the field comment. It is redundant with type-level comment.

prover-ray/wiop/compilers/lookuptologderivsum/lookuptologderivsum.go
```diff
@@ -411,22 +411,28 @@ func viewExprs(cols []*wiop.ColumnView) []wiop.Expression {
...
+	// corresponding LogDerivativeSum query's Result as required to be zero.
+	Ld *wiop.LogDerivativeSum
```
ivokub: Now when exported it should have more descriptive field name, we could embed i.e.:
```go
type ResultIsZeroVerifierAction struct {
    *wiop.LogDerivativeSum
}
```
What we have to keep in mind when embedding is that `*wiop.LogDerivativeSum` also defines `Check(wiop.Runtime)` method, but we overwrite it here.

ivokub (changes requested):

prover-ray/wiop/compilers/lookuptologderivsum/lookuptologderivsum.go
ivokub: Perhaps just to be on the safe side we should do:
```go
type ResultIsZeroVerifierAction struct {
    LogDerivSum *wiop.LogDerivativeSum
}
```
in this case we don't inherit the Check method from the embedded field.

prover-ray/wiop/compilers/logderivativesum/logderivativesum.go
ivokub: Similar comment, for safety just in case we shouldn't embed perhaps to avoid inheritance of Check method of the wiop.LogDerivativeSum.

prover/backend/invalidity/prove_test.go
YaoJGalteland: this was caught by a CI check.

Without `MaxL2Logs` in `Config`, `circuit.MaxL2Logs` would be `0`. The panic guard:

```go
config.MaxL2Logs <= 0 { panic(...) }
```

would catch the invalid circuit config.

`16` comes from `config.GetTestTracesLimits().BlockL2L1Logs()`.

ivokub: Thanks! I was just wondering why the CI didn't fail before on that, seems odd. But in that case wouldn't it be better to use directly `config.GetTestTracesLimits().BlockL2L1Logs()` instead of hardcoding the value?

ivokub (approved):
Thanks! Looks perfect!

arijitdutta67 (approved):
LGTM

Issue #3331, opened 2026-06-09
[RV verifier] - check LookupToLogDerivativeSum and LogDerivativeSum verifier implementation


ivokub:
Previously we would have needed separate verifier step in ray verifier as it used directly local opening, but now perhaps it works with general vanishing polynomial check and lagrange evaluation. Double check what the queries compile to and how we need to implement on the verifier side.

PR #3357, opened 2026-06-12
feat(verifier-ray): add LogDerivativeSum verifier

YaoJGalteland:
This PR implements issue(s) #3331

## What this adds

A complete `LogDerivativeSum` sub-verifier for verifier-ray, covering both the Go codegen layer and the Zig runtime verifier.

It implements the one remaining boundary check that the `logderivativesum` compiler leaves to the verifier:

```text
Σᵢ Zᵢ[n-1] == Result
```

(and `Result == 0` for lookup-reduced queries).

`ZFinal` and `Result` are read from `ctx.rounds` at verify time, i.e. from the Fiat-Shamir transcript, rather than being baked in as honest-prover constants.

This is a soundness requirement: fixed-position cell openings must be committed before the evaluation coin is derived.


verifier-ray/codegen/logderivativesum_zig.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// WriteLogDerivSystemZig writes the Zig source for a single LogDerivSystem,
+// emitting `system_<index>_logderiv` (plus its backing arrays). It emits data
+// only; the Zig sub-verifier owns the boundary-check implementation.
+func WriteLogDerivSystemZig(w io.Writer, index int, system LogDerivSystem) error {
+	return WriteLogDerivSystemZigWithOptions(w, index, system, defaultLogDerivZigOptions())
+}
+
+func WriteLogDerivSystemZigWithOptions(w io.Writer, index int, system LogDerivSystem, opts LogDerivZigOptions) error {
```
ivokub, line 31: I would keep similar style code generation as in `spec_zig.go` and `vanishing_zig.go` i.e. using `text/template`

verifier-ray/src/query/logderivativesum.zig
```diff
@@ -0,0 +1,62 @@
...
+
+        // The final-sum identity links the Z endpoints to the claimed result.
+        if (!sum.eql(result)) return error.FinalSumMismatch;
+
+        // Lookup queries reduce to a LogDerivativeSum whose result must be 0.
+        if (query.result_is_zero and !result.isZero()) return error.LookupResultNonZero;
+    }
+}
+
+fn scalarToExt(value: protocol.Scalar) ext.Ext {
```
ivokub: Perhaps move to `field/value.zig` and make as a method on protocol.Scalar?

ivokub: And I think later on we could perhaps concretize the check in `verify` side -- we always lift to extension, but there is no point if the scalars all come in as base elements.

Lets keep it as is for now, but we should later on imo definitely think about it, perhaps as a TODO for later pass?

verifier-ray/src/verifier.zig
```diff
@@ -86,11 +92,12 @@ pub fn verify(
     };
 
     // Step 3 — dispatch each sub-verifier with ctx + its own claims.
-    // TODO(new-sub-verifier): add dispatch call here — step 4 above.
+    try logderivativesum.verify(systems.logderivativesum, ctx);
```
ivokub: A small nitpick - the ordering doesn't matter in practice, but in `Systems` logderivative.System comes after vanishing.System

verifier-ray/test/logderivativesum_test.zig
```diff
@@ -0,0 +1,67 @@
...
+// both at (round=0, index=0) and (round=0, index=1) respectively.
+fn oneQuerySystem(comptime result_is_zero: bool) logderivativesum.System {
+    return .{ .queries = &[_]logderivativesum.Query{.{
+        .z_final_refs = &[_]logderivativesum.ScalarRef{.{ .round = 0, .index = 0 }},
+        .result_ref = .{ .round = 0, .index = 1 },
+        .result_is_zero = result_is_zero,
+    }} };
+}
+
+fn baseScalar(v: u32) protocol.Scalar {
```
ivokub, line 32: The function is defined here but then not used below to define five, seven etc.

verifier-ray/test/logderivativesum_test.zig
```diff
@@ -0,0 +1,67 @@
...
+fn makeCtx(comptime cells: []const protocol.Scalar) protocol.Context {
+    const rounds: []const protocol.RoundMessage = &[_]protocol.RoundMessage{
+        .{ .columns = &.{}, .cells = cells },
+    };
+    return .{ .all_coins = &.{}, .rounds = rounds };
+}
+
+// oneQuerySystem returns a System with one query whose z_final and result are
+// both at (round=0, index=0) and (round=0, index=1) respectively.
+fn oneQuerySystem(comptime result_is_zero: bool) logderivativesum.System {
```
ivokub, line 24: Perhaps we should also have tests where we have multiple refs? Just for completeness.

verifier-ray/testdata/generate/main.go
```diff
@@ -833,15 +816,20 @@ func writeVanishingFixtures(cases []vanishingFixtureCase, systems []codegen.Name
...
 		}
-		if err := codegen.WriteVanishingSystemZigWithOptions(&out, i, systems[i], codegen.VanishingZigOptions{
+		if err := codegen.WriteVanishingSystemZigWithOptions(&out, i, systems[i].Vanishing, codegen.VanishingZigOptions{
 			FieldImport:     "verifier_ray.field.koalabear",
 			VanishingImport: "verifier_ray.query.vanishing",
 		}); err != nil {
 			return err
 		}
-		writeVanishingScenario(&out, i, cases[i])
+		if err := codegen.WriteLogDerivSystemZigWithOptions(&out, i, systems[i].LogDeriv, codegen.LogDerivZigOptions{
```
ivokub: Perhaps we should have in `codegen` a helper method which calls `WriteSpec`, `WriteVanishingSystem` and `WriteLogDerivSystem` etc., and it automatically figures out if to emit headers etc.

verifier-ray/testdata/generated/vanishing.zig
```diff
@@ -50,10 +51,16 @@ const system_0 = vanishing.System{
     .total_quotient_claims = 1,
 };
 
-const vanishing_scenario_0 = VanishingScenario{
+const system_0_logderiv_queries = [_]logderivativesum.Query{};
+
+// logderiv system: "bool-col"
```
ivokub: Perhaps it would be better if we have the comment above `system_%d_logderiv_queries`?

verifier-ray/testdata/generated/vanishing.zig
```diff
@@ -4,20 +4,21 @@ const verifier_ray = @import("verifier_ray");
...
+pub const Scenario = struct { name: []const u8, spec: protocol.Spec, system: vanishing.System, logderiv: logderivativesum.System = .{}, honest: VanishingProofView, invalid: ?VanishingProofView = null };
 
 pub const system_0_spec = protocol.Spec{
     .round_coin_counts = &[_]usize{ 0, 1, 1 },
     .round_coin_offsets = &[_]usize{ 0, 0, 1 },
     .total_round_coins = 2,
 };
 
-// scenario: "Vanishing/BooleanColumn"
+// scenario: "bool-col"
```
ivokub, line 21: Why change the identifier? It is not important though, but I think the scenario name should come from the prover-ray side and correspond to Scenario.name?

ivokub (commented):
Looks good. A few comments and after we update the prover-ray depenendency then should be good to merge.

verifier-ray/testdata/generate/main.go
YaoJGalteland: good idea, then the three-call block in writeCompiledFixtures is replaced with one WriteCompiledSystemZig call.

ivokub (approved):
Thanks!

verifier-ray/testdata/generated/vanishing.zig
arijitdutta67: Does this concern got missed somehow? The explicit scenario names seem to be useful.

arijitdutta67 (approved):
Looks good to me. Thanks for the work.

PR #3362, opened 2026-06-15
perf(prover): limitless prover performance optimization

gusiri:
# Results
- ~15% faster on average vs prod, on mainnet traces
- gains scale with conflation weight: lighter conflations see less, heavier conflations up to ~20%.
- peak memory ~400 GB (no changes vs prod)

# Changes

- Sub-prover concurrency = 5 (prove.go). GL/LPP segment proving is memory-bandwidth-bound; 5 concurrent jobs is the measured sweet spot (faster than 4 and than 6, with a lower peak RSS). 
- Per-module GL circuit cache (circuit_cache.go). Each module's compiled GL circuit is loaded once and shared across its segments instead of reloaded per segment.
- Outer-circuit setup loaded in background (prove.go). Overlaps the outer setup load with segment proving instead of blocking after conglomeration.
- FLATTEN base sizing (limitless.go). Size .*FLATTEN discovery bases to NextPow2(traceLimit * 8) so segmentation matches the flattened dimensions.
- Cleanup: removed perf_log benchmarking instrumentation.

# Note on GOMEMLIMIT config
Run with GOMEMLIMIT=300GiB.
The prover holds ~120 GiB of off-heap mmap (circuit chunks, arenas, witnesses) that GOMEMLIMIT does not count, so without a cap the heap grows until heap + off-heap exceeds the 512 GB box and it swap-thrashes. GOMEMLIMIT=300 keeps peak memory ~400 GB (≈100 GB margin).

# Test Environment
- r8a.16xlarge (64 cores, 512GB)
- mainnet traces
- execution-limitless prover

### Checklist

  PR.


AlexandreBelling (approved):
LGTM, thanks for the update

srinathln7 (approved):

PR #3394, opened 2026-06-17
feat(verifier-ray): add benchmarking framework for the verifier

arijitdutta67:
This PR implements issue(s) #3122

### Checklist

  PR.


verifier-ray/bench/verifier_profile/main.go
```diff
@@ -0,0 +1,492 @@
...
+	caseIndex int
+	mode      string
+	input     string
+	metadata  caseMetadata
+	stats     traceStats
+}
+
+func main() {
+	var (
+		casesFlag = flag.String("cases", "0", "case selector: all, N, A-B, or comma-separated selectors")
```
ivokub: Lets simplify calling and benchmark running - imo there is no use of running the benchmarks for non-profiled case, we want to get information how much different step takes cycles.

And we don't care about runtime of invalid case, this doesn't give a lot of information imo as the verifier can fail for different reasons, so the cycle count depends on which invalid case we choose.

And lets always run all cases. So we can remove `casesFlag`, `inputFlag` and `modeFlag`. And we can also remove all the parsing and validation and keep everything straightforward. This also means we can remove the corresponding columns from the output.

verifier-ray/bench/verifier_profile/main_test.go
```diff
@@ -0,0 +1,106 @@
...
+	if markerTranscript.cycle != 2 || markerTranscript.value != 5 {
+		t.Fatalf("marker: got %#v, want cycle=2 value=5", markerTranscript)
+	}
+	markerVanishing := stats.markers[markVanishingDone]
+	if markerVanishing.cycle != 4 || markerVanishing.value != 8 {
+		t.Fatalf("marker: got %#v, want cycle=4 value=8", markerVanishing)
+	}
+}
+
+func TestRenderCSV(t *testing.T) {
```
ivokub: I'm not sure we need the test.

verifier-ray/bench/verifier_profile/main_test.go
```diff
@@ -0,0 +1,106 @@
+package main
+
+import (
+	"encoding/csv"
+	"strings"
+	"testing"
+)
+
+func TestParseCaseSelector(t *testing.T) {
```
ivokub: I don't think we need this test as well when we remove different case handling and always handle all the cases.

verifier-ray/build.zig
```diff
@@ -1,7 +1,28 @@
...
+        "Enable comptime inclusion of profiling counters",
+    ) orelse false;
+    const r5_marks_arg = (b.option(bool, "r5-marks", "Enable R5 phase markers") orelse false) and r5; // only allow R5 marks when building for R5 target
+    // The `embedded-spec` option gives the index of the embedded specification from the fixtures. Later we can make this option more
+    // flexible to also allow specifying a path to a spec file, but for now we just embed the spec from the fixtures. This is only used
+    // for execution target, not for any test fixtures or library.
+    const embedded_spec = b.option(usize, "embedded-spec", "Embedded specification index") orelse 0;
+    // The `embedded-input` option is used to embed the input file into the binary to avoid needing to pass it in at runtime as we don't have
+    // input serialization yet. This is only used for execution target, not for any test fixtures or library.
+    const embedded_input = b.option(EmbeddedInputType, "embedded-input", "Embed the input file into the binary") orelse EmbeddedInputType.none;
```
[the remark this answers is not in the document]
ivokub: It is fine for now. Later on when we have serialization we'll start loading the input from memory.

verifier-ray/src/main.zig
```diff
@@ -119,21 +88,42 @@ extern fn open(path: [*:0]const u8, flags: c_int) c_int;
...
     const fd = open(native_input_path.ptr, o_rdonly);
     if (fd < 0) exitNative(1);
 
-    const mapped_addr = mmap(null, raw_input_len, prot_read, map_private, fd, 0);
+    const mapped_addr = mmap(null, @sizeOf(verifier.ProofData), prot_read, map_private, fd, 0);
     if (@intFromPtr(mapped_addr) == map_failed) exitNative(1);
 
     const mapped_bytes: [*]const u8 = @ptrCast(mapped_addr);
     return @ptrCast(@alignCast(mapped_bytes));
 }
```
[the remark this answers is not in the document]
ivokub: Lets keep as is for now.

verifier-ray/src/main.zig
```diff
@@ -119,21 +88,42 @@ extern fn open(path: [*:0]const u8, flags: c_int) c_int;
...
+        return &embedded_input;
+    }
+    // TODO: we have kept the compatibility with the old way of loading input, but we don't have serialization
+    // so it will fail if the input is not embedded.
+
+    // the input is linked into the binary at compile time using the
+    // `_in_start` symbol defined in the linker script, so we can just take its
+    // address and cast it to our structured input type
+    return @ptrCast(@alignCast(&_in_start));
+}
```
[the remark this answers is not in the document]
ivokub: Lets keep as is for now.

verifier-ray/src/main.zig
```diff
@@ -72,38 +54,25 @@ pub export fn r5_main() noreturn {
         unreachable;
     }
 
-    // the input is linked into the binary at compile time using the
-    // `_in_start` symbol defined in the linker script, so we can just take its
-    // address and cast it to our structured input type
-    const input: *const Input = @ptrCast(@alignCast(&_in_start));
+    // the input is linked into the binary ype
```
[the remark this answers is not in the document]
ivokub: Fixed.

verifier-ray/testdata/generate/main.go
```diff
@@ -942,6 +945,269 @@ func writeVanishingRuntimeTraceCell(out *bytes.Buffer, cell runtimeTraceCell, in
...
+	writeVerifyProof(out, fmt.Sprintf("verify_case_%d", idx), tc.honest)
+	if tc.invalid != nil {
+		writeVerifyProof(out, fmt.Sprintf("verify_case_%d_failing", idx), *tc.invalid)
+	}
+	fmt.Fprintf(out, "const verify_case_%d_systems = verifier.Systems{ .vanishing = system_%d };\n", idx, idx)
+	fmt.Fprintln(out)
+}
+
+func writeVerifyProof(out *bytes.Buffer, prefix string, proof vanishingProofView) {
+	fmt.Fprintf(out, "var %s_witness_claims = [_]ext.Ext{\n", prefix)
```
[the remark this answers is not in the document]
ivokub: Fixed.

verifier-ray/testdata/generate/main.go
```diff
@@ -942,6 +945,269 @@ func writeVanishingRuntimeTraceCell(out *bytes.Buffer, cell runtimeTraceCell, in
...
+	for _, claim := range proof.witnessClaims {
+		fmt.Fprintf(out, "    %s,\n", extValueLiteral(claim))
+	}
+	fmt.Fprintln(out, "};")
+	fmt.Fprintln(out)
+
+	fmt.Fprintf(out, "var %s_quotient_claims = [_]ext.Ext{\n", prefix)
```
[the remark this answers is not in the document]
ivokub: Fixed.

verifier-ray/testdata/generate/main.go
```diff
@@ -942,6 +945,269 @@ func writeVanishingRuntimeTraceCell(out *bytes.Buffer, cell runtimeTraceCell, in
...
+	for _, claim := range proof.quotientClaims {
+		fmt.Fprintf(out, "    %s,\n", extValueLiteral(claim))
+	}
+	fmt.Fprintln(out, "};")
+	fmt.Fprintln(out)
+
+	fmt.Fprintf(out, "var %s_module_sizes = [_]usize%s;\n", prefix, intArrayLiteral(proof.moduleSizes))
```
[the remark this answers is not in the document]
ivokub: Fixed.

verifier-ray/testdata/generate/main.go
```diff
@@ -942,6 +945,269 @@ func writeVanishingRuntimeTraceCell(out *bytes.Buffer, cell runtimeTraceCell, in
...
+	fmt.Fprintf(out, "    .module_sizes = &%s_module_sizes,\n", prefix)
+	fmt.Fprintln(out, "};")
+	fmt.Fprintln(out)
+}
+
+func writeVerifyRoundData(out *bytes.Buffer, prefix string, roundIdx int, round runtimeTraceRound) {
+	for columnIdx, column := range round.columns {
+		switch {
+		case column.publicBaseValues != nil:
+			fmt.Fprintf(out, "var %s_round_%d_column_%d_base = [_]field.Element{\n", prefix, roundIdx, columnIdx)
```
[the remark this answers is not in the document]
ivokub: Fixed.

verifier-ray/testdata/generate/main.go
```diff
@@ -942,6 +945,269 @@ func writeVanishingRuntimeTraceCell(out *bytes.Buffer, cell runtimeTraceCell, in
...
+			for _, value := range column.publicBaseValues {
+				fmt.Fprintf(out, "    %s,\n", fieldValueLiteral(value))
+			}
+			fmt.Fprintln(out, "};")
+			fmt.Fprintln(out)
+		case column.publicExtValues != nil:
+			fmt.Fprintf(out, "var %s_round_%d_column_%d_ext = [_]ext.Ext{\n", prefix, roundIdx, columnIdx)
```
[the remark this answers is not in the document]
ivokub: Fixed.

verifier-ray/testdata/generate/main.go
```diff
@@ -942,6 +945,269 @@ func writeVanishingRuntimeTraceCell(out *bytes.Buffer, cell runtimeTraceCell, in
...
+		case column.publicExtValues != nil:
+			fmt.Fprintf(out, "    .{ .public_column = .{ .ext = &%s_round_%d_column_%d_ext } },\n", prefix, roundIdx, columnIdx)
+		default:
+			panic("runtime trace column has no data variant")
+		}
+	}
+	fmt.Fprintln(out, "};")
+	fmt.Fprintln(out)
+
+	fmt.Fprintf(out, "var %s_round_%d_cells = [_]protocol.Scalar{\n", prefix, roundIdx)
```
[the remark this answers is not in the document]
ivokub: Fixed.

ivokub (commented):
Fixed some of the Copilot issues.

verifier-ray/Makefile
```diff
@@ -27,31 +43,36 @@ help:
...
+		'  profile-zkc-all       Profile every generated verifier case' \
 		'  clean                 Remove Zig build artifacts' \
 		'' \
 		'input variables:' \
-		'  INPUT_FILE            Binary input fixture for native and zkc runs, default testdata/inputs/passing.bin' \
-		'  NATIVE_INPUT_FILE     Fixed native input path; must match main.zig' \
-		'  IN_BYTES_FILE         Big-endian hex transport file for zkc' \
+		'  EMBEDDED_INPUT        Embedded input fixture kind: valid or invalid, default valid' \
+		'  EMBEDDED_SPEC         Embedded fixture index, default 0' \
+		'  PROFILE_CASES         Profiling selector: all, N, A-B, or comma-separated selectors, default 0' \
```
ivokub: When we remove the flags from bench executor, then can also remove from here.

verifier-ray/Makefile
```diff
@@ -27,31 +43,36 @@ help:
...
+		'  PROFILE_INPUT         Profiling input fixture kind: valid or invalid, default valid' \
+		'  PROFILE_MODE          Profiling mode: raw, profiled, or both, default profiled' \
+		'  PROFILE_OUT           Profiling CSV output path, default bench/verifier-profile.csv' \
```
ivokub: I think we don't need a flag to define output location. Lets just put it into fixed location which we document.

For me the issue with having so many different flags is that it confuses the end-user. It is easy for the agents to figure out, but we should target the implementation for humans.

verifier-ray/Makefile
```diff
@@ -63,34 +84,39 @@ fmt-codegen:
...
 
-fmt-testdata-generate:
+fmt-testdata-generator:
 	@if [[ -n "$$(gofmt -l testdata/generate)" ]]; then \
 		echo "please run gofmt"; \
 		gofmt -l testdata/generate; \
 		exit 1; \
 	fi
 
+fmt-profile:
```
ivokub: Lets move all Go fmt commands into a single target fmt-go

verifier-ray/Makefile
```diff
@@ -142,5 +173,15 @@ zkc-verify-failing-expected:
 		echo "zkc-verify-failing failed as expected"; \
 	fi
 
+profile-zkc:
```
ivokub: Lets keep one target which benchmarks all cases.

verifier-ray/Makefile
```diff
@@ -1,20 +1,36 @@
...
+# Legacy file-based input plumbing (unused for now, kept for reintroduction).
 PASSING_INPUT_FILE := testdata/inputs/passing.bin
 FAILING_INPUT_FILE := testdata/inputs/failing.bin
 INPUT_FILE ?= $(PASSING_INPUT_FILE)
 IN_ORIGIN ?= 0x08800000
 NATIVE_INPUT_FILE := zig-out/input.bin
 IN_BYTES_FILE ?= zig-out/input.hex
 
-ci: fmt build-release run-passing run-failing-expected test zkc-verify-passing zkc-verify-failing-expected
+ci: generate-testdata ci-checks
```
ivokub: Why changing the target? Lets keep as is

verifier-ray/bench/verifier_profile/main.go
```diff
@@ -0,0 +1,492 @@
...
+	"os/exec"
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+const (
+	defaultInput       = "valid"
+	defaultInputOrigin = "0x08800000"
```
ivokub: Lets not leak the values and lets try to keep the entry points to a single source of truth which I think right now can stay at Makefile. So instead of here calling directly `elf-to-json` and `zkc exec` etc we can just call the corresponding makefile targets. Then when we change something there the benchmark runner will inherit all the correct defaults.

verifier-ray/bench/verifier_profile/main.go
arijitdutta67: Removed. Thanks

verifier-ray/Makefile
arijitdutta67: The separation was intended to run the read-only checks after the test data generation. I tried to make it more explicit. Please let me know if this is ok.

verifier-ray/bench/verifier_profile/main_test.go
arijitdutta67: It was just for codepath verification. But we can remove it. Removed.

verifier-ray/testdata/generate/main.go
```diff
@@ -923,6 +988,285 @@ func writeTraceCell(out *bytes.Buffer, cell runtimeTraceCell, indent string) {
...
+	return expressionCount, bucketCount, vanishingCount
+}
+
+func fieldValueLiteral(value field.Element) string {
+	return fmt.Sprintf(".{ .value = %d }", u(value))
+}
+
+func extValueLiteral(value field.Ext) string {
+	a0, a1, b0, b1, c0, c1 := field.ExtToUint64s(&value)
+	return fmt.Sprintf(
```
ivokub, line 1243: We can use:
```go
	return fmt.Sprintf(
		"ext.Ext.fromUints(%d, %d, %d, %d, %d, %d)", a0, a1, b0, b1, c0, c1,
	)
```
instead. This makes the fixture way smaller (2x).

verifier-ray/Makefile
ivokub: And what does it solve? CI runtime or local checking runtime?

ci target is essentially meant for the user to check everything works before pushing a PR, it mimics the ci behaviour.

ivokub (dismissed):

YaoJGalteland (dismissed):

YaoJGalteland (approved):

PR #3405, opened 2026-06-18
test(e2e): disable partial-prover forced-tx test

gusiri:
Raise the event wait to 90m, the jest per-test timeout to 100m, and the CI step timeout to 110m

This PR implements issue(s) #

### Checklist

  PR.

### DIFF
--- a/.github/workflows/partial-prover-run-e2e-tests-in-parallel.yml
+++ b/.github/workflows/partial-prover-run-e2e-tests-in-parallel.yml
@@ -33,8 +33,6 @@ jobs:
         include:
           - test: local
             archive-name: local
-          - test: forced-transactions:local
-            archive-name: forced-transactions-local
 
     name: Run test:partial-prover:${{ matrix.test }} e2e tests
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/partial-prover-run-e2e-tests-in-parallel.yml
```diff
@@ -33,8 +33,6 @@ jobs:
         include:
           - test: local
             archive-name: local
-          - test: forced-transactions:local
-            archive-name: forced-transactions-local
```
jonesho, line 37: Why disabled it? is it just temporary for testing purpose on new prover image?

gusiri: Since we decided to disable FTX in prod, there's no need to keep running the FTX partial test.

The test is currently failing due to a coordinator-side bug, and because we don't plan to enable the FTX feature in prod, I don't expect us to fix the bug.

Given that, I think it's better to simply disable the test.

jonesho (approved):

PR #3415, opened 2026-06-22
feat(prover-ray): fri pcs api proposal

ThomasPiellard:
API proposal for the full FRI pcs.

There is one opinionated choice in my proposal -> I made the choice that the pcs is made for batch opening many polynomials of different sizes, living in different fields, on one SINGLE (possibly shifted) point. So for opening, we cannot give many random points for each polynomials, but only many shifts, associated to a shared random point. The reason is that for checking the AIR relations, the verifier needs the claimed values of the polynomials only at a single point, possibly shifted

### DIFF
--- a/prover-ray/crypto/koalabear/fri/pcs.go
+++ b/prover-ray/crypto/koalabear/fri/pcs.go
@@ -51,16 +51,16 @@
 // both produce identical OpeningProofs.
 //
 //   - One-shot:   pcs.Open(in OpenInputs) (OpeningProof, error)
-//                 pcs.Verify(in VerifyInputs, proof OpeningProof) error
+//     pcs.Verify(in VerifyInputs, proof OpeningProof) error
 //
 //     For callers that have all challenges + query positions ready
 //     up-front (precomputed transcript, tests, integration recipes).
 //
 //   - State-machine: NewOpenerState(...) ->
-//        ComputeClaimedValues(zeta) ->
-//        CommitDeepQuotient(alphaDEEP) ->
-//        Fold(alpha_j)        // numRounds times
-//        Open(queryPositions)
+//     ComputeClaimedValues(zeta) ->
+//     CommitDeepQuotient(alphaDEEP) ->
+//     Fold(alpha_j)        // numRounds times
+//     Open(queryPositions)
 //
 //     Mirrors [fri.ProverState]'s coin-fed pattern: caller binds each
 //     returned root to its transcript and derives the next challenge.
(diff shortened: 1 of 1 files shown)


prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -0,0 +1,614 @@
...
+// OpeningProof. Must be called after all folds are done.
+//
+// For each (query, batch), the method opens the batch's commitment
+// tree at the folded query position via PairedBranch, then packages
+// the raw conjugate-pair row data at every present size in the batch.
+func (os *OpenerState) Open(queryPositions []int) OpeningProof {
+	panic("TODO(pcs): OpenerState.Open")
+}
+
+// =============================================================================
```
AlexandreBelling, line 590: Move that to PR description

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -0,0 +1,614 @@
+// Package fri's PCS layer wraps the existing multi-degree FRI primitives
+// (Commit / RSEncoder / Tree / ProverState / Proof) into a batch
+// polynomial-commitment scheme with an Open/Verify surface.
+//
+// This file is a DESIGN PROPOSAL — types and signatures only. Bodies
+// panic with "TODO(pcs)". Review the layout, the OpeningProof shape,
+// and the open questions at the bottom of this doc before any
+// implementation lands.
+//
+// =============================================================================
```
AlexandreBelling, line 10: This should go either in the PR desc or in a separate markdown documentation.

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -0,0 +1,614 @@
+// Package fri's PCS layer wraps the existing multi-degree FRI primitives
+// (Commit / RSEncoder / Tree / ProverState / Proof) into a batch
+// polynomial-commitment scheme with an Open/Verify surface.
+//
+// This file is a DESIGN PROPOSAL — types and signatures only. Bodies
```
AlexandreBelling, line 5: This is not true since you will implement it afterward :)

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -0,0 +1,614 @@
...
+//   - (iii) empty shift list is an error (every committed row is
+//           opened at least once). OPEN QUESTION 1 below.
+//   - (iv)  duplicate shifts inside a row's shift list is an error.
+//   - (v)   no cross-batch dedup; caller is responsible.
+//   - (vi)  caller picks batch order. Convention: setup batches at
+//           the front, AIR-quotient batch at the back, witness rounds
+//           in between -- though the PCS itself doesn't care, only
+//           that prover and verifier agree on the order.
+//
+// =============================================================================
```
AlexandreBelling, line 102: Should go in the PR description

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -0,0 +1,614 @@
...
+// Open questions for review
+// =============================================================================
+//
+// 1. Empty shift lists. Loom rejects them; should this PCS too? An
```
AlexandreBelling: Yes, you can safely assume that if that happened, there would be a serious problem in the upstream wiop.

AlexandreBelling: And therefore panic is good defense

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -0,0 +1,614 @@
...
+//    empty shift list means "committed but not opened" -- the row's
+//    value is still authenticated by the Merkle path, but it doesn't
+//    contribute to the DEEP quotient. Allowing this is more flexible
+//    but adds a dead-code-detection failure mode (typos in the shift
+//    schedule become silent commitments). Default proposal: REJECT
+//    empty shift lists, matching loom.
+//
+// 2. Multi-size conjugate-pair openings. The underlying [Tree] is
```
AlexandreBelling: It is not obvious but when opening a branch, in the tree. The deepmost sibling is the conjugate of the opened value. This is a consequence of the codewords being stored in bit-reversed order.

AlexandreBelling: (it is what I implemented in the refactoring PR, that's why there is no "Pair" structure anymore).

One remark also, to construct the FRI trees, I directly map the fext elements into an octuplet without hashing. This saves hash cycles and also let you recover the original value.

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -0,0 +1,614 @@
...
+//          shape today; the multi-size tree's aux-leaf machinery is
+//          unused.
+//
+//    Default proposal: option (b) -- a new "PairedBranch" type that
+//    pairs the existing Branch with sibling-aux octuplets at every
+//    level, plus raw row data at both s and s^1 for the bottom and at
+//    both (s>>d) and (s>>d)^1 for each aux level. WMerkleOpening
+//    below reflects this. Worth re-validating on cost vs option (c).
+//
+// 3. State-machine vs one-shot API. Both proposed; concrete decision
```
AlexandreBelling: One comment with the one-shot is that without adding an explicit dependency over the FS API, it will be unsound. Still OK, to implement it in the test-side of the package.

The WIOP will use the state-machine approach.

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -0,0 +1,614 @@
...
+//    point for the team is: which one ships in the v1 implementation,
+//    or do both ship simultaneously? Default proposal: ship the state-
+//    machine variant (mirrors the existing fri.ProverState pattern);
+//    add the one-shot wrapper if a caller asks for it.
+//
+// 4. Where does the canonical-name -> (batchIdx, sizeLog2, rowIdx, isExt)
```
AlexandreBelling: This column naming thing must not exists in the crypto package as this is a matter for the wiop package.

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -0,0 +1,614 @@
...
+//    mapping live? Caller-side, per the precedent set by loom (the PCS
+//    doesn't know column names). Worth documenting an example caller
+//    that builds this mapping at compile time.
+//
+// 5. Encoders + Params relationship. [NewPCS] currently takes both. We
```
AlexandreBelling: Better to derive the encoder from the parameters. This removes completely the invariant that the caller always provides consistent values.

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -0,0 +1,614 @@
...
+}
+
+// =============================================================================
+// Batch / Shape / Shifts -- caller-facing types
+// =============================================================================
+
+// Batch is one batch of polynomials committed via a single [Commit]
+// call. It's an alias for MultiSizeTable so callers see the same
+// witness shape they already build for commitment.
+type Batch = MultiSizeTable
```
AlexandreBelling, line 202: Ok, but that's already a local type. We could rename the type [MultiSizeTable] into something more suitable if you see something better. Batch is too generic of a term: it won't be clear what that is to a reader.

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -0,0 +1,614 @@
...
+	Shift   int
+	Entries []deepEntry
+}
+
+type sizeBundle struct {
+	SizeLog2 int
+	Bundles  []shiftBundle
+}
+
+type layout []sizeBundle
```
AlexandreBelling: The layout is like an explanation of which of the claims are used to compute a DEEP quotient. But another question is how you map the polynomials to their respective location in the commitment Merkle Tree. I think both need a solution.

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -0,0 +1,614 @@
...
+type SizedShifts struct {
+	Base [][]int
+	Ext  [][]int
+}
+
+// =============================================================================
+// Layout -- internal canonical enumeration
+// =============================================================================
+//
+// Mirrors loom's canonicalLayout. Producer of the alpha_DEEP power
```
AlexandreBelling, line 240: Good to add in the PR description. But cleaner not to talk about Loom in the code itself.

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -0,0 +1,614 @@
...
+// OpeningProof bundles everything Verify needs to check that every
+// polynomial in every committed Batch evaluates to the listed values
+// at zeta and at the rotation shifts in BatchShifts.
+type OpeningProof struct {
+	// ClaimedValues[b] mirrors shifts[b] exactly. The outer protocol
+	// reads these to evaluate its constraints at zeta and to bind into
+	// the alpha_DEEP transcript challenge.
+	ClaimedValues []BatchClaimedValues
+
+	// DeepQuotientRoots: one Merkle root per distinct native size in
```
AlexandreBelling, line 291: Shouldn't that be in the FRI proof?

AlexandreBelling: Or do you commit to the DEEP quotient separately? From my perspective, we only need to do that for the first one. But after, they get fold in the FRI trees as a we fold the running codeword during FRI.

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -0,0 +1,614 @@
...
+	ConjugatePairsBySize []SizedConjugatePair
+
+	// Path is the extended multi-size Merkle branch. See PairedBranch.
+	Path PairedBranch
+}
+
+// SizedConjugatePair holds the conjugate-pair row values at one size
+// for one query: {row(omega^j), row(omega^(j ^ 1))} where j is the
+// folded query position at this size's tree level.
+type SizedConjugatePair struct {
```
AlexandreBelling, line 349: I think this structure overcomplexify things. With the bit-reversed order. You only need to open one position in the sized-Batch. Not even, the conjugate. And that's for the best because the "column" hashing work will likely dominate the verifier costs (still to check that over a realistic work).

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -0,0 +1,614 @@
...
+//	    fs.absorb(deepRoots OR running-layer root produced by fold j-1)
+//	    fs.sample(FoldAlphas[j])
+//	fs.absorb(final polynomial)
+//	for k in 0..NumQueries-1:
+//	    fs.sample(QueryPositions[k]) // mod N/2
+//
+// (The "absorb running-layer root" step at j=0 is actually
+// absorbing the DEEP roots, since the running polynomial AT round 0
+// is the largest DEEP codeword.)
+func (pcs *PCS) Open(in OpenInputs) (OpeningProof, error) {
```
AlexandreBelling: You could implement it in the test package if needed. In the main API, this is dangerous becuase it either draw an extra dependency or corresponds to an insecure implementation.

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -0,0 +1,614 @@
...
+	panic("TODO(pcs): Open")
+}
+
+// VerifyInputs bundles every parameter pcs.Verify needs. Mirrors
+// OpenInputs without the witnesses and with per-batch roots / shapes
+// in their place.
+type VerifyInputs struct {
+	Roots  []field.Octuplet
+	Shapes []Shape
```
AlexandreBelling, line 444: IMO, Shape could be considered integral part of the proof. 

I think it clean, that we already separate the commitment (Root) and the coins (Zeta, AlphaDeep, FoldAlphas, QueryPositions).

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -0,0 +1,614 @@
...
+
+// ComputeClaimedValues evaluates every (batch, size, row, shift)
+// declared in shifts at zeta * omega_N^shift and returns the per-
+// batch claimed values in the order the caller should bind them to
+// the transcript (canonical layout order).
+//
+// Returns a snapshot the caller can both bind and embed into the
+// final OpeningProof; OpenerState.Open will reuse the same snapshot
+// for the bridge construction.
+func (os *OpenerState) ComputeClaimedValues(zeta field.Ext) []BatchClaimedValues {
```
AlexandreBelling: It is a leak of concern. 

The WIOP does that already. I think you can assume that the claims are already given to you by the time you call NewOpener.

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -0,0 +1,614 @@
...
+//  2. Encodes each codeword to size rate*N via the matching encoder.
+//  3. Commits each codeword as a FRI level (a paired-leaf Merkle
+//     tree built via buildTreeExt).
+//  4. Returns the per-size roots in DESCENDING native-size order
+//     (largest first), ready for the caller to bind and to use as
+//     levelRoots in fri.Verify on the verifier side.
+//
+// The state machine transitions to the FRI commit phase: subsequent
+// calls to Fold expect the FRI fold challenges in order.
+func (os *OpenerState) CommitDeepQuotient(alphaDEEP field.Ext) []field.Octuplet {
```
AlexandreBelling: You should not need to commit to the DEEP quotient except for the first round. And I'd push strongly toward that because, more trees = more merkle proofs to open = more verifier time = less efficient recursion. Given that this is protocol level optimization, i'd prefer seeing it in the initial implementation (it's hard to change once people rely on it).

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -0,0 +1,614 @@
...
+//
+// Wraps the inner [fri.ProverState.Fold].
+func (os *OpenerState) Fold(alpha field.Ext) field.Octuplet {
+	panic("TODO(pcs): OpenerState.Fold")
+}
+
+// FinalPolyExt returns the FRI final polynomial after all folds are
+// done. Used by the caller to bind into the transcript before
+// deriving the query positions.
+func (os *OpenerState) FinalPolyExt() []field.Ext {
```
AlexandreBelling: Could be renamed FinalPoly

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -0,0 +1,614 @@
...
+// Verifier-side helpers
+// =============================================================================
+//
+// A verifier-side VerifierState mirroring OpenerState is an option but
+// is NOT proposed for v1. Reason: verification is more linear than
+// proving (no Merkle commitments to produce mid-stream, just checks),
+// so the one-shot pcs.Verify is sufficient. If a use case emerges for
+// a coin-fed verifier (e.g. incremental verification), add it then.
+
+// =============================================================================
```
AlexandreBelling, line 600: Can go in the PR documentation

AlexandreBelling (dismissed):

prover-ray/crypto/koalabear/fri/pcs.go
ThomasPiellard: the shape is part of the public data, the source of truth shouldn't come from the proof

gusiri (approved):

arijitdutta67 (approved):

PR #3416, opened 2026-06-22
fix(besu-plugins): Removed mutual exclusiveness of the txpool and blo…

Filter94:
…ck import validator plugins

This PR implements issue(s) #

### Checklist

  PR.

### DIFF
--- a/linea-besu/plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txvalidation/LineaBlockTransactionValidatorPlugin.java
+++ b/linea-besu/plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txvalidation/LineaBlockTransactionValidatorPlugin.java
@@ -36,10 +36,13 @@ import org.hyperledger.besu.plugin.services.TransactionValidatorService;
 public class LineaBlockTransactionValidatorPlugin extends AbstractLineaRequiredPlugin {
 
   /**
-   * Set once this plugin has been registered. Lets {@link
-   * net.consensys.linea.sequencer.txpoolvalidation.LineaTransactionPoolValidatorPlugin} detect that
-   * protocol-level transaction-type validation is active and skip its own redundant pool-level type
-   * validator, since this plugin's rule already runs at pool admission.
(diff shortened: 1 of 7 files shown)


linea-besu/plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txvalidation/LineaBlockTransactionValidatorPlugin.java
```diff
@@ -25,12 +24,26 @@
  * types (e.g. blob, delegate code) are accepted at the protocol level.
  *
  * <p>Note: Besu's {@link TransactionValidatorService} rules also run during transaction pool
- * admission (RPC/P2P). Pool-level type validation is explicitly handled by {@link
- * net.consensys.linea.sequencer.txpoolvalidation.LineaTransactionPoolValidatorPlugin}.
+ * admission (RPC/P2P), so the type checks here overlap with the pool-level type validation in
+ * {@link net.consensys.linea.sequencer.txpoolvalidation.LineaTransactionPoolValidatorPlugin}. The
+ * two plugins may be enabled together; to avoid validating transaction types twice at pool
+ * admission, the pool plugin skips its own type validator when this plugin is registered (detected
+ * via {@link #registered}). This plugin remains the source of type enforcement during block import
```
fluentcrafter, line 31: > to avoid validating transaction types twice at pool admission, the pool plugin skips its own type validator when this plugin is registered (detected via {@link #registered}).
 
If I understand correctly, this is a regression. At least in the RPC nodes, we want to validate all transaction types so that users, when calling `eth_sendRawTransaction`, get an error right away, instead of accepting and dropping a block building time only. 

Or am I misunderstanding the comment?

Filter94: The comment is saying that if there was no special treatment for the case when both of these plugins are enabled, one of the validation would run twice. But it's avoided by checking the registration status

fluentcrafter (approved):

PR #3325, opened 2026-06-09
feat(lineth-stack): add quickstart

Othryades:
## Summary
- Adds a guided Linea stack quickstart with Sepolia and local-L1 modes.
- Generates runtime accounts and Sepolia deployer material without requiring a raw private key in `.env`.
- Adds host-backed artifacts, service config rendering, Web3Signer/Postman signing setup, status/links/export tooling, traffic/smoke helpers, and optional ForcedTransactionGateway deployment.

## Validation
- shell syntax checks
- TypeScript helper tests
- quickstart static checks
- Sepolia Docker Compose config render
- local-L1 Docker Compose config render
- local-L1 fresh boot validation
- Sepolia fresh boot validation
- bridge smoke and traffic checks

## Notes
- `./scripts/start.sh --tail` remains the v0 user entrypoint.
- A wizard/CLI can wrap this flow later; it is intentionally not introduced in this PR.

---

> [!NOTE]
> **Medium Risk**
> Changes L1 rollup deploy fees and optional gateway deployment (nonce-order sensitive); the quickstart is large and dev-only but touches funding, signing, and contract wiring paths used at boot.
> 
> **Overview**
> Introduces **`docs/getting-started/lineth-stack/`** as a guided dev/demo stack: `./scripts/start.sh --tail`, Compose profiles for Sepolia L1 or optional local Besu+Teku L1, host-backed **`artifacts/`** (wallets, genesis, rendered configs, deployments), Web3Signer-backed coordinator/postman signing, status/links/export/smoke/traffic helpers, and operator docs (README, `.env.example`, `AGENTS.md`).
> 
> **Contract deploy changes** support that flow: new **`contracts/common/helpers/feeOverrides.ts`** (single legacy vs EIP-1559 fee model, `L1_DEPLOY_GAS_PRICE_WEI`, unit tests) wired into **`deployPlonkVerifierAndLineaRollupV8.ts`**; **`DEPLOY_FORCED_TRANSACTION_GATEWAY`** (default **true** for existing behavior, quickstart opts **false**) gates Mimc/ForcedTransactionGateway deploy and role grant; stricter **`L1_NONCE`** / boolean env parsing.
> 
> **.gitignore`** updates repo path ignores and gitignores quickstart runtime output under `lineth-stack/artifacts` and `lineth-output/`.
> 
[...]

PR #3434, opened 2026-06-24
refactor(contracts): centralize deploy helpers and validate nonce env vars

Othryades:
## Summary

This PR is the first of three small refactors suggested after review of the Linea stack quickstart work (#3325). It moves repeated deploy-script utilities into `contracts/common/helpers/` and updates local deploy scripts to use them.

**What moved into shared helpers**

- **`loadArtifactFromDirectory`** — loads a compiled contract JSON from a folder (replaces many copy-pasted `findContractArtifacts` functions).
- **`getDeployNonceFromEnv`** — reads `L1_NONCE` / `L2_NONCE` with strict validation (must be a non-negative integer string). Replaces `parseInt(process.env.*_NONCE)`, which could silently accept bad values like `"5x"` or `"1e3"`.
- **`getBooleanEnvVar` / `getBooleanEnvVarOrDefault`** — strict `"true"` / `"false"` parsing for deploy flags.
- **`LOCAL_L2_DEPLOY_FEE_OVERRIDES`** — shared constant for local L2 deploy fee settings (exported from the helpers barrel).

**What changed in deploy scripts**

All scripts under `contracts/local-deployments-artifacts/` (and a few Hardhat deploy tasks) now import these helpers instead of keeping local copies. Script-specific logic (which nonce offset to use, deployment order, etc.) stays at the call site.

**Nonce fallback fix (included in this PR)**

During review we found that the new helper briefly applied the `offset` argument even when the env var was **unset** and the code fell back to the wallet’s live nonce. That would skip a nonce compared to the old behaviour. The fix keeps offsets **only when the env var is explicitly set** — unset/empty still returns the wallet nonce unchanged.

## Why

- Less duplication across deploy scripts (easier to maintain, fewer drift bugs).
- Safer env parsing for nonces and booleans during local/quickstart deploys.
- Addresses feedback to refactor generic deploy mechanics out of individual scripts and reuse them elsewhere.

## Test plan

- [x] `ts-node contracts/common/helpers/deployments.test.ts`
- [x] `ts-node contracts/common/helpers/environment.test.ts`
- [x] `ts-node contracts/common/helpers/feeOverrides.test.ts`
- [ ] `pnpm -F contracts run lint:ts`
- [ ] `pnpm -F contracts run build`
- [ ] Optional: run a local deploy script that uses `L1_NONCE` / `L2_NONCE` overrides

## Related

- Follow-up to quickstart review context in #3325
- Companion PRs (separate branches): quickstart TS lib consolidation and shell helper consolidation

### DIFF
--- a/contracts/common/helpers/deployments.test.ts
+++ b/contracts/common/helpers/deployments.test.ts
@@ -1,0 +1,123 @@
+import { AbstractSigner } from "ethers";
+import fs from "fs";
+import * as assert from "node:assert/strict";
+import os from "os";
+import path from "path";
+
+import { getDeployNonceFromEnv, loadArtifactFromDirectory } from "./deployments";
+
+type TestCase = {
+  name: string;
+  run: () => Promise<void> | void;
+};
+
+const ENV_NAME = "LINETH_DEPLOY_NONCE_TEST_NONCE";
+
+function withEnv(value: string | undefined, run: () => Promise<void> | void): Promise<void> | void {
+  const previous = process.env[ENV_NAME];
+  if (value === undefined) {
+    delete process.env[ENV_NAME];
+  } else {
+    process.env[ENV_NAME] = value;
+  }
+  const restore = () => {
+    if (previous === undefined) {
+      delete process.env[ENV_NAME];
+    } else {
+      process.env[ENV_NAME] = previous;
+    }
+  };
+  try {
+    const result = run();
+    if (result instanceof Promise) {
+      return result.finally(restore);
+    }
+    restore();
+    return result;
+  } catch (error) {
+    restore();
+    throw error;
+  }
+}
+
+function mockWallet(liveNonce: number): AbstractSigner {
+  return {
+    getNonce: async () => liveNonce,
+  } as unknown as AbstractSigner;
+}
+
+const tests: TestCase[] = [
+  {
+    name: "getDeployNonceFromEnv uses the env value plus offset when set",
+    run: async () => {
+      await withEnv("10", async () => {
+        assert.equal(await getDeployNonceFromEnv(mockWallet(999), ENV_NAME), 10);
+        assert.equal(await getDeployNonceFromEnv(mockWallet(999), ENV_NAME, 5), 15);
+      });
+    },
+  },
+  {
+    name: "getDeployNonceFromEnv falls back to the live wallet nonce plus offset when unset or empty",
+    run: async () => {
+      await withEnv(undefined, async () => {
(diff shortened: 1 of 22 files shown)


VGau:
**Style nit**

`getBooleanEnvVarOrDefault` doesn't reuse `getOptionalEnvVar` as the review suggested — it reads `process.env` directly. This is safe (the important part: it avoids `getEnvVarOrDefault`), so it's a style nit only.

---

**Deferred #3325 findings (flag, not block)**

These blocking/high-priority items from the original review are not in this PR and, per the PR's own "fix order", belong to later steps — calling them out so they aren't lost:

- **L1 TokenBridge fee policy still empty.** In `deployBridgedTokenAndTokenBridgeV1_1.ts`, the `deployTokenBridgeOnL1` branch still leaves `fees = {}`, while the L2 branch now gets `LOCAL_L2_DEPLOY_FEE_OVERRIDES`. The original review asked to apply a pinned L1 gas policy (`L1_DEPLOY_GAS_PRICE_WEI`) here. Confirm this is intentionally deferred to the gas-policy step.
- **Address-env validation** (`validateAddressEnvVar` / `parseEnvAddressList`) in the TokenBridge fork is untouched — also appears deferred.

Othryades:
> **Style nit**
> 
> `getBooleanEnvVarOrDefault` doesn't reuse `getOptionalEnvVar` as the review suggested — it reads `process.env` directly. This is safe (the important part: it avoids `getEnvVarOrDefault`), so it's a style nit only.
> 
> **Deferred #3325 findings (flag, not block)**
> 
> These blocking/high-priority items from the original review are not in this PR and, per the PR's own "fix order", belong to later steps — calling them out so they aren't lost:
> 
> * **L1 TokenBridge fee policy still empty.** In `deployBridgedTokenAndTokenBridgeV1_1.ts`, the `deployTokenBridgeOnL1` branch still leaves `fees = {}`, while the L2 branch now gets `LOCAL_L2_DEPLOY_FEE_OVERRIDES`. The original review asked to apply a pinned L1 gas policy (`L1_DEPLOY_GAS_PRICE_WEI`) here. Confirm this is intentionally deferred to the gas-policy step.
> * **Address-env validation** (`validateAddressEnvVar` / `parseEnvAddressList`) in the TokenBridge fork is untouched — also appears deferred.

@VGau 
The direct process.env read is intentional/safe because it avoids getEnvVarOrDefault, but I can switch it to getOptionalEnvVar if you prefer.

VGau (approved):

PR #3436, opened 2026-06-24
refactor(lineth-stack): share quickstart TypeScript infra helpers in internal/lib (#3325 follow-up B)

Othryades:
## What this PR does (in plain language)

This is **PR B of 3** follow-up cleanups from Grant's review of the quickstart work (#3325). His ask was: stop copy-pasting helper functions across scripts and give them one shared home.

Before this PR, several small "infrastructure" helpers in the quickstart TypeScript scripts (`docs/getting-started/lineth-stack/scripts/internal/`) were **defined more than once** — sometimes byte-for-byte identical. That's risky: if someone fixes one copy, the others silently drift. The worst offender was the security redaction helper, which existed twice.

This PR creates a new shared folder `scripts/internal/lib/` and moves the generic mechanics there, so each helper is defined **exactly once** and imported everywhere it's needed. Domain scripts keep their domain logic; only the plumbing moved.

## New shared modules (`scripts/internal/lib/`)

| Module | What it holds | Why it matters |
|--------|---------------|----------------|
| `errors.ts` | `sanitizeExternalError` (redacts URLs + 32-byte hex from logs) and `sanitizeSecrets` (redacts a caller-supplied secret list) | **Security helper was duplicated** — now one source of truth so redaction can't drift |
| `fs.ts` | `ensureDir`, `writeFileAtomic` (temp file → rename → chmod) | Atomic-write pattern was reimplemented in 4 places |
| `env.ts` | `.env`/`process.env` parsing: `envValue`, `requiredEnvValue`, `requiredProcessEnv`, `envNumber`, `parseDecimalWei`, `parseBoolean`, `readDotEnvFile` | Env parsing was scattered; `sepolia-policy.ts` was wrongly acting as an env-helper "barrel" |
| `timing.ts` | `appendDeployTimingRecord` (one JSONL deploy-timing line) | The same JSONL writer + rounding lived in two files |

## Key behaviour preserved (no functional change)

- **File modes are unchanged on purpose.** Secret files (keystores) stay `0o600`; container-readable runtime files stay `0o644`. These were NOT unified — the distinction is intentional and is now passed explicitly.
- Exact error-message wording, regexes, and JSONL field order/rounding were copied verbatim from the originals.
- `sepolia-policy.ts` now owns **policy only** and no longer re-exports env helpers; all importers were updated to import from `lib/` directly.

## Tests

- New unit tests for all four modules: `lib/{errors,fs,env,timing}.test.ts` (16 cases) — **all passing locally** (Node 24.14.1).
  - errors: URL + 64-hex redaction; secret-list redaction incl. case variants
  - fs: atomic write creates file with requested mode, leaves no `.tmp`; `ensureDir` idempotent; `0o600` vs `0o644` both pinned
  - env: number/boolean/wei parsing valid+invalid paths; required-value throw paths
  - timing: one JSON line with expected fields + 3-decimal rounding, negative-duration clamp
- The static guard `check-quickstart-static.sh` was extended to **forbid re-introducing** the moved helpers outside `lib/` (no second `sanitizeExternalError` / `ensureDir` / `writeFileMode`; `sepolia-policy.ts` may not re-export env helpers).

## Scope

- Touches **only** `docs/getting-started/lineth-stack/scripts/internal/*.ts` (+ their tests) and the static-check script. No `contracts/` (that's PR A) and no shell (that's PR C).

## Merge order

Part of: **PR A → Wizard PR → PR B → PR C**. This PR (B) should land **after** the wizard PR merges, then be rebased. It is independent of PR A (#3434).

## Deferred (intentionally not here)

`lib/keystore.ts`, `lib/validation.ts`, `lib/format.ts`, `lib/logger.ts` — lower payoff / need more judgment. Left for a later cleanup once the `lib/` layout is proven.


docs/getting-started/lineth-stack/scripts/internal/lib/env.test.ts
```diff
@@ -0,0 +1,92 @@
+import * as assert from "node:assert/strict";
+
+import {
+  envNumber,
+  envValue,
+  parseBoolean,
+  parseDecimalWei,
+  readDotEnvContents,
+  readDotEnvFile,
```
thedarkjester, line 9: not sure I understand the usage / reason why you need the reading of the file or contents - surely we can access the data with `dotenv()`?

docs/getting-started/lineth-stack/scripts/internal/account-setup.ts
```diff
@@ -327,18 +311,18 @@ function writeRuntimeKeysEnv(wallets: Record<string, RuntimeWallet>) {
     ...runtimeRoles.map((role) => `${role.envName}='${wallets[role.envName].privateKey}'`),
     "",
   ];
-  writeFileMode(OUT_RUNTIME_KEYS_ENV, lines.join("\n"), CONTAINER_READABLE_FILE_MODE);
+  writeFileAtomic(OUT_RUNTIME_KEYS_ENV, lines.join("\n"), CONTAINER_READABLE_FILE_MODE);
   log(`Wrote ${OUT_RUNTIME_KEYS_ENV}`);
```
thedarkjester, line 315: what data is this logging?

docs/getting-started/lineth-stack/scripts/internal/account-setup.ts
```diff
@@ -327,18 +311,18 @@ function writeRuntimeKeysEnv(wallets: Record<string, RuntimeWallet>) {
...
 }
 
 function writeKeystorePasswordFile(password: string) {
-  writeFileMode(OUT_KEYSTORE_PASSWORD_FILE, `${password}\n`, CONTAINER_READABLE_FILE_MODE);
+  writeFileAtomic(OUT_KEYSTORE_PASSWORD_FILE, `${password}\n`, CONTAINER_READABLE_FILE_MODE);
   log(`Wrote ${OUT_KEYSTORE_PASSWORD_FILE}`);
```
thedarkjester, line 320: what data is this logging?

Othryades: only logs the file path string (OUT_RUNTIME_KEYS_ENV = artifacts/accounts/runtime-keys.env), not the contents

docs/getting-started/lineth-stack/scripts/internal/account-setup.ts
Othryades: it only logs the file path string

docs/getting-started/lineth-stack/scripts/internal/lib/env.test.ts
Othryades: This parser was only moved. It already lived in deployer-wallet.ts; this PR just relocates it into lib/env.ts to dedupe, so it isn't new code. Keeping the quickstart footprint minimal (no extra deps) was one of the quickstart project 'goal'.

Can update if you'd prefer.

thedarkjester: still trying to understand its usage - can you access the same data with the `dotenv()` functions?

Othryades: it reads a .env file from a specific path on disk and returns a plain JS object { KEY: "value", ... } so the script can work with it directly.
Yes, dotenv.parse(fs.readFileSync(path)) would do the same thing.
I kept the custom parser mostly because dotenv is not a dependency for these scripts today, so adding it felt like new scope.
The parser is tiny (~20 lines) and covered by tests, but happy to switch it to dotenv.parse() in a follow-up, or in this PR if you prefer.

thedarkjester: it is fine, more about building more code to maintain if we already have tools. Extracting and transforming might be the same, so we can leave it.

docs/getting-started/lineth-stack/scripts/internal/quickstart-preflight.ts
```diff
@@ -3,13 +3,12 @@ import * as path from "node:path";
 import { JsonRpcProvider } from "ethers";
 
 import { resolveL1DeployerConfig } from "./deployer-wallet";
+import { envValue, readDotEnvFile } from "./lib/env";
+import { sanitizeExternalError } from "./lib/errors";
 import {
   buildSepoliaPolicyConfig,
-  envValue,
   LOCAL_L1_CHAIN_ID,
```
VGau: This variable is not exported from the `./sepolia-policy` file. There is a typescript error. Could you fix it please ?

VGau (approved):

PR #3440, opened 2026-06-25
feat(prover-ray): PCS wrapper for FRI

Tabaie:
Implements the quotient PCS foundation over multi-degree FRI in `prover-ray`.

- Adds multi-tree FRI levels and query openings for virtual PCS levels.
- Introduces the PCS API surface, canonical layout, claim evaluation, and alpha_DEEP batching.
- Reconstructs virtual DEEP quotient levels on the prover side and wires them into the existing FRI prover state.
- Adds verifier-side reconstruction from authenticated row openings, including level-0 pair reconstruction and auxiliary level values.
- Completes PCS `Open` / `Verify` for the current standalone FRI package.
- Adds regression coverage for layout invariants, reconstruction arithmetic, multi-tree openings, honest Open/Verify flow, and negative/mutation cases.

The design from https://github.com/LFDT-Lineth/lineth-monorepo/pull/3415 is largely followed, with the main exception that this implementation does not commit to the DEEP quotient/codewords. This almost halves the number of hashes done by the prover at no cost to soundness. It also simplifies the state machine aspect of the protocol, in that the PCS layer does not add any rounds to the protocol: The main call to the PCS layer simply instantiates a FRI protocol which is run by the ambient.

Another point of divergence is that for a single polynomial, the quotient is built simply as $\sum_j \frac{f(X)-y_j}{X-z_j}$ with no $\alpha_{DEEP}$ batching coefficient (still used to batch together codewords for different polynomials.) Since the ambient protocol guarantees that query points are all distinct, this does not hurt soundness either.

The PR also anticipates the possibility that a FRI level might have incoming polynomials committed into distinct trees, which can come up in the ambient protocol if columns of the same size are introduced in different rounds.


prover-ray/crypto/koalabear/fri/pcs_test.go
```diff
@@ -0,0 +1,577 @@
...
+	)
+	if err != nil {
+		t.Fatalf("reconstruct aux query value: %v", err)
+	}
+	if want := started.levels[1].Evals[auxBase]; !aux.Equal(&want) {
+		t.Fatalf("aux reconstructed value mismatch\ngot:  %s\nwant: %s", aux.String(), want.String())
+	}
+}
+
+func encodeCanonicalTestPoly(poly []field.Ext, encoder *RSEncoder) []field.Ext {
```
ThomasPiellard: I think you can completely drop the canonical -> Lagrange conversion for the tests, you can directly create random polynomials which are interpreted in Lagrange. In the prover workflow the columns are directly interpreted in Lagrange basis

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -319,283 +525,793 @@ type SizedClaimedValues struct {
...
 
-// Open produces an OpeningProof in one call.
-//
-// The caller is responsible for deriving all challenges via Fiat-
-// Shamir from the appropriate prefix of the transcript. The
-// documented derivation order (which Verify expects mirrored on its
-// side) is:
+// Open computes the claimed evaluations at zeta. The caller must absorb the
+// returned values into the transcript before deriving alpha_DEEP.
+func (pcs *PCS) Open(
```
ThomasPiellard: Suggestion: Open should be renamed 'ComputeClaimedValues', it's what the function does. The method to actually compute the opening is 'Open' on *ProverState

ThomasPiellard (commented):
The API is a bit confusing. There should be essentially 3 methods exposed: `Commit` (commit to a batch = list of polynomials grouped by size), `Open` (=compute the proof + the claimed values) and `Verify` (yes/no, takes as input the proof + the commitments).
I think Commit, Open, Verify should be methods on a shared object, or static functions. But here those methods are scattered accross Pcs, ProverState, and Commit is static, it's a bit weird

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -319,283 +525,793 @@ type SizedClaimedValues struct {
...
+		return nil, err
+	}
+	claimed, err := pcs.claimedValues(witness, shifts, zeta)
+	if err != nil {
+		return nil, err
+	}
+	pcs.openings = append(pcs.openings, pcsOpening{
+		layout:    layout,
+		committed: committed,
+		claimed:   claimed,
```
ThomasPiellard, line 583: claimed's pointer is shared between pcs.openings and the return value. I think claimed should appear either once, in the proof only, or if really we want to separate it from the proof we should copy the slice

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -319,283 +525,793 @@ type SizedClaimedValues struct {
...
-//	for each (b, sizeLog2, rowIdx, isExt, shift) in canonical layout order:
-//	    fs.absorb(claimed value at this entry)
-//	fs.sample(AlphaDeep)
-//	for j in 0..numRounds-1:
-//	    fs.absorb(deepRoots OR running-layer root produced by fold j-1)
-//	    fs.sample(FoldAlphas[j])
-//	fs.absorb(final polynomial)
-//	for k in 0..NumQueries-1:
-//	    fs.sample(QueryPositions[k]) // mod N/2
+//	F(X) = Σ_i alpha_DEEP^i · Σ_j (f_i(X) - y_ij)/(X - z_ij)
```
ThomasPiellard, line 709: so, as discussed this is not 100% clear. You don't use alphaDeep for folding the same polynomial at various shifts of zeta, you just sum the terms ( e.g. you do (P-P(zeta))/(X-zeta) + (P-P(w^k zeta))/(X - w^k zeta) + ..  instead of (P-P(zeta))/(X-zeta) + alphaDeep*(P-P(w^k zeta))/(X - w^k zeta) + alphaDeep^2*.. . The security of this relies on SZ, assuming that the poles (zeta , w^k zeta , ..) are sampled at random over E6. But here there are not independent over E6: once zeta is fixed, the other poles are sampled in the set {zeta, w*zeta, w^2*zeta, ... } with w of order at most like 2^22 (=koala bear's two acidity). This set is much smaller than E6.
-> The question is: does that impact the security ? Do we need alphaDeep ?

Tabaie: The soundness argument is not statistical at all. Say the rational function $F(X):=\frac{f(X)-y}{X-z}+\frac{f(X)-y'}{X-z'}$ is a polynomial and hence can be evaluated at $z$. Then by formally clearing the denominators you get $(X-z)(X-z')F(X)=(X-z')\big(f(X)-y\big)+(X-z)\big(f(X)-y'\big)$. Evaluating at $z$ gives $(z-z')\big(f(z)-y\big)=0$. So to deduce $f(z)=y$ we only need $z\ne z'$ which the shifts already guarantee.

Tabaie: Basically there is no way for distinct poles to cancel each other out even if they are highly correlated.

prover-ray/crypto/koalabear/fri/pcs.go
Tabaie: I changed it to `AddOpening` so that it is not confused with the existing `Open` functions. Added some clearer documentation of what it does as well. Does that work?

prover-ray/crypto/koalabear/fri/pcs.go
Tabaie: The logic for this was that, obviously the caller needs the evaluations. But the PCS should also keep it to have a record of all openings when we want to compute the proof. The fact that they share the reference I guess can be a bit wonky, so I documented that the caller shouldn't modify the slice.

prover-ray/crypto/koalabear/fri/fri.go
```diff
@@ -102,29 +102,43 @@ type domainLight struct {
...
+	x.Exp(domain.generator, big.NewInt(int64(exponent)))
+	return x
+}
+
+func domainPointExt(domain domainLight, position int) field.Ext {
+	return field.Lift(domainPoint(domain, position))
+}
+
+// QueryLayer holds one Merkle branch per tree backing one folding level. The
+// running FRI layers use one branch; virtual PCS levels may use several.
```
AlexandreBelling, line 119: Why does the virtual PCS may use several branches?

prover-ray/crypto/koalabear/fri/fri.go
```diff
@@ -523,68 +578,93 @@ func openQueryExt(s int, layers [][]field.Ext, trees []*Tree, numRounds int) Que
...
 
-		q[j] = QueryLayer(path)
+		q[j] = QueryLayer{path}
 	}
 
 	return q
 }
 
-func checkQueryExt(s int, fq Query,
+type queryValueProvider interface {
```
AlexandreBelling: If I understand correctly, the interface is not meant to be implemented multiple time or in a separate package. Can you use static unexported helper function here?

prover-ray/crypto/koalabear/fri/fri.go
```diff
@@ -523,68 +578,93 @@ func openQueryExt(s int, layers [][]field.Ext, trees []*Tree, numRounds int) Que
...
+
+func (merkleQueryValues) checkRoundShape(_ int, opening QueryLayer, roots QueryLayerRoots, numLeaves int) error {
+	return checkQueryLayerShape(opening, roots, numLeaves, true)
+}
+
+func (merkleQueryValues) checkLevelShape(_ int, opening QueryLayer, roots QueryLayerRoots, numLeaves int) error {
+	return checkQueryLayerShape(opening, roots, numLeaves, true)
+}
+
+func (merkleQueryValues) queryPair(_ int, round, base int, opening QueryLayer, roots QueryLayerRoots) (
```
AlexandreBelling: What is the point of passing an `_ int` parameter?

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -1,11 +1,8 @@
...
+// polynomial-commitment scheme.
 //
-// This file is a DESIGN PROPOSAL — types and signatures only. Bodies
-// panic with "TODO(pcs)". Review the layout, the OpeningProof shape,
-// and the open questions at the bottom of this doc before any
-// implementation lands.
+// This file defines the PCS-facing types and the canonical column layout.
 //
 // =============================================================================
 // Design overview
```
AlexandreBelling, line 8: For a next PR, we should move that in a documentation Readme or just delete it.

prover-ray/crypto/koalabear/fri/pcs.go
AlexandreBelling: We don't need this function as the caller has already been evaluating the claims by himself as we call the PCS.

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -319,295 +532,768 @@ type SizedClaimedValues struct {
...
 	Roots  []field.Octuplet
 	Shapes []Shape
 	Shifts []BatchShifts
+	Zetas  []field.Ext
 
-	Zeta           field.Ext
-	AlphaDeep      field.Ext
-	FoldAlphas     []field.Ext
-	QueryPositions []int
+	Challenges Challenges
```
AlexandreBelling, line 1033: This just adds indirection in the code. It is not helpful abstraction

prover-ray/crypto/koalabear/fri/reedsolomon.go
```diff
@@ -65,12 +64,13 @@ func (enc *RSEncoder) Encode(p []field.Element, fftOpt ...fft.Option) []field.El
...
-	// coefficients directly in N-bit-reversed order and use a DIT FFT, avoiding
-	// the two explicit BitReverse passes previously needed for normal order.
+	// IFFT(DIF) consumes natural-order Lagrange input and returns n-bit-reversed
+	// coefficients. The n-to-N size change needs one normalization to natural
+	// coefficient order, after which FFT(DIF) emits the N-bit-reversed codeword
+	// directly.
 	enc.smallDomain.FFTInverse(_p[:n], fft.DIF, fftOpt...)
-	scatterBitReversedCoeffs(_p, n, int(N))
-	enc.Domain.FFT(_p, fft.DIT, fftOpt...)
+	gutils.BitReverse(_p[:n])
```
AlexandreBelling, line 72: But then, how is that an optimization? Your version does more bit reversing

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -75,103 +91,43 @@
 // For each native size N == 2^sizeLog2 in DESCENDING order, within each
 // size:
 //
-//	for shift s in ASCENDING order:
```
AlexandreBelling, line 78: It's really not useful to have this type of code paraphrasing. It will be unmaintained and outdated as we change the code.

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -183,13 +139,59 @@ import (
...
+	pcs.openings = pcs.openings[:0]
 }
 
 // NewPCS validates the encoder schedule against Params and returns a
-// ready-to-use PCS. Wraps [assertValidMultiEncoder] for the
-// cross-encoder consistency check.
-func NewPCS(params Params, encoders []*RSEncoder) (*PCS, error) { //nolint:revive // design stub
-	panic("TODO(pcs): NewPCS")
+// ready-to-use PCS.
+func NewPCS(params Params, encoders []*RSEncoder) (*PCS, error) {
```
AlexandreBelling, line 160: Panic on failure. Failures here cannot be reasonnably expected. If it happens, it's a compilation-time bug, it's simpler and more useful to just panic. 

How do you want to handle an error here anyway?

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -183,13 +139,59 @@ import (
...
+	if len(encoders) != params.numRounds+1 {
+		return nil, fmt.Errorf("fri: NewPCS: got %d encoders, want %d", len(encoders), params.numRounds+1)
+	}
+	if len(encoders) == 0 {
```
AlexandreBelling, line 164: Isn't that unreachable as per the above check?

prover-ray/crypto/koalabear/fri/fri.go
AlexandreBelling, line 20: It does not makes sense to specify N and D in that way since we use a multi-size PCS. There are many pairs (N, D)

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -183,13 +139,59 @@ import (
...
+		return nil, fmt.Errorf("fri: NewPCS: no encoders")
+	}
+
+	for i, encoder := range encoders {
```
AlexandreBelling, line 168: Shouldn't you check the size of the domain as well here?

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -183,13 +139,59 @@ import (
...
+		}
+	}
+
+	inverseRate := encoders[0].InverseRate()
+	wantInverseRate := params.N / params.D
+	if inverseRate != wantInverseRate {
+		return nil, fmt.Errorf("fri: NewPCS: inverse rate %d, want %d", inverseRate, wantInverseRate)
+	}
+
+	for i, encoder := range encoders {
```
AlexandreBelling, line 187: You could fuse that in the above loop

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -238,73 +239,285 @@ type SizedShifts struct {
 // =============================================================================
 //
 // Mirrors loom's canonicalLayout. Producer of the alpha_DEEP power
```
AlexandreBelling, line 241: Can we remove Loop as a reference in the documentation? The package reader won't have the context of what this is.

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -238,73 +239,285 @@ type SizedShifts struct {
...
+// Used by both AddOpening (with shapes derived from witnesses) and Verify
 // (with shapes passed in directly).
-func canonicalLayout(shapes []Shape, shifts []BatchShifts) (layout, error) { //nolint:revive,unused // design stub
-	panic("TODO(pcs): canonicalLayout")
+func canonicalLayout(shapes []Shape, shifts []BatchShifts) (layout, error) {
+	if len(shapes) != len(shifts) {
+		return nil, fmt.Errorf("fri: canonicalLayout: got %d shapes, %d shifts", len(shapes), len(shifts))
+	}
+
+	maxSizeLog2 := -1
```
AlexandreBelling, line 271: Aren't we supposed to check or assert that maxSizeLog2 is no longer set to -1 after the loop below?

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -319,295 +532,768 @@ type SizedClaimedValues struct {
...
-// can be added without breaking existing callers.
-type OpenInputs struct {
-	Witnesses []Batch
-	Committed []CommitterState
-	Shifts    []BatchShifts
-
-	Zeta           field.Ext
+
+// Challenges bundles the Fiat-Shamir values supplied by the caller.
+type Challenges struct {
```
AlexandreBelling, line 540: Can we leave this in the verifier inputs? And not try to do FS derivation in the current package as this is already the responsibility of the wiop package.

AlexandreBelling (approved):

PR #3445, opened 2026-06-25
feat(prover-ray): implement the FRI PCS compiler

AlexandreBelling:
This PR implements issue(s) #

### Checklist

  PR.


prover-ray/wiop/wiop_runtime.go
```diff
@@ -52,6 +52,13 @@ type Runtime struct {
 	// commitment are for committed columns (and not for FRI commitments). The
 	// precomputed columns commitment is stored in the [System].
 	Commitments map[int]field.Octuplet
+	// PCSOpeningProof carries the polynomial-commitment opening proof produced by
+	// the PCS compiler's opening action (a *fri.OpeningProof, stored as an opaque
+	// value so the core wiop package need not depend on the FRI package). It is
+	// nil until the PCS opening action runs; [System.Prove] copies it into the
+	// [Proof] and [System.Verify] restores it here before running verifier
+	// actions. Mirrors the transport role of [Runtime.Commitments].
+	PCSOpeningProof any
```
ivokub: It looks like we could define the correct type `*fri.OpeningProof` and it doesn't create any import cycles. I would prefer to have concrete strong typing for safety.

prover-ray/wiop/compilers/pcs/pcs.go
```diff
@@ -0,0 +1,644 @@
...
+		return
+	}
+
+	c := &compiled{}
+
+	// Commit the static precomputed round once, if it owns columns. A throwaway
+	// runtime exposes the (static) precomputed assignments; its encoders are a
+	// prefix of the static schedule so the root is stable across proof runs.
+	if len(sys.PrecomputedRound.Columns) > 0 {
+		st := commitToRound(friInverseRate, &sys.PrecomputedRound.Round, wiop.NewRuntime(sys))
```
ivokub, line 159: Why do we create runtime every time? It doesn't seem to be mutated?

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -552,47 +542,102 @@ func (pcs *PCS) Commit(witness MultiSizeTable) CommitterState {
...
+//
+// The FRI schedule is restricted to the largest size actually opened, so the
+// fold count follows the witness rather than the (possibly larger, static)
+// Params: a size-2^k top level folds k times regardless of Params.D >= 2^k.
 func (pcs *PCS) NewProverState(alphaDeep field.Ext) (*ProverState, error) {
 	if len(pcs.openings) == 0 {
 		return nil, fmt.Errorf("fri: NewProverState: AddOpening must be called first")
 	}
 
+	pcs, err := pcs.restrictToOpenings()
```
ivokub, line 636: Perhaps use another name for new variable `pcs`, it currently shadows the method target and is a bit confusing.

ivokub (commented):
Mostly looks fine, but it seems the global compiler which samples a separate evaluation point for every module, but FRI PCS assumes that the evaluation points are all same. So for example the test:
```go
func TestCompileGlobalQuotientMultiModuleUsesDifferentZetas(t *testing.T) {
	sys := wiop.NewSystemf("pcs-global-multi-module")
	r0 := sys.NewRound()
	mod2 := sys.NewSizedModule(sys.Context.Childf("mod2"), 2, wiop.PaddingDirectionNone)
	mod8 := sys.NewSizedModule(sys.Context.Childf("mod8"), 8, wiop.PaddingDirectionNone)
	col2 := mod2.NewColumn(sys.Context.Childf("col2"), wiop.VisibilityOracle, r0)
	col8 := mod8.NewColumn(sys.Context.Childf("col8"), wiop.VisibilityOracle, r0)

	mod2.NewVanishing(
		sys.Context.Childf("col2-is-3"),
		wiop.Sub(col2.View(), wiop.NewConstantVector(mod2, elem(3))),
	)
	mod8.NewVanishing(
		sys.Context.Childf("col8-is-5"),
		wiop.Sub(col8.View(), wiop.NewConstantVector(mod8, elem(5))),
	)

	global.Compile(sys)
	Compile(sys)

	_ = sys.Prove(func(rt *wiop.Runtime) {
		rt.AssignColumn(col2, baseVec(2, 3))
		rt.AssignColumn(col8, baseVec(8, 5))
	})
}
```
fails now (`pcs: every LagrangeEval must share the same evaluation point`). Is it expected and fixed in follow-up PR?

Tabaie:
> FRI PCS assumes that the evaluation points are all same

Oh that shouldn't be. Let me take a look.

Update: The one-zeta limitation has been introduced in this branch. Perhaps there is a reason I'm not aware of.

YaoJGalteland:
> it seems the global compiler which samples a separate evaluation point for every module

Yes. This seems to conflict with the FRI design, where we expect a single zeta across all modules.

In global.Compile:

```
for i, m := range sys.Modules {
    if len(m.Vanishings) == 0 {
        continue
    }
    mCtx := compCtx.Childf("m%d", i)
    compileModule(sys, m, mCtx, quotientRound, evalRound) // evalCoin := evalRound.NewCoinField(...)
}
```

Since evalCoin is created inside compileModule, each module gets its own evaluation coin. I think moving evalCoin out of the loop that calls compileModule could potentially solve the per-module eval point problem.

prover-ray/wiop/compilers/pcs/pcs.go
AlexandreBelling: It is only once at initialization for the precomputed column. It is a trick to be able to use the same function as for the actual runtime commits.

prover-ray/crypto/koalabear/fri/pcs.go
AlexandreBelling: On contrary, I think overshadowing is the right practice here. As this guarantees we may no longer access the receiver. I can add a comment to clarify.

AlexandreBelling:
I'll get to a unique coin in the global constraints as it just simplifies everything. It was like this in the old prover, I just assumed to would still be the same.

prover-ray/wiop/compilers/pcs/pcs.go
```diff
@@ -0,0 +1,640 @@
...
+func (a *openingProverAction) Run(rt *wiop.Runtime) {
+	proof := a.c.open(rt)
+	rt.PCSOpeningProof = &proof
+}
+
+// openingVerifierAction replays the opening transcript and checks the proof.
+type openingVerifierAction struct{ c *compiled }
+
+func (a *openingVerifierAction) Check(rt *wiop.Runtime) error {
+	return a.c.verify(rt, *rt.PCSOpeningProof)
```
ivokub, line 250: We don't check `rt.PCSOpeningProof` is non-nil before dereferencing.

prover-ray/wiop/compilers/pcs/pcs.go
ivokub: Oh indeed, I misread `if` to be a loop

prover-ray/crypto/koalabear/fri/pcs.go
ivokub: I think a comment would be useful - otherwise if later anyone does some changes not looking at this context it may cause unexpected behaviour.

ivokub (approved):
Thanks! It now looks good. We could add some cosmetic changes (explicit nil check and a comment), but no blocker in general.

YaoJGalteland (approved):
Thank you for the implementation, this is a substantial piece of cryptographic infrastructure and the core protocol is sound.

The absorbed roots pin every committed table before any challenge exists. Merkle authentication forces every query answer to come from those tables, so the prover cannot substitute values after seeing ζ, α_DEEP, or query positions. The verifier reconstructs F_N pointwise from the pinned row openings and the claimed y_ij. If every table is a valid codeword and every claim is correct, F_N is low-degree and all 229 fold checks pass. If any claim is wrong or any table is not a codeword, F_N is far from low-degree and the proximity gap theorem guarantees at least one query fails.

PR #3451, opened 2026-06-25
feat(riscv-guests): refactor the guests to use common riscv-guests build scripts

ivokub:
Applied for arithmetization tests and verifier-ray. Particularly, the arithmetization tests now use exactly the providers through riscv-guests so that we test against the actual implementations what the actual guests would use.

### Checklist

  PR.


riscv-guests/build_common/linker_script.ld
```diff
@@ -0,0 +1,44 @@
+ENTRY(_start)
```
lorenzogentile404, line 1: Do we want a copy of the default generated linker script stored in the repo? It may be convenient, but I'd like to understand if there are other reasons.

arithmetization/src/test/Makefile
```diff
@@ -3,7 +3,10 @@
...
 
 # Declare all targets as phony to prevent conflicts with files of the same name
-.PHONY: exec debug compile elf-exec elf-debug elf-to-json install-zkc zkc-exec zkc-debug clean clean-all linker-script verify-elf vector-build vector-json vector-exec keccak-rust-build keccak-rust-json keccak-rust-exec keccak-zig-build keccak-zig-json keccak-zig-exec blake-rust-build blake-rust-json blake-rust-exec act4-build act4-exec require-test require-src require-bin-ext require-vector-file require-n-vectors require-vector-build-artifacts require-vector-json-mode require-json
+.PHONY: exec debug compile elf-exec elf-debug install-zkc zkc-exec zkc-debug clean clean-all vector-build vector-json vector-exec keccak-rust-build keccak-rust-json keccak-rust-exec keccak-zig-build keccak-zig-json keccak-zig-exec blake-rust-build blake-rust-json blake-rust-exec act4-build act4-exec require-test require-src require-vector-file require-n-vectors require-vector-build-artifacts require-vector-json-mode require-json
+
+# set explicitly the default target so the include below (which has linker-script as first target) does not steal the default goal.
+.DEFAULT_GOAL := exec
```
lorenzogentile404, line 10: Is this a keyword in make to potentially define a default target that is not the first one? Nice.

arithmetization/src/test/Makefile
```diff
@@ -16,12 +19,23 @@
...
+#  * the ELF→ZKC-JSON conversion
+#     - ELF_TO_JSON
+#     - ELF_TO_JSON_GEN
+#     - elf-to-json
+#     - LINKER_SCRIPT
+#     - IN_BYTES
+#     - BIN_EXT
+#     - JSON_EXT
+include $(ARITHMETIZATION_DIR)/../riscv-guests/build_common/Makefile
+
```
lorenzogentile404, line 34: Does this copy the other Makefile inside this one, so we can treat those targets as non-external one? Nice.

arithmetization/src/test/Makefile
```diff
@@ -56,40 +66,9 @@ BIN  := $(MAKEFILE_DIR)$(if $(filter .s,$(EXT)),asm/bin/,$(if $(filter .zig,$(EX
 JSON := $(BIN).json
 # Note: in the future the JSON file will be removed and main.zkc will receive directly the ELF file
 
-# Hex big-endian input or `@path/to/in_bytes`
-IN_BYTES        ?= ""
-
-# If true, Zig tests using keccak_provide use the Linea keccak wrapper instead of Zesu stdlibs_accel
-KECCAK_ACCEL    ?= false
-
```
lorenzogentile404, line 64: Was this one removed as we always assume `KECCAK_ACCEL = true` from now on? I created a benchmark to compare `zkc keccak` with `riscv keccak` in the testing environmetn, so we probably want to keep that. See https://github.com/LFDT-Lineth/lineth-monorepo/pull/3409.

arithmetization/src/test/Makefile
```diff
@@ -117,7 +96,7 @@ compile: require-src linker-script
 		riscv64-unknown-elf-gcc -march=rv64im -mabi=lp64 -nostdlib -T$(LINKER_SCRIPT) -o $(BIN) $(SRC); \
 	elif [ "$(EXT)" = ".zig" ]; then \
 		echo "\nGenerating $(BIN) from $(SRC)"; \
-		zig build --build-file $(MAKEFILE_DIR)zig/build.zig -Dpath=$(BASENAME) $(if $(filter true,$(KECCAK_ACCEL)),-Dkeccak-accel=true); \
```
lorenzogentile404, line 120: Same comment as above.

arithmetization/src/test/Makefile
```diff
@@ -183,104 +153,8 @@ clean-all:
...
-		'    _input_start = _in_start;' \
-		'    _input_end = _in_end;' \
-		'' \
-		'    /* heap grows upward from this address by program convention */' \
-		'    _heap_start = $(HEAP_ORIGIN);' \
-		'}' \
-		> $(LINKER_SCRIPT)
-
-# Verify that the generated ELF matches the configured memory layout
-verify-elf: require-test
```
lorenzogentile404, line 240: Shall we keep this in the `build_common/Makefile` or here or get rid of it?

arithmetization/src/test/Makefile
```diff
@@ -425,11 +299,11 @@ keccak-rust-exec: keccak-rust-json
 keccak-zig-build:
 	$(MAKE) -f $(MAKEFILE_DIR)Makefile vector-build \
 		KECCAK_N_VECTORS=$(if $(filter -1,$(KECCAK_N_VECTORS)),$(shell awk 'END { print NR }' $(KECCAK_ALL_FILE)),$(KECCAK_N_VECTORS)) \
-		TEST=keccak/keccak_with_provider_batched_inputs.zig
+		TEST=keccak/keccak_batched_input_native.zig
 
 keccak-zig-json: keccak-zig-build
 	$(MAKE) -f $(MAKEFILE_DIR)Makefile vector-json \
-		TEST=keccak/keccak_with_provider_batched_inputs.zig \
+		TEST=keccak/keccak_batched_input_native.zig \
```
lorenzogentile404, line 309: the batched was meant to emphasize that we concatenate several keccak input in a single json given to the interpreter, that means we invoke keccak several times within the same program.

lorenzogentile404 (changes requested):
I'd just like to have some clarification, LGTM overall.

riscv-guests/build_common/linker_script.ld
ivokub: For me having the default linker script in the repo is for convenience - when we checkout repo and `zig build` in the guests then it just works. In case for testing if you would want to change it then you can just make linker-script with overridden parameters and you just have it locally.

arithmetization/src/test/Makefile
ivokub: Yes https://tool.oschina.net/uploads/apidocs/make-manual/Special-Variables.html.

Though I personally don't like relying on the default targets and rather would prefer to use explicit targets. In this case I just overrode the default target to keep the existing behaviour if it is used anywhere.

And I think it is also better to have `make help` target which outputs the defined targets etc. instead of having them defined in Makefile, see for example https://github.com/LFDT-Lineth/lineth-monorepo/blob/main/verifier-ray/Makefile#L29-L62

arithmetization/src/test/Makefile
ivokub: Yes https://web.mit.edu/gnu/doc/html/make_3.html#SEC15

arithmetization/src/test/Makefile
ivokub: No, we don't assume KECCAK_ACCEL = true. It is rather that keccak acceleration flag was defined in multiple places and imo it just creates confusion. Particularly that this test file is defined to be `_native.zig` i.e. it should always run with native keccak implementation, so forcing it through a build flag didn't make sense.

I refactored that the non-accelerated tests (keccak_native.zig and keccak_batched_input_native.zig) use the keccak implementation from l2-execution package which already imports the native (non-accelerated) keccak implementation from zesu.

So now the native keccak test imports l2-execution package https://github.com/LFDT-Lineth/lineth-monorepo/blob/chore/accelerator-refactor/arithmetization/src/test/zig/src/keccak/keccak_native.zig#L4-L14 and uses the resolved zkvm_keccak from there.

We have defined the import to be always using non-accelerated version https://github.com/LFDT-Lineth/lineth-monorepo/blob/chore/accelerator-refactor/arithmetization/src/test/zig/build.zig#L31-L33

And you can see that it falls back to zesu implementation https://github.com/LFDT-Lineth/lineth-monorepo/blob/chore/accelerator-refactor/riscv-guests/l2-execution/src/zkvm_provide.zig#L29-L31 and https://github.com/LFDT-Lineth/lineth-monorepo/blob/chore/accelerator-refactor/riscv-guests/l2-execution/src/zkvm_provide.zig#L65-L69

arithmetization/src/test/Makefile
ivokub: Explained above

arithmetization/src/test/Makefile
ivokub: I removed it from here and didn't add to build_common neither. Imo it was too explicit and added some inline bash parsing which is difficult to read and understand.

The main reason for removing the target is mainly because we just generate it with our own linker-script target and imo it is too verbose if we have another target to check that it is correctly generated. If the template is incorrect, then the linker fails anyway, so no point in adding hard-to-read tests.

arithmetization/src/test/Makefile
ivokub: Though we still have the `batched_input` in the file name indicating exactly the same? I only wanted to rename "with_provider" part which imo is ambiguous (is it accelerated or not). "native" suffic  - non-accelerated native implementation, "zkc" - accelerator zkc implementation.

riscv-guests/build_common/linker_script.ld
lorenzogentile404: Ok, let's keep it.

arithmetization/src/test/Makefile
lorenzogentile404: Yes, I created an issue for that. At the moment they are documented in `arithmetization/src/test/README.md.`

arithmetization/src/test/Makefile
lorenzogentile404: My bad, I missed it was still there.

lorenzogentile404: provider meant that you can select either the accelerated (zkc) one or the riscv (from zesu).

lorenzogentile404 (dismissed):
LGTM @ivokub. Could you update the branch and double check the job defined here still pass before merging?

riscv-guests/build_common/Makefile
```diff
@@ -0,0 +1,100 @@
...
+# STACK_ORIGIN
+# SP
+# PROGRAM_ORIGIN
+# IN_ORIGIN
+# HEAP_ORIGIN
+# in arithmetization/src/test/Makefile have default values but can be overriden.
+# Generate the linker script with the specified memory layout
+linker-script:
+	@echo "\nGenerating $(LINKER_SCRIPT)"
+	@printf '%s\n' \
```
Filter94, line 49: Is it possible to move it to a separate file and render it as a template? That would be more maintainable

riscv-guests/lineth-accelerators/include/zkvm_accelerators.h
Filter94, line 1: Can we instead fetch them from the [source](https://github.com/eth-act/zkvm-standards/blob/main/standards/c-interface-accelerators/zkvm_accelerators.h)? Same for Io

lorenzogentile404 (approved):
LGTM

arijitdutta67 (approved):

PR #3455, opened 2026-06-26
feat(verifier-ray): add Poseidon2 acceleration.

ivokub:
This PR implements issue(s) #

### Checklist

  PR.


.github/workflows/arithmetization-keccak-zkc-vs-reference-benchmark.yml
```diff
@@ -1,22 +1,6 @@
 name: Arithmetization zkc keccak vs riscv keccak benchmark
 
 on:
-  pull_request:
-    branches:
-      - main
```
amkCha, line 6: I think a rebase will fix the issue with the benchmark no ? Why do you delete the trigger on the PRs ?

.github/workflows/riscv-guests-zkc-interpreter-run.yml
```diff
@@ -103,5 +103,7 @@ jobs:
       # non-zero under these flags.
       # ZKC_REF= (empty) ensures the install-zkc prereq inside elf-exec is a guaranteed no-op
       # (zkc is already on PATH from the step above; an empty ZKC_REF triggers the skip path).
+      #
+      # TODO: gogen is disabled until it start supporting native arithmetic
       - name: Run l2-execution guest under zkc (gogen, fast, accel keccak)
-        run: make -C riscv-guests/l2-execution exec KECCAK_ACCEL=true ZKC_EXEC_FLAGS="--quiet --gogen --fast"
```
amkCha, line 107: Maybe we could open an issue to remind Gautam to put the flag back when native fields are fixed in gogen ?

arithmetization/src/main/lib/poseidon2/poseidon2.zkc
```diff
@@ -0,0 +1,179 @@
...
+        poseidon2_state[o + 0] = t01 + t01123
+        poseidon2_state[o + 2] = t23 + t01233
+    }
+}
+
+// mat_mul_internal applies the internal (partial-round) linear layer in place:
+//
+//     σ'[i] = (Σ_j σ[j]) + internal_diag[i] · σ[i]
+fn mat_mul_internal<poseidon2_state>() {
+    var sum:𝔽
```
amkCha: can be written var sum:F = 0

arithmetization/src/main/lib/poseidon2/poseidon2.zkc
```diff
@@ -0,0 +1,179 @@
...
+        sum = sum + poseidon2_state[i]
+    }
+
+    for i:u5 = 0; i<STATE_WIDTH; i = i + 1 {
+        poseidon2_state[i] = sum + (internal_diag[i] * poseidon2_state[i])
+    }
+}
+
+// Poseidon2 state: where the permutation takes place. An array of 16
+// KOALABEAR field elements (felts): [ σ₀, σ₁, ..., σ₁₅ ].
```
amkCha: Do you mind putting memories at the top of the file ? So we keep consistency with the other zkc structures

arithmetization/src/main/lib/poseidon2/poseidon2.zkc
```diff
@@ -0,0 +1,179 @@
...
+    for r:u5 = 0; r<FULL_ROUNDS_HALF; r = r + 1 {
+        full_round(r + FULL_ROUNDS_HALF + PRTL_ROUNDS)
+    }
+}
+
+// sbox computes the 3rd power of its input
+fn sbox(x:𝔽) -> (res:𝔽) {
+    res = x * x
+    res = res * x
+    return
```
amkCha: you don't need to add return, same for the sbox function.  Also, it wasn't possible to write res = (x * x ) * x ?

arithmetization/src/main/lib/poseidon2/poseidon2.zkc
```diff
@@ -0,0 +1,179 @@
...
+// Merkle–Damgård hashing, …) is left to the caller.
+//
+// Parameters:
+// * width 16
+// * S-box degree 3
+// * 6 full + 21 partial rounds.
+//
+// ===================================================================================
+
+fn write_to_state<poseidon2_state>(address:u5, value:𝔽) -> () {
```
amkCha, line 22: There is a feature #[inline] that you can put on top of 1-lines function and it inlines them when they are called to remove so exec steps. You can try it, I think it would have an impact. See PR where I'm going to use it https://github.com/LFDT-Lineth/lineth-monorepo/pull/3485

arithmetization/src/main/lib/poseidon2/poseidon2.zkc
```diff
@@ -0,0 +1,179 @@
...
+    poseidon2_state[address] = value
+}
+
+fn read_from_state<poseidon2_state>(address:u5) -> (value:𝔽) {
```
amkCha, line 26: Same here, I think we can inline

.github/workflows/riscv-guests-zkc-interpreter-run.yml
ivokub: Added comment https://github.com/LFDT-Lineth/zkc/issues/1892#issuecomment-4863839204

.github/workflows/arithmetization-keccak-zkc-vs-reference-benchmark.yml
ivokub: So, your PR https://github.com/LFDT-Lineth/lineth-monorepo/pull/3396 is already included and fixed the issue I had previously. However another PR https://github.com/LFDT-Lineth/lineth-monorepo/pull/3409 which also does gogen mode benchmarks reintroduced the same problem elsewhere. And as it has a flow CI workflow -> Go program -> Makefile, so if I would make a change in the Makefile then it would affect all users using it for calling `zkc` :)

So to unblock my Poseidon2 PR I'm currently not running the benchmark at every commit and PR, only when instructed to.

For me it makes regardless of the gogen issue as well. Right now we run keccak benchmarks during every commit and merge which touches `arithmetization/*`, and when we work on something else then it doesn't make sense to wait for that in CI (it is 4-5 minutes imo). I kept the `workflow_dispatch`, so it is still alway manually triggeable.

I guess alternative approach is as well to make the benchmark running criteria more specific -- run on changes in `arithmetization/src/main/keccak/*` only for example.

arithmetization/src/main/lib/poseidon2/poseidon2.zkc
ivokub: When I inline here I get panic:
```
panic: Width() called on native register "read_from_state_value$13"
```

I'm not sure if it is a bug or expected.

arithmetization/src/main/lib/poseidon2/poseidon2.zkc
ivokub: Added.

arithmetization/src/main/lib/poseidon2/poseidon2.zkc
ivokub: Changed:
* removed return to be implicit
* `(x * x) * x` inline

arithmetization/src/main/lib/poseidon2/poseidon2.zkc
ivokub: Fixed

arithmetization/src/main/lib/poseidon2/poseidon2.zkc
ivokub: Done

ivokub (commented):
Thanks! Addressed comments except for the second `#[inline]` which makes zkc exec panic.

arithmetization/src/main/riscv/utils/constants.zkc
```diff
@@ -157,6 +157,8 @@ const CUSTOM_1:Opcode = 0b0101011 // custom-1
 // Extraneous precompiles
 const FUNCT3_KECCAK:Funct3 = 0b000
 const FUNCT7_KECCAK:Funct7 = 0b0000000
+const FUNCT3_POSEIDON2:Funct3 = 0b001
+const FUNCT7_POSEIDON2:Funct7 = 0b0000000
```
lorenzogentile404, line 162: https://github.com/LFDT-Lineth/lineth-monorepo/blob/main/arithmetization/src/main/lib/README.md @OlivierBBB and I wrote a README to agree on the custom opcode we use for EVM precompiles and external programs + the value of funct3 and funct7.

riscv-guests/lineth-accelerators/src/poseidon2.zig
```diff
@@ -0,0 +1,33 @@
...
+pub fn lineth_zkvm_poseidon2_permutation(input: [*c]const zkvm_bytes_64, output: [*c]zkvm_bytes_64) callconv(.c) zkvm_status {
+    if (input == null or output == null) {
+        lineth_std.panic();
+    }
+
+    // invoke custom opcode for the Poseidon2 permutation. Kept in sync with arithmetization/src/main/riscv/utils/constants.zkc
+    // opcode format: opcode(0x2b = custom-1) | funct3(0b001) | funct7(0b0000000) | rd(output_offset) | rs1(input_offset) | rs2(unused)
+    asm volatile (
+        \\.insn r 0x2b, 0b001, 0b0000000, %[out], %[in], %[unused]
+        :
```
lorenzogentile404, line 24: See comment above about agreed https://github.com/LFDT-Lineth/lineth-monorepo/blob/main/arithmetization/src/main/lib/README.md.

arithmetization/src/test/zkc/poseidon2/Makefile
```diff
@@ -0,0 +1,41 @@
+# Runs the Poseidon2 permutation unit test.
```
lorenzogentile404, line 1: For complex logic we started using go scripts instead of Makefiles, as I believe it is more readable and easier to maintain.

lorenzogentile404: See `arithmetization/src/test/scripts/keccak_zkc_vs_reference_speedup/main.go`.

lorenzogentile404 (changes requested):
Left a few comments.

arithmetization/src/main/riscv/utils/constants.zkc
ivokub: Included in README

riscv-guests/lineth-accelerators/src/poseidon2.zig
ivokub: Added

arithmetization/src/test/zkc/poseidon2/Makefile
ivokub: Yes, for complex logic it makes sense. Though in this Makefile we don't do anything complex, it just reads the zkc input JSON line-by-line and runs zkc exec on the input. We don't do any output parsing or writing results etc.

Imo it is not worth adding Go file for it, locally scoped Makefile is way more descriptive and easier to understand.

ivokub (commented):
Addressed comments. Imo it is not worth adding Go script for a small loop which runs `zkc exec` for all inputs in the test.

arithmetization/src/test/zkc/poseidon2/Makefile
lorenzogentile404: `vector-exec` then you probably can use this target from `arithmetization/src/test/Makefile`.

lorenzogentile404 (commented):
See reply to comment.

lorenzogentile404 (approved):

verifier-ray/src/crypto/fiat_shamir.zig
```diff
@@ -20,7 +20,15 @@ pub const Transcript = struct {
 
     pub fn updateExt(self: *Transcript, values: field_value.ExtSlice) void {
         for (values) |ext_value| {
-            self.hasher.writeElements(&.{ ext_value.B0.a0, ext_value.B0.a1, ext_value.B1.a0, ext_value.B1.a1, ext_value.B2.a0, ext_value.B2.a1 });
+            // Absorb the six base limbs directly. Routing through writeElements
```
arijitdutta67, line 23: This is very interesting. Was curious about the diff of the improvement.

verifier-ray/src/crypto/poseidon2.zig
```diff
@@ -15,17 +17,35 @@ pub fn zeroDigest() Digest {
...
+// out of `compress` (no by-value arguments and return value). `state`
+// is read into the local permutation buffer before it is overwritten, so it is
+// fine for `state` to be the running hash; `right` must not alias `state`.
+pub fn compressInPlace(state: *Digest, right: *const Digest) void {
     profiling.poseidon2Compress();
-    var state: [16]field.Element = undefined;
-    @memcpy(state[0..block_size], &left);
-    @memcpy(state[block_size..], &right);
+    // `align(8)` matches `zkvm_bytes_64`, letting `permutationAccel16` reinterpret
+    // this buffer in place (see there). The array assignments lower to word stores,
```
arijitdutta67, line 30: "see there" -> better to give the riscv_guest location and explain a bit more. One who is not aware of risv_guest might get confused.

verifier-ray/src/crypto/poseidon2.zig
```diff
@@ -137,6 +175,21 @@ fn permutation(comptime width: usize, state: *[width]field.Element) void {
...
+// expects in `zkvm_bytes_64`. The bit casts are therefore plain reinterpretations.
+fn permutationAccel16(state: *[constants.width]field.Element) void {
+    // `[16]Element` is bit-identical to `zkvm_bytes_64` (16 LE u32 words = 64 bytes),
+    // and the accelerator permits `input`/`output` to alias, so reinterpret the state
+    // buffer in place. This avoids the two 64-byte copies (in/out staging) that the
+    // previous `@bitCast` round-trip lowered to byte-wise `memcpy` on R5.
+    // Safe because `compress` declares its state `align(8)`.
+    const buf: *lineth_accel.zkvm_bytes_64 = @ptrCast(@alignCast(state));
+    _ = lineth_accel.lineth_zkvm_poseidon2_permutation(buf, buf);
+}
```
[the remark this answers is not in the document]
arijitdutta67: Why not address this @ivokub? We can maybe panic with a error message also.

verifier-ray/src/crypto/poseidon2.zig
ivokub: It is the `permutationAccel16` in the same file, called through `permutation` which calls either permutationNative or permutationAccel16 depending if the accelerator is linked

verifier-ray/src/crypto/fiat_shamir.zig
ivokub: I don't have the R5 cycle diff at hand, but it was probably a few hundred cycles as we can avoid additional function calls (Zig compiler inlines the `writeElement` calls.

verifier-ray/src/crypto/poseidon2.zig
arijitdutta67: Ah I got confused

arijitdutta67 (approved):

PR #3465, opened 2026-06-29
fix(prover): filter accumulator-summary <-> state-summary inclusions

gusiri:
## Problem

On a conflation that fills the accumulator to 100% of its rows (8192), the state-manager fails the `LOOKUP_STATE_MGR_ACCUMULATOR_SUMMARY_TO_STATE_SUMMARY_REVERSED` inclusion.

partial prover error:
```
panic: The prover did not pass: Several errors were encountered:
failed LOOKUP_STATE_MGR_ACCUMULATOR_SUMMARY_TO_STATE_SUMMARY_REVERSED - row 3652 was not found in the including table : [STATE_SUMMARY_ACC_STATEMENT_INITIAL_ROOT_0=0 ... STATE_SUMMARY_ACC_STATEMENT_FINAL_HVAL_7=0 STATE_SUMMARY_ACCUMULATOR_STATEMENT_IS_READ_NON_ZERO=0 STATE_SUMMARY_ACCUMULATOR_STATEMENT_IS_READ_ZERO=0 STATE_SUMMARY_ACCUMULATOR_STATEMENT_IS_INSERT=0 STATE_SUMMARY_ACCUMULATOR_STATEMENT_IS_UPDATE=0 STATE_SUMMARY_ACCUMULATOR_STATEMENT_IS_DELETE=0]
... (rows 3652-8191, all all-zero)
```

## Root cause

`accumulatorsummary.ConnectToStateSummary` defines two plain inclusions between the accumulator-summary and state-summary tables, comparing the full columns including padding. The state summary is zero-padded to its size, and accumulator columns are zeroed at inactive rows, so the accumulator-summary contains an all-zero row only when it has at least one inactive row. When the accumulator is exactly full, it has none, so the reverse inclusion fails on the state summary's zero-padding rows.

## Fix

Filter both inclusions by each side's active flag using `InsertInclusionDoubleConditional`: `accumulator.Cols.IsActiveAccumulator` and `statesummary.Module.IsActive`. Both columns already exist and are constrained (`IsActiveAccumulator` is binary and zeroed at inactive rows with exactly one `Is*` set when active; `IsActive` gates the statement via `HKey = IsActive * keyHash`). Padding rows no longer participate, so a fully-packed accumulator matches the zero-padded state summary. This mirrors the already-filtered inclusions in `arithmetization_connector.go`. No new columns or assignments.

## Soundness

The check is not weakened. The filters are constrained columns tied to the real active regions, so a prover cannot zero a filter on a real row to exclude it. The bidirectional inclusion still asserts every active accumulator statement appears in the state summary and vice versa, which is the original intent of the reverse check.

## Testing

- `accumulator`, `accumulatorsummary`, `statesummary` and the rest of the state-manager suites pass.
- Full/partial prover on the failing conflation passes without `--large` (previously failed at this lookup).

## Note on setup

This changes the constraint system, so the prover assets (`zkevm.bin`, blueprints, verification keys) must be regenerated

## Conflation with this error

`30054884-30054999` (mainnet backtest) fills `SHOMEI_MERKLE_PROOFS` (8192) and triggers it. Builds on the bit-decomposition fix (#<bitdecompose PR>), which resolved the earlier recombination and Merkle-root failures on the same conflation.

### Checklist

  PR.

### DIFF
--- a/prover/protocol/dedicated/bits/bitdecompose.go
+++ b/prover/protocol/dedicated/bits/bitdecompose.go
@@ -79,43 +79,26 @@ func BitDecompose(comp *wizard.CompiledIOP, packed []ifaces.Column, numBits int)
 // Run implements the [wizard.ProverAction] interface and assigns the bits
 // columns
 func (bd *BitDecomposed) Run(run *wizard.ProverRuntime) {
-	bits := make([][]field.Element, len(bd.Bits))
 
-	// Obtain packed elements from
-	elements := make([][]field.Element, 0, len(bd.Packed))
+	size := bd.Packed[0].Size()
+
+	// Limbs are compacted independently, so combine them by absolute row over a
+	// shared range rather than by per-limb compact position.
+	assignments := make([]smartvectors.SmartVector, len(bd.Packed))
 	for i, packed := range bd.Packed {
-		// ind mirrors the mapping in BitDecompose's constraint loop:
-		// packed[len-1] (LSB limb) → IsPackedLimbNotZero[0], etc.
-		ind := len(bd.Packed) - 1 - i
-
-		v := packed.GetColAssignment(run)
-		var packedElements []field.Element
-		var packedElementsIsZero []field.Element
-
-		for colElement := range v.IterateCompact() {
-			packedElements = append(packedElements, colElement)
-
-			isPackedLimbNotZero := field.One()
-			if colElement.IsZero() {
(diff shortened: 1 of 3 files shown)


arijitdutta67 (dismissed):
LGTM

AlexandreBelling (dismissed):

arijitdutta67 (approved):

PR #3467, opened 2026-06-29
feat(prover-ray): adding public inputs

### LINKED ISSUE #3207: Add support for the public-inputs in the WIOP
Implement support for public-input in the wizard-IOP framework:

- [ ] Design the interface at the WIOP level. E.g. how do we want to declare and represent public-inputs in the WIOP.
- [ ] Design the interface with the arithmetization
- [ ] Implement the required constraining if any is needed
- [ ] Implement the required compilation if any is needed
- [ ] Testing at the wiop-level

Soleimani193:
This PR implements issue(s) #3207 
close #3207 

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> Breaking changes to core `Prove`/`Verify` affect all prover-ray callers; public-input binding is security-relevant but covered by new and updated tests.
> 
> **Overview**
> Adds a **public-input / statement** path to WIOP: registered cells are no longer part of the proof blob and must be supplied explicitly at verification time.
> 
> **`System.RegisterPublicInputs`** and **`PublicInputs`** let you declare statement cells (typically via **`ColumnPosition.Open`**). **`Prove`** now returns **`(Proof, PublicInput)`**; registered cells are omitted from **`proof.Cells`** and their values are returned in **`PublicInput`** in registration order. **`Verify(proof, pub)`** assigns those values from **`pub`**, checks length against the registry, and errors if a public-input cell appears in the proof or if any registered input was not consumed.
> 
> Call sites across compiler tests, pipeline tests, PCS tests, and **`zkcdriver`** are updated to the new signatures. **`public_input_test.go`** covers honest/tampered statements, dynamic modules, and duplicate registration panics.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f02dd9b17484b4364e225b55411c3690ac0f9235. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


prover-ray/wiop/wiop_prove_verify.go
```diff
@@ -172,8 +240,13 @@ func (sys *System) Verify(proof Proof) error {
 	}
 
 	// This checks that all the items of the proof have been used. Meaning all
-	// cells and all columns of the proof are read.
+	// cells and all columns of the proof are read. It also enforces the
+	// no-overlap invariant: a proof item must not be a registered public input.
```
AlexandreBelling, line 256: I am actually writing PR to remove support for Public column (I won't need them, so we can simplify the code).

prover-ray/wiop/wiop_system.go
```diff
@@ -34,6 +38,12 @@ type System struct {
 	// via [System.NewMessageBusSend] and [System.NewMessageBusReceive], in
 	// declaration order.
 	MessageBuses []*MessageBus
+	// PublicInputs maps a public-input name to the [ObjectID] of the [Cell] or
+	// [Column] that stores its value. It is populated via
```
AlexandreBelling: Can we set that, only Cell may be public-inputs? It is to simplify on the framework side

prover-ray/wiop/wiop_system.go
```diff
@@ -69,10 +79,51 @@ func NewSystemf(msg string, args ...any) *System {
...
+//
+// Panics if name is already registered, if id does not identify a column or
+// cell, or if id is a non-public column.
+func (sys *System) RegisterPublicInput(name string, id ObjectID) {
+	if _, ok := sys.PublicInputs[name]; ok {
+		panic(fmt.Sprintf("wiop: RegisterPublicInput: public input %q already registered", name))
+	}
+
+	switch id.Kind() {
+	case KindColumn:
```
AlexandreBelling: Can we remove that case?

prover-ray/wiop/wiop_system.go
Soleimani193: Yes, we can, but from arithmetization we obtain public modules, i.e., a set of public columns. If we want only cells, we have to open all the positions in those columns to cells. Does not seem very practical? What do you think?

prover-ray/wiop/wiop_prove_verify.go
Soleimani193: so two questions:
-   can I assume that there is no public column in the proof? 
-  for cells we have to avoid overlapping between proof and public inputs as cells are any way public. right?

AlexandreBelling:
Can you fix the DCO check also azam? You will need to rebase the branch using the --signoff option.

Also, can you fix the linting issues?

Soleimani193:
> Can you fix the DCO check also azam? You will need to rebase the branch using the --signoff option.
> 
> Also, can you fix the linting issues?

Sure, just wanted to confirm the proposal first, then complete the PR.

AlexandreBelling:
I think that after my comments are adressed it will be LGTM. 

I have prepped a PR #3476 where I do what I mentionned about. 

Let's prioritize merging #3441 and #3445 first.

prover-ray/wiop/wiop_prove_verify.go
AlexandreBelling: Yes, no public columns in the proof as well. For the scope of your PR you can just treat the "no columns in PI".

prover-ray/wiop/wiop_system.go
AlexandreBelling: That's what I understood to be the plan. The arithmetization expose public column  as it is convenient for them and we only actually expose cells to the verifier. The good news is that we already have a way to construct opening very easily now. It automatically reduce to a local constraint.

AlexandreBelling (dismissed):

ThomasPiellard:
@Soleimani193 why does `Prove` return public inputs ? Normally it's the other way around, `Prove` should take as argument a statement (= modules description + public inputs) + a private witness, and return only a proof. If the public inputs that you are talking about in your PR are local openings (= certain cells of a private witness like the last entry of a logup column), then those values should be part of the proof (even if the mechanism for the verifier is the same)

prover-ray/wiop/wiop_system.go
```diff
@@ -75,10 +84,40 @@ func NewSystemf(msg string, args ...any) *System {
...
 
+// RegisterPublicInput records that cell stores the public input named name.
+// Public inputs are always cells: a public column from the arithmetization is
+// exposed by opening the desired position into a cell (see
+// [ColumnPosition.Open]), and that cell is registered here. The named cell
+// becomes part of the protocol statement — its value is carried in a
+// [PublicInput] separately from the [Proof].
+//
+// Panics if cell is nil or if name is already registered.
+func (sys *System) RegisterPublicInput(name string, cell *Cell) {
```
ivokub: Should we also check that we wouldn't have two distinct `cell.Context.ID` registered under different names?

prover-ray/wiop/wiop_system.go
```diff
@@ -75,10 +84,40 @@ func NewSystemf(msg string, args ...any) *System {
...
+	if cell == nil {
```
ivokub: Couldn't we possibly have a large number of public inputs? I.e. in case of R5 execution the public input would be the transaction data. In this case what would be the public input be, some hash of the data or the data itself?

Perhaps we could register a slice of cells as public inputs?

prover-ray/wiop/wiop_system.go
```diff
@@ -75,10 +84,40 @@ func NewSystemf(msg string, args ...any) *System {
 	// Wire the back-reference after the System pointer is stable.
 	sys.PrecomputedRound.system = sys
 	sys.Annotations = make(Annotations)
+	sys.PublicInputs = make(map[string]ObjectID)
 
 	return sys
 }
 
+// RegisterPublicInput records that cell stores the public input named name.
```
ivokub: And also, a Cell already has a "Annotation" field, why do we need to have it named twice? I would just have `RegisterPublicInputs(cells []*Cell)`

prover-ray/wiop/wiop_prove_verify.go
```diff
@@ -100,7 +119,15 @@ func (sys *System) Prove(assign func(rt *Runtime)) Proof {
 		proof.DynamicSizes[k] = v
 	}
 
-	return proof
+	// Capture the registered public-input cells into a separate PublicInput.
+	// GetCellValue resolves lazily-assigned openings, so a cell opened from a
+	// public column resolves to that column's value at prove time.
+	pub := make(PublicInput, len(sys.PublicInputs))
+	for _, id := range sys.PublicInputs {
```
ivokub: Maps in go are not deterministic, i.e. when we loop over `sys.PublicInputs` then we get the keys/values in random order. Similarly, because the returned PublicInput is a map as well, then its order is as well not deterministic.

prover-ray/wiop/wiop_system.go
```diff
@@ -75,10 +84,40 @@ func NewSystemf(msg string, args ...any) *System {
...
+// [ColumnPosition.Open]), and that cell is registered here. The named cell
+// becomes part of the protocol statement — its value is carried in a
+// [PublicInput] separately from the [Proof].
+//
+// Panics if cell is nil or if name is already registered.
+func (sys *System) RegisterPublicInput(name string, cell *Cell) {
+	if cell == nil {
+		panic("wiop: RegisterPublicInput requires a non-nil Cell")
+	}
+	if _, ok := sys.PublicInputs[name]; ok {
```
YaoJGalteland: is it possible to register the same cell twice with different names? like:

RegisterPublicInput("a", cell)
RegisterPublicInput("b", cell)

ivokub (changes requested):
I think the idea is good, but I would restructure it differently. Right now the public inputs what the verifier would consume are quite coupled to the system and prover, but imo normally the verifier should be able to construct the public inputs itself, receiving from the prover only the proof.

But the current PublicInput structure is a `map[ObjectID]field.Gen` which means that the verifier would have to know the internal indexing of the cells of the system. Imo this is particularly problematic for the verifier-ray where we don't follow the same indexing etc.

Another issue imo right now is that the public input and the system is not in detereministic order (because we use Go maps internally).

Rather, I would simplify -- RegisterPublicInputs wouldn't take a "name string" as input because imo we wouldn't need it (if we want to, we can already store some descriptive name in Cell.Annotations). And we should be able to register many cells at a time.

Now, internally we should just keep the public input registry in the system as `[]*Cell`, so we have a deterministic order and it depends only on the order of registration.

Now, on the prover side, we don't need to provide public inputs as inputs, they are already present when `assign` function runs. And it also shouldn't output the public inputs. If we want to simplify tests, then perhaps we could add a separate method to System which allows to extract the flattened public inputs we can feed into verifier (but the assumption is, that in production the verifier would construct them themselves).

Now, imo for public inputs it makes sense if we have it as `[]field.Gen` instead (i.e. flattened list of cell values). And the verifier takes this as input.

prover-ray/wiop/wiop_system.go
```diff
@@ -75,10 +84,40 @@ func NewSystemf(msg string, args ...any) *System {
...
+		panic(fmt.Sprintf("wiop: RegisterPublicInput: public input %q already registered", name))
+	}
+	sys.PublicInputs[name] = cell.Context.ID
```
YaoJGalteland: there seems to be an ID collision problem.

The ID only encodes `(kind, slot, position)`. There is no system pointer or global counter in it.

For example:

```go
sys1 := wiop.NewSystemf("sys1")
r1 := sys1.NewRound()
cell1 := r1.NewCell(ctx1, false) // ID = {kind=Cell, slot=0, pos=0}

sys2 := wiop.NewSystemf("sys2")
r2 := sys2.NewRound()
cell2 := r2.NewCell(ctx2, false) // ID = {kind=Cell, slot=0, pos=0} — same
```

Then this is accepted silently:

```go
sys1.RegisterPublicInput("foo", cell2)
```

Since `cell2`’s ID happens to equal `cell1`’s ID, `piSet` is built from `sys1.PublicInputs` as `{slot=0, pos=0}`. That means `piSet` also contains `cell1`’s ID.

So in the prove loop:

```go
if _, ok := piSet[cell.Context.ID]; ok {
    continue // cell1 is skipped here because its ID matches piSet
}
proof.Cells[cell.Context.ID] = rt.GetCellValue(cell)
```

`cell1` gets skipped from `proof.Cells`, and its value is never captured into the proof. So `cell1` silently disappears from `proof.Cells`.

Maybe we should add an ownership check:

```go
if cell.round.system != sys {
    panic("wiop: RegisterPublicInput: cell does not belong to this system")
}
```

YaoJGalteland:
right now everything runs in one process, so the ownership of `PublicInput` doesn’t matter much.

But I agree that public inputs should probably be owned by the external caller / verifier, rather than being produced as part of the prover output.

If the prover outputs `PublicInput`, a buggy or malicious prover could provide a value that doesn’t correspond to the actual computation it ran, and the verifier wouldn’t catch that. The proof only establishes internal consistency; it doesn’t prove that `PublicInput` matches the real-world input the verifier actually cares about.

prover-ray/wiop/wiop_system.go
AlexandreBelling: From my pov, the name is only here for decoration purpose as the cells already bears an identifier. That's fine for me, but we should certainly non-duplicate check

Soleimani193: We would use Contex.lable as the human-readable name, so no collision

prover-ray/wiop/wiop_system.go
Soleimani193: This is the matter of "garbage in, garbage out" sys is PIOP and we have a single PIOP per compile

Soleimani193:
> Now, imo for public inputs it makes sen

We prefer to keep PI in the proof () output as it is more efficient to go over the rounds once and collect Proof and PI simultaneously. The Verifier can ignore it and use its  reconstructed PI (as long as it is an explicit input to the verifier)

ivokub (dismissed):

ivokub (approved):

AlexandreBelling (approved):

Issue #3418, opened 2026-06-22
Issue: Benchmark and Optimize field-op RISC-V cycle cost


YaoJGalteland:
### Background

Each `Poseidon2.compress()` call costs ~32,000 R5 cycles. A single permutation has:

- ~465 `mul` calls per permutation
- ~900+ `add`/`sub` calls per permutation
- `matMulInternalInPlace` is called **21× per permutation**, making it the dominant hot path

Optimization so far has been based on **static analysis only** — instruction counting from Zig/LLVM output, not measured cycle costs under `zkc`. This is unreliable: a micro-benchmark shows `mul` (8 cycles) is cheaper than `add` (12 cycles) in `zkc`, which inverts the usual assumption and means some current spellings are suboptimal.

---

### Observed issues

| Location | Current spelling | Issue |
|---|---|---|
| various | `halve()` | Compiles to ~5 instructions (ANDI, branch, ADD, SRLI, jump); `mul(inv2Exp1)` may be ~2 instructions (MUL + REMU) |
| `mul2ExpNegN` | n× `halve()` loop | n sequential halves instead of a single `mul(inv2ExpN)` |
| `double()` | delegates to `add(self, self)` | Likely fine but unverified against `mul` cost |
| `add()` | promotes to u64 before compare | On RV64 may cost 1–2 extra instructions; unverified |
| `sub()`, `neg()` | data-dependent branches | Branch cost under `zkc` not measured |

We already replaced `halve()` in `matMulInternalInPlace` but the function still exists and may be called elsewhere.

---

#### 1. Benchmark individual field ops under `zkc`

Write a minimal R5 fixture that calls each field operation N times in a tight loop and counts cycles via the marker mechanism. Record cycles-per-call for:

`add`, `sub`, `neg`, `double`, `mul`, `square`, `halve`, `pow`

This gives ground truth instead of LLVM-output estimates.

#### 2. Identify cheaper equivalents

From the benchmark data, for each operation in `matMulInternalInPlace`, `matMulExternalInPlace`, and `matMulM4InPlace`, check whether a cheaper spelling exists:

| Current | Candidate | Cheaper if |
|---|---|---|
| `halve()` | `mul(inv2Exp1)` | `mul` < 5 cycles |
| `x.mul(field.Element.init(3))` | `x.add(x).add(x)` | 2× `add` < `mul` + REMU |
| `x.double().double()` | stays as-is | already 2× `add` |
[...]

PR #3471, opened 2026-06-29
perf(verifier-ray): optimize KoalaBear field operations

YaoJGalteland:
This PR implements issue(s) #3418

## Summary

This PR adds an R5 micro-benchmark for KoalaBear field operations and optimizes several verifier field hot paths.

The benchmark builds a small guest program, runs it through `zkc`, and reports per-operation cycle counts for base-field arithmetic. The current benchmark reports **fast interpreter cycles only**. AIR/proof-cost measurement is still blocked until register splitting is available for the traced path.

## Base field (F_p)

| op               | main   | field-opt | change          |
|------------------|--------|-----------|-----------------|
| add              | 8.02   | 2.01      | 4.0× faster ✅  |
| sub              | 5.01   | 3.01      | 1.7× faster ✅  |
| neg              | 4.00   | 3.00      | 1.3× faster ✅  |
| double           | 5.02   | 2.01      | 2.5× faster ✅  |
| mul              | 4.01   | 4.01      | —               |
| square           | 4.01   | 4.01      | —               |
| `pow(x, 2^20)` | 154.03 | 154.03 | — |
| powComptime(2^20)| 42.02  | 42.02     | —               |
| inverse          | 277.00 | 107.00    | 2.6× faster ✅  |
| div              | 281.01 | 111.01    | 2.5× faster ✅  |

## Extension field (F_{p^6})

| op            | main    | field-opt | change          |
|---------------|---------|-----------|-----------------|
| ext/add       | 99.03   | 12.03     | 8.2× faster ✅  |
| ext/sub       | 96.03   | 18.03     | 5.3× faster ✅  |
| ext/mul       | 626.03  | 199.03    | 3.1× faster ✅  |
| ext/square    | 626.03  | 118.03    | 5.3× faster ✅  |
| ext/inverse   | 1020.03 | 333.03    | 3.1× faster ✅  |
| ext/div       | 1634.03 | 520.03    | 3.1× faster ✅  |
| ext/mulByBase | 18.02   | 18.02     | —               |

The benchmark also shows a large gap between runtime exponentiation and compile-time exponentiation. For exponents known at compile time, `powComptime` uses about **3.8× fewer R5 cycles** than `pow`. As a follow-up/codegen rule, verifier code should prefer `powComptime` whenever the exponent is a compile-time value. This avoids paying the runtime exponentiation overhead for constant exponents.


verifier-ray/bench/field_op_bench/main.zig
```diff
@@ -0,0 +1,208 @@
...
+//   90 = start div,      91 = end div
+
+const kb = @import("koalabear");
+const Element = kb.Element;
+
+const N: u64 = 1000;
+
+// ── write syscall ─────────────────────────────────────────────────────────────
+
+fn writeR5(bytes: []const u8) void {
```
ivokub: Perhaps we can use the methods from `src/profiling.zig`?

verifier-ray/bench/field_op_bench/main.zig
```diff
@@ -0,0 +1,208 @@
...
+    // stays nonzero, so repeated division never produces or divides by zero.
+    acc = a;
+    emitMark(90, 0);
+    i = 0;
+    while (i < N) : (i += 1) {
+        acc = acc.div(b);
+    }
+    emitMark(91, acc.value);
+
+    asm volatile (
```
ivokub: We can also use the `zkvm_exit` call. See https://github.com/LFDT-Lineth/lineth-monorepo/pull/3455/changes#diff-6fffc46722ba1a084d11a31227e661250d430c232d176e1493d8335df11ad65aR147.

This means we would need to add dependency on the lineth-accelerators packages for this bench package.

verifier-ray/src/field/koalabear.zig
```diff
@@ -63,16 +63,11 @@ pub const Element = extern struct {
...
 
     pub fn sub(self: Element, rhs: Element) Element {
-        if (self.value >= rhs.value) {
-            return .{ .value = self.value - rhs.value };
-        }
-        return .{ .value = modulus - (rhs.value - self.value) };
+        return .{ .value = (self.value + modulus - rhs.value) % modulus };
     }
 
     pub fn neg(self: Element) Element {
```
ivokub, line 73: It seems for neg it is rather more efficient to:
```zig
pub fn neg(self: Element) Element {
    return .{ .value = (modulus - self.value) % modulus };
}
```
to avoid the if check

verifier-ray/bench/field_op_bench/main.zig
```diff
@@ -0,0 +1,208 @@
...
+
+    // div — b is nonzero (so div's internal inverse is well defined) and acc
+    // stays nonzero, so repeated division never produces or divides by zero.
+    acc = a;
+    emitMark(90, 0);
+    i = 0;
+    while (i < N) : (i += 1) {
+        acc = acc.div(b);
+    }
+    emitMark(91, acc.value);
```
ivokub: Dunno if it would be interesting to also add Ext benchmarks as well?

ivokub (dismissed):
Thanks! Looks good, but see the comments as well.

verifier-ray/src/field/koalabear.zig
YaoJGalteland: good idea! it reduced to 3

verifier-ray/bench/field_op_bench/main.zig
YaoJGalteland: `ext/mul` costs ~ 500 cycles, and `ext/inverse` is ~ 700 cycles.

Let me break them down a bit more to see if there is room to improve.

YaoJGalteland: we can expand the operations and defer the % p fold in add and sub until just before the value is used as input to the next (base) multiply.

ivokub (dismissed):
Thanks! Looks good

arijitdutta67 (dismissed):
LGTM

ivokub (approved):

arijitdutta67 (approved):

PR #3450, opened 2026-06-25
fix(sdk-viem): bound eth_getLogs queries to avoid provider block-range limits issue

VGau:
Rate-limited RPC providers can reject eth_getLogs requests spanning more than 10,000 blocks. Several sdk-viem actions issued unbounded earliest..latest log queries that would fail on such providers.

- getMessageProof: try the full-range L2MessagingBlockAnchored query first and, on a range-limit rejection, fall back to a binary search over currentL2BlockNumber() to locate a <=10k window; chunk the L2 MessageSent range query into <=10k windows.
- getL2ToL1MessageStatus: determine anchoring via the currentL2BlockNumber() view instead of scanning L2MessagingBlockAnchored logs (O(1) eth_call).
- getMessageByMessageHash / getTransactionReceiptByMessageHash: add optional messageBlockNumber to pin the lookup to a single block.
- claimOnL1: add optional messageL2BlockNumber forwarded to getMessageProof.

Robust provider range-limit detection via viem's typed errors and narrow pre-deployment zero-data handling. Full unit-test coverage added.

This PR implements issue(s) #

### Checklist

* [ ] I wrote new tests for my new core changes.
* [ ] I have successfully ran tests, style checker and build against my new changes locally.
* [ ] If this change is deployed to any environment (including Devnet), E2E test coverage exists or is included in this
  PR.
* [ ] I have informed the team of any breaking changes if there are any.

---

> [!NOTE]
> **Medium Risk**
> Changes L2→L1 status and proof resolution logic and RPC behavior; incorrect anchoring or chunk ordering could misreport claimability or build wrong Merkle proofs, though behavior is heavily tested.
> 
> **Overview**
> **sdk-viem** no longer relies on unbounded `earliest`..`latest` `eth_getLogs` for bridge lookups, which many RPCs reject beyond ~10k blocks.
> 
> **`getMessageProof`** tries a full-range `L2MessagingBlockAnchored` query first; on any failure it binary-searches `currentL2BlockNumber()` to find a ≤10k L1 window, then fetches L2 `MessageSent` events in ascending ≤10k chunks for Merkle proof construction. Pre-deployment `eth_call` zero-data is treated as “not finalized”; other RPC errors propagate.
> 
> **`getL2ToL1MessageStatus`** replaces an L1 log scan with comparing the message’s L2 block to `currentL2BlockNumber()` plus `isMessageClaimed`.
> 
[...]

Issue #3435, opened 2026-06-24
Coordinator - adapt usage of eth_getLogs to max of 10K blocks


fluentcrafter:
# Description

Infura is limiting `eth_getLogs` to a maximum of 10K blocks, including indexed logs with 1 matching result. Coordinators is making some queries with `eth_getLogs(startBlock=0, endBlock=LATEST)`  which will stop working. 

Potential areas:
- Message anchoring L1 -> L2
- Message relaying L2 -> L1
- Forced transactions

We shall make sure that from now on, `eth_getLogs` use bounded queries.

PR #3473, opened 2026-06-30
fix(coordinator): bound eth_getLogs queries to provider block-range limits

gauravahuja:
## Description

#3435

Rate-limited RPC providers (e.g. Infura) reject `eth_getLogs` requests spanning more than 10,000 blocks — **including indexed single-result queries**. The coordinator issued several unbounded `earliest..latest` log queries that would stop working on such providers.

This ports the approach of the sdk-viem fix (#3450) to the coordinator.

## Changes

Replaced the four unbounded point lookups with bounded binary searches via `EthLogsSearcher.findLog`. Each drops the indexed-number topic (an exact-match filter gives the binary search no directional signal) and instead compares the decoded message/ftx number in the search predicate — valid because these numbers increase monotonically with block height:

- `MessageAnchoringService.getRollingHash` — `L1RollingHashUpdated` (message anchoring L1→L2)
- `L1MessageSentEventsFetcher.findL1RollingHashUpdatedEvent` — `L1RollingHashUpdated` (message anchoring L1→L2)
- `ForcedTransactionsL1EventsFetcher.getStartBlockNumberToListenForLogs` — `ForcedTransactionAdded` (forced txs)
- `InvalidityProofAssembler.getPrevFtxRollingHash` — `ForcedTransactionAdded` (forced txs)

Added a tunable config field **`ethLogsSearchMaxBlockRange`** (default `10_000`, TOML key `eth-logs-search-max-block-range`), surfaced to each app as `l1EventSearchMaxBlockRange` and used as the binary-search chunk size. It is kept **separate** from the rolling-forward `ethLogsSearchBlockChunkSize`: a point lookup wants the widest legal block span (provider cap) per probe, whereas the rolling-forward scan uses a small chunk to bound result *count*. Reusing the small chunk made the binary search degrade to a near-linear walk on sparse events.

Already-bounded queries were left unchanged: `getLogsRollingForward` (chunked) and `ZkProofCreationCoordinatorImpl.getBridgeLogs` (single-block).

## Testing

- `coordinator:ethereum:message-anchoring` and `coordinator:ethereum:forced-transactions` test suites pass, including the sparse "spread apart" resilience test (which configures a tiny `ethLogsSearchBlockChunkSize` and now relies on `ethLogsSearchMaxBlockRange` to converge), the restart/resume path, and the `BadNonce` invalidity-proof path exercising `getPrevFtxRollingHash`.
- `coordinator:app` config v2 parsing tests updated and passing (new key added to full-config TOMLs + reference TOMLs).
- Spotless clean on all changed modules.

---

> [!NOTE]
> **Medium Risk**
> Changes how the coordinator resolves L1 events for message anchoring and forced transactions; mis-tuned range or search logic could miss events or delay processing, but behavior is configurable and covered by existing module tests.
> 
> **Overview**
> Replaces **unbounded** `earliest..latest` L1 log point lookups (which break on providers like Infura that cap `eth_getLogs` spans at 10k blocks) with **bounded binary search** via `EthLogsSearcher.findLog`, using event topic only and comparing decoded message/FTX numbers in the search predicate instead of indexed-number topics.
> 
> Adds tunable **`eth-logs-search-max-block-range`** (default `10_000`, wired as `l1EventSearchMaxBlockRange`) for message anchoring and forced-transactions L1 scraping—**separate** from the smaller rolling-forward chunk size so sparse-event searches stay efficient.
> 
> **Message anchoring:** rolling-hash resolution in `MessageAnchoringService` and `L1MessageSentEventsFetcher` uses the new search; anchoring also **caches the last found block** to narrow follow-up lookups.
> 
> **Forced transactions:** resume-block discovery in `ForcedTransactionsL1EventsFetcher` and prev rolling-hash in `InvalidityProofAssembler` use the same pattern; rolling hash is **read from the DB when available**, with bounded on-chain search only after prune.
> 
> Config/TOML, docker reference configs, and parsing tests are updated; new unit tests cover `InvalidityProofAssembler.getPrevFtxRollingHash`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ab9787b3174a0b20f1b3b67a8710ce1c5c943c91. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


coordinator/app/src/test/kotlin/linea/coordinator/config/v2/ForcedTransactionsConfigParsingTest.kt
```diff
@@ -42,6 +42,7 @@ class ForcedTransactionsConfigParsingTest {
       polling-timeout = "PT10S"
       eth-logs-search-success-backoff-delay = "PT0.5S"
       eth-logs-search-block-chunk-size = 500
+      eth-logs-search-max-block-range = 5000
```
fluentcrafter, line 45: What's the difference between:
- A) eth-logs-search-block-chunk-size
- B) eth-logs-search-max-block-range

?

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/ForcedTransactionsL1EventsFetcher.kt
```diff
@@ -51,20 +54,26 @@ internal class ForcedTransactionsL1EventsFetcher(
...
+            val foundFtxNumber = ForcedTransactionAddedEvent.fromEthLog(ethLog).event.forcedTransactionNumber
+            when {
+              foundFtxNumber < finalizedForcedTransactionNumber -> SearchDirection.FORWARD
+              foundFtxNumber > finalizedForcedTransactionNumber -> SearchDirection.BACKWARD
+              else -> null
+            }
+          }
+            .thenApply { eventLog ->
+              if (eventLog == null) {
+                throw IllegalStateException(
```
fluentcrafter, line 73: This is a super efficient approach. Once we make the first binary search, the subsequent searches can safely be from that block (inclusive) onwards

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/conflation/InvalidityProofAssembler.kt
```diff
@@ -88,22 +88,25 @@ class InvalidityProofAssembler(
...
+        topics = listOf(ForcedTransactionAddedEvent.topic),
+      ) { ethLog ->
+        val foundFtxNumber = ForcedTransactionAddedEvent.fromEthLog(ethLog).event.forcedTransactionNumber
+        when {
+          foundFtxNumber < prevFtxNumber -> SearchDirection.FORWARD
+          foundFtxNumber > prevFtxNumber -> SearchDirection.BACKWARD
+          else -> null
+        }
+      }.thenApply { ethLog ->
+        if (ethLog == null) {
```
fluentcrafter, line 122: After these new eth_getLogs constraints, I believe that we should extend the ForcedTransactionRecord with an extra field: `l1BlockNumber: ULong`

This would allow for significantly shortening the searches. Additionally, we can Inject the FTX DAO here and first check the DB; if not found, we make a blockchain query.

coordinator/ethereum/message-anchoring/src/main/kotlin/linea/anchoring/MessageAnchoringService.kt
```diff
@@ -80,25 +81,26 @@ class MessageAnchoringService(
...
-          L1RollingHashUpdatedEvent.topic,
-          messageNumber.toHexStringUInt256(),
-        ),
-      )
-      .thenApply { rawLogs ->
-        val events = rawLogs.map(L1RollingHashUpdatedEvent::fromEthLog)
-        require(events.size == 1) {
-          "Expected exactly 1 event RollingHashUpdated(messageNumber=$messageNumber) " +
-            "but got ${events.size} events. events=$events"
+        topics = listOf(L1RollingHashUpdatedEvent.topic),
```
fluentcrafter: Likewise here, we should optimise the search given that both move monotonically.

coordinator/ethereum/message-anchoring/src/main/kotlin/linea/anchoring/clients/L1MessageSentEventsFetcher.kt
```diff
@@ -88,16 +89,21 @@ internal class L1MessageSentEventsFetcher(
...
+      chunkSize = l1EventSearchMaxBlockRange.toInt(),
       address = l1SmartContractAddress,
-      topics = listOf(
-        L1RollingHashUpdatedEvent.topic,
-        messageNumber.toHexStringUInt256(),
-      ),
-    ).thenApply {
-      it.firstOrNull()?.let { log -> L1RollingHashUpdatedEvent.fromEthLog(log) }
+      topics = listOf(L1RollingHashUpdatedEvent.topic),
+    ) { ethLog ->
```
fluentcrafter, line 98: Likewise here, we should optimise the search given that both move monotonically.

fluentcrafter (commented):
IMO the solution is suboptimal. It's blindly replaces the query with a *full binary search* which is can be heavy. 

We only need a full binary search on the 1st query, then one the `fromBlock` can the latest event's blockNumber.

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/conflation/InvalidityProofAssembler.kt
gauravahuja: I've implemented checking the DB approach. On the specific l1BlockNumber field: for this path it wouldn't add anything, for two reasons:
  - When the previous FTX's record exists, we already get the rolling hash directly from db, so there's no chain query to shorten.
  - When it's absent (table is pruned), an l1BlockNumber column wouldn't help either as it would have lived on that same pruned row.

coordinator/ethereum/message-anchoring/src/main/kotlin/linea/anchoring/MessageAnchoringService.kt
gauravahuja: Done, I keep track of the fromBlock

coordinator/ethereum/message-anchoring/src/main/kotlin/linea/anchoring/clients/L1MessageSentEventsFetcher.kt
gauravahuja: Done, I keep track of the fromBlock

coordinator/app/src/test/kotlin/linea/coordinator/config/v2/ForcedTransactionsConfigParsingTest.kt
gauravahuja: - eth-logs-search-block-chunk-size (default 1000)
  - used by the rolling-forward / streaming scans (getLogsRollingForward and the ForcedTransactionAdded polling subscription). 
  - Here we walk forward collecting all matching events, so
  the chunk bounds the number of logs returned per getLogs and gives the loop fine-grained timeout / stopAfterTargetLogsCount granularity. 
  - Small is good.

- eth-logs-search-max-block-range (default 10 000)
  - used as the chunk size for the binary-search point lookups (findLog: locate one RollingHashUpdated / ForcedTransactionAdded by its indexed number). 
  - Here we want the widest span 
  the provider allows per probe so convergence is ~log2(range / chunk), 
  - Large is good (bounded to the provider cap, e.g. Infura's 10k).

coordinator/ethereum/message-anchoring/src/main/kotlin/linea/anchoring/MessageAnchoringService.kt
```diff
@@ -79,26 +83,51 @@ class MessageAnchoringService(
...
+    val fromBlock = rollingHashSearchFromBlock.get()
+    return findRollingHashUpdatedEvent(messageNumber, fromBlock)
+      .thenCompose { ethLog ->
+        if (ethLog != null || fromBlock == BlockParameter.Tag.EARLIEST) {
+          SafeFuture.completedFuture(ethLog)
+        } else {
+          // Cached window overshot (unexpected: message numbers to anchor increase monotonically).
+          // Fall back to a full-range search so an existing event is never missed.
+          findRollingHashUpdatedEvent(messageNumber, BlockParameter.Tag.EARLIEST)
+        }
```
fluentcrafter, line 102: I don't understand this branching, can you please illustrate when we may fall in the else branch?

coordinator/ethereum/forced-transactions/src/main/kotlin/linea/ftx/conflation/InvalidityProofAssembler.kt
fluentcrafter: > On the specific l1BlockNumber field: for this path it wouldn't add anything, for two reasons:

By protocol, the `parentFtx` event will be before the `currentFtx`, so we can limit the `toBlock = currentFtx.l1BlockNumber` because the event will be before this one.

fluentcrafter (approved):

coordinator/ethereum/message-anchoring/src/main/kotlin/linea/anchoring/MessageAnchoringService.kt
gauravahuja: The else is reached only when both hold:
  1. the bounded search returned null (no RollingHashUpdated for messageNumber in [cachedFromBlock, LATEST]), and
  2. fromBlock != EARLIEST i.e. we'd already advanced the cached high-water-mark from a previous lookup.

It only triggers when the event we're looking for sits at a block earlier than rollingHashSearchFromBlock. Under normal operation this never happens. The else is purely a defensive approach. We can throw an exception in th else or make one full window search just to be safe.

PR #3475, opened 2026-06-30
fix(prover): fix the bitdecompose.go by removing IsPackedLimbNotZero

AlexandreBelling:
This PR fixes a missing constraints following a broken (and unneeded constraint filter). This breaks the setup.

@gusiri Let's not include in the current version update, but to the next one to avoid disruption on your end.

There is some risks with this one as I have not fully grasped where this extra column was supposed to achieve.

### Checklist

* [x] If this change is deployed to any environment (including Devnet), E2E test coverage exists or is included in this
  PR.

### DIFF
--- a/prover/protocol/dedicated/bits/bitdecompose_test.go
+++ b/prover/protocol/dedicated/bits/bitdecompose_test.go
@@ -84,3 +84,75 @@ func TestBitDecomposeHeterogeneousLimbs(t *testing.T) {
 	proof := wizard.Prove(comp, prove)
 	require.NoError(t, wizard.Verify(comp, proof), "recombination constraints must hold")
 }
+
+// TestBitDecomposeRejectsMismatchedBits is the soundness counterpart to the
+// happy-path test: it asserts that the bit-recombination constraint binds the
+// committed packed limbs to the decomposed bits unconditionally.
+//
+// Previously the recombination was multiplied by a committed, otherwise
+// unconstrained IsPackedLimbNotZero gate; a prover could zero that gate to make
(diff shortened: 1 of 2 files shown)


gusiri (approved):

prover/protocol/dedicated/bits/bitdecompose_test.go
```diff
@@ -84,3 +84,75 @@ func TestBitDecomposeHeterogeneousLimbs(t *testing.T) {
...
+// Previously the recombination was multiplied by a committed, otherwise
+// unconstrained IsPackedLimbNotZero gate; a prover could zero that gate to make
+// the identity vacuous and assign bits unrelated to the packed value. For the
+// Merkle gadget that decouples the committed accumulator position from the path
+// actually walked (false (non-)membership / state-root forgery). Here the
+// prover commits one position (committedPos) but assigns boolean bits that
+// decompose a different value (pathPos); verification must reject.
+func TestBitDecomposeRejectsMismatchedBits(t *testing.T) {
+
+	const (
```
arijitdutta67, line 101: Just a cosmetic one, we can merge the two const blocks.

arijitdutta67 (approved):
Even if we add constraint for `bd.IsPackedLimbNotZero` to be zero when the input is zero, it does not serve much of a  purpose. So I agree with the change.

AlexandreBelling:
Shall we merge that one?

arijitdutta67:
> Shall we merge that one?

@Soleimani193 this is the fix for the bug you were mentioning to me. Just adding you in the loop :)

Issue #3247, opened 2026-06-01
Maru - BUG - multi validator - chain fork due to round number in QBTD block header hash


fluentcrafter:
# Bug: Chain fork when COMMIT quorum is collected after round change timer but before ROUND_CHANGE quorum

## Summary

In a multi-validator Maru QBFT setup, a chain fork can occur when validators experience message delivery delays that cause them to commit a block at different rounds. Because Maru includes the round number in the block header hash (unlike Besu), two validators can end up with different, irreconcilable blocks at the same height — both signed by a quorum.

## Environment

- 4 validators
- Local K3S cluster
- 1-second block time

## Steps to Reproduce

1. Start a 4-validator QBFT network.
2. Introduce message delivery delays such that:
   - Validators 0–2 trigger a `ROUND_CHANGE` on timer.
   - Validator 3 collects a `COMMIT` quorum **after** the round timer fires, but **before** seeing a `ROUND_CHANGE` quorum.
   - `ROUND_CHANGE` messages from Validators 0–2 arrive at Validator 3 even later.
3. Validators 0–2 collect the `ROUND_CHANGE` quorum before their `COMMIT` quorum — so they do **not** accept the block for round 0 and advance to round 1.
4. Validator 3 accepts the block for round 0 (with a valid quorum of `COMMIT` signatures).

## Expected Behavior

All validators commit the same block at the same height. No fork occurs.

## Actual Behavior

The chain forks:
- Validators 0–2 commit a block at **round 1**.
- Validator 3 commits a block at **round 0**.

Both blocks for the same height carry valid quorum signatures. Because Maru includes the round number in the block header hash, the two blocks have different hashes and **the fork cannot be resolved retrospectively**.

> Note: This is Maru-specific. Besu excludes the round number from the block header hash, so the same scenario would not produce a fork there (both blocks would hash identically).

## Root Cause

[...]

PR #3493, opened 2026-07-02
feat(maru): Added a new CL phase in which Beacon block's chain identi…

Filter94:
…ty doesn't depend on the round on which consensus was achieved. Related major refactoring to the serialization module

This PR implements issue(s) #3247

### Checklist

  PR.


maru/app/src/integrationTest/kotlin/maru/app/MaruClForkUpgradeTest.kt
```diff
@@ -0,0 +1,173 @@
...
+        assertThat(beaconChain.getLatestBeaconState().beaconBlockHeader.number)
+          .isGreaterThanOrEqualTo(targetHeight)
+      }
+  }
+
+  @Test
+  fun `network upgrades live from QBFT_PHASE0 to QBFT_PHASE1 without losing chain agreement`() {
+    val stackStartupMargin = 40UL
+    val blocksBeforeSwitch = 5UL
+    val currentTimestamp = (System.currentTimeMillis() / 1000).toULong()
```
fluentcrafter, line 87: `currentTimestamp` is captured before both Maru stacks start and before `awaitTillMaruHasPeers` completes (lines 93–113). If startup + peering takes longer than the 40 s `stackStartupMargin` — plausible on a loaded CI machine running 4 JVMs — `phase1ActivationTimestamp` will already be in the past when block production begins. Every block ends up post-PHASE1, `preSwitchHeaders` is empty, and the test fails with *"Expected at least one block before the PHASE1 activation"* — a misleading failure that hides the real cause.

Fix: move the timestamp capture to **after** peering is established so the margin is measured from a known-ready state:

```kotlin
followerStack.maruApp.awaitTillMaruHasPeers(1u)
validatorStack.maruApp.awaitTillMaruHasPeers(1u)
val currentTimestamp = (System.currentTimeMillis() / 1000).toULong() // ← move here
val phase1ActivationTimestamp = currentTimestamp + blocksBeforeSwitch
```

maru/config/src/main/kotlin/maru/config/consensus/JsonFriendlyForksSchedule.kt
```diff
@@ -93,7 +93,7 @@ object ForkConfigDecoder : Decoder<JsonFriendlyForksSchedule> {
               )
             }.toSet(),
           fork = ChainFork(
-            clFork = ClFork.QBFT_PHASE0,
+            clFork = obj["clfork"].valueOrNull()?.let { ClFork.valueOf(it) } ?: ClFork.QBFT_PHASE0,
```
fluentcrafter: `ClFork.valueOf(it)` throws a raw `IllegalArgumentException` on unknown enum names **before** a `ConfigResult` is produced, so it escapes Hoplite's structured `ConfigFailure` path entirely. An operator config with a typo (e.g. `clfork: QBFT_PHASE2`) gets a JVM stack trace instead of a readable config error. The same pre-existing issue exists for `ElFork.valueOf` on the line below — since you're adding a new config key here, this is a good moment to fix both:

```kotlin
clFork = obj["clfork"].valueOrNull()
  ?.let { name ->
    ClFork.entries.find { it.name == name }
      ?: return (ConfigFailure.Generic("Unknown clfork value '$name'") as ConfigFailure).invalid()
  }
  ?: ClFork.QBFT_PHASE0,
```

maru/app/src/integrationTest/kotlin/maru/app/MaruMultiValidatorTest.kt
```diff
@@ -57,10 +58,18 @@ class MaruMultiValidatorTest {
 
   private val log = LogManager.getLogger(this.javaClass)
 
-  private val maruFactory0 = MaruFactory(validatorPrivateKey = key0)
-  private val maruFactory1 = MaruFactory(validatorPrivateKey = key1)
-  private val maruFactory2 = MaruFactory(validatorPrivateKey = key2)
-  private val maruFactory3 = MaruFactory(validatorPrivateKey = key3)
+  private var maruFactory0 = MaruFactory(validatorPrivateKey = key0)
```
fluentcrafter: Changing these from `val` to `var` + mutating them in `useClForkForAllValidators` works correctly under JUnit 5's default per-method lifecycle (fresh instance per test), but the isolation relies on that guarantee implicitly — there is no `@BeforeEach` reset.

If a `@TestInstance(Lifecycle.PER_CLASS)` annotation is ever added to this class (common for slow integration tests to avoid re-running `@BeforeAll`), the PHASE1 test would silently contaminate every test that runs after it in the same instance.

Prefer making the PHASE1 test self-contained by declaring its four factories locally inside the test body, keeping the `val` fields for all other tests.

maru/core/src/main/kotlin/maru/core/BeaconBlockHeader.kt
```diff
@@ -22,16 +22,38 @@ data class BeaconBlockHeader(
 ) {
   val hash by lazy { headerHashFunction(this) }
```
fluentcrafter: As raised before, I am concerned that the `BeaconBlockHeader` data class includes the `headerHashFunction`, which is not a true hash function because different inputs (round number) yield the same hash. 

This was a big design flaw, IMO. The current implementation aims to fix a bug, but at the expense of other tradeoffs.

I'm now wondering if this has 2 responsibilities mixed together: 
- `BeaconBlock`Header - the beacon block that consensus was reached upton.
- QBFT message of (BeaconBlock) - which represents the QBFT message of a given beacon block. 

Because we decided to save the rounds (for future purposes),  are they blended together?

Anyway, I think we should clearly segregate the 2 concepts:
- **hash**Function - which is a true hash, inline equals/hashcode contract 
- `beaconBlockIdHash` - which is what `hash` & `headerHashFunction` seem to be doing. Maybe we shall rename the field for clarity?

fluentcrafter (dismissed):
full disclosure: review assisted by claude :)

My main concern is the `hash` field as explained above.

maru/app/src/integrationTest/kotlin/maru/app/MaruClForkUpgradeTest.kt
Filter94: It doesn't make sense. We need to assign the switch time before we create Maru instances so they're using the same genesis file

maru/core/src/main/kotlin/maru/core/BeaconBlockHeader.kt
Filter94: > which is not a true hash function

You mean it's not a proper implementation of a `hashCode`? It's not meant to be, since we're using a separate `hash` accessor for the `BeaconBlockHeader`'s on-chain identity

> This was a big design flaw, IMO

What was the design flaw exactly?

> Anyway, I think we should clearly segregate the 2 concepts

They are segregated. I can rename the `headerHashFunction` if it will make it more obvious

> I'm now wondering if this has 2 responsibilities mixed together

I think you can say so. In [Dafny spec](https://github.com/Consensys/qbft-formal-spec-and-verification/blob/main/dafny/spec/L1/types.dfy#L58) there is the same duality, `RawBlock` vs `Block`. The difference is that `Block` doesn't have the signatures and the round

fluentcrafter (approved):

Issue #2817, opened 2026-04-14
Coordinator - send blobs with EIP-7594 support


fluentcrafter:

gauravahuja:
https://github.com/LFDT-web3j/web3j/pull/2263 hasn't been merged into web3j yet.

PR #3514, opened 2026-07-06
feat(coordinator)!: Web3j upgrade to onboard 7594 support

Filter94:
This PR implements issue(s) #2817

### Checklist

  PR.


jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/ExecutionWitnessResponseParser.kt
```diff
@@ -1,8 +1,8 @@
 package linea.web3j.ethapi
 
-import com.fasterxml.jackson.databind.JsonNode
 import linea.ethapi.ExecutionWitness
 import linea.kotlin.decodeHex
+import tools.jackson.databind.JsonNode
```
Filter94, line 5: Newer jackson API

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/L1SubmissionConfigToml.kt
```diff
@@ -129,9 +129,11 @@ data class L1SubmissionConfigToml(
...
     val submissionTickInterval: Duration = 12.seconds,
     val maxSubmissionTransactionsPerTick: UInt = 2u,
-    // eip-7691 on Prague fork allows up to 9 blobs per transaction.
-    // however, Geth nodes fail with "transaction too large" error. only 7 blobs are accepted
-    val targetBlobsPerTransaction: UInt = 7u,
+    // 6 is the current EIP-7594 (Fusaka/PeerDAS) network wrapper cap on blobs per transaction
+    // (org.web3j.crypto.transaction.type.Transaction4844.MAX_BLOBS_PER_TRANSACTION). Not enforced
+    // here so a future EIP raising the cap doesn't require a config-validation change; a value
+    // above the library's cap fails fast when the blob transaction is built.
+    val targetBlobsPerTransaction: UInt = 6u,
```
gauravahuja, line 136: When this is released do we need to make sure that the configs also update to 6? If this is a cap it is better to enforce that here

fluentcrafter (approved):

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/L1SubmissionConfigToml.kt
Filter94: All configs are already at 6. 

> If this is a cap it is better to enforce that here

That was my first thought, but I deliberately chose not to enforce it to avoid a code change when/if blob cap is raised

Issue #3518, opened 2026-07-06
Bound eth_getLogs in EventBasedContractDeploymentBlockNumberProvider


gauravahuja:
## Description

Follow-up to #3435 (and PR #3473). While auditing remaining unbounded `eth_getLogs` calls, one latent case remains:

`EventBasedContractDeploymentBlockNumberProvider.getDeploymentBlock()` — `jvm-libs/linea/clients/linea-contract-clients/src/main/kotlin/linea/contract/ContractDeploymentBlockNumberProvider.kt:32`

```kotlin
ethApiClient.getLogs(
  fromBlock = BlockParameter.Tag.EARLIEST,
  toBlock = BlockParameter.Tag.LATEST,
  address = contractAddress,
  topics = listOf(Upgraded.topic),
)
```

It locates the contract's deployment block by finding the earliest `Upgraded` event via an `EARLIEST..LATEST` scan. Rate-limited RPC providers (e.g. Infura) reject `eth_getLogs` spanning more than 10,000 blocks, so this will fail on such providers.

## Why it's lower severity (not urgent)

- **Conditional fallback**: only used by `Web3JL2MessageServiceSmartContractClient` when `smartContractDeploymentBlockNumber` is *not* configured; otherwise a `StaticContractDeploymentBlockNumberProvider` is used and no `getLogs` is issued. Setting the deployment block number in config avoids this path entirely.
- **Cached**: the result is cached (`deploymentBlockNumberCache`), so it runs at most once per instance, not on a poll loop.
- Runs against the **L2 message-service** client, not the L1 finalization path that first surfaced the limit.

## Proposed fix

Unlike the indexed monotonic lookups fixed in #3473, this is a *first-occurrence* search, so binary search does not apply. Use a bounded rolling-forward scan that stops at the first match:

```kotlin
ethLogsSearcher.getLogsRollingForward(
  fromBlock = BlockParameter.Tag.EARLIEST,
  toBlock = BlockParameter.Tag.LATEST,
  address = contractAddress,
  topics = listOf(Upgraded.topic),
  chunkSize = <= 10_000,
  stopAfterTargetLogsCount = 1u,
).thenApply { it.logs.firstOrNull()?.blockNumber }
```

## Notes / blast radius

- Requires switching `EventBasedContractDeploymentBlockNumberProvider` from `EthApiClient` to `EthLogsSearcher` (interface already available in the module).
[...]

PR #3520, opened 2026-07-06
fix(coordinator): bound eth_getLogs in deployment-block lookup

gauravahuja:
## Description

#3518. Stacked on #3519 (base branch `fix/bound-finalized-state-getlogs`) because the added test reuses the `testFixtures(interfaces)` test dependency introduced there.

`EventBasedContractDeploymentBlockNumberProvider.getDeploymentBlock` issued an unbounded `getLogs(EARLIEST..LATEST)` query to find the earliest `Upgraded` event (the contract deployment block). Rate-limited providers (e.g. Infura) reject `eth_getLogs` spanning more than 10,000 blocks.

## Change

This is a *first-occurrence* search (not an indexed monotonic lookup), so binary search does not apply. Replaced it with a bounded forward-chunk scan over `EthApiClient`: walk `[EARLIEST..LATEST]` in `<= ethLogsSearchMaxBlockRange` (default 10_000) windows and return the earliest `Upgraded` event from the first non-empty chunk. The result stays cached, so it runs at most once.

Kept `EthApiClient` (self-contained recursive scan, no `Vertx`) rather than switching to `EthLogsSearcher` — this avoids a DI cascade through the widely-constructed L2 message-service `create`/`createReadOnly` factories, and the new max-range is a defaulted constructor param, so **no call sites change**.

## Testing

New `EventBasedContractDeploymentBlockNumberProviderTest`: bounded multi-chunk search hit + not-found. Full `linea-contract-clients` suite passes; spotless clean.

---

> [!NOTE]
> **Low Risk**
> Localized RPC lookup change with the same semantics and cache; no auth or call-site wiring changes.
> 
> **Overview**
> Fixes coordinator/L2 contract startup against rate-limited RPCs by **replacing one unbounded `getLogs(EARLIEST..LATEST)`** in `EventBasedContractDeploymentBlockNumberProvider` with a **forward scan in windows of at most `ethLogsSearchMaxBlockRange` blocks** (default **10_000**).
> 
> After resolving the chain span via `getAbsoluteBlockNumbers`, `findFirstUpgradedBlock` walks chunks until the first `Upgraded` log appears and returns the **earliest** block in that chunk; behavior and caching are unchanged, and existing constructors keep working via the defaulted range param.
> 
> Adds **`EventBasedContractDeploymentBlockNumberProviderTest`** for multi-chunk success and the not-found `IllegalStateException` path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 049a0281f908969aa83abdebea04e3bdc17ebba3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/jvm-libs/linea/clients/linea-contract-clients/src/main/kotlin/linea/contract/ContractDeploymentBlockNumberProvider.kt
+++ b/jvm-libs/linea/clients/linea-contract-clients/src/main/kotlin/linea/contract/ContractDeploymentBlockNumberProvider.kt
@@ -2,7 +2,9 @@ package linea.contract
 
 import linea.contract.events.Upgraded
 import linea.domain.BlockParameter
+import linea.domain.toBlockParameter
 import linea.ethapi.EthApiClient
+import linea.ethapi.extensions.getAbsoluteBlockNumbers
 import org.apache.logging.log4j.LogManager
 import org.apache.logging.log4j.Logger
 import tech.pegasys.teku.infrastructure.async.SafeFuture
@@ -21,6 +23,7 @@ class StaticContractDeploymentBlockNumberProvider(
 class EventBasedContractDeploymentBlockNumberProvider(
   private val ethApiClient: EthApiClient,
   private val contractAddress: String,
+  private val ethLogsSearchMaxBlockRange: UInt = 10_000u,
   private val log: Logger = LogManager.getLogger(EventBasedContractDeploymentBlockNumberProvider::class.java),
 ) : ContractDeploymentBlockNumberProvider {
   private val deploymentBlockNumberCache = AtomicReference<ULong>(0UL)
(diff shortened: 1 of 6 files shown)


jvm-libs/linea/clients/linea-contract-clients/src/main/kotlin/linea/contract/ContractDeploymentBlockNumberProvider.kt
```diff
@@ -21,36 +23,54 @@ class StaticContractDeploymentBlockNumberProvider(
...
+        toBlock = chunkEnd.toBlockParameter(),
+        address = contractAddress,
+        topics = listOf(Upgraded.topic),
+      ).thenCompose { logs ->
+        // Scanning forward, so the first chunk with any Upgraded event holds the earliest one.
+        val earliestInChunk = logs.minByOrNull { it.blockNumber }
+        if (earliestInChunk != null) {
+          SafeFuture.completedFuture(earliestInChunk.blockNumber)
+        } else {
+          findFirstUpgradedBlock(fromBlock = chunkEnd + 1UL, toBlock = toBlock)
```
fluentcrafter: Recursive algorithms are problematic with Futures; they generate deeply nested future handles that hang... 

why not use FindLogs?

gauravahuja: I need the first block, find logs will return any one

gauravahuja: Actually i think i can simply use getLogsRollingForward, will update

gauravahuja: can you review again

fluentcrafter (approved):

PR #3534, opened 2026-07-07
feat(coordinator): Add inital block number config for finalized state search

gauravahuja:
This PR implements issue(s) #

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> Mis-set deployment block could skip `FinalizedStateUpdated` events and break finalization/FTX state; behavior when unset matches the previous full-chain search.
> 
> **Overview**
> Adds optional **`protocol.l1.contractDeploymentBlockNumber`** (TOML → config, same pattern as L2) so L1 rollup log scans for **`FinalizedStateUpdated`** do not start at chain genesis when the contract was deployed later.
> 
> **`Web3JLineaRollupSmartContractClientReadOnly`** gains **`finalizedStateSearchInitialBlockParameter`** (default **`EARLIEST`**). The forward-search cache and full-range fallback now use that bound instead of always **`EARLIEST`**, shrinking **`getLogs`** ranges for rate-limited RPC providers.
> 
> **`L1DependentApp`** (finalization monitor) and **`ConflationApp`** (forced-transactions contract client) pass **`configs.protocol.l1.contractDeploymentBlockNumber ?: EARLIEST`**. Internal **`finalisedBlockNumber`** parameters in the finalized-state search path are renamed to **`finalizedBlockNumber`**; protocol parsing tests are updated for the new L1 field.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a20921351f958546f2888cf104a31eac0cf490a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/ProtocolToml.kt
+++ b/coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/ProtocolToml.kt
@@ -36,6 +36,7 @@ data class ProtocolToml(
   data class Layer1Config(
     val contractAddress: String,
     val blockTime: Duration = 12.seconds,
+    val contractDeploymentBlockNumber: ULong?,
   )
 
   data class Layer2Config(
@@ -55,6 +56,8 @@ data class ProtocolToml(
       ProtocolConfig.Layer1Config(
         contractAddress = this.l1.contractAddress,
         blockTime = this.l1.blockTime,
+        contractDeploymentBlockNumber = this.l1.contractDeploymentBlockNumber
+          ?.let { BlockParameter.BlockNumber(it) },
       ),
       l2 =
       ProtocolConfig.Layer2Config(
--- a/jvm-libs/linea/clients/linea-contract-clients/src/main/kotlin/linea/contract/l1/Web3JLineaRollupSmartContractClientReadOnly.kt
+++ b/jvm-libs/linea/clients/linea-contract-clients/src/main/kotlin/linea/contract/l1/Web3JLineaRollupSmartContractClientReadOnly.kt
@@ -27,6 +27,7 @@ open class Web3JLineaRollupSmartContractClientReadOnly(
   val contractAddress: String,
   private val ethLogsSearcher: EthLogsSearcher,
   private val l1EventSearchMaxBlockRange: UInt = 10_000u,
+  private val finalizedStateSearchInitialBlockParameter: BlockParameter? = null,
   private val log: Logger = LogManager.getLogger(Web3JLineaRollupSmartContractClientReadOnly::class.java),
 ) : LineaRollupSmartContractClientReadOnly,
   LineaRollupSmartContractClientReadOnlyFinalizedStateProvider,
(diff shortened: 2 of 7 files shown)


coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/ProtocolToml.kt
```diff
@@ -36,6 +36,7 @@ data class ProtocolToml(
   data class Layer1Config(
     val contractAddress: String,
     val blockTime: Duration = 12.seconds,
+    val contractDeploymentBlockNumber: ULong?,
```
Filter94, line 39: To prevent poor performance, shall we instead make it a mandatory config? It would force the deployer to set it explicitly

jvm-libs/linea/clients/linea-contract-clients/src/main/kotlin/linea/contract/l1/Web3JLineaRollupSmartContractClientReadOnly.kt
```diff
@@ -27,6 +27,7 @@ open class Web3JLineaRollupSmartContractClientReadOnly(
   val contractAddress: String,
   private val ethLogsSearcher: EthLogsSearcher,
   private val l1EventSearchMaxBlockRange: UInt = 10_000u,
+  private val finalizedStateSearchInitialBlockParameter: BlockParameter? = null,
```
Filter94: To me 0/BlockParameter.Tag.EARLIEST makes more as a default

Filter94 (dismissed):

jvm-libs/linea/clients/linea-contract-clients/src/main/kotlin/linea/contract/l1/Web3JLineaRollupSmartContractClientReadOnly.kt
gauravahuja: If I set default to BlockParameter.Tag.EARLIEST, do you want me to still keep the type nullable? Having the type nullable makes the callsite easier since the config is also nullable.

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/ProtocolToml.kt
gauravahuja: I'm not sure how to take a call here. The coordiantor works without the config. Is performance optimization a reason to make a config mandatory? We have a similar `Layer2Config` with optional `contractDeploymentBlockNumber`

jvm-libs/linea/clients/linea-contract-clients/src/main/kotlin/linea/contract/l1/Web3JLineaRollupSmartContractClientReadOnly.kt
Filter94: Yes, that's what I mean. I think we don't need nullability here since 0/EARLIEST === null

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/ProtocolToml.kt
Filter94: I think it is, since for production, the time difference can be 100 or 1000 times. I didn't measure, but the performance implications are severe enough IMO

Filter94 (approved):

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/ProtocolToml.kt
fluentcrafter: There is a significant difference between L1 and L2.

- L2 - MessageService contract is expected to be in the first few blocks <10, so a Rolling Forward Query makes it in 1 go;
- L1 was deployed whenever. So I believe that making it mandatory would be the best...

PR #3541, opened 2026-07-09
refactor(prover-ray): FRI cleanups

Tabaie:
This PR dismantles FRI as a standalone low-degree test in favor of simplifying the PCS flow.
Other small tightenings:
* A single Merkle path at query position `s` yields every size's aligned leaf for that tree, so the per-(level, tree) re-openings are dropped.
* Replace all leaf digests in the proof by rows, as the former can and must be recomputed by the verifier
* Remove certain redundant sibling or conjugate fields in the proof, making it non-malleable


prover-ray/crypto/koalabear/fri/pcs.go
ivokub: Documentation stale, there doesn't seem to be LastRoot method anymore

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -81,7 +81,7 @@
 //	// Absorb the final polynomial; squeeze query positions; open.
 //	transcript.Absorb(friState.FinalPoly())
```
ivokub: Similarly, there isn't FinalPoly() method it seems?

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -903,42 +910,100 @@ func (pcs *PCS) claimsForBatchEntry(
...
+	return OpeningProof{
+		InputQueries: pcs.openInputQueries(queryPositions),
+		FRIProof:     state.Open(queryPositions),
+	}
+}
+
+func (pcs *PCS) openInputQueries(queryPositions []int) []InputQuery {
 	restricted, err := pcs.restrictToOpenings()
 	if err != nil {
 		panic(err)
```
ivokub, line 910: I think we shouldn't panic this easily -- usually the panics are reserved for "package/library inconsistency", not when inputs are incorrect. This could crash the program early when the input is somehow incorrect.

I'd recommend returning an error here and then return error also in `PCS.Open`

prover-ray/crypto/koalabear/fri/fri_prover_state.go
```diff
@@ -131,56 +131,12 @@ func (st *ProverState) Open(openedPositions []int) Proof {
 		panic("fri: ProverState.Open: called before all folding rounds were consumed")
```
ivokub, line 131: Return error instead of panicing.

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -1065,14 +1201,17 @@ type VerifyInputs struct {
 	Challenges Challenges
 }
```
ivokub, line 1191: Perhaps we should add a constructor for creating `VerifyInputs`? In tests right now you use `shapesFromBatches` helper utility which is not publicly available. So using PCS verifier right now outside of the package is complicated as the API user would have to know how to construct the `VerifyInput.Shapes` itself.

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -903,42 +910,100 @@ func (pcs *PCS) claimsForBatchEntry(
...
 	}
-	return res
+	return inputs
+}
+
+func openInputBranch(p Params, committed CommitterState, queryPosition int) InputBranch {
+	tree := committed.Tree
+	numLeaves := tree.NumLeaves()
+	if numLeaves > p.N || p.N%numLeaves != 0 {
+		panic("fri: openInputBranch: tree size incompatible with domain size")
```
ivokub: Again -- please do not panic here but instead return an error.

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -903,42 +910,100 @@ func (pcs *PCS) claimsForBatchEntry(
...
+	}
+	input.Siblings = siblings
+	for sizeLog2 := range committed.EncodedTable {
+		table := committed.EncodedTable[sizeLog2]
+		if table.NumRows() == 0 || table.Size() == numLeaves {
+			continue
+		}
+		levelSize := table.Size()
+		if levelSize > numLeaves || numLeaves%levelSize != 0 {
+			panic("fri: openInputBranch: level size incompatible with tree size")
```
ivokub: panic vs error

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -483,25 +484,28 @@ func powers(x field.Ext, length int) []field.Ext {
...
+// InputQuery holds the PCS input-tree openings for one FRI query.
+type InputQuery []InputBranch
+
+// InputBranch is a Merkle branch whose path leaves are opened as row preimages.
+// ConjugateLeaf is the row at the round-0 fold-conjugate position s^1; it is
+// non-nil iff this branch backs the top level (the only level folded as a
+// conjugate pair straight out of an input tree). Its Merkle digest is derived
+// from it rather than transmitted, so for a top branch Siblings holds one
+// fewer entry than AuxSiblings.
+type InputBranch struct {
```
ivokub: The type name is a bit ambiguous right now. Perhaps InputTreeOpening is better? (this is an opening proof)

prover-ray/crypto/koalabear/fri/fri.go
```diff
@@ -164,13 +166,10 @@ type Level struct {
...
 // commits to those polynomials before invoking FRI).
 type Proof struct {
-	// LevelQueries[l-1][k] = openings for levels[l].Evals at outer query k.
-	LevelQueries [][]QueryLayer
-
 	// Running-polynomial FRI path
-	FRIRoots     []field.Octuplet // Merkle roots for running poly T_1..T_{r-1}
-	FinalPolyExt []field.Ext
-	FRIQueries   []Query // len = NumQueries
+	FRIRoots       []field.Octuplet // Merkle roots for running poly T_1..T_{r-1}
```
ivokub: Perhaps there is inconsistency here as well. You renamed `FRIQueries` to `RunningQueries`, but we still have `FRIRoots` field. Perhaps just `Roots`? We're inside `fri` package already.

ivokub (changes requested):
In general looks good. Though I think we should rely less on panics and use errors instead. In that case it is more explicit that something can go wrong.

prover-ray/crypto/koalabear/fri/fri.go
```diff
@@ -268,175 +267,99 @@ func checkLevelTrees(label string, trees []*Tree) error {
...
-
+// checkOpeningProofShape validates prf's structure against p and the
+// challenge lengths before any authentication or reconstruction runs, so a
+// malformed proof can never cause an out-of-bounds access later.
+func checkOpeningProofShape(p Params, prf Proof, foldAlphas []field.Ext, positions []int) error {
 	wantFRIRoots := p.numRounds - 1
 	if p.numRounds <= 1 {
 		wantFRIRoots = 0
 	}
+	wantRunningQueries := wantFRIRoots
```
arijitdutta67: Maybe `wantLayersPerRunningQuery` instead of `wantRunningQueries`, renaming makes more sense to me when I read line 294 and 295.

prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -1006,14 +1141,15 @@ func reconstructQueryValueAt(
...
 		}
-		sizedRow := rows[entry.BatchIdx][entry.SizeLog2]
-		row := sizedRow.Leaf
+		branch := opening[inputIndexByBatch[entry.BatchIdx]]
+		var row RowOpening
 		if sibling {
-			if !sizedRow.HasSibling {
-				return field.Ext{}, fmt.Errorf("fri: reconstructQueryValueAt: batch %d size %d missing sibling row",
-					entry.BatchIdx, entry.SizeLog2)
+			row = *branch.ConjugateLeaf
```
arijitdutta67, line 1136: Please add a nil guard before dereferencing.

prover-ray/crypto/koalabear/fri/pcs.go
arijitdutta67: Had a comment on this philosophy. In the old prover, we followed the offensive programming strategy (panic instead of error) (proposed originally by @AlexandreBelling). We wanted to fail immediately when some input is wrong. Because we shall never get a correct proof out of it and detecting it immediately helps debugging.

But we did use error wrapping when it makes sense to propagate and collect errors as much as possible when step by step bug fixing is time consuming (e.g. limitless prover when running a test takes a lot of time). But here, for local pcs bug fixing, panic sometimes makes more sense to me than returning error.

arijitdutta67 (dismissed):
Overall looks good to me.

Tabaie:
Much of the feedback is regarding the heavy use of panics rather than the more Go-idiomatic error return-parameters. After oral discussions with @ivokub and @arijitdutta67 we decided to get rid of most panics in the verifier, as a verifier should not crash easily.
The prover is already panic-heavy. So for consistency's sake, we will keep the panics there. The panic vs error question in the prover is a deeper "coding philosophy" question and will have to involve @AlexandreBelling and other prover stakeholder.

prover-ray/crypto/koalabear/fri/pcs.go
Tabaie: I gave it a go but couldn't come up with anything that was nontrivial and did much more than filling the structure with its arguments. Made a public `Shape()` method instead.

prover-ray/crypto/koalabear/fri/pcs.go
Tabaie: Good idea. Implemented.

prover-ray/crypto/koalabear/fri/fri.go
Tabaie: It does sound redundant. Although I still think it means something because it is referring to the underlying low-degree testing protocol. Renamed to `RoundRoots`.

arijitdutta67 (approved):

ivokub (approved):
Thanks! Changes look good!

PR #3542, opened 2026-07-10
fix(prover): Bind `parentStateRoot` to `parentShnarf` preimage

Tabaie:
This PR tightens the state root binding of the aggregated proof seen by L1.

Now the parent shnarf (to which the aggregation PI is already bound) is opened in the PI interconnection circuit and ensured to match the parent state root of the first execution proof. This requires no nontrivial change to the verifier contract, but it does involve updates to the coordinator as the prover backend now needs the parent shnarf pre-image.

It also removes the Koalabear migration state root fallback, which is no longer necessary.

-----

I am not well versed in the coordinator or the Kotlin language so we will probably need some hands-on involvement from the coordinator team.


AlexandreBelling:
Any blocker for merging this?

Tabaie:
> Any blocker for merging this?

The team suggested a more "surgical" set of changes to the coordinator that I haven't got around to implementing yet.

PR #3543, opened 2026-07-12
docs(linea-besu): Linea-Besu plugin options generator + doc.linea PR workflow

Othryades:
## Summary
- Extracts Linea-Besu plugin options through the existing Java reflection module and renders neutral MDX partials.
- Generated MDX, manifest, and report files are ephemeral and ignored by Git; pull-request CI generates and uploads them for review.
- `.github/workflows/besu-plugin-options-docs.yml` validates relevant PRs and automatically opens or updates a scoped `Consensys/doc.linea` PR after relevant changes land on `main`.
- Publication uses the `DOC_LINEA_PR_*` GitHub App secrets and can modify only `docs/stack/reference/_generated/besu/`.

## Companion documentation PR
- [Consensys/doc.linea#1644](https://github.com/Consensys/doc.linea/pull/1644) atomically moves the current Besu partials and wrapper imports into the component-owned namespace. Merge it before this publisher is enabled.

## Test plan
- [x] Node 24 renderer tests: 8 passed, 8 extractor-dependent tests skipped locally
- [x] focused Prettier and workflow YAML validation
- [x] `git diff --check`
- [ ] Java extraction on CI (local machine has JDK 17; repository requires JDK 25)

---

> [!NOTE]
> **Low Risk**
> Mostly new docs tooling and CI; runtime Besu behavior is unchanged aside from two CLI option description/label fixes for published reference accuracy.
> 
> **Overview**
> Adds an end-to-end pipeline so **Linea-Besu `--plugin-*` CLI options** stay in sync with Java `@Option` sources and land in **Consensys/doc.linea** as MDX partials under `docs/stack/reference/_generated/besu/`.
> 
> A new Gradle module **`besu-plugin-options-docgen`** reflects sequencer/tracer `*CliOptions` classes via picocli and writes ephemeral JSON manifest + report. The standalone **`scripts/besu-plugin-options`** tool renders neutral MDX tables (with MDX-safe escaping, hidden options as **Advanced**, forced-tx group excluded), validates wrapper import completeness, and keeps generated output out of git.
> 
> **`.github/workflows/besu-plugin-options-docs.yml`** runs generate/check/tests/Prettier on relevant PRs; on `main` (or manual publish) it copies only the Besu namespace into **doc.linea** and opens/updates a PR via a GitHub App, with path guards and a provenance partial.
> 
> Small doc-source fixes: corrected **`LineaProfitabilityCliOptions`** description for compressed-tx cache size; **`LineaL1L2BridgeSharedCliOptions`** `paramLabel` changed to `<BYTES_32>`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bac10985d4122bd974bbe43ca06b03f613f39445. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/besu-plugin-options/lib.js
```diff
@@ -0,0 +1,1093 @@
...
+  // A single literal is preserved verbatim (minus the type suffix) so decimals
+  // like `1.0` are not collapsed to `1`.
+  const literal = cleaned.match(/^([-+]?\d+(?:\.\d+)?(?:[eE][-+]?\d+)?)[lLfFdD]?$/);
+  if (literal) return { resolved: true, display: literal[1] };
+
+  // Otherwise only evaluate genuine arithmetic (digits and operators only).
+  const arithmetic = cleaned.replace(/([0-9])[lLfFdD]\b/g, "$1");
+  if (!/^[\d.\s*+\-/()]+$/.test(arithmetic)) return { resolved: false };
+  try {
+    const value = Function('"use strict";return (' + arithmetic + ")")();
```
kyzooghost: resolveNumeric uses the `Function()` constructor to evaluate Java default-value expressions. That is a code-eval primitive - it compiles and runs whatever string it is handed.

The regex above it `(/^[\d.\s*+\-/()]+$/)` makes this safe right now: only digits, dots, whitespace, and arithmetic operators can reach it, so no executable JS can be formed. The regex is the only barrier.

The risk is that the barrier is fragile. The next person who wants hex literals, scientific notation, or a qualified constant name will widen the regex - and `Function()` will then run attacker-shaped text in CI. There is nothing in the code that stops that slip.

**Suggestion** 
Drop the eval. A small arithmetic parser over the same character set does the same job with no execution sink to weaken.

scripts/besu-plugin-options/.npmrc
kyzooghost, line 1: The dependency tree is currently simple: just Prettier, with no install scripts or native builds. pnpm handles this safely by default.

However, `.npmrc` does not explicitly state that install scripts should never run. Add:

```ini
only-built-dependencies=[]
```

This makes the policy clear and blocks future dependencies from running build scripts without an intentional config change.

Low priority - this is a precaution, not a current issue.

kyzooghost:
I wonder about parsing Java with JS. We are effectively re-implementing a Java parser by hand. 

Could we consider the following options:

**1. Reflection via a tiny Java harness (most correct).** Build the plugin modules with the Gradle the repo already mandates (JDK 25), then run a small Java main that uses reflection to enumerate @Option fields and dump JSON. The JVM resolves 1024 * 1024 * 16L, BigDecimal.ONE, and String.format(...) natively - the entire resolveValueExpr/resolveNumeric layer, including the eval, disappears. Cost: a Gradle task + build time. This is the structurally right answer if correctness matters more than build speed.

**2. picocli's own model.** These annotations are picocli. A Java harness can construct CommandLine and walk CommandSpec.options() - including hidden ones - getting names, defaults, descriptions, and ${DEFAULT-VALUE} resolution for free, straight from the library that defines the semantics. Same cost as #1, cleaner data.

scripts/besu-plugin-options/lib.js
Othryades: Agreed — the regex was the only gate. Replaced `Function()` with a small recursive-descent arithmetic evaluator over the same allowlisted character set, so there is no eval sink to weaken later. Covered by the existing `DEFAULT_B = 1024 * 1024 * 16L` case plus a direct `evalArithmetic` unit test.

scripts/besu-plugin-options/.npmrc
Othryades: Done — added `only-built-dependencies=[]` to this tool’s `.npmrc` so future deps cannot run install scripts without an intentional config change.

Othryades:
Addressed Grant + Jeff review feedback in 51a10d5:

1. **MDX sanitization** — `escapeCell` now escapes `\`, backticks, `{`, `}`, `<`, `>`, `|`; applied to descriptions, names, defaults, types, and config-key lines. Fixture test asserts malicious `@Option` text cannot leave raw `{`/`}`/backticks in generated partials.
2. **Publish branch guard** — `publish-doc-linea` now fails unless `refs/heads/main` or `refs/tags/linea-besu-*` (mirrors `sdk-release.yml`). Still dry-run by default.
3. **No `Function()` eval** — `resolveNumeric` uses a pure arithmetic parser.
4. **Install-script lock** — `only-built-dependencies=[]` in the tool `.npmrc`.

On JS vs Java parsing (from Slack): keeping static JS for this POC — no monorepo compile in the docs job, matches the earlier pilot, and every publish still lands as a reviewed doc.linea PR with drift checks. Happy to spike a Java extractor later if we hit brittle parses when expanding to Prover/Postman/Coordinator.

Othryades:
Follow-up to Jeff’s JS-vs-Java point: extract step is now Java reflection.

- New module: `:linea-besu:plugins:besu-plugin-options-docgen` (`generateBesuPluginOptionsManifest`)
- Reads `@Option` + resolves defaults via `Field.get` / `create()` (no hand-rolled Java expression evaluator)
- `scripts/besu-plugin-options/lib.js` is now MDX-only (~343 lines, down from ~1.1k)
- Workflow installs JDK 25 + Go (for go-corset) before generate/check

`pnpm run generate` runs the Gradle extractor then renders MDX.

scripts/besu-plugin-options/lib.js
Othryades: Done in a broader follow-up: extract is now Java reflection (`:linea-besu:plugins:besu-plugin-options-docgen`). The `Function()` / JS semantics layer is gone; defaults come from `Field.get` after `create()`.

kyzooghost (dismissed):

Othryades:
@thedarkjester @kyzooghost Could I get an approval?

Othryades:
@thedarkjester  @kyzooghost Could I get an approval?

.github/workflows/besu-plugin-options-docs.yml
```diff
@@ -0,0 +1,222 @@
...
+        with:
+          name: besu-plugin-options-docs-${{ github.sha }}
+          path: scripts/besu-plugin-options/output/
+          if-no-files-found: error
+          retention-days: 7
+
+  publish-doc-linea:
+    if: >-
+      ${{ (github.event_name == 'push' && github.ref == 'refs/heads/main') ||
+      (github.event_name == 'workflow_dispatch' && inputs.publish == true) }}
```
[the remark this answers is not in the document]
Othryades: Closing as not applicable. GitHub documents that the `inputs` context preserves Boolean values; only `github.event.inputs` converts them to strings: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_dispatchinputs. This workflow uses `inputs.publish`, so comparison with `true` is correct.

linea-besu/plugins/besu-plugin-options-docgen/src/main/java/net/consensys/linea/besu/docgen/CliOptionsManifestGenerator.java
```diff
@@ -0,0 +1,518 @@
...
+        defaultDisplay = null;
+        defaultResolved = false;
+      }
+
+      String description = descriptionRaw;
+      if (description.contains("${DEFAULT-VALUE}") && defaultDisplay != null) {
+        description = description.replace("${DEFAULT-VALUE}", defaultDisplay);
+      }
+      // Docs have a Default column; drop picocli-help "(default: …)" from the description text.
+      description = stripEmbeddedDefault(description);
```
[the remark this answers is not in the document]
Othryades: Closing as already handled by Picocli. `OptionSpec.description()` returns interpolated descriptions, including completion candidates. The generator output from [58ee55f08](https://github.com/LFDT-Lineth/lineth-monorepo/blob/58ee55f080830d13af14b7c091ec6d581f1ab150/scripts/besu-plugin-options/output/linea-besu-plugin-options.json#L721-L723) contains `SEQUENCER, RPC, P2P`; the compressor candidates are also expanded, and the generated report has an empty `unresolvedTokens` list.

thedarkjester (approved):

PR #3552, opened 2026-07-13
feat(jvm-libs): unify signing behind an opaque Signer contract in signer-interfaces

nadeemb53:
Unifies the repo's signing seam into a single backend-neutral contract, `linea.crypto.Signer`, hosted in a new `jvm-libs:linea:core:signer-interfaces` module. This lets alternative KMS/HSM signing backends be developed against a stable interface without living in this repo, while keeping exactly one `Signer` in the codebase.


jvm-libs/linea/core/signer-api/src/main/kotlin/linea/signing/EcdsaSignerService.kt
```diff
@@ -0,0 +1,19 @@
+package linea.signing
+
+import tech.pegasys.teku.infrastructure.async.SafeFuture
+
+/**
+ * A secp256k1 signing service whose private key never enters this process (remote KMS/HSM).
+ */
+interface EcdsaSignerService {
```
fluentcrafter: This feels more like a Client than a Service, nevertheless in this case I would just remove such suffixes it hurts more than it helps. 

Also, we already have a [Signer](https://github.com/LFDT-Lineth/lineth-monorepo/blob/main/coordinator/core-interfaces/src/main/kotlin/linea/crypto/Signer.kt) interface. We cannot have 2, it defeats the purpose.

If we go with this new interface, we should migrate the code to it.

jvm-libs/linea/core/signer-interfaces/build.gradle
fluentcrafter, line 1: folder name shall be `signer-interfaces` to align with other modules.

jvm-libs/linea/core/signer-api/src/main/kotlin/linea/signing/SignatureRS.kt
```diff
@@ -0,0 +1,15 @@
+package linea.signing
+
+import java.math.BigInteger
+
+/**
+ * A canonical secp256k1 ECDSA signature. Canonical means low-s (EIP-2): s <= n/2, so each message
+ * has a single valid signature encoding and downstream recovery-id derivation is well-defined.
+ * Backends must normalize before constructing this type; the constructor enforces it.
+ */
+data class SignatureRS(val r: BigInteger, val s: BigInteger) {
```
fluentcrafter: As already raised before, if this is specific for Secp256k1 so it should be reflected in the name, given that it does no.

jvm-libs/linea/core/signer-api/src/main/kotlin/linea/signing/EcdsaSignerService.kt
Filter94: Another note is that `EcdsaSigner`'s scope looks too narrow. We want a `Signer` to `verify` signatures. User doesn't have to be aware of the R and S and if we change the curve or the signature algorithm, business logic also has to stay the same

nadeemb53: a single Signer interface like this then?
```kotlin
interface Signer {
  fun publicKey(): ByteArray
  fun sign(bytes: ByteArray): SafeFuture<ByteArray>
}
```

coordinator/clients/web3signer-client/src/main/kotlin/linea/crypto/Web3SignerRestClient.kt
```diff
@@ -8,36 +8,38 @@ import io.vertx.ext.web.client.impl.HttpResponseImpl
...
-      client.post(path, buffer).get().map {
-        it as HttpResponseImpl<*>
-        it.body().toString()
-      }
+    return client.post(path, buffer).thenApply { response ->
+      when (val body = response.map { (it as HttpResponseImpl<*>).body().toString() }) {
+        is Ok -> {
+          val signature = body.value.decodeHex()
+          val rSize = 32
+          val sSize = 32
```
Filter94: Minor: Can we import this constant from somewhere else? If not, can we at least define them statically so they're not instantiated on every call? Also, it's clearly unnecessary to have separate constants for rSize and sSize

coordinator/ethereum/test-utils/src/main/kotlin/linea/crypto/ECKeypairSigner.kt
```diff
@@ -1,11 +1,16 @@
...
+import tech.pegasys.teku.infrastructure.async.SafeFuture
 
 class ECKeypairSigner(private val keyPair: ECKeyPair) : Signer {
-  override fun sign(bytes: ByteArray): Pair<BigInteger, BigInteger> {
-    val signature = keyPair.sign(bytes)
-    return signature.r to signature.s
+  override fun publicKey(): ByteArray = Numeric.toBytesPadded(keyPair.publicKey, 64)
+
+  override fun sign(bytes: ByteArray): SafeFuture<ByteArray> {
+    // web3j's ECKeyPair.sign signs its input directly, so hash here to honor the Signer contract.
```
Filter94: Minor: I'd rather remove this comment

jvm-libs/linea/core/signer-interfaces/src/main/kotlin/linea/crypto/Secp256k1Signature.kt
```diff
@@ -0,0 +1,35 @@
+package linea.crypto
+
+import java.math.BigInteger
+
+/**
+ * A canonical secp256k1 ECDSA signature. Canonical means low-s (EIP-2): s <= n/2, so each message
+ * has a single valid signature encoding and downstream recovery-id derivation is well-defined.
+ * Backends must normalize before constructing this type; the constructor enforces it.
+ */
+data class Secp256k1Signature(val r: BigInteger, val s: BigInteger) {
```
Filter94, line 10: I know it was an issue before your changes, but I feel like we're missing a layer of abstraction here. If the signer returns plain ByteArray:
1. User of the `Signer` interface has to be aware of the intricacies of the elliptic curve used by the signer
2. Type safety binding the signer to the correct signature verification algorithm for its output signatures is missing

jvm-libs/linea/core/signer-interfaces/src/main/kotlin/linea/crypto/Secp256k1Signature.kt
```diff
@@ -0,0 +1,35 @@
...
+  /** Encodes as the 64-byte r || s wire format, each component left-padded to 32 bytes. */
+  fun toBytes(): ByteArray = padTo32(r) + padTo32(s)
+
+  companion object {
+    const val SIZE_BYTES = 64
+
+    fun fromBytes(bytes: ByteArray): Secp256k1Signature {
+      require(bytes.size == SIZE_BYTES) { "signature must be $SIZE_BYTES bytes (r || s), got ${bytes.size}" }
+      return Secp256k1Signature(
+        r = BigInteger(1, bytes.copyOfRange(0, 32)),
```
Filter94: `32` used here are rSize and sSize that are not reused here, but hardcoded

jvm-libs/linea/core/signer-interfaces/src/main/kotlin/linea/crypto/Secp256k1Signature.kt
nadeemb53: The reason sign returns ByteArray is actually your earlier point: if we change the curve or algorithm, business logic shouldn't change.

Filter94: Secp256k1Signature is in the interfaces package, meaning that the user is aware of the specific elliptic curve. It means that a change in the elliptic curve will inevitably lead to changes in the business logic. The way to overcome this is to introduce an interface allowing verification of a signature

nadeemb53: I could move this into a new package like `jvm-libs/linea/core/crypto-secp256k1` next to signer-interfaces but I would rather keep `Secp256k1Signature` in signer-interfaces as the canonical wire format and revisit when a second curve or a verifying consumer actually appears. Open to thoughts on this.

jvm-libs/linea/core/signer-api/src/main/kotlin/linea/signing/EcdsaSignerService.kt
nadeemb53: changed the Signer interface to be like this after discussion:
```kotlin
interface Signer<T> {
  fun publicKey(): ByteArray
  fun sign(bytes: ByteArray): SafeFuture<T>
}
```
where [T] is the signature type the backend produces

jvm-libs/linea/core/signer-interfaces/src/main/kotlin/linea/crypto/Secp256k1Signature.kt
nadeemb53: as discussed, making Signer generic over its signature type was agreed

Filter94 (approved):

fluentcrafter (approved):

PR #3555, opened 2026-07-13
feat(misc): Added details on GP VK anchoring and update management

Filter94:
This PR implements issue(s) #

### Checklist

  PR.


rollup_spec/src/rollup_spec/rollup.py
```diff
@@ -464,15 +493,30 @@ def run_rollup_guest(rollup_input: RollupProofPrivateInput) -> RollupProof:
     )
 
 
-def verify_l2_execution_proof(proof: L2ExecutionProof) -> None:
+def recursive_stark_verify(program_vk: Hash32, proof: bytes) -> None:
```
Filter94, line 513: @ivokub Do we know the exact signature of the verification precompile function? Should it be the same as `recursive_stark_verify` + the public input hash?

rollup_spec/tests/test_proof_io_v1.py
```diff
@@ -442,6 +461,10 @@ def test_decode_aggregation_request_maps_all_fields() -> None:
     assert int(proof.public_inputs.end_block_number) == 1000520
     assert proof.l2_l1_roots == [Hash32(bytes([0x77]) * 32), Hash32(bytes([0x88]) * 32)]
     assert proof.filtered_addresses == [Address(bytes([0x01]) * 20)]
+    # §ProgramVK anchoring: the rollup proof's own VK, plus its single combined
+    # program_vks list (here the exec VK it verified).
+    assert proof.program_vk == _ROLLUP_VK
+    assert proof.public_inputs.program_vks == [_EXEC_VK]
```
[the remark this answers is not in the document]
Filter94: We don't have a legacy usage and adding it to every single field by the same logic doesn't make sense to me

rollup_spec/src/rollup_spec/l2_execution.py
```diff
@@ -286,13 +292,22 @@ class L2ExecutionProof:
...
+    VK). It is populated by the rollup request codec; the zero-hash default is
+    only a placeholder for reference-model construction.
     """
     public_inputs: L2ExecutionProofPublicInput
     start_block_number: U64
     proof: bytes = b""
     l2_l1_messages: List[Hash32] = field(default_factory=list)
     tx_froms: List[Address] = field(default_factory=list)
     filtered_addresses: List[Address] = field(default_factory=list)
+    program_vk: Hash32 = ZERO_PROGRAM_VK  # runtime input supplied by the coordinator
```
jonesho: Can you brief a bit more on how does `program_vk` being a runtime input supplied by the coordinator? e.g. the `l2-execution` guest program VK is gonna be a coordinator config, and how does it pass down to l2-execution guest program? There's no corresponding field in l2-execution proof request...

Also why there's no `program_vk` (to denote `rollup-aggregation` guest program VK) added in `FinalizationSubmission`?

Filter94: The cleanest way would be introduce 2 structures:
1. L2ExecutionProof that doesn't have the `program_vk`. It's not sufficient for proof verification, but it's what is actually generated by the prover
2. L2ExecutionProof that has the `program_vk`. Can be passed over to Rollup guest program and verified

Let me separate them and embed one into the other so it doesn't bring this confusion

Filter94: > guest program VK is gonna be a coordinator config, and how does it pass down to l2-execution guest program?

Coordinator is in charge of the upgrade logic and it should define which VK should be executed by the Prover. I think it has to be a part of the Json request, but inside the Prover it will be omitted from the GP input. Let me add it to the json request files

Filter94: > Also why there's no program_vk (to denote rollup-aggregation guest program VK) added in FinalizationSubmission?

That's because Aggregation program's VK is embedded into the PlonkVerifier contract on L1. Currently we're indexing it with an index in the contract, but maybe we should use programVK's hash as an index

Filter94: > Let me add it to the json request files

That's actually wrong. I see that `guestProgramId` is a part of the prover client's request, so it's passed already. I don't know what Id is though and how it's converted to the program VK, maybe it is the program VK. I let you define it

jonesho: > > Let me add it to the json request files
> 
> That's actually wrong. I see that `guestProgramId` is a part of the prover client's request, so it's passed already. I don't know what Id is though and how it's converted to the program VK, maybe it is the program VK. I let you define it

From the last conversation with prover team, `guestProgramId` should be the hash of the program VK, maybe @srinathln7 and @AlexandreBelling can confirm

Filter94: So it's passed outside of the json

rollup_spec/src/rollup_spec/rollup.py
ivokub: When I'm looking at the spec right now, then here `recursive_start_verify` takes as input only `program_vk` and `proof`, where the proof is the raw STARK proof without any public input pinning.

Normally, the verifier must get exactly the same public inputs as the prover gets/provides (depending if we use read_input/write_output). If it is plain public input then the verifier must also get it plain.

I'm trying to follow where the prover is called but it seems it is assumed implicitly (for example in `run_l2_execution_guest` and `run_rollup_aggregation_guest`). In that case, the verifier should get the same public input what is constructed there (i.e. `L2ExecutionProofPublicInput` or `RollupPublicInput`). And both of these are relatively small inputs (few hash digests and U64 values), so could be provided directly.

ivokub: So the `recursive_stark_verify` should take additional typed public input `L2ExecutionProofPublicInput`/`RollupPublicInput`

Filter94: Is my understanding correct that we'll have a tailored function for a GP verification (l2-execution, rollup) which will hide a low-level "precompile" verification call?

Another note on plain PI. If we pass on just a PI hash, it's easier to verify during upgrades than plain PI values. If the set of PI is changing between GP versions (which is possible as protocol evolves), we'll have to maintain 2 signatures of the verifier function too.

rollup_spec/src/rollup_spec/prover_io/schemas/getZkRollupAggregationProofV1.request.schema.json
```diff
@@ -85,14 +92,16 @@
         "startBlockNumber",
         "publicInputs",
         "l2L1Roots",
-        "filteredAddresses"
+        "filteredAddresses",
+        "programVk"
```
jonesho, line 96: @Filter94 to simplify thing, is it possible to provide the `rollup` programVk right away from the `getZkRollupProofV1.response.json` such that the coordinator doesn't need to memorize the rollupProof programVk and provide it when assembling the `getZkRollupAggregationProofV1.request.json`, or I'm missing something that would prevent us from doing that?

rollup_spec/src/rollup_spec/prover_io/schemas/getZkRollupProofV1.request.schema.json
```diff
@@ -116,15 +116,17 @@
         "publicInputs",
         "l2L1Messages",
         "txFroms",
-        "filteredAddresses"
+        "filteredAddresses",
+        "programVk"
```
jonesho, line 120: @Filter94 similar as above, is it possible to provide the `l2-execution` programVk right away from the `getZkL2ExecutionProofV1.response.json` such that the coordinator doesn't need to memorize the`l2-execution` programVk and provide it when assembling the `getZkRollupProofV1.request.json`?

rollup_spec/src/rollup_spec/prover_io/schemas/getZkRollupAggregationProofV1.request.schema.json
Filter94: I think it might be a bad idea, because Prover depending on the Program VK may choose a json serialization strategy

rollup_spec/src/rollup_spec/rollup.py
ivokub: From verifier point of view, the public input is contextless -- it is just an array of field elements. However, what it checks is that some particular column at some particular row has the value defined in the public input array.

Now the question rather is, how are the columns/rows determined. This is actually defined by the R5 interpreter not by the guest program it runs. So it does make sense indeed that we have only a single public input (a hash) and it is up to the guest program itself to represent it. Then we would have completely guest-program agnostic circuits and verifier.

Imo, for me this motivates, as you suggest, to just use hash of public input as input to verifier (in addition to programVK and proof, of course)

ivokub: But in this case imo same should happen when proving the execution of the guest. But this is still being designed afaik -- perhaps indeed how it will work in practise is that zkvm executor outputs the full public input, but internally it only constraints the hash to it. And then the hashing consistency is handled in the R5 interpreter already.

Filter94: cc: @Soleimani193

jonesho (approved):

PR #3565, opened 2026-07-14
feat(prover-ray): fri documentation

arijitdutta67:
This PR implements issue(s) #

### Checklist

  PR.


AlexandreBelling:
There should already be a doc in here: https://github.com/LFDT-Lineth/lineth-monorepo/tree/main/prover-ray/docs

Could you consolidate it instead?

The rest of the changes in the comments are good to me however.

arijitdutta67:
The existing doc is already good, this can only serve as additional reference. We can reconsider opening it once everything is settled and matured a bit.

PR #3573, opened 2026-07-15
refactor(prover-ray): Aux leaves as pairs

Tabaie:
This PR simplifies FRI by unifying the top and auxiliary trees. It restructures FRI so auxiliary levels sit one level up the commitment tree as conjugate pairs, folded a round later exactly like the deepest leaves, unifying the aux and top-level paths and simplifying branches/trees.

* Removes the separate `alphaDeep` challenge: each level's alphaDeep is the square of its own introduction-round fold challenge.
* `FinalPoly` sent as coefficients, not a codeword; new `LogFinalPolySize` option to stop folding early. This serves as a soundness fix as previously the final codeword's low-degree bound was not enforced.
* `checkFolds` computes fold points by one exponentiation + per-round squaring.
* `Params` sizes are now log2-based (`LogCodewordSize`/`LogPlainTextSize`)


ivokub (approved):
Otherwise looks good, but there is a Copilot consideration (created issue https://github.com/LFDT-Lineth/lineth-monorepo/issues/3642) and another one: https://github.com/LFDT-Lineth/lineth-monorepo/issues/3641

I think it is good to merge now as is, lets start implementing the verifier in Zig and then fix in a follow-up PR.

AlexandreBelling (approved):

YaoJGalteland (approved):
The PR looks good; the optimizations align the commitment structure with the fold structure.

- The fold operates on conjugate pairs → commit pairs, not individual rows → ConjugateLeaf removed
- The fold challenge is the natural batching randomness → batch inside the fold step, not before it → alphaDeep removed

PR #3576, opened 2026-07-16
feat(ci): updated the release auto changelog and version bump process…

jonesho:
… to be both commit scope and path gated

This PR implements issue(s) #

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> Changes how semver bumps and release tags are computed across all components; misconfigured scopes or regex could skip legitimate commits or suppress releases until commit messages are fixed.
> 
> **Overview**
> Release **version bumps and changelogs** now require commits to match **both** a component’s conventional-commit **scopes** and its **`--include-path`** filters, instead of path-only (or global) git-cliff rules.
> 
> A shared **`cliff.template.toml`** replaces root **`cliff.toml`**. **`run-git-cliff-commands`** takes a required **`scopes`** input, renders the template by substituting **`@@SCOPES@@`**, and uses that config for bump/changelog steps. Reusable release, changelog-upload, milestone metadata, and milestone prepend workflows pass per-component scope alternations (e.g. `coordinator|deps|misc`); the milestone prepend loop switches entry parsing to **`#`** delimiters and renders the template per component.
> 
> **README** documents how to reproduce CI locally with the same scopes and paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6accc4d52f63666524742be5c8565881310126c0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


coordinator/cliff.toml
Filter94, line 1: Is all of this copypaste necessary and different from the default toml?

Othryades:
As per the investigation on this I made yesterday, there may be 2 issues
First, linea-besu/package/cliff.toml accepts only commits labelled linea-besu, tracer, or sequencer. The actual Bouncy Castle update was: chore(deps): align bouncycastle catalog pin to 1.84
Because it is labelled deps, the new configuration would skip it completely. It would not appear in the Besu changelog and would not trigger a Besu version bump IIUC. 
Therefore, it prevents the incorrect major bump but also hides small genuine dependency update that caused the release.
Second, the PR changes the root cliff.toml, which calculates the version and changelog for the combined Linea milestone release. With the new rules, real changes labelled deps, misc, or ci would be ignored I think.

jonesho:
> As per the investigation on this I made yesterday, there may be 2 issues First, linea-besu/package/cliff.toml accepts only commits labelled linea-besu, tracer, or sequencer. The actual Bouncy Castle update was: chore(deps): align bouncycastle catalog pin to 1.84 Because it is labelled deps, the new configuration would skip it completely. It would not appear in the Besu changelog and would not trigger a Besu version bump IIUC. Therefore, it prevents the incorrect major bump but also hides small genuine dependency update that caused the release. Second, the PR changes the root cliff.toml, which calculates the version and changelog for the combined Linea milestone release. With the new rules, real changes labelled deps, misc, or ci would be ignored I think.

@Othryades I see your point, I don't have a perfect solution for now, but I think it makes sense to add `deps` into the scope of `linea-besu`, such that commits like "The actual Bouncy Castle update" would not be skipped in version bump and changelog update of `linea-besu-package` release, however, this would also mean any future commits with `deps` in scopes that touched `lib.versions.toml` would be included as well, given that the painpoint is that there's no way to tell which library version update in `lib.versions.toml` is related to `linea-besu-package`, but imo it's still better than missing the positive updates in the release.

coordinator/cliff.toml
jonesho: Originally I was trying to keep the usage by having a seperate cliff toml file for each component, as for easier to invoke git cliff command, especially for local, and for potential future case where each component might have different git-cliff rules. But I agree, for now, given that all component shares the same rules with different scopes, we can consolidate all of them into one template toml file and inject the corresponding scopes in workflow.

Othryades (approved):

PR #3597, opened 2026-07-20
feat(coordinator): support custom transaction signers

nadeemb53:
- add a coordinator `SignerFactory` returning the shared digest `Signer` contract
- use one Web3j transaction-signing path for local keys, Web3Signer, and custom backends
- send digests to Web3Signer with `applyHash=false` and remove the old custom extension seam

E2E requires a Web3Signer image containing Consensys/web3signer#1213. The current `26.3.0` pin hashes the digest again and cannot produce a recoverable transaction signature.


coordinator/app/src/main/kotlin/linea/coordinator/app/SignerFactory.kt
```diff
@@ -0,0 +1,148 @@
...
+ * coordinator's transaction encoding path.
+ */
+fun interface SignerFactory {
+  fun create(
+    vertx: Vertx,
+    signerConfig: SignerConfig,
+  ): Signer<Secp256k1Signature>
+}
+
+/** Creates the OSS local-key and Web3Signer backends. */
```
Filter94: Is OSS notion relevant here?

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/SignerConfigToml.kt
```diff
@@ -47,10 +54,22 @@ data class SignerConfigToml(
       return when (this) {
         WEB3J -> SignerConfig.SignerType.WEB3J
         WEB3SIGNER -> SignerConfig.SignerType.WEB3SIGNER
+        CUSTOM -> SignerConfig.SignerType.CUSTOM
       }
     }
   }
 
+  data class CustomConfig(
+    @param:ConfigDoc("Logical signer name resolved by the injected signer factory.")
```
Filter94, line 63: Enumeration of available `SignerType` options should be here too

coordinator/clients/web3signer-client/src/main/kotlin/linea/crypto/Web3SignerRestClient.kt
```diff
@@ -40,5 +47,7 @@ class Web3SignerRestClient(
 
   companion object {
     const val WEB3SIGNER_SIGN_ENDPOINT = "/api/v1/eth1/sign/"
+    private const val DIGEST_SIZE_BYTES = 32
```
Filter94, line 50: Can we reuse this constant with `LocalKeySigner` somehow?

Filter94 (dismissed):

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/SignerConfigToml.kt
nadeemb53: The available values are already listed in the type field’s `ConfigDoc: WEB3J, WEB3SIGNER, and CUSTOM`. Is there another generated-doc location you had in mind?

coordinator/clients/web3signer-client/src/main/kotlin/linea/crypto/Web3SignerRestClient.kt
nadeemb53: Both constants are private and live in separate modules. Sharing them would require exposing a public constant purely to avoid duplicating the protocol value 32. I’d prefer to keep them local unless digest size becomes part of the shared signer contract.

coordinator/app/src/main/kotlin/linea/coordinator/app/SignerFactory.kt
nadeemb53: I'll update the wording

coordinator/app/src/main/kotlin/linea/coordinator/config/v2/toml/SignerConfigToml.kt
Filter94: No, I think I misinterpreted the code here. What I meant is that it would be good to have the acceptable options for the name listed here. But since it's not about `SignerType` and names are resolved at runtime, it's probably not possible. Maybe it has to be a responsibility of the `SignerFactory` to list available options on failure

Filter94 (approved):

PR #3598, opened 2026-07-20
refactor(prover-ray): refactor backend core and add unit tests

gusiri:
This PR :
- Add blobs_test.go: tests covering sszBlobs framing, ELF section extraction, the pub-input byte layout, and BuildZkcInputs integration. 
- Extract loadELFBlobs and encodeInputs as standalone functions to make the encoding logic directly testable. encodeInputs builds the pub-input map directly; no JSON round trip in the production path.
- Rename blob to memoryBlob to avoid confusion with EIP-4844 blobs.
- Precompute ELF blobs once in Core.New() instead of re-parsing the ELF on every Prove call.
- Compile the test circuit from zkc_01.zkc at test time and drop the stale checked-in zkc_01.bin/zkc_02.bin, which no longer decode after the zkc update in #3580.
- 

### Checklist

  PR.


prover-ray/backend/core.go
```diff
@@ -86,8 +93,11 @@ func (c *Core) Prove(ctx context.Context, job Job) Result {
 }
 
 // buildInputs converts a Job's Payload into the three ZkC pub-input blobs.
+// ELF blobs are pre-extracted in [New] and reused across calls; only the SSZ
+// blobs differ per job.
 func (c *Core) buildInputs(job Job) (map[string][]byte, error) {
-	return BuildZkcInputs(c.elf, decodePayload(job), c.cfg.inOrigin())
+	blobs := append(append([]memoryBlob{}, c.elfBlobs...), sszBlobs(c.cfg.inOrigin(), decodePayload(job))...)
+	return encodeInputs(blobs, c.elfEntry)
```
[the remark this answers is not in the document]
gusiri: fixed

prover-ray/backend/core.go
```diff
@@ -18,18 +18,19 @@ var ErrNotImplemented = errors.New("not yet implemented")
...
 }
 
 // New loads the circuit binary and the guest ELF, calls [zkcdriver.NewZkCDriver]
 // to define all columns and constraints, and returns a [Core] ready to prove.
 //
 // Compiler passes (rangecheck → lookup → logderiv → localvanishing → global)
-// and wiop.Materialize are not yet wired — they must be added before the
-// system can produce cryptographically sound proofs (see backend-overview.md §4).
+// and wiop.Materialize are not yet wired. They must be added before the
+// system can produce sound proofs (see wiki backend-overview.md §4).
```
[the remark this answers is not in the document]
gusiri: fixed

prover-ray/backend/core.go
```diff
@@ -120,7 +130,7 @@ func EncodeStatelessInput(_ []byte) ([]byte, error) {
 // SerializeProof encodes a wiop.Proof into the wire bytes the coordinator
 // expects in the "proof" field of the response.
 //
-// Wire format not yet decided (backend-overview.md §6).
+// Wire format not yet decided (wiki backend-overview.md §6).
```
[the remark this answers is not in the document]
gusiri: fixed

prover-ray/backend/blobs.go
```diff
@@ -27,36 +28,49 @@ type blob struct {
...
 //
-// elfBytes is the raw guest ELF. sszInput is the raw SSZ StatelessInput (not
-// yet framed — BuildZkcInputs adds the [u64 LE len] prefix). inOrigin is the
-// guest RAM address where the SSZ input is placed (use [DefaultINOrigin]).
+// elfBytes is the raw guest ELF. sszInput is the framed StatelessInput the
+// guest reads at _in_start: the 0x0001 schema id followed by the SSZ body
+// (the guest's deserializer strips the schema id). BuildZkcInputs prepends
+// only the [u64 LE len] prefix; it neither adds nor strips the schema id.
+// inOrigin is the guest RAM address where the input is placed
+// (use [DefaultINOrigin]).
```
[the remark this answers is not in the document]
gusiri: This is wrong. The guest's own sample input (riscv-guests/l2-execution/test/testdata/stateless_input.ssz) starts with 00 01, and the spec decoder (decode_stateless_input_ssz) rejects input without the 0x0001 prefix. Unframed SSZ makes the guest fail to decode.

prover-ray/backend/blobs.go
```diff
@@ -84,38 +98,38 @@ func elfBlobs(ef *elf.File) ([]blob, error) {
...
+// buildJSON encodes memory blobs and entryPoint into the compact JSON hex format that
+// zkc_util.ParseJsonInputFile decodes. Blob boundaries are separated by "____"
+// (four underscores); offset and size within one entry use a single "_".
+func buildJSON(memBlobs []memoryBlob, entryPoint uint64) []byte {
+	offsetSizeParts := make([]string, len(memBlobs))
+	dataParts := make([]string, len(memBlobs))
+	for i, b := range memBlobs {
 		offsetSizeParts[i] = fmt.Sprintf("%016x_%016x", b.offset, len(b.data))
 		dataParts[i] = hex.EncodeToString(b.data)
 	}
```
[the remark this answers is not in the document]
gusiri: This is wrong. len() is never negative, %016x pads identically everywhere, and the arith reference printJson uses bare len(...) with no cast; we stay byte-for-byte faithful to it.

prover-ray/backend/blobs.go
```diff
@@ -84,38 +98,38 @@ func elfBlobs(ef *elf.File) ([]blob, error) {
...
-
-	// Blob boundaries are separated by "____" (four underscores); individual
-	// fields within one entry use a single "_" as a visual separator.
-	json := []byte(`{` +
-		`"entry_point_and_blobs_count":"0x` + fmt.Sprintf("%016x_%016x", entryPoint, len(blobs)) + `",` +
+	return []byte(`{` +
+		`"entry_point_and_blobs_count":"0x` + fmt.Sprintf("%016x_%016x", entryPoint, len(memBlobs)) + `",` +
```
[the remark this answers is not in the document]
gusiri: This is wrong. len() is never negative.

prover-ray/backend/core.go
```diff
@@ -41,6 +42,11 @@ func New(cfg Config) (*Core, error) {
 		return nil, fmt.Errorf("reading guest ELF %q: %w", cfg.GuestELFPath, err)
 	}
 
+	precomputed, entry, err := loadELFBlobs(elfBytes)
+	if err != nil {
+		return nil, fmt.Errorf("extracting ELF blobs: %w", err)
+	}
+
 	sys := wiop.NewSystemf("linea-riscv")
```
ivokub: would make "linea-riscv" as a package level constant

prover-ray/backend/blobs.go
```diff
@@ -84,38 +98,38 @@ func elfBlobs(ef *elf.File) ([]blob, error) {
...
-// elf_to_json_gen/main.go produces, ensuring compatibility with the
-// existing zkcdriver input reader.
-func encodeInputs(blobs []blob, entryPoint uint64) (map[string][]byte, error) {
-	offsetSizeParts := make([]string, len(blobs))
-	dataParts := make([]string, len(blobs))
-	for i, b := range blobs {
+// buildJSON encodes memory blobs and entryPoint into the compact JSON hex format that
+// zkc_util.ParseJsonInputFile decodes. Blob boundaries are separated by "____"
+// (four underscores); offset and size within one entry use a single "_".
+func buildJSON(memBlobs []memoryBlob, entryPoint uint64) []byte {
```
ivokub: Do we need the buildJSON at all? Later on we will call binFile.Trace which takes `map[string][]byte` as input, so we would be decoding the json again into map. I don't think we need it.

ivokub: You can have a look at `zkcdriver_test.go` on main how we would provide the inputs to the zkc tracer.

prover-ray/backend/blobs_test.go
```diff
@@ -0,0 +1,419 @@
+package backend
```
ivokub, line 1: Do the tests come from existing elf-to-json tests? A lot of buildInputs tests just test that we have correct buildJSON implementation which is essentially just JSON encoding.

ivokub (changes requested):
Have a look - imo we don't need the JSON encoding, it just adds overhead ssz -> JSON -> map[string][]byte. I recommend having a look at zkcdriver_test.go on main (and also zkcdriver/example_test.go) how to provide inputs to zkc tracer

prover-ray/backend/core.go
gusiri: I removed the JSON part now.
I added that JSON originally to test against the arith team's elf_to_json_gen output. since the map's byte layout is only defined implicitly by main.zkc.
I kept one test (TestEncodeInputs_MatchesReferenceJSON) that checks the direct map is byte-identical to ParseJsonInputFile on the reference JSON.

gusiri: >would make "linea-riscv" as a package level constant

updated.

prover-ray/backend/blobs_test.go
gusiri: No, elf_to_json_gen has no tests, so I wrote these to cover the JSON format. Now that the JSON is gone, those assertions are replaced with assertions on the raw map bytes.

prover-ray/backend/blobs.go
gusiri: Removed JSON part.

prover-ray/backend/job.go
```diff
@@ -28,7 +28,7 @@ type Job struct {
 	// SSZ-encoded StatelessInput that the guest reads at _in_start.
 	//
 	// Multi-block conflation encoding is not yet decided (open question #1
```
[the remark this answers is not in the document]
gusiri: Updated.

prover-ray/backend/blobs.go
```diff
@@ -84,38 +94,48 @@ func elfBlobs(ef *elf.File) ([]blob, error) {
 	return result, nil
 }
 
-// sszBlobs splits ssz into the two blobs that linea_zkvm_io expects at
+// sszBlobs splits ssz into the two memory blobs that linea_zkvm_io expects at
 // _in_start: an 8-byte LE length prefix followed by the raw SSZ payload.
 // The split matches elf_to_json_gen's sszInputBlobs (commit 09fcdb42).
-func sszBlobs(inOrigin uint64, ssz []byte) []blob {
+func sszBlobs(inOrigin uint64, ssz []byte) []memoryBlob {
```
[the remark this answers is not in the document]
gusiri: Updated.

prover-ray/backend/blobs_test.go
```diff
@@ -0,0 +1,429 @@
...
+
+func TestSszBlobs_PayloadAtInOriginPlus8(t *testing.T) {
+	ssz := []byte{0xAA, 0xBB, 0xCC}
+	got := sszBlobs(DefaultINOrigin, ssz)
+	require.Len(t, got, 2)
+
+	// Second memory blob: raw SSZ bytes at inOrigin+8.
+	assert.Equal(t, DefaultINOrigin+8, got[1].offset, "payload memory blob offset must be inOrigin+8")
+	assert.Equal(t, ssz, got[1].data, "payload memory blob must contain the raw SSZ bytes")
+}
```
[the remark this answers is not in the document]
gusiri: Fixed.

prover-ray/backend/core.go
```diff
@@ -85,9 +92,14 @@ func (c *Core) Prove(ctx context.Context, job Job) Result {
 	}
 }
 
-// buildInputs converts a Job's Payload into the three ZkC pub-input blobs.
-func (c *Core) buildInputs(job Job) (map[string][]byte, error) {
-	return BuildZkcInputs(c.elf, decodePayload(job), c.cfg.inOrigin())
+// buildInputs converts a Job's Payload into the three ZkC pub-input memory blobs.
+// ELF memory blobs are pre-extracted in [New] and reused across calls; only the
+// SSZ memory blobs differ per job.
```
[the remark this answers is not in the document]
gusiri: Fixed.

prover-ray/backend/core.go
```diff
@@ -41,7 +45,12 @@ func New(cfg Config) (*Core, error) {
 		return nil, fmt.Errorf("reading guest ELF %q: %w", cfg.GuestELFPath, err)
 	}
 
-	sys := wiop.NewSystemf("linea-riscv")
+	precomputed, entry, err := loadELFBlobs(elfBytes)
+	if err != nil {
+		return nil, fmt.Errorf("extracting ELF blobs: %w", err)
+	}
```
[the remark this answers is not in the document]
gusiri: Applied.

prover-ray/backend/blobs_test.go
gusiri: Removed JSON and related tests.

prover-ray/backend/blobs.go
```diff
@@ -27,36 +24,49 @@ type blob struct {
...
+// only the [u64 LE len] prefix; it neither adds nor strips the schema id.
+// inOrigin is the guest RAM address where the input is placed
+// (use [DefaultINOrigin]).
 func BuildZkcInputs(elfBytes, sszInput []byte, inOrigin uint64) (map[string][]byte, error) {
-	ef, err := elf.NewFile(bytes.NewReader(elfBytes))
+	memBlobs, entry, err := loadELFBlobs(elfBytes)
 	if err != nil {
-		return nil, fmt.Errorf("parsing guest ELF: %w", err)
+		return nil, err
 	}
```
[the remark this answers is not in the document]
gusiri: Leaving as-is. loadELFBlobs already wraps its failures with descriptive context (parsing guest ELF, no loadable sections), and BuildZkcInputs has no path or other context to add

prover-ray/backend/blobs.go
```diff
@@ -27,36 +24,49 @@ type blob struct {
...
-// guest RAM address where the SSZ input is placed (use [DefaultINOrigin]).
+// elfBytes is the raw guest ELF. sszInput is the framed StatelessInput the
+// guest reads at _in_start: the 0x0001 schema id followed by the SSZ body
+// (the guest's deserializer strips the schema id). BuildZkcInputs prepends
+// only the [u64 LE len] prefix; it neither adds nor strips the schema id.
+// inOrigin is the guest RAM address where the input is placed
+// (use [DefaultINOrigin]).
 func BuildZkcInputs(elfBytes, sszInput []byte, inOrigin uint64) (map[string][]byte, error) {
-	ef, err := elf.NewFile(bytes.NewReader(elfBytes))
+	memBlobs, entry, err := loadELFBlobs(elfBytes)
```
Soleimani193: As I understood, one of the objectives was to decode ELF to blobs only once, but here you are doing it for every job.

prover-ray/backend/blobs.go
```diff
@@ -84,38 +94,50 @@ func elfBlobs(ef *elf.File) ([]blob, error) {
...
-		`"blobs_data":"0x` + strings.Join(dataParts, "____") + `"` +
-		`}`)
+	var dataLen int
+	for _, b := range memBlobs {
+		dataLen += len(b.data)
+	}
+	offsetAndSize := make([]byte, 0, 16*len(memBlobs))
+	data := make([]byte, 0, dataLen)
+	for _, b := range memBlobs {
+		offsetAndSize = binary.BigEndian.AppendUint64(offsetAndSize, b.offset)
```
Soleimani193, line 149: You can even define a struct for ELF that contains blobs and also offsetAndSize and other fields, probably, so that here you just append to the existing fields, like this, the ELF part is processed only once.

prover-ray/backend/blobs.go
gusiri: The real hot path is `Core.buildInputs`, which resues the ELF blobs parsed once in `Core.New`

`BuildZkcInputs` was misleading. It is a test-only helper function. I changed it's name to `buildZkcInputs` and marked it a helper function.

prover-ray/backend/core.go
```diff
@@ -14,22 +14,26 @@ import (
 // ErrNotImplemented is returned by stubs that are not yet wired up.
 var ErrNotImplemented = errors.New("not yet implemented")
 
+// wiopSystemName names the wiop constraint system built in [New].
+const wiopSystemName = "linea-riscv"
```
ivokub: We should use "Lineth" instead of Linea instead now for the OSS components.

prover-ray/backend/blobs.go
gusiri: I added a struct `elfInputs` in `core` for only ELF related parts. This will improve readability and we can add more elf related stuffs to this struct in the future

This won't improve performance though:
The ELF is already processed once. The expensive part (parsing the ELF file) happens a single time in Core.New, and the per-job path just reuses the parsed blobs.

On "define a struct so you just append to the existing fields": 
the reason that doesn't help is that the prover needs the ELF bytes and the per-job SSZ bytes as **one contiguous array** (`blobs_data` is a single []byte). So we can't literally append the SSZ onto a shared, pre-serialized ELF field and hand it over. Each job has to `copy` the ELF bytes into its own array with the SSZ appended anyway.

prover-ray/backend/blobs_test.go
```diff
@@ -0,0 +1,449 @@
...
+		`"entry_point_and_blobs_count":"0x` + fmt.Sprintf("%016x_%016x", entryPoint, len(memBlobs)) + `",` +
+		`"blobs_offset_and_size":"0x` + strings.Join(offsetSizeParts, "____") + `",` +
+		`"blobs_data":"0x` + strings.Join(dataParts, "____") + `"` +
+		`}`)
+}
+
+// TestEncodeInputs_MatchesReferenceJSON pins encodeInputs to the reference
+// wire format: building the map directly must be byte-identical to encoding
+// the elf_to_json_gen JSON and decoding it with zkc's own parser.
+func TestEncodeInputs_MatchesReferenceJSON(t *testing.T) {
```
ivokub: I think the test is redundant -- it tests the json decoding of `zkc_util.ParseJsonInputFile`. But my understanding is that we would never use it and we provide the `map[string][]byte` as input to zkc `Trace` already.

prover-ray/backend/blobs_test.go
```diff
@@ -0,0 +1,449 @@
...
+// helper.
+func compileZKCBin(t *testing.T, srcPath string) string {
+	t.Helper()
+
+	srcBytes, err := os.ReadFile(srcPath)
+	require.NoError(t, err)
+
+	src := source.NewSourceFile(srcPath, srcBytes)
+	zkcField := field.KOALABEAR_16
+	zkcCfg := codegen.DEFAULT_CONFIG.SplitRegisters(true).Quiet(true)
```
ivokub: Register splitting is already default for `DEFAULT_CONFIG`, currently setting it here is no-op.

And newer zkc has removed `Quiet` option afaik, I think with https://github.com/LFDT-Lineth/lineth-monorepo/pull/3575 it will break now.

prover-ray/backend/core.go
gusiri: updated: "lineth-riscv"

prover-ray/backend/core.go
```diff
@@ -14,22 +14,25 @@ import (
...
 // New loads the circuit binary and the guest ELF, calls [zkcdriver.NewZkCDriver]
 // to define all columns and constraints, and returns a [Core] ready to prove.
 //
 // Compiler passes (rangecheck → lookup → logderiv → localvanishing → global)
-// and wiop.Materialize are not yet wired — they must be added before the
-// system can produce cryptographically sound proofs (see backend-overview.md §4).
+// and wiop.Materialize are not yet wired. They must be added before the
+// system can produce sound proofs (see wiki backend-overview.md §4).
 func New(cfg Config) (*Core, error) {
 	binBytes, err := os.ReadFile(cfg.CircuitBinPath)
```
ivokub: We don't need to read binBytes and then `bytes.NewReader`, we can just open the file and then provide `*os.File` below.

prover-ray/backend/blobs_test.go
gusiri: removed `.SplitRegisters(true).Quiet(true)`

prover-ray/backend/blobs_test.go
gusiri: Removed JSON related tests.

prover-ray/backend/blobs.go
```diff
@@ -4,59 +4,82 @@ import (
...
+// [New] and reused for every job; only the per-job SSZ blobs differ.
+type elfInputs struct {
+	blobs []memoryBlob
+	entry uint64
+}
+
+// loadELFInputs parses elfBytes and returns the ELF's memory blobs and entry
+// point. Callers that process many jobs from the same ELF should call this
+// once at startup and cache the result on [Core].
+func loadELFInputs(elfBytes []byte) (elfInputs, error) {
```
ivokub: Also, we don't need to pass bytes here -- the `elf.NewFile` takes as input `io.Reader` so we could pass only `io.Reader` as input and then we can avoid reading everything in-memory and creating new buffer here.

Soleimani193 (dismissed):

prover-ray/backend/core.go
gusiri: `New` now uses `os.Open` to get `binFile` and passes the `*os.File` to `NewZkCDriver`.

ivokub (dismissed):
Small comments here and there, but otherwise looks good. See if you want to fix, otherwise I'm good with merging. I think the `io.Reader` things are minor optimizations though -- there files won't be too big I think (a la tens of MBs top), so it is fine if we duplicate in memory.

prover-ray/backend/blobs.go
gusiri: Done. loadELFInputs now takes io.ReaderAt and New passes the opened *os.File directly

prover-ray/backend/blobs_test.go
```diff
@@ -0,0 +1,451 @@
...
+	}
+
+	ssz1 := []byte{0x01, 0x02}
+	ssz2 := []byte{0xFF, 0xFE}
+
+	inputs1 := c.buildInputs(Job{Payload: ssz1})
+	inputs2 := c.buildInputs(Job{Payload: ssz2})
+
+	assert.NotEqual(t, inputs1["blobs_data"], inputs2["blobs_data"], "different SSZ must produce different blobs_data")
+	assert.Equal(t, inputs1["entry_point_and_blobs_count"], inputs2["entry_point_and_blobs_count"], "same ELF must produce identical entry_point_and_blobs_count")
```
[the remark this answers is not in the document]
gusiri: Updated.

prover-ray/backend/core.go
```diff
@@ -108,9 +118,9 @@ func (c *Core) runProve(
 	return proof, pub, nil
 }
 
-// EncodeStatelessInput SSZ-encodes the coordinator's per-block payload into
-// the byte slice the guest reads at _in_start. The [u64 LE len] frame is
-// added later by [BuildZkcInputs]; this returns raw SSZ only.
+// EncodeStatelessInput encodes the coordinator's per-block payload into the
+// framed StatelessInput the guest consumes: the 0x0001 schema id followed by
+// the SSZ body. [buildZkcInputs] later prepends the [u64 LE len] prefix.
```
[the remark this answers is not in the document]
gusiri: Fixed.

prover-ray/backend/core.go
```diff
@@ -85,9 +90,14 @@ func (c *Core) Prove(ctx context.Context, job Job) Result {
 	}
 }
 
-// buildInputs converts a Job's Payload into the three ZkC pub-input blobs.
-func (c *Core) buildInputs(job Job) (map[string][]byte, error) {
-	return BuildZkcInputs(c.elf, decodePayload(job), c.cfg.inOrigin())
+// buildInputs converts a Job's Payload into the three ZkC pub-input values,
+// keyed by name (see [encodeInputs]). ELF memory blobs are pre-extracted in
+// [New] and reused across calls; only the SSZ memory blobs differ per job.
```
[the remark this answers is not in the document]
gusiri: Fixed.

prover-ray/backend/blobs.go
```diff
@@ -84,38 +107,50 @@ func elfBlobs(ef *elf.File) ([]blob, error) {
...
+// encodeInputs builds the keyed byte map that [zkcdriver.PreReadInputs]
+// expects, one entry per pub-input key:
+//
+//   - "entry_point_and_blobs_count": [8 BE entry point][8 BE blob count]
+//   - "blobs_offset_and_size":       per blob, [8 BE offset][8 BE size]
+//   - "blobs_data":                  all blob bytes concatenated
+//
+// The layout is byte-identical to zkc_util.ParseJsonInputFile applied to the
+// JSON that the reference elf_to_json_gen tool emits, without the JSON round
+// trip; TestEncodeInputs_MatchesReferenceJSON pins the equivalence.
```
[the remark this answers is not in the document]
gusiri: Fixed.

Soleimani193 (approved):

ivokub (approved):

PR #3600, opened 2026-07-20
feat(prover-ray): filesystem job adapter

gusiri:
This PR implements the minimal filesystem adapter and introduces the Runner boundary that a future gateway adapter can reuse.

## What changed

- Added the L2 execution job adapter for `getZkL2ExecutionProofV1`.
- Decode coordinator requests, SSZ-encode payloads, build `backend.Job`, and call the prover.
- Added filesystem polling/claiming, response writing, and request archiving.
- Added V1-shaped success responses and provisional failure responses.
- Added focused request, runner, and filesystem adapter tests.

## Notes

- Multi-block requests and forced transactions are rejected as not implemented.
- Real proof bytes/public inputs are expected from follow-up backend work.

### Checklist

  PR.


prover-ray/backend/jobadapter/request.go
```diff
@@ -0,0 +1,467 @@
...
+}
+
+// DecodeRequest parses a getZkL2ExecutionProofV1 request body and SSZ-encodes
+// each payload's statelessInput, porting proof_io_v1.py::decode_request and
+// _decode_payload: it injects {chainId, forkName} into each payload's
+// statelessInput and converts the already-validated empty executionRequests
+// list into the empty container shape expected by [ssz.EncodeStatelessInput].
+// It also validates and preserves rollupExtension.forcedTransactions for the
+// Runner capability check.
+func DecodeRequest(data []byte) (*Request, error) {
```
srinathln7: Does it make sense to rename as `DecodeL2ExecutionRequest` since the decoding logic applies specifically to L2 execution?

prover-ray/backend/jobadapter/request.go
```diff
@@ -0,0 +1,467 @@
...
+	BlockNumber        uint64
+	FramedSSZ          []byte
+	ForcedTransactions []json.RawMessage
+}
+
+// Request is a decoded getZkL2ExecutionProofV1 request: routing metadata, the
+// range-level chain identity, and one [DecodedPayload] per block. The block
+// range is implied by the payloads (their executionPayload.blockNumber), as in
+// the reference decoder.
+type Request struct {
```
srinathln7: May be rename to `L2ExecutionRequest` similar to the counterpart in `response.go` ?

prover-ray/backend/jobadapter/filesystem/adapter.go
```diff
@@ -0,0 +1,263 @@
...
+	if err != nil {
+		return fmt.Errorf("jobadapter: encoding response for %s: %w", name, err)
+	}
+	if err := writeFileAtomic(filepath.Join(a.responsesDir(), name), data, filePerm); err != nil {
+		return fmt.Errorf("jobadapter: writing response for %s: %w", name, err)
+	}
+	return nil
+}
+
+func writeFileAtomic(path string, data []byte, perm fs.FileMode) error {
```
srinathln7, line 222: May be not necessary at this stage, but in the future, we may consider moving to a generic `util` or some `files` pkg. What do u think?

prover-ray/backend/jobadapter/runner.go
```diff
@@ -0,0 +1,118 @@
...
+	if proverVersion == "" {
+		return nil, fmt.Errorf("jobadapter.NewRunner: proverVersion must be set")
+	}
+	return &Runner{prover: prover, proverVersion: proverVersion}, nil
+}
+
+// Run runs one raw request body. Decode, validation, and proof failures all map
+// to a response body rather than a returned error.
+func (r *Runner) Run(ctx context.Context, id string, data []byte) RunResult {
+	req, err := DecodeRequest(data)
```
srinathln7: Can we design the interface so that the runner remains as generic as possible, rather than being tightly coupled to L2 execution requests? I understand this may not be a strong requirement at this stage since we're primarily defining the skeleton, but I think it's worth noting it down for future.

prover-ray/backend/jobadapter/runner.go
```diff
@@ -0,0 +1,118 @@
...
+	payload := req.Payloads[0]
+	if len(payload.ForcedTransactions) != 0 {
+		return failedRunResult(id, FailureCodeInvalidInput, fmt.Errorf(
+			"forced transactions are not supported (got %d): %w",
+			len(payload.ForcedTransactions), backend.ErrNotImplemented))
+	}
+
+	result := r.prover.Prove(ctx, backend.Job{
+		ID:         id,
+		Type:       backend.ProofTypeL2Execution,
```
srinathln7: Same point as on R74. May be we accept the job type as param? What do u think?

prover-ray/backend/jobadapter/request.go
gusiri: I renamed it to `DecodeL2ExecutionRequest`

prover-ray/backend/jobadapter/request.go
gusiri: I renamed `Request` to `L2ExecutionRequest`

srinathln7 (approved):

PR #3609, opened 2026-07-21
chore(coordinator): riscv blob data skeleton

fluentcrafter:
This PR implements issue(s) #

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> Changes coordinator blob persistence abstractions and how rollup proof data is stored in Postgres, but the V1 path is preserved and V2 appears test-only for now.
> 
> **Overview**
> Introduces **RISC-V–oriented blob persistence** alongside the existing `BlobRecord` path, without changing current production wiring.
> 
> New domain types **`BlobRecordV2`** and **`BlobData`** model block intervals with parent/end shnarfs and a list of per-blob hash, compressed payload, and batch counts. **`BlobsRepository` / `BlobsDao`** are generalized to **`BlobsRepositoryG<T>` / `BlobsDaoG<T>`**, with **`BlobsRepositoryV2`** and **`BlobsDaoV2`** for V2 records.
> 
> Postgres access is refactored: shared SQL and CRUD live in **`BlobsPostgresDaoG`**, while **`BlobsPostgresDao`** only maps V1 rows and **`BlobsPostgresDaoV2`** maps V2 by serializing **`BlobsInfo`** into the existing `blob_compression_proof` JSON column (with empty placeholders for legacy `blob_hash` / `expected_shnarf` until a later schema migration). Matching **`BlobsRepositoryImplV2`** and **`RetryingBlobsPostgresDaoV2`** wrap the new DAO.
> 
> Integration tests move common blob DAO scenarios into **`BlobsPostgresDaoTestBase`**, with **`BlobsPostgresDaoV2Test`** exercising the V2 mapper; a **`createBlobRecordV2`** test fixture supports that.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1d5fcd62c6fd7e72232781a15927ccad06613989. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


coordinator/persistence/src/main/kotlin/linea/persistence/conflation/BlobsDao.kt
```diff
@@ -1,22 +1,27 @@
...
 
-interface BlobsDao {
-  fun saveNewBlob(blobRecord: BlobRecord): SafeFuture<Unit>
+interface BlobsDaoG<T> {
+  fun saveNewBlob(blobRecord: T): SafeFuture<Unit>
 
   fun getConsecutiveBlobsFromBlockNumber(
     startingBlockNumberInclusive: ULong,
     endBlockCreatedBefore: Instant,
   ): SafeFuture<List<BlobRecord>>
```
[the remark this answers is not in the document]
gauravahuja: This is relevant comment

coordinator/persistence/src/main/resources/db/V002__add_blob_and_aggregation_schema.sql
```diff
@@ -16,13 +16,13 @@ create table if not exists blobs (
...
-    blob_hash                      varchar(256),
+    blob_hash                      varchar(256), -- remove after RISC-V
     status                         smallint,
     start_block_timestamp          bigint,
     end_block_timestamp            bigint,
     batches_count                  integer,
-    expected_shnarf                varchar(256),
-    blob_compression_proof         jsonb,
+    expected_shnarf                varchar(256), -- remove after RISC-V
+    blob_compression_proof         jsonb, -- rename to blob_info after RISC-V
```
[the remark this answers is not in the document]
gauravahuja: I am not sure if we look at the checksums, but best to remove the comments and just make a note elsewhere.

jonesho: I'd suggest to just make a new sql db migration file for the db schema update, e.g. V006__xxx,  but keep our coordinator config `schemaVersion` to be 5

coordinator/persistence/src/main/kotlin/linea/persistence/conflation/BlobsPostgresDaoG.kt
```diff
@@ -0,0 +1,183 @@
...
+    """
+      .trimIndent()
+
+  protected val insertSql =
+    """
+     insert into $TableName
+     (created_epoch_milli, start_block_number, end_block_number,
+     blob_hash, status, start_block_timestamp, end_block_timestamp,
+     batches_count, expected_shnarf, blob_compression_proof)
+     VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, CAST($10::text as jsonb))
```
jonesho, line 87: This sql would need to overridable for `BlobsPostgresDaoV2` as some of the db column names will be different

coordinator/persistence/src/integrationTest/kotlin/linea/persistence/conflation/BlobsPostgresDaoV2Test.kt
```diff
@@ -0,0 +1,312 @@
...
+import org.junit.jupiter.api.assertThrows
+import org.junit.jupiter.api.extension.ExtendWith
+import tech.pegasys.teku.infrastructure.async.SafeFuture
+import java.util.concurrent.ExecutionException
+import kotlin.time.Clock
+import kotlin.time.Duration.Companion.seconds
+import kotlin.time.toJavaDuration
+
+@ExtendWith(VertxExtension::class)
+class BlobsPostgresDaoV2Test : CleanDbTestSuiteParallel() {
```
jonesho: This class `BlobsPostgresDaoV2Test` is very similar to the `BlobsPostgresDaoTest`, can we refactor the duplicated coded into a base class and let those two test classes extends the base?

jonesho: @gauravahuja I'd pushed a new commit `a7307e8` done by Claude to refactor the duplicate test code to a base class, please review and see if it makes sense, thanks!

coordinator/persistence/src/main/resources/db/V002__add_blob_and_aggregation_schema.sql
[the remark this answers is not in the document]
jonesho: Yes, the migration mechanism uses checksum to determine whether the schema update sql file should be applied or not, so we should remove the comments, and maybe just simply create e.g. `V006__xxx` for risc-v schema update, but keep our coordinator config `schemaVersion` as 5

coordinator/core-interfaces/src/main/kotlin/linea/domain/Blob.kt
```diff
@@ -154,6 +154,73 @@ data class BlobRecord(
...
+    result = 31 * result + totalBatchesCount.hashCode()
+    result = 31 * result + blobsData.hashCode()
+    return result
+  }
+}
+
+data class BlobData(
+  val blobHash: ByteArray,
+  val compressedData: ByteArray,
+  val batchesCount: UInt,
```
jonesho, line 201: @gauravahuja should we add a field called `val blobFinalBlockHash: ByteArray` in BlobData? The reason is that it is needed in `BlobsSubmissionV9` as `val blobFinalBlockHashes: List<ByteArray>` (which would be passed to the `submitBlobs` contract function`FunctionBuildersV9.buildSubmitBlobsFunctionV9`), wdyt?

coordinator/persistence/src/main/kotlin/linea/persistence/conflation/BlobsPostgresDaoG.kt
gauravahuja: column names will remain same, we will only remove or rename the columns once we switch to risc v completely.

coordinator/persistence/src/main/resources/db/V002__add_blob_and_aggregation_schema.sql
gauravahuja: There is no schema update. we are using the same columns. This is just a comment for future, like a todo

coordinator/core-interfaces/src/main/kotlin/linea/domain/Blob.kt
gauravahuja: We can, but do we get this from the proof response?

coordinator/persistence/src/integrationTest/kotlin/linea/persistence/conflation/BlobsPostgresDaoV2Test.kt
gauravahuja: I had thought having a base class but it couples to the two tests together, I wasn't sure how each of them would evolve so didn't do it. Anyways, if you refactored it already it is also good.

coordinator/persistence/src/main/resources/db/V002__add_blob_and_aggregation_schema.sql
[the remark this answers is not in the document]
gauravahuja: removed the comments here and added a todo in code

coordinator/core-interfaces/src/main/kotlin/linea/domain/Blob.kt
jonesho: No, they are not provided from the proof response, as you can see from  the`RollupProofResponseV1` data class (which is derived from `rollup_spec/src/rollup_spec/prover_io/schemas/getZkRollupProofV1.response.schema.json`

gauravahuja: Even if we add it to the database, how do we get the blob hash in the first place?

jonesho: My thought was that the blob hash would be calculated from `BlobShnarfCalculator` as it is now, no?

coordinator/core-interfaces/src/main/kotlin/linea/persistence/BlobsRepository.kt
```diff
@@ -1,22 +1,26 @@
 package linea.persistence
 
 import linea.domain.BlobRecord
+import linea.domain.BlobRecordV2
 import tech.pegasys.teku.infrastructure.async.SafeFuture
 import kotlin.time.Instant
 
-interface BlobsRepository {
-  fun saveNewBlob(blobRecord: BlobRecord): SafeFuture<Unit>
+interface BlobsRepositoryG<T> {
```
Filter94, line 8: What does G stand for?

Filter94 (approved):

coordinator/core-interfaces/src/main/kotlin/linea/domain/Blob.kt
gauravahuja: Shnarf calculator will not work for the riscv flow where we might have to process multiple blobs per rollup proof.

coordinator/core-interfaces/src/main/kotlin/linea/persistence/BlobsRepository.kt
gauravahuja: Generic

coordinator/core-interfaces/src/main/kotlin/linea/domain/Blob.kt
jonesho: > Even if we add it to the database, how do we get the blob hash in the first place?

Okay, now I found that it is mentioned [here](https://github.com/LFDT-Lineth/lineth-monorepo/blob/main/rollup_spec/Readme.md?plain=1#L551):

> **Per-blob shnarf computation.** For each blob carried by the calling transaction, the contract reads the blob's versioned hash via the EIP-4844 `blobhash(i)` opcode...

jonesho: @gauravahuja But still, back to my question, should we add a field called val `finalBlockHash: ByteArray` in `BlobData`? Such that the blob submitter supporting `BlobsSubmissionV9` can submit the blob transaction with the given `BlobRecordV2` object alone?

gauravahuja: There is blobHash and blobFinalBlockHash. I don't see blobHash in the submission interface. I am not sure what blobFinalBlockHash mean, is it hash of the last block in the blob?

gauravahuja: Anyway, since changes to the blob info does not require schema changes, we can add this field when needed.

PR #3619, opened 2026-07-23
fix(prover-ray): adding limit checks for the lookups and permutations

bogdanbear:
This PR adds per-query row-limit checks to the lookup (lookuptologderivsum) and permutation (grandproduct) compilers, guarding against small-field accumulator overflow: each check runs as both a prover action (panics) and a verifier action (rejects), registered on the witness round so it fires before any rows are walked.

Introduces the budget primitives in wiop: MaxLookupRows (2^30) and MaxPermutationRows (2^58), enforced via TableRelationQuery.CheckRowLimit / ValidateRowLimit, which bound each side of a query independently.
Establishes how the budget is shared: lookups sharing an including (B) table drain one common accumulator, so their limit is divided by the group's lookup count; permutations and message-bus handles use independent per-module accumulators, so their full budget applies per query/handle.

### Checklist


prover-ray/wiop/query_table_relation.go
```diff
@@ -92,6 +92,91 @@ func (t Table) Round() *Round {
...
+// lookup is a fatal programming error rather than a proof to reject gracefully.
+func (tr *TableRelationQuery) CheckRowLimit(rt *Runtime, limit uint64) {
+	if err := tr.ValidateRowLimit(rt, limit); err != nil {
+		panic(err)
+	}
+}
+
+// checkTablesRowLimit sums the runtime row counts of every fragment in
+// tables and returns an error if the total reaches limit.
+func checkTablesRowLimit(path, side string, tables []Table, rt *Runtime, limit uint64) error {
```
Soleimani193, line 164: It should not be a run-time decision; it should be a compile-time guarantee. 

- shards are uniform; compiled WIOP is the same for all the shards, it is just that the trace/input changes per shard
- if it is a run-time decision/check, some shards generate proof, some not

To make it a compile-time guarantee, the compiler can collect and unify the queries up to a limit. As it is a compile-time guarantee, we have to use static parameters; the max size of a column is 2^22, so that you can unify lookup queries up to 2^8 S tables.

bogdanbear: As discussed yesterday, I have added both compile and runtime checks, as the verifier runtime check offers the clearest soundness guarantees.

ThomasPiellard (dismissed):

prover-ray/wiop/query_table_relation.go
```diff
@@ -92,6 +92,148 @@ func (t Table) Round() *Round {
...
+//
+// The limit is enforced on both sides: the prover panics
+// ([TableRelationQuery.CheckRowLimit]) since it is trusted code about to build
+// an unsound witness, while the verifier rejects the proof with an error
+// ([TableRelationQuery.ValidateRowLimit]).
+//
+// The bound is a deliberately loose upper bound: it sums whole fragment heights
+// and ignores selectors, so rows that a selector would exclude are still
+// counted.
+const MaxLookupRows uint64 = 1 << 30
```
AlexandreBelling, line 119: You could set this to field.Modulus - 1

bogdanbear: I chose it like this on purpose, to make sure that even if all the terms are moved to one side of the logderiv equation, we still do not have overflows. There might be weird attacks when we play on both sides of the equation (and not just one side as we did in the hackmd). 
With this parameter choice, these potential attacks are prevented.

AlexandreBelling (dismissed):

ThomasPiellard (approved):

AlexandreBelling (approved):

PR #3632, opened 2026-07-24
chore(misc): seed gradle cache on main and clean up stale pr caches

nadeemb53:
## Problem

All PRs are failing with `Received status code 429 from server: Too Many Requests` from `repo.maven.apache.org` during dependency resolution.

Root cause chain:

1. The repo's Actions cache sits at exactly **10.0 GB, the hard GitHub cap**, dominated by duplicated per-PR entries (5x 344MB zig caches, 2x 677MB CodeQL, 3x 229MB node caches on `refs/pull/*` refs).
2. GitHub LRU-evicts to stay under the cap, and the Gradle caches get wiped within hours of being written. This morning's PR jobs logged `Gradle User Home cache not found. Will initialize empty` while **zero** Gradle cache entries existed in the repo, even though yesterday's main run at 20:53 UTC saved them successfully.
3. Every job then downloads the full dependency graph from Maven Central. The self-hosted runner fleet egresses through shared NAT IPs, so Sonatype's per-IP rate limit trips fleet-wide.
4. Once evicted, the cache cannot come back on its own: PR jobs are read-only, and main only writes when a push touches Gradle-built paths. #3629 set `cache-read-only` to setup-gradle's existing default, and its own merge commit skipped the Gradle test jobs entirely, so nothing changed.

## Fix

**`gradle-cache-seed.yml`** reseeds a warm Gradle cache from main every 4 hours (plus on pushes touching Gradle files, plus manual dispatch). It uses a lenient resolve-only init script that downloads every resolvable configuration of every project without compiling, covering runtime and test configurations too. Validated with `--dry-run`: the task registers across all projects including the linea-besu tree.

**`actions-cache-cleanup.yml`** reduces the eviction pressure itself: deletes a PR's caches when it closes (`pull_request_target` on the closed event runs in base context with no PR code checked out), and a daily sweep removes any `refs/pull/*` cache not accessed for 3 days. This frees several GB so main's caches stop being evicted.

## After merging

Trigger `gradle-cache-seed` once via workflow_dispatch to seed immediately instead of waiting for the first cron tick.

### DIFF
--- a/.github/gradle/resolve-all-dependencies.init.gradle
+++ b/.github/gradle/resolve-all-dependencies.init.gradle
@@ -1,0 +1,16 @@
+// Registers a resolveAllDependencies task in every project so the cache seed
+// workflow can download the full dependency graph without compiling anything.
+allprojects {
+  tasks.register("resolveAllDependencies") {
+    notCompatibleWithConfigurationCache("resolves configurations eagerly")
+    doLast {
+      configurations.matching { it.canBeResolved }.each { cfg ->
+        try {
+          cfg.incoming.artifactView { lenient(true) }.files.each { }
+        } catch (Exception e) {
+          logger.lifecycle("Skipped ${project.path}:${cfg.name}: ${e.message}")
+        }
+      }
+    }
+  }
+}
--- a/.github/gradle/resolve-all-dependencies.init.gradle
+++ b/.github/gradle/resolve-all-dependencies.init.gradle
@@ -1,5 +1,7 @@
 // Registers a resolveAllDependencies task in every project so the cache seed
-// workflow can download the full dependency graph without compiling anything.
+// workflow can download every configuration's dependency artifacts without
+// running any compilation. Only loaded when passed via --init-script, so it
+// has no effect on regular builds.
 allprojects {
   tasks.register("resolveAllDependencies") {
     notCompatibleWithConfigurationCache("resolves configurations eagerly")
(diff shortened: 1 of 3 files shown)


.github/gradle/resolve-all-dependencies.init.gradle
```diff
@@ -0,0 +1,16 @@
+// Registers a resolveAllDependencies task in every project so the cache seed
+// workflow can download the full dependency graph without compiling anything.
+allprojects {
+  tasks.register("resolveAllDependencies") {
+    notCompatibleWithConfigurationCache("resolves configurations eagerly")
```
fluentcrafter, line 7: 1. why do we need to do this eagerly?

2. Our tasks are already big and slow 😭 What's the performance impact of this?

> workflow can download the full dependency graph without compiling anything.
3. Is this comment correct? Gradle needs to make the dependency graph before compiling the code...

gauravahuja: I don't think this runs this tasks in every CI run or gradle invocation. Isn't this meant to run on a cron job?

nadeemb53: No impact on regular builds since the task only exists when the init script gets passed in explicitly via `--init-script`, and the only thing doing that is cron on main every 4h. 

The "eagerly" part is just about that seed run: it forces resolution of every resolvable configuration in every project, not just the ones the requested tasks would need so all the compile/runtime/test jars end up in the cache.

Fair point on the wording. Gradle always resolves the graph before compiling. What I meant was it downloads all the dependency artifacts without actually running any compile tasks on top. Fixed the wording.

fluentcrafter (approved):

Issue #3631, opened 2026-07-24
E2E Tests traffic generator keeps running after CTRL+C cancellation


pajicf:
## Impact
Environment: __local__

## Describe the bug
If end-to-end tests are interrupted with CTRL+C, the background process of sending transactions keeps running.

### Steps to Reproduce
Steps to reproduce the bug:
1. Go to project root
2. Prepare environment for running e2e tests as per guide [here](https://github.com/LFDT-Lineth/lineth-monorepo/blob/ad3483d4a3aae6a144e6e02b8ed1b00ae9fe9a34/e2e/README.md)
3. Run `pnpm -F e2e run test:local-core`
4. Wait for tests to start
5. Cancel using CTRL+C

PR #3633, opened 2026-07-24
fix(e2e): run jest teardown on SIGINT and SIGTERM

pajicf:
This PR implements issue(s) #3631

### Details

When e2e tests are ran, during the [global setup](https://github.com/LFDT-Lineth/lineth-monorepo/blob/cf75766ec29006561528f4d7f031c4afbb86613b/e2e/src/config/jest/global-setup.ts) inside Jest, a background service for generating traffic on the L2 is started.
This background process currently ends only after Jest finishes with the whole test suite and runs the `stopL2TrafficGeneration()` function inside [`global-teardown.ts`](https://github.com/LFDT-Lineth/lineth-monorepo/blob/cf75766ec29006561528f4d7f031c4afbb86613b/e2e/src/config/jest/global-teardown.ts).

The test suite should be aware of forced process cancellations such as SIGTERM and SIGINT. Only these two are explicitly handled on purpose as they are the only termination signals which leave room for proper resource cleanup, others lead to forced termination. (see more [here](https://nodejs.org/api/process.html#signal-events))

In case of a hurry and repeated initiation of SIGINT (CTRL+C spamming), the code allows for termination without proper cleanup [here](https://github.com/LFDT-Lineth/lineth-monorepo/blob/cf75766ec29006561528f4d7f031c4afbb86613b/e2e/src/config/jest/global-setup.ts#L15).

### Checklist

  PR.

### DIFF
--- a/e2e/src/config/jest/global-setup.ts
+++ b/e2e/src/config/jest/global-setup.ts
@@ -9,15 +9,15 @@ import { globalTeardown } from "./global-teardown";
 
 const logger = createTestLogger();
 
-let isForcefullyShuttingDown = false;
+let shutdownAlreadyTried = false;
 async function handleForcedShutdown(signal: NodeJS.Signals) {
   // In case of a forced shutdown, prevent another cleanup being run
-  if (isForcefullyShuttingDown) {
+  if (shutdownAlreadyTried) {
     process.exit(constants.signals[signal] ?? 1);
   }
 
   logger.info("Forcefully shutting down, stopping background processes...");
-  isForcefullyShuttingDown = true;
+  shutdownAlreadyTried = true;
 
   try {
     await globalTeardown();
--- a/e2e/src/config/jest/global-setup.ts
+++ b/e2e/src/config/jest/global-setup.ts
@@ -1,10 +1,32 @@
+import { constants } from "node:os";
+
 import { createTestLogger } from "../logger";
 import { ensureOnceOffPrerequisites } from "./prerequisites";
 import { startL2TrafficGeneration } from "./traffic";
 import { setStopL2TrafficGeneration } from "./traffic-state";
 import { createTestContext } from "../setup";
+import { globalTeardown } from "./global-teardown";
 
 const logger = createTestLogger();
+
+let isForcefullyShuttingDown = false;
+async function handleForcedShutdown(signal: NodeJS.Signals) {
+  // In case of a forced shutdown, prevent another cleanup being run
+  if (isForcefullyShuttingDown) {
+    process.exit(constants.signals[signal] ?? 1);
+  }
+
+  logger.info("Forcefully shutting down, stopping background processes...");
+  isForcefullyShuttingDown = true;
+
+  try {
+    await globalTeardown();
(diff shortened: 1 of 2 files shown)


e2e/src/config/jest/global-setup.ts
```diff
@@ -16,4 +38,7 @@ export default async (): Promise<void> => {
   logger.info("L2 traffic generation started.");
 
   setStopL2TrafficGeneration(stopPolling);
+
+  process.on("SIGINT", (signal) => handleForcedShutdown(signal));
+  process.on("SIGTERM", (signal) => handleForcedShutdown(signal));
```
fluentcrafter, line 43: What if `startL2TrafficGeneration` does not return? The handlers won't be set properly. These statements shall be on line 32, no?

e2e/src/config/jest/global-setup.ts
```diff
@@ -1,11 +1,33 @@
...
 import { createTestLogger } from "../logger";
 import { ensureOnceOffPrerequisites } from "./prerequisites";
 import { startL2TrafficGeneration } from "./traffic";
 import { setStopL2TrafficGeneration } from "./traffic-state";
 import { createTestContext } from "../setup";
+import { globalTeardown } from "./global-teardown";
 
 const logger = createTestLogger();
 
+let isForcefullyShuttingDown = false;
```
fluentcrafter: nit: 
It took a while to understand what it does.. I think this naming helps a little 
```suggestion
let shutdownAlreadyTried = false;
```

e2e/src/config/jest/global-setup.ts
pajicf: This is handled inside [`traffic-state.ts`](https://github.com/LFDT-Lineth/lineth-monorepo/blob/58ba1699a3bf8042e1eecbf05a39ab55cd4968f2/e2e/src/config/jest/traffic-state.ts#L12)

It's handled via a global singleton, It is indeed kinda confusing since there's no direct correlation connecting these two from code perspective but has to be done this way since `global-setup.ts` and `global-teardown.ts` are independently called by Jest behind the scenes and can't be connected directly otherwise

e2e/src/config/jest/global-setup.ts
pajicf: Just to make sure we're on the same page, I've described above in the PR on purpose of it

> In case of a hurry and repeated initiation of SIGINT (CTRL+C spamming), the code allows for termination without proper cleanup [here](https://github.com/LFDT-Lineth/lineth-monorepo/blob/cf75766ec29006561528f4d7f031c4afbb86613b/e2e/src/config/jest/global-setup.ts#L15).

It's basically describing a forced shutdown, but yeah, might be bad naming without the full context. I'll rename to your suggestion, thanks! :)

fluentcrafter (approved):

pajicf:
@fluentcrafter I've closed this PR and opened a new one [here](https://github.com/LFDT-Lineth/lineth-monorepo/pull/3643) with proper commit messages as the repo wasn't allowing me to rebase to make a tidy history

PR #3636, opened 2026-07-27
feat(prover-ray): Bypass PCS for short columns

Tabaie:
A commitment to a column cannot be verified via FRI when its height falls under the length of the final plaintext size. In particular for scalar cells the column will be of height 1. If the final plaintext is meant to be constant as well, this is covered by https://github.com/LFDT-Lineth/lineth-monorepo/pull/3610. But if a larger final plaintext size is chosen, as is commonly done, that method won't do - there is no longer a natural way to mix the codeword in.

This PR modifies the compiler and runtime to handle cases with columns of length 1 to 8 fundamentally separately. The column is sent to the verifier as a polynomial in canonical format in the clear, and the verifier checks any evaluation claim itself directly. This will result in fewer operations for the prover and more significantly the verifier, shrink the proof size, and simplify all FRI logic, especially on the verifier side.


ivokub (changes requested):
Lets postpone for now

AlexandreBelling:
I understand the advantage of this. But at the same time, I am skeptical the ratio: speed-up vs added-complexity is good. My plan going forward was more to remove support for public columns as it creates edge-cases in the compilation suite.

PR #3637, opened 2026-07-27
feat(arithmetization): memory tweaks for RISC-V interpreter

DavePearce:
This puts through a number of tweaks to the RISC-V interpreter which dramatically reduce the cost of tracing.  At this stage, these tweaks should be considered experimental.


arithmetization/src/main/riscv/ram/read.zkc
```diff
@@ -1,72 +1,92 @@
...
-            return
+    var word:u32
+    //
+    woffset::boffset = address
+    // read enclosing word
+    word = ram[woffset as u64]
+    //
+    switch boffset {
+        case 0: {
+            trash::value = word
```
letypequividelespoubelles, line 16: did you check the perf diff with value && 0xFF ?

arithmetization/src/main/riscv/ram/write.zkc
```diff
@@ -1,75 +1,49 @@
...
-// 5. Insert the byte(s) value(s) we want to write into the masked word
-// 6. Write the word back to RAM at the calculated word index word_idx
-
 fn write_8<ram>(address:Address, value:u8) -> () {
-    var word_idx:Address = address / 4
-    var offset:Address = address % 4
-    switch offset {
+    var boffset:u2
+    var woffset:u62
+    var tmp0:u8, tmp1:u8, tmp2:u8, tmp3:u8
```
letypequividelespoubelles, line 7: I guess 3 tmp re enough ?

arithmetization/src/main/riscv/ram/read.zkc
DavePearce: No, I did not.  It might be worth checking.

arithmetization/src/main/riscv/ram/write.zkc
DavePearce: Well, not if you write this as I did:
```
tmp3::tmp2::tmp1::tmp0 = ram[woffset as u64]
```
So, there is only one lookup for reading `ram` in the function.  Otherwise, you have four lookups.  Its not completely clear to me whether it matters, but we do have lookup selectors to consider, etc.

FlorianHuc (approved):

gusiri (approved):

arijitdutta67 (approved):

Issue #3088, opened 2026-05-13
[RISCV] - Coordinator - conflation tidyup


fluentcrafter:
# Description

Decompose coordinator L1DependentApp into smaller chunks to allow better code reuse between the current proving system and RISC-V.

Key areas:
- L2 Pricing - shall be isolated/independent
- Conflation and Proof coordination
- L1Relaying (blob and aggregation submission)

PR #3640, opened 2026-07-27
refactor(coordinator): L1RelayingAppV1 to better allow RISC-V extension

fluentcrafter:
This PR implements issue(s) #3088 

---

> [!NOTE]
> **Low Risk**
> Structural refactor for extensibility; minor behavioral tweak to tracker seed values using lastFinalizedBlock instead of zero.
> 
> **Overview**
> Refactors **`L1RelayingAppV1`** so a RISC-V variant can subclass and override wiring without duplicating blob/finalization coordination logic.
> 
> The class is now **`open`**, with **`open val`** on gas cap providers and both submission coordinators, plus an **`open fun init()`** that registers blob/aggregation gauges. Metrics (block trackers, delay histograms, and event consumers) move to class-level **`val`**s shared by the coordinators instead of being created inside each coordinator’s `run` block. **`start()`** calls **`init()`** before starting services.
> 
> Tracker initial values now use **`lastFinalizedBlock`** rather than **`0UL`** for blob and finalization “latest submitted” trackers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 423e29b312d7987273f1ccf936602388da7e5dce. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/coordinator/app/src/main/kotlin/linea/coordinator/app/L1RelayingAppV1.kt
+++ b/coordinator/app/src/main/kotlin/linea/coordinator/app/L1RelayingAppV1.kt
@@ -133,12 +133,57 @@ class L1RelayingAppV1(
     }
   }
 
-  private val highestAcceptedBlobTracker = HighestULongTracker(lastFinalizedBlock).also {
+  // Metrics Setup
+  val highestAcceptedBlobTracker = HighestULongTracker(lastFinalizedBlock)
+  val latestBlobSubmittedBlockNumberTracker = LatestBlobSubmittedBlockNumberTracker(lastFinalizedBlock)
+  val latestFinalizationSubmittedBlockNumberTracker = LatestFinalizationSubmittedBlockNumberTracker(lastFinalizedBlock)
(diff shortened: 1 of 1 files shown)


jonesho (dismissed):

jonesho (approved):

jonesho (approved):

coordinator/app/src/main/kotlin/linea/coordinator/app/L1RelayingAppV1.kt
```diff
@@ -51,7 +51,7 @@ import kotlin.time.Clock
  *   - submit blobs
  *   - submit aggregations/finalizations
  */
-class L1RelayingAppV1(
+open class L1RelayingAppV1(
```
Filter94, line 54: What's the idea behind making it an open class? What are we trying to achieve that is unachievable by composition?

PR #3647, opened 2026-07-27
fix(coordinator): dead code, SSRF gap, finalization handler isolation, concurrency races

PrazwalR:
- Removes verified-dead `finalizeBlocksEthCall` and a private-key fragment leaking into a test-utils exception message
- Fixes a real bug in `FinalizationMonitorImpl`: async handler failures bypassed the isolation try/catch, silently skipping every subsequent handler and never retrying that finalization update
- Closes an SSRF gap in `conflation_createProverRequests` by validating caller-supplied API hosts against the coordinator's own already-trusted config
- Adds input validation preventing a `ULong` underflow from caller-supplied block ranges
- Fixes two concurrency races where Prometheus gauge callbacks read mutable collections without the lock their mutators use (`GlobalAggregationCalculator`, `ConflationServiceImpl`)
- Adds an observable failure counter to `EventDispatcher`

### DIFF
--- a/coordinator/app/src/main/kotlin/linea/coordinator/api/dto/ConflationCreateProverRequestJsonDto.kt
+++ b/coordinator/app/src/main/kotlin/linea/coordinator/api/dto/ConflationCreateProverRequestJsonDto.kt
@@ -30,6 +30,18 @@ data class ConflationCreateProverRequestJsonDto(
   )
 
   fun toDomainObject(): ConflationBacktestingConfig {
+    // startBlockNumber/endBlockNumber/batchesFixedSize are caller-supplied Longs/Ints converted to
+    // ULong/UInt below. That conversion is a bit-reinterpretation, not a range check, so a negative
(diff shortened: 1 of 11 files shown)


nadeemb53:
Thanks for putting this together! We're going to close this PR though, it bundles a bunch of unrelated changes together without regression tests or any prior discussion on the issues, so it's tough to review as-is. A few of the underlying concerns are legit, but the fixes really need their own separate design and review:

- the SSRF mitigation only checks hostnames, ports, schemes, paths, and cross-service endpoints are still caller-controlled
- unsigned request-limit conversions still aren't validated
- the finalization change isolates handlers, but doesn't actually give the retry behavior the description claims
- none of the behavioral or concurrency changes have tests

In general, big audit-style PRs that touch a bunch of unrelated coordinator subsystems at once aren't really mergeable in this form. Would love to see these split into smaller, focused PRs with tests so we can actually review them properly.

PR #3650, opened 2026-07-27
perf(prover-ray): speed up FRI PCS commit/open/verify (27–82×)

gbotrel:
Optimizes the koalabear FRI batch PCS. **Format-preserving** — roots, fold values and proofs are bit-identical, so no protocol or in-circuit-verifier change (mutation/soundness + root-equality tests stay green).

## Perf — before → after (large config: `sizes 2^8..2^12`, 400 base + 400 ext, rate 4, 32 queries, 32 cores)

| phase | before | after | speedup |
|---|---:|---:|---:|
| **Commit** | 5.53 s | **67 ms** | **82×** |
| **Open** | 5.88 s | **106 ms** | **55×** |
| **Verify** | 283 ms | **10 ms** | **28×** |

```
            before                              after      speedup
Commit  5.53s ██████████████████████████████  →   67 ms      82×
Open    5.88s ████████████████████████████████ →  106 ms      55×
Verify  0.28s █▌                               →   10 ms      28×
```

Small config (`2^8..2^10`, 64+64): Commit 208→5.5 ms, Open 210→9.7 ms, Verify 28→1.6 ms.

> **Update (#3656 merged into `perf/fri`):** a further Open pass aliases the committed codewords in `reconstructLevels` instead of copying/widening a fresh `[]E6` per column, and parallelizes leaf hashing over SIMD groups. Open dropped **156 ms → 106 ms** and Open allocations **621 MB → 12 MB (−98%)** on this 32-core box; Commit stays 67 ms (the SIMD-group refinement is a further win on many-core machines, neutral here — no regression). Still bit-identical (Verify + root-equality + `-race` green).

## What changed
- **Commit** — Merkle leaves hashed 16-wide with the AVX-512 batch Poseidon2 kernel in **column-major** layout (`gnark Compressx16Columns`): the kernel loads instead of gathering, and base columns are filled by bulk copy. The RS encode is **parallelized across rows** (`fft.WithNbTasks(1)`) — the serial encode was the dominant bottleneck. Leaf hashing parallelizes over 16-leaf SIMD groups (not individual leaves) so full groups stay on the SIMD path on many-core machines (#3656).
- **Open** — resolve claim-point indices once + parallelize `EvalsAt`; replace the dense N×P denominator-inverse table with **one length-N vector per level** via the rotation identity `1/(x − ζωˢ) = ω⁻ʳˢ·1/(ωᵉ⁻ʳˢ − ζ)`. `reconstructLevels` now aliases the committed codeword per column (no per-column `4N`-length copy, no up-front base→`E6` widening — base elements are lifted on the stack inside `EvalsAt`), cutting Open allocations 621 MB → 12 MB (#3656).
- **Verify** — hoist the query-independent claim computation out of the per-query loop + parallelize the loop (45× fewer allocs).

Needs `gnark-crypto` master (koalabear `Compressx16Columns`).

## vs Plonky3 (`TwoAdicFriPcs`, same machine, matched params)

| phase | ours | Plonky3 | |
|---|---:|---:|---|
| Commit | 67 ms | 63 ms | parity |
| Open | 106 ms | 22 ms | P3 ~5× |
| Verify | 10 ms | 43 ms | **ours ~4×** |

**Comparison caveats / differences** (both KoalaBear, Poseidon2 width-16/rate-8 AVX-512 — measured at parity, ~675 ns/16-perm):

| | ours | Plonky3 |
|---|---|---|
| challenge extension | **degree-6** (`E6`, ≈186-bit) | degree-4 (≈124-bit) |
| blowup / rate | 4 (`log_blowup=2`) | 4 |
| queries · fold arity | 32 · 2 | 32 · 2 |
| commit layout | multi-size DEEP-FRI, **base + ext** poly rows | base-field matrices |
| open points | 1–3 rotations/row (DEEP) | 1 OOD point |

~~The Open gap is expected: our degree-6 extension makes every `E6×E6` mul heavier.~~ **Correction:** the gap was structural, not a degree-6 tax — `EvalsAt` computed the DEEP quotient per (column, claim) in full `E6×E6` arithmetic, where the standard formulation combines columns first (ext×base) and divides once per distinct point. #3657 restructures it combine-then-divide (bit-identical proofs) on gnark-crypto's new AVX-512 `VectorE6` kernels: Open drops to **22.5 ms** on the production-like shared-points workload — Plonky3 parity at our higher (≈186 vs ≈124 bit) soundness — restoring Open < Commit. On the restructured hot loop the degree-6 extension costs ~1.5× vs degree-4, not 5×. Note the shared-points config is also the representative one: with per-row random shifts (this PR's original headline) the old code was actually 35% slower (149.6 ms).


prover-ray/crypto/koalabear/fri/pcs.go
```diff
@@ -1269,125 +1300,196 @@ func (pcs *PCS) Verify(in VerifyInputs, proof OpeningProof) error {
...
-		// round's running pair (rq.Rounds[0] is left at its zero value, so
-		// round 0 needs no special case).
-		for j := uint8(1); j < pcs.Params.numRounds(); j++ {
-			opening := proof.FRIProof.RunningQueries[queryIdx][j-1]
-			if err = checkQueryLayerShape(
-				opening, runningRoots[j], 1<<(pcs.Params.LogCodewordSize-j), true); err != nil {
-				return fmt.Errorf("fri: pcs.Verify: query %d round %d: %w", queryIdx, j, err)
-			}
-			branch, err := authenticateQueryLayer(j, opening, runningRoots[j], queryPosition>>j)
+	// A level's entry claims (claim points + values) do not depend on the
```
AlexandreBelling, line 1321: @ivokub @YaoJGalteland  @arijitdutta67  

Any conflict with your ongoing piece of work?

AlexandreBelling (dismissed):

prover-ray/crypto/koalabear/fri/pcs.go
YaoJGalteland: Look at [verify.zig:474-477](vscode-webview://1pthp2p2fjqqc1juoqhmvrqegcd2st6mce28t0qm7r9dja8djrtp/verifier-ray/src/pcs/verify.zig#L474-L477):


```
const claims = inputs.entry_claims[g + ei];   // ← claimed VALUES: precomputed input ✓
for (entry.shifts, 0..) |shift, k| {
    claim_store[ei][k] = .{
        .point = try reconstruct.shiftedPoint(entry.size_log2, shift, inputs.zeta),  // ← claim POINT: computed here
        .value = claims[k],
    };
```
So shiftedPoint runs:

num_queries × Σ_levels Σ_entries (shifts per entry) times.

The claim points are identical across all queries — only x_self/x_sib and the running pair vary per query. So the entire claim_store point computation is redundant (num_queries − 1)/num_queries of the time.

I built a micro-benchmark, ran it, and traced the R5 cost model. Here's what I found. With production friNumQueries = 229 and a ~200-column layout (600 total shifts). The hoist eliminates 98.8% of the shiftedPoint work — 228 of every 229 calls are redundant. 

I think this would be a nice-to-have optimization for verifier-ray. @ivokub

ivokub: It definitely is!

ThomasPiellard:
There is something off I think
Commit -> 67ms
Open -> 156ms ?? (are the queries parallelised ?)
Open should be much cheaper than Commit.
As discussed also the ratio time(Commit) / time(Verify) is a bit weird (compared to e.g. Plonky3)

AlexandreBelling:
The opening part contains the FRI commitment part. To me, it's not non-sensical that the opening time is bigger than the commitment time.

ThomasPiellard (approved):

ivokub (approved):

PR #3651, opened 2026-07-27
feat(verifier-ray): FRI Low-Degree Testing Functions

Tabaie:
This PR implements the basic FRI machinery, tested against generated vectors from the prover-ray side. It does not implement a complete polynomial commitment scheme.


verifier-ray/src/query/fri.zig
```diff
@@ -0,0 +1,244 @@
...
+    FoldMismatch,
+    FinalPolyMismatch,
+    BoundaryAuxNotConstant,
+};
+
+/// FRI configuration shared by every check in this module. Comptime, since it
+/// is fixed at protocol-compile time and shared with the prover. Mirrors
+/// prover-ray's `fri.Params`, restricted to what the low-degree-test core
+/// needs; the DEEP/PCS layer carries its own encoder schedule on top.
+pub const Params = struct {
```
YaoJGalteland, line 34: `logPlaintextSize` is gone here, while `prover-ray`’s `fri.Params` includes it and uses it in `restrictTo`. Do we expect PCS verification to need the same `restrictTo`? If so, we may need to add `logPlaintextSize` back.

Tabaie: Good point. This structure had `num_rounds` instead which essentially contained the same information, but it's better to have the Go and Zig codes match closely.

verifier-ray/src/query/fri.zig
```diff
@@ -0,0 +1,271 @@
+const std = @import("std");
```
ivokub: I'd avoid importing `std` - this can bring in unwanted side effects/imports. Our goal would be to have as small and fast as possible implementation (so that we can execute verifier in zkVM for proof recursion).

I see down that there are essentially only a few places it is being used and I think we can avoid std.

verifier-ray/src/query/fri.zig
```diff
@@ -0,0 +1,271 @@
...
+    if (query_branches.len != num_rounds - 1) return Error.InvalidRunningLayerShape;
+    if (round_roots.len != num_rounds - 1) return Error.InvalidRunningLayerShape;
+
+    for (1..@as(usize, num_rounds)) |j| {
+        const branch = query_branches[j - 1];
+        const want_siblings = params.log_codeword_size - j;
+        if (branch.siblings.len != want_siblings) return Error.InvalidRunningLayerShape;
+
+        const recovered = try branch.recoverRoot(shr(position, j));
+        if (!std.meta.eql(recovered, round_roots[j - 1])) return Error.MerkleProofInvalid;
```
ivokub: Here - we could just check that two digests are equal, not use `std.meta.eql`. Right now it seems we don't have a dedicated "equals" method on `poseidon2.Digest`, but it can be added nicely.

verifier-ray/src/query/fri.zig
```diff
@@ -0,0 +1,271 @@
...
+}
+
+fn shr(value: usize, amount: usize) usize {
+    const shift: std.math.Log2Int(usize) = @intCast(amount);
+    return value >> shift;
+}
+
+/// Converts a Poseidon2 digest into an extension element. Expects
+/// coordinates 6 and 7 to be zero, matching prover-ray's `octupletToExt`.
+fn octupletToExt(digest: poseidon2.Digest) Error!ext.Ext {
```
ivokub, line 260: I'd move it into poseidon2.zig to be a method on Digest.

verifier-ray/src/query/fri.zig
```diff
@@ -0,0 +1,271 @@
...
+    // field.rootOfUnityBy bounds every domain log-size to <= max_order_root
+    // (24), so the low `width` bits of `value` always fit in a u32 with room
+    // to spare.
+    const v: u32 = @intCast(value);
+    const reversed: u32 = @bitReverse(v);
+    const shift: std.math.Log2Int(u32) = @intCast(32 - width);
+    return reversed >> shift;
+}
+
+fn shr(value: usize, amount: usize) usize {
```
ivokub: Perhaps we don't need the Log2Int here. `amount` parameter is essentially comptime known (it is a loop index), so perhaps we can just define
```zig
fn shr(value: usize, comptime amount: usize) usize {
    return value >> shift;
}
```
and then it becomes obvious enoufh that we can just inline it in the calling place? It is called only once imo.

verifier-ray/src/query/fri.zig
```diff
@@ -0,0 +1,271 @@
...
+        }
+    }
+}
+
+// 2^{-1} mod p (p = 2_130_706_433 = 2^31 - 2^24 + 1). Duplicated from
+// crypto/poseidon2.zig's inv2Exp1, which is private to that module.
+const inv_two: field.Element = .{ .value = 1_065_353_217 };
+
+fn fullDomainGenerator(comptime params: Params) field.Element {
+    comptime {
```
ivokub: I don't understand why we have comptime check first and then we return runtime. Essentially, the generator is already known at compile time (from the FRI params), so this could be completely inlined at compile time (i.e. the generator will be embedded in the code already)

verifier-ray/src/query/fri.zig
```diff
@@ -0,0 +1,271 @@
...
+        _ = field.rootOfUnityBy(@as(usize, 1) << params.log_codeword_size) catch
+            @compileError("fri: log_codeword_size exceeds the supported KoalaBear root-of-unity order");
+    }
+    return field.rootOfUnityBy(@as(usize, 1) << params.log_codeword_size) catch unreachable;
+}
+
+/// x = g^{bitrev_{log_size}(position)}, where g generates the size-2^log_size
+/// subgroup. Matches prover-ray's `domainPoint`: the codeword is stored
+/// bit-reversed so that FRI conjugate pairs land at adjacent positions.
+fn domainPoint(log_size: u8, generator: field.Element, position: usize) field.Element {
```
ivokub: log_size is also comptime known imo. This comes from FRI params.

ivokub: And we can possibly use u32 or u64 for the position directly, then we don't need the type casts below. Considerations:
* u64 -- we target R64IM anyway, so there is no difference in we use u32 vs u64 -- we would be using same instructions (I think there are no dedicated u32 arithmetic in R5-64IM). And pow takes u64 as input, so its more convenient (but I think it should take u32 due to 2-adicity anyway)
* u32 -- the codeword size is bounded by 2-adicity anyway (imo 24), so cannot be bigger than u32. And perhaps this would also allow later to instead use R32 for example if we want to try out some more efficient arithmetization for recursion even

verifier-ray/src/query/fri.zig
```diff
@@ -0,0 +1,271 @@
...
+    /// log2 of the number of coefficients the final polynomial reveals
+    /// (`fri.Params.logFinalPolySize`, private in Go); this length is the
+    /// enforced low-degree bound on the final layer.
+    log_final_poly_size: u8 = 0,
+    /// Number of independent FRI queries.
+    num_queries: usize,
+
+    /// Number of folding rounds: `fri.Params.numRounds()`.
+    pub fn numRounds(comptime self: Params) u8 {
+        comptime {
```
ivokub, line 48: Perhaps we can add comptime check that num_queries != 0. Otherwise any proof verifies. Normally it is always set, but then it ensures that in case there is some serialization bug etc we wouldn't have problems.

verifier-ray/src/query/fri.zig
```diff
@@ -0,0 +1,271 @@
...
+    resolved: []const ResolvedQuery,
+    fold_alphas: []const ext.Ext,
+    positions: []const usize,
+) Error!void {
+    const num_rounds = params.numRounds();
+    if (resolved.len != params.num_queries) return Error.InvalidResolvedQueryCount;
+    if (fold_alphas.len < num_rounds) return Error.InsufficientFoldAlphas;
+    if (positions.len < resolved.len) return Error.InsufficientPositions;
+
+    const generator = fullDomainGenerator(params);
```
ivokub: You can instead `const generator = comptime fullDomainGenerator(params);`. Then the runtime is a bit faster.

prover-ray/crypto/koalabear/fri/vectors_gen_test.go
```diff
@@ -0,0 +1,418 @@
+//go:build gen_vectors
```
ivokub: Imo the preferred approach would be instead of doing "push-based" test vector generation from prover-ray to rather be pull-based in verifier-ray. So for example for rest of testdata generation (but also the code generation we use to generate the Zig representation of Go-side `wiop.System`) we just read the `prover-ray` structures and parse the information.

This way:
* we can pin the supported prover-ray version in verifier-ray (i.e. we have in `go.mod` for the codegen/fixturegen concrete prover-ray commits what we're using). This allows the prover to iterate while not caring about implementing same updates on the verifier side. Later on when things are stable we can probably keep the prover and verifier synced, but imo for now it has worked well
* we can avoid doing some serialization round (i.e. right now everything goes through JSON). It is fine to have it, but still an unnecessary step what for tests adds overhead. For example, when we do pull-based approach (i.e. here we just compute the proof), then we can just write Zig source with the data, no serialization needed. See for example the vanishing/verify tests where we directly use templates to write output. Have a look how the test fixtures there look and how we generate.

When there is something which is not public on the prover side, then we can just create a separate PR there making things public (i.e. NewCompleteBinaryTree imo) and then we can use the pinned version here.

ivokub (changes requested):
In principle looks good. There are a few optimizations here and there what we could do, but I'm also fine if we postpone the optimization work for later (i.e. comptime usage etc.). But Imo minimally two things I would prefer to do:
* define the fixture generator here on the verifier-ray side. If something is not public prover side then lets add a small PR which makes methods/types public what we need.
* for fixtures lets write the Zig source directly (i.e. lets not do JSON serialization/deserialization).
* avoid std import in `fri.zig`. I think we can completely avoid that (its only an equality of arrays and Log2Int method which imo we don't need).

For the rest of the comments -- I'm fine to postpone for a later bigger optimization round.

verifier-ray/test/fri_fold_cases.zig
```diff
@@ -0,0 +1,48 @@
+// FROZEN test vectors: NOT auto-generated. See fri_test.zig's doc comment
```
ivokub, line 1: We should have created a separate PR for exporting the private methods interfaces and then use code generation.

ivokub (approved):

YaoJGalteland (approved):

PR #3654, opened 2026-07-28
fix(contracts): use correct sender when reporting yield

thedarkjester:
Uses the yield manager as the message sender as this is where the funds are technically coming from

---

> [!NOTE]
> **Medium Risk**
> Changes how yield is represented on the L1→L2 bridge (message hash and events); incorrect attribution could break L2 accounting, but scope is limited to the yield reporting path.
> 
> **Overview**
> **`reportNativeYield`** now records the **yield manager** as the L1 sender when building the synthetic cross-chain message (`MessageSent` and its hash), instead of treating the rollup / message service as the source of funds.
> 
> The PR also adds **pinned `LineaRollup` deployment bytecode** under `contracts/deployments/bytecode/2026-07-29/` (Hardhat artifact + metadata: `CONTRACT_VERSION` 8.0, Solidity 0.8.33, optimizer 200 runs, EVM Osaka) so deploy/verify can use the build that includes this fix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 16ec6a3314b332bc6b2fb4d32afff7ea6fa6821f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kyzooghost (approved):
approved - I wonder if we should create an E2E test case specifically for this case of "Postman claims the yield message"

PR #3655, opened 2026-07-28
feat(verifier-ray): PCS

Tabaie:
Building on top of https://github.com/LFDT-Lineth/lineth-monorepo/pull/3651, this PR implements a standalone PCS, with test-vectors against prover-ray. The PCS is not integrated into the protocol.


verifier-ray/src/query/pcs.zig
```diff
@@ -0,0 +1,412 @@
...
+    /// `claimed_values[e.claim_offset..][0..e.shifts.len]`.
+    claimed_values: []const ext.Ext,
+    zeta: ext.Ext,
+    fold_alphas: []const ext.Ext,
+    query_positions: []const usize,
+    proof: OpeningProof,
+};
+
+pub fn verify(comptime system: System, input: VerifyInput) Error!void {
+    const params = system.params;
```
YaoJGalteland: System.params stays the unrestricted max-size config. There is no restrictTo call anywhere in pcs.zig or fri.zig. It derives numRounds, the expected round-root count, and layer counts straight from system.params. Do we also need a restricted version of the params as prover-ray?

YaoJGalteland (commented):
The core verification logic looks sound and correctly mirrors `prover-ray`. I only found one completeness issue related to the unrestricted params.

verifier-ray/src/crypto/merkle.zig
```diff
@@ -0,0 +1,202 @@
...
+fn absorbLeafHeader(hasher: *poseidon2.MDHasher, base_width: usize, ext_width: usize) void {
+    hasher.writeElements(&.{
+        field.Element.init(leaf_domain_tag),
+        field.Element.init(@as(u64, base_width)),
+        field.Element.init(@as(u64, ext_width)),
+    });
+}
+
+/// One committed row's preimage: prover-ray's `RowOpening`.
+pub const RowOpening = struct {
```
ivokub, line 84: Instead of defining new, we can use field.Vector which is a union type (either one or another).

verifier-ray/src/query/pcs.zig
```diff
@@ -0,0 +1,423 @@
...
+}
+
+/// Whether `point` lands in the size-2^log_size multiplicative subgroup.
+/// Every domain point is a base-field root of unity, so an extension-valued
+/// point that isn't itself a lifted base element can never coincide with
+/// one. Mirrors prover-ray's `pointInDomain`.
+fn pointInDomain(point: ext.Ext, log_size: u8) bool {
+    if (!point.isBase()) return false;
+    const shift: std.math.Log2Int(u64) = @intCast(log_size);
+    const powered = point.B0.a0.pow(@as(u64, 1) << shift);
```
ivokub: I think this can be a comptime pow instead. log_size can be comptime I think. This reduces runtime cost as the exponentiation is done with known exponent.

verifier-ray/testdata/generate/.gitignore
```diff
@@ -0,0 +1 @@
+/generate
```
ivokub, line 1: I don't understand what does this filer? Do we need the gitignore?

verifier-ray/testdata/generate/fri/.gitignore
```diff
@@ -0,0 +1 @@
+/fri
```
ivokub, line 1: Do we need the gitignore?

verifier-ray/testdata/generate/fri/go.mod
```diff
@@ -0,0 +1,23 @@
+module github.com/consensys/linea-monorepo/verifier-ray/testdata/generate/fri
```
ivokub, line 1: Why separate go.mod for fri fixture generation only? Would other fixture generators break right now?

verifier-ray/src/query/pcs.zig
```diff
@@ -0,0 +1,423 @@
...
+        const query_position = input.query_positions[query_idx];
+        const opening = input.proof.input_queries[query_idx];
+        try authenticateInputQuery(params, opening, input.roots, info, query_position);
+
+        if (num_rounds > 0) {
+            const running_query = input.proof.fri_proof.running_queries[query_idx];
+            try fri.resolveRunningLayers(params, input.proof.fri_proof.round_roots, running_query, query_position, rounds_buf[query_idx][0..num_rounds]);
+        }
+
+        final_buf[query_idx] = fri.domainPointExt(params.log_codeword_size - num_rounds, query_position >> num_rounds);
```
ivokub: I'm not sure we should use `final_buf` as a temporary variable here. It is a bit confusing. Imo we can just define a new variable instead?

verifier-ray/src/query/pcs.zig
```diff
@@ -0,0 +1,423 @@
...
+    }
+
+    try fri.checkFolds(params, &resolved, input.fold_alphas, input.query_positions);
+}
+
+/// Evaluates the revealed final-polynomial coefficients at `x` via Horner's
+/// method, rather than expanding the whole final-domain codeword via FFT
+/// (prover-ray's approach): the verifier only ever needs one or two point
+/// evaluations per query, never the full codeword.
+fn evalFinalPoly(coeffs: []const ext.Ext, x: ext.Ext) ext.Ext {
```
ivokub: We already have polynomial package in verifier-ray. Cannot we reuse it?

ivokub (commented):
I think it looks good. Though because its based on #3651 then lets merge that before and I'll approve this one after that.

Tabaie:
> I think it looks good. Though because its based on #3651 then lets merge that before and I'll approve this one after that.

I think with stacked PRs you can approve separately and then merge the whole thing all at once. Though this doesn't seem to be set up the right way anymore. The base used to be https://github.com/LFDT-Lineth/lineth-monorepo/pull/3651 and now it is `main`.

verifier-ray/testdata/generate/fri/go.mod
Tabaie: Yes at least at the time of development it did. I'll check again.

verifier-ray/testdata/generate/.gitignore
Tabaie: I think these are codegen makefile targets which we shouldn't commit. I'll double check If they are necessary.

verifier-ray/src/query/pcs.zig
```diff
@@ -0,0 +1,412 @@
...
+    query_position: usize,
+) Error!void {
+    if (opening.len != mapping.distinct_count or roots.len != mapping.distinct_count) return Error.InputTreeCountMismatch;
+
+    const codeword_size = @as(usize, 1) << params.log_codeword_size;
+    for (opening, roots) |branch, root| {
+        const num_levels = branch.leaves.len;
+        if (num_levels == 0) return Error.InputTreeShapeMismatch;
+        const num_leaves = @as(usize, 1) << @as(std.math.Log2Int(usize), @intCast(num_levels));
+        if (num_leaves > codeword_size or codeword_size % num_leaves != 0) return Error.InputTreeShapeMismatch;
```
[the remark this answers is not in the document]
YaoJGalteland: the bounds check is Fixed in the next PCS integration PR

verifier-ray/src/query/pcs.zig
```diff
@@ -0,0 +1,463 @@
...
+
+    // One challenge per intermediate layer root, absorbing the root between
+    // squeezes; then the final round's challenge, with no root after it. When a
+    // protocol never folds (num_rounds == 0) the buffer is empty and this whole
+    // block is elided at comptime rather than indexing into a [0]ext.Ext.
+    if (comptime num_rounds > 0) {
+        for (fri_proof.round_roots, 0..) |root, i| {
+            challenges.fold_alphas[i] = transcript.randomExt();
+            transcript.updateElements(&root);
+        }
```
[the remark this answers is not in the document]
YaoJGalteland: the deriveChallenges indexes fold_alphas without shape check is Fixed in the next PCS integration PR

verifier-ray/src/query/pcs.zig
```diff
@@ -0,0 +1,508 @@
...
+const canonical = @import("../polynomial/canonical.zig");
+const fiat_shamir = @import("../crypto/fiat_shamir.zig");
+const fri = @import("fri.zig");
+
+/// The PCS/DEEP-quotient layer: input-tree authentication, per-level Horner
+/// DEEP-quotient reconstruction, and the boundary-round / D=1 special cases,
+/// ported from prover-ray's `pcs.go` `PCS.Verify`. Produces the
+/// `fri.ResolvedQuery` records `fri.checkFolds` needs.
+///
+/// Unlike prover-ray, `System.layout` is not recomputed per proof from raw
```
YaoJGalteland, line 15: Does this design run into trouble with dynamic modules?

Prover-ray supports dynamic modules (Module.isDynamic). A dynamic module's size is supplied per Runtime — it can differ between the 100-tx proof and the 200-tx proof produced by the same compiled prover. If you bake the layout at codegen time from one representative run, you get a verifier that rejects any proof where a dynamic module was sized differently. That means:

- Either you restrict to static-only circuits (losing expressiveness)
- Or you regenerate and recompile verifier-ray for every distinct dynamic-module sizing (operationally painful)

YaoJGalteland: @Tabaie Could you confirm if this is an issue? as the fix involves a major update?

If so, I think the fix doesn’t belong in this PR. Since this PR only targets standalone PCS tests with fixed shapes, it’s fine to keep the current approach here.

The fix should live in the PCS integration PR, where we wire PCS into the full protocol verifier. That’s the right place to:

- treat module sizes as runtime inputs
- bind them into the transcript
- and make sure the verifier doesn’t assume a fixed layout

Tabaie: Yes you're right that it's overly comp time and can't handle dynamic modules. I noticed that you already fixed it in the integration PR.

ivokub (dismissed):

YaoJGalteland (dismissed):

ivokub (approved):

Issue #868, opened 2025-04-11
Monitor forced L1 transactions


FlorianHuc:
## Description
System to monitor forced transaction on L1 and alerting for those that are not included and soon to expire

TBD: could be part of coordinator for deployment facility, or outside to give better uptime.

## Tasks
- [ ] gets the list of forced transactions from L1
- [ ] for every forced tx validate if it has been included on Linea (looking by tx hash for instance)
- [ ] for non included tx that was posted more than Xh ago, check if is invalid
- [ ] for non included tx with assigned block less than Y blocks away and which we don't know it is invalid or included, raise an alert.

## Acceptance criteria
- [ ]

## Risks
- [ ] we must know when there is an issue to prove a tx is invalid and that it cannot be included.

PR #3657, opened 2026-07-28
perf(prover-ray): combine-then-divide DEEP quotient in FRI Open (5x, bit-identical)

gbotrel:
Follow-up to #3650, closing the Open gap vs Plonky3. **Format-preserving**: exact regrouping of field arithmetic, so roots, fold values and proofs stay bit-identical (equality-tested against the previous implementation, kept as the reference in `fri_evalsat_test.go`; mutation/root-equality suites and `-race` green).

## The key algorithmic change

`Level.EvalsAt` computed the batched DEEP quotient **quotient-per-claim**: per (position, column, claim) one full `E6×E6` mul by the denominator inverse, plus one more `E6×E6` per (position, column) for the Horner α-ladder — ~2400–3200 `E6×E6` muls per position; this made Open `E6×E6`-bound and *slower than Commit*.

It now computes **combine-then-divide** (the structure Plonky3 uses): columns sharing a claim-point set are combined once into `G(X) = Σ_c α^c·f_c(X)` — one multiply-accumulate per (position, column), vectorized — and the division runs once per (position, **distinct point**):

```
F(X) = running(X)·α^C + Σ_g Σ_{z∈g} (G_g(X) − Y_g,z)/(X − z),   Y_g,z = Σ_{c∈g} α^c·y_c,z
```

Production protocols open every row at the same handful of points, so there is one group and the per-position cost drops from ~3200 scalar `E6×E6` muls to ~800 AVX-512 mul-accumulates + 3 `E6×E6` muls.

The combine pass runs on gnark-crypto's new AVX-512 `VectorE6` kernels (`ScalarMulAccByElement` / `ScalarMulAcc` / `ScalarMul`, Consensys/gnark-crypto#867 → Consensys/gnark-crypto#868), currently pinned to commit `500ec11`; **to re-pin once #868 is merged to master**. Off AVX-512 hardware the kernels fall back to the equivalent scalar loops.

## Perf (large config, 32 cores)

| shift schedule | Open before | restructure only (scalar) | + AVX kernels | |
|---|---:|---:|---:|---|
| shared points `{0,1,2}` (production shape) | 149.6 ms | 30.7 ms | **22.5 ms** | **6.6×** |
| per-row random shifts | 111 ms | 85 ms | 81 ms | 1.4× |

Open (22.5 ms) is now well below Commit (67 ms), as FRI cost structure predicts — and at **Plonky3 parity (22 ms) despite our degree-6 vs their degree-4 extension** (≈186- vs ≈124-bit soundness): the hot loop is ext×base, where the extension degree costs 1.5×, not 5×. Commit/Verify unchanged.

## Benchmark headline

The shared-shift-set configs are now the headline benchmarks: production protocols open all rows at a shared handful of points (ζ, ζω, …). The previous default — 1–3 uniformly-random shifts per row, now the `/shifts=random` variant — is adversarial for the grouped combine pass (every column its own group) and is kept to track the worst case.

### DIFF
--- a/prover-ray/crypto/koalabear/fri/bench_test.go
+++ b/prover-ray/crypto/koalabear/fri/bench_test.go
@@ -39,20 +39,24 @@ type benchConfig struct {
 
 func (c benchConfig) name() string {
 	suffix := ""
-	if c.sharedShifts != nil {
-		suffix = "/shifts=shared"
+	if c.sharedShifts == nil {
+		suffix = "/shifts=random"
 	}
 	return fmt.Sprintf("sizes=%d..%d/base=%d/ext=%d/rate=%d/queries=%d%s",
 		c.minLog2, c.maxLog2, c.basePolys, c.extPolys, c.rate, c.numQueries, suffix)
 }
 
-// benchConfigs mirrors bench/main.go's defaults (the large one) plus a small
-// config for quick iteration, and a production-like variant of the large
-// config whose rows all share a tiny fixed shift set {0,1,2}.
+// benchConfigs: the headline configs open every row at the shared shift set
+// {0,1,2}, matching production protocols where all rows of a proof share a
+// handful of claim points (ζ, ζω, …) — the shape Level.EvalsAt's grouped
+// combine pass is optimized for. The /shifts=random variant draws 1–3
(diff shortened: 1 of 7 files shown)


YaoJGalteland (approved):
LGTM

PR #3665, opened 2026-07-29
feat(verifier-ray): deriving pcs challenges from Fiat-Shamir

YaoJGalteland:
Building on top of #3655.

This PR derives the PCS opening challenges—`zeta`, the FRI fold challenges, and the query positions—from the verifier’s Fiat–Shamir transcript, then wires them into `verifier.verify`.

Previously, the PCS layer in `query/pcs.zig` only exposed a `verify` function that received these challenges as caller-supplied inputs. In tests, they were read directly from the golden vectors.

### PCS integration Follow-ups

* Make PCS mandatory in `verifier.zig` by removing the optional `Systems.pcs` and `Proof.pcs_opening` fields once every protocol includes a commitment.
* Prove the link between the PCS and vanishing claims. `witness_claims` and `quotient_claims` should be re-sliced from the FRI-authenticated `entry_claims`, rather than supplied independently by the proof.
* Double-check whether the PCS batch roots need to be bound explicitly to the transcript. `pcs.verify` currently authenticates against the prover-supplied `pcs_opening.roots`, but those roots are not yet tied to the oracle commitments absorbed by the transcript. As a result, `zeta` may not be provably bound to the root used for the opening. --> Wire verifier.zig PCS path to use bound roots instead of proof-supplied roots; remove roots from PcsOpening
* Cleanup tests

### DIFF
--- a/verifier-ray/src/crypto/fiat_shamir.zig
+++ b/verifier-ray/src/crypto/fiat_shamir.zig
@@ -70,6 +70,10 @@ pub const Transcript = struct {
     /// exactly. Squeezing continues from the current transcript state, so callers
     /// must have absorbed everything up to the query-derivation point.
     pub fn randomManyIntegers(self: *Transcript, out: []usize, comptime upper_bound: usize) void {
+        comptime {
+            if (upper_bound == 0 or (upper_bound & (upper_bound - 1)) != 0)
+                @compileError("fiat_shamir.randomManyIntegers: upper_bound must be a non-zero power of two");
+        }
         var i: usize = 0;
         while (i < out.len) {
             const digest = self.randomDigest();
--- a/verifier-ray/src/crypto/fiat_shamir.zig
+++ b/verifier-ray/src/crypto/fiat_shamir.zig
@@ -1,4 +1,3 @@
-const std = @import("std");
 const field = @import("../field/koalabear.zig");
 const ext = @import("../field/koalabear_ext.zig");
 const field_value = @import("../field/value.zig");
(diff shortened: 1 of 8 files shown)


verifier-ray/src/crypto/fiat_shamir.zig
```diff
@@ -60,4 +61,23 @@ pub const Transcript = struct {
...
         };
     }
+
+    /// Fills `out` with integers uniformly reduced modulo `upper_bound` (which
+    /// must be a power of two), consuming one Poseidon2 digest at a time and
+    /// taking its eight base limbs in order. Mirrors prover-ray's
+    /// `RandomManyIntegers`, used to derive FRI query positions from the live
+    /// transcript. Squeezing continues from the current transcript state, so
+    /// callers must have absorbed everything up to the query-derivation point.
+    pub fn randomManyIntegers(self: *Transcript, out: []usize, upper_bound: usize) void {
```
ivokub: In practice `upper_bound` is known, so can use comptime. And we can remove the `std.debug.assert` as it will be optimized away in small and fast build modes (which we use for R5 target anyway).

verifier-ray/src/crypto/fiat_shamir.zig
```diff
@@ -60,4 +61,23 @@ pub const Transcript = struct {
...
+        std.debug.assert(upper_bound != 0 and (upper_bound & (upper_bound - 1)) == 0);
+        var i: usize = 0;
+        while (i < out.len) {
+            const digest = self.randomDigest();
+            for (digest) |element| {
+                out[i] = @as(usize, element.value) % upper_bound;
```
ivokub, line 81: Is this also used in prover? Imo taking a modulus is not uniform? Usually for uniform sampling we compare that the value is less than the bound and if not then sample a new one.

ivokub (changes requested):
In principle looks good. Please see one small comment (removing std package and using comptime upper bound for randomManyIntegers

verifier-ray/src/crypto/fiat_shamir.zig
YaoJGalteland: We need to match prover-ray, where the sampled value is reduced modulo the upper_bound rather than using rejection sampling.

verifier-ray/src/verifier.zig
```diff
@@ -98,6 +126,26 @@ pub fn verify(
...
+    // the Fiat-Shamir challenges. Challenge derivation touches the transcript;
+    // the check is pure arithmetic.
+    if (comptime systems.pcs) |pcs_system| {
+        const opening = proof.pcs_opening orelse return error.MissingPcsOpening;
+        const pcs_challenges = try pcs.deriveChallenges(pcs_system, &transcript, opening.proof.fri_proof);
+        try pcs.verify(pcs_system, .{
+            .roots = opening.roots,
+            .claimed_values = opening.claimed_values,
+            .zeta = all_coins[pcs_system.zeta_coin_index],
+            .fold_alphas = &pcs_challenges.fold_alphas,
```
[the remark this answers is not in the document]
YaoJGalteland: defer this to the PCS integration.

verifier-ray/src/verifier.zig
```diff
@@ -98,6 +126,26 @@ pub fn verify(
...
+            .query_positions = &pcs_challenges.query_positions,
+            .proof = opening.proof,
+        });
```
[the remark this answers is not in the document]
YaoJGalteland: The fix has already been incorporated in the next PCS integration PR, where we handle all checks and ensure values are properly bound to trusted values.

PR #3666, opened 2026-07-29
feat(linea-besu): support custom liveness signers

nadeemb53:
- add a backend-neutral `SignerService` extension point for externally packaged Besu plugins
- inject the configured signer directly instead of retaining a named signer registry in Besu
- refactor liveness transaction signing onto the shared digest signer path
- retain Web3Signer as the default backend and preserve existing Web3Signer configuration and constructor behavior
- add a separately packaged fake signer plugin to verify the real Besu service/classloader boundary

Existing deployments continue to use Web3Signer without configuration changes. Custom signing is opt-in through `signer-type=custom`.


.github/workflows/maven-release.yml
```diff
@@ -56,14 +57,29 @@ jobs:
         env:
           LIB_TO_RELEASE: ${{ github.event.inputs.libToRelease }}
           GITHUB_TOKEN: ${{ secrets._GITHUB_TOKEN_RELEASE_ACCESS }}
-        run: ./gradlew "${LIB_TO_RELEASE}:build"
+        run: |
+          if [[ "$LIB_TO_RELEASE" == ':linea-besu:plugins:sequencer-interfaces' ]]; then
```
Filter94: Why does `:linea-besu:plugins:sequencer-interfaces` need a special treatment?

linea-besu/plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/liveness/Web3SignerDigestSigner.java
Filter94, line 1: Can't we reuse the SignerFactory used in Coordinator? Same question about the Signer implementation itself, can't we reuse the `Web3SignerRestClient`

linea-besu/plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorPlugin.java
```diff
@@ -135,6 +147,48 @@ public void doStart() {
             blobCompressorSelector));
   }
 
+  static Signer<Secp256k1Signature> resolveLivenessSigner(
```
Filter94: I think this logic is complicated enough to be injected into the `LineaTransactionSelectorPlugin`. We're adding the whole testsuite that is focused on the instantiation of the `LineaTransactionSelectorPlugin` with the liveness plugin because signer instantiation is not trivial. From responsibilities PoV that's wrong, so I suggest we move this logic into a separate class if it's possible to inject it here

Filter94: Looking at the source code, other objects injected into `LineaTransactionSelectorPlugin` via `ServiceManager`. And having the signer as a service might make sense and be future proof if we use the signer for some other purpose later on

linea-besu/plugins/linea-sequencer/sequencer/src/test/java/net/consensys/linea/sequencer/liveness/LineaLivenessTxBuilderTest.java
```diff
@@ -0,0 +1,151 @@
...
+            config, () -> Optional.of(Wei.ONE), BigInteger.valueOf(1337), signer);
+
+    assertThatThrownBy(() -> builder.buildUptimeTransaction(true, 1234, 0))
+        .isInstanceOf(IOException.class)
+        .hasMessageContaining("Failed to sign liveness transaction")
+        .hasRootCauseMessage("KMS unavailable");
+  }
+
+  @Test
+  void restoresInterruptFlagWhenSigningIsInterrupted() throws Exception {
```
Filter94: What is the purpose of this test?

linea-besu/plugins/linea-sequencer/build.gradle
```diff
@@ -14,6 +14,13 @@ plugins {
 }
 
 allprojects {
+  repositories {
```
Filter94: What new dependencies do we need from teku?

linea-besu/plugins/sequencer-interfaces/build.gradle
```diff
@@ -8,8 +8,14 @@ buildscript {
   }
 }
 
+apply plugin: 'java-library'
+
 repositories {
   mavenCentral()
+  maven {
+    url 'https://artifacts.consensys.net/public/teku/maven/'
```
Filter94: If it's really needed, we should define it in the root of linea-besu plugins and apply to subprojects

linea-besu/plugins/linea-sequencer/signer-test-plugin/build.gradle
Filter94, line 1: If it's for the test, why don't we define it in the acceptance tests sourceset?

.github/workflows/maven-release.yml
nadeemb53: `sequencer-interfaces` exposes types from `signer-interfaces` through `NamedSignerProviderService`. Gradle doesn’t publish project dependencies automatically, so the workflow stages both artifacts with the same version before releasing them.

linea-besu/plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/liveness/Web3SignerDigestSigner.java
nadeemb53: SignerFactory isn’t reusable directly because it depends on coordinator-specific config and Vertx. Web3SignerRestClient also assumes the signing key ID is the public key, and the liveness config supports aliases and resolves the public key using the configured address. So neither looks like a replacement to me.

linea-besu/plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/txselection/LineaTransactionSelectorPlugin.java
nadeemb53: Okay, I'll make this change

linea-besu/plugins/linea-sequencer/sequencer/src/test/java/net/consensys/linea/sequencer/liveness/LineaLivenessTxBuilderTest.java
nadeemb53: This covers the interruption handling in signTransaction. When waiting for the async signer is interrupted, we wrap the error in IOException but must restore the thread’s interrupt flag so cancellation doesn't just disappear

linea-besu/plugins/linea-sequencer/build.gradle
nadeemb53: `tech.pegasys.teku.internal:async`
The shared Signer.sign() API returns SafeFuture, so consumers of signer-interfaces need that artifact

linea-besu/plugins/linea-sequencer/signer-test-plugin/build.gradle
nadeemb53: The separate project is intentional because the test needs a standalone plugin JAR copied into besu.plugins.dir. Putting it on the normal acceptance-test source set would place it on the test classpath and wouldn’t verify the real external plugin/service classloader boundary. A custom isolated source set could work, but it needs the same packaging and classpath separation.

linea-besu/plugins/linea-sequencer/sequencer/src/main/java/lineth/config/LineaLivenessServiceCliOptions.java
```diff
@@ -234,21 +248,15 @@ public static LineaLivenessServiceCliOptions fromConfig(
   @Override
   public LineaLivenessServiceConfiguration toDomainObject() {
     if (enabled) {
-      if (contractAddress == null
-          || signerUrl == null
-          || signerKeyId == null
-          || signerAddress == null) {
+      if (contractAddress == null || signerAddress == null) {
```
jonesho, line 241: Is signerAddress needed for custom signer? if no, we should remove the `signerAddress == null` check above

nadeemb53: It's needed: https://github.com/LFDT-Lineth/lineth-monorepo/blob/main/linea-besu/plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/liveness/LineaLivenessService.java#L109

linea-besu/plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/config/LineaLivenessServiceCliOptions.java
jonesho, line 122: If this `signerAddress` is needed in both web3jsigner and custom, the message should be updated too, its currently stated that it's related to web3signer only

linea-besu/plugins/linea-sequencer/sequencer/src/main/java/lineth/config/LineaLivenessServiceCliOptions.java
```diff
@@ -32,6 +33,8 @@ public class LineaLivenessServiceCliOptions implements LineaCliOptions {
 
   public static final String CONTRACT_ADDRESS = "--plugin-linea-liveness-contract-address";
 
+  public static final String SIGNER_TYPE = "--plugin-linea-liveness-signer-type";
+  public static final String SIGNER_NAME = "--plugin-linea-liveness-signer-name";
   public static final String SIGNER_URL = "--plugin-linea-liveness-signer-url";
```
jonesho, line 37: As an UX suggestion: as an user, I found it's not quite clear to see which configs are related to web3signer or custom or both, it'd be great if it could tell by the config name, e.g. `--plugin-linea-liveness-custom-signer-name` and `--plugin-linea-liveness-web3-signer-url`, the naming is up to you, wdyt?

linea-besu/plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/config/LineaLivenessServiceCliOptions.java
nadeemb53: makes sense

linea-besu/plugins/linea-sequencer/sequencer/src/main/java/lineth/config/LineaLivenessServiceCliOptions.java
nadeemb53: The generic naming's intentional here. It's part of the shared external-signer contract. signer-name should work for any signer from a custom provider. The URL, key ID, and TLS options are legacy Web3Signer-specific stuff, but we can't rename them without breaking existing configs. And if we only prefix the new option with `custom`, I think config looks inconsistent. I can update the descriptions for these if it works.

.github/workflows/maven-release.yml
```diff
@@ -56,14 +57,29 @@ jobs:
...
+            ./gradlew \
+              ':jvm-libs:linea:core:signer-interfaces:build' \
```
jonesho: I think the `:jvm-libs:linea:core:signer-interfaces:build` is redundant here, I tested it and it should not be needed as `sequencer-interfaces` has explicitly declared the api dependency to `signer-interfaces`

However, as you said, agree that the `:jvm-libs:linea:core:signer-interfaces:publish` is needed below for `:linea-besu:plugins:sequencer-interfaces:publish`:

Cusror:

> publish tasks are not transitive across project dependencies — publishing sequencer-interfaces will not publish signer-interfaces. Since it's an api dependency, consumers pulling sequencer-interfaces from the Maven repo need signer-interfaces published there too. So the explicit :jvm-libs:linea:core:signer-interfaces:publish is genuinely required there.

nadeemb53: Hmm the project dependency does get Gradle to build the signer JAR that sequencer-interfaces needs, so it's not strictly required just to compile. But it doesn't run signer-interfaces' full verification lifecycle like no tests, no Spotless. Since this workflow publishes both artifacts, I'd rather keep the explicit `signer-interfaces:build` in as a release gate. wdyt?

jonesho: Okay ic, can u put some comment on that, such that the next reviewers would not have the same question regarding this `:jvm-libs:linea:core:signer-interfaces:build` step

linea-besu/plugins/linea-sequencer/gradle/dist.gradle
```diff
@@ -38,7 +38,7 @@ jar {
   // Kotlin standard library classes are available at runtime when the plugin is loaded by Besu
   from {
     configurations.runtimeClasspath
-      .filter { it.name.startsWith('kotlin-stdlib') }
+      .filter { it.name == "kotlin-stdlib-${libs.versions.kotlin.get()}.jar" }
       .collect { zipTree(it) }
   }
```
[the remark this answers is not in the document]
nadeemb53: This repository resolves Kotlin 2.4.0

.github/workflows/maven-release.yml
fluentcrafter: I don't see this approach as sustainable or scalable. If tomorrow we have another dependency, this becomes a bit of a mess... 

would calling `${LIB_TO_RELEASE}:buildNeeded` solve that?

.github/workflows/maven-release.yml
```diff
@@ -56,14 +57,32 @@ jobs:
...
+          fi
 
       - name: Deploy to Local File System
         env:
           LIB_TO_RELEASE: ${{ github.event.inputs.libToRelease }}
           INPUT_VERSION: ${{ github.event.inputs.version }}
           GITHUB_TOKEN: ${{ secrets._GITHUB_TOKEN_RELEASE_ACCESS }}
-        run: ./gradlew "${LIB_TO_RELEASE}:publish" "-Pversion=${INPUT_VERSION}"
+        run: |
+          if [[ "$LIB_TO_RELEASE" == ':linea-besu:plugins:sequencer-interfaces' ]]; then
```
fluentcrafter: IMO, instead of adding ad hoc special cases in `maven-release.yml`, we should try to configure linea-besu:plugins:sequencer-interfaces in Gradle.build release task.

linea-besu/plugins/build.gradle
```diff
@@ -0,0 +1,17 @@
...
+ * This file is dual-licensed under either the MIT license or Apache License 2.0.
+ * See the LICENSE-MIT and LICENSE-APACHE files in the repository root for details.
+ *
+ * SPDX-License-Identifier: MIT OR Apache-2.0
+ */
+
+subprojects {
+  repositories {
+    maven {
+      url 'https://artifacts.consensys.net/public/teku/maven/'
```
fluentcrafter: why do we need this given it's already added by net.consensys.zkevm.kotlin-common-minimal-conventions.gradle ?

linea-besu/plugins/linea-sequencer/gradle/dist.gradle
```diff
@@ -38,7 +38,7 @@ jar {
   // Kotlin standard library classes are available at runtime when the plugin is loaded by Besu
   from {
     configurations.runtimeClasspath
-      .filter { it.name.startsWith('kotlin-stdlib') }
+      .filter { it.name == "kotlin-stdlib-${libs.versions.kotlin.get()}.jar" }
```
fluentcrafter: What's the need to change this?

linea-besu/plugins/sequencer-interfaces/src/main/java/linea/signing/NamedSignerProviderService.java
```diff
@@ -0,0 +1,31 @@
...
+ * coordinates {@code x || y}.
+ */
+public interface NamedSignerProviderService extends BesuService {
+  /**
+   * Resolves a signer by its provider-defined logical name.
+   *
+   * @param name logical signer name
+   * @return initialized secp256k1 signer
+   */
+  Signer<Secp256k1Signature> create(String name);
```
fluentcrafter: suggestion: `create` suggests that a new one may be created everytime, however the service provider my have cached service instances, and it should given that some of them have instantiation overhead due to key retrieval 

```suggestion
  Signer<Secp256k1Signature> get(String name);
```

fluentcrafter (commented):
We plan to migrate besu-plugins to Kotlin. So we should write new code in Kotlin instead of Java. 

Can you please do that, at lease for the easy bits?

linea-besu/plugins/build.gradle
nadeemb53: Needed currently because `signer-test-plugin` does not apply the Kotlin convention plugin, and repositories are project-scoped. But now that we're talking about the rewrite to kotlin, this would change.

linea-besu/plugins/linea-sequencer/gradle/dist.gradle
nadeemb53: Needed because the PR turns the sequencer into a Kotlin module, which pulls `kotlin-stdlib`, `kotlin-stdlib-jdk7`, and `kotlin-stdlib-jdk8` (2.4.0) into its runtime classpath.

I checked, and keeping `startsWith('kotlin-stdlib')` causes the jar task to blow up with a duplicate `META-INF/versions/9/module-info.class` error. Using the exact filter makes sure only the core 2.4.0 stdlib gets embedded.

Besu is still on Kotlin 2.2.21 for now. Once besu#10983 is merged and pulled in, we can drop this shading block completely.

jonesho (dismissed):
LGTM

linea-besu/plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/liveness/Web3SignerDigestSigner.java
Filter94: I'll rephrase the question then. Is there anything reusable between the two? Like the endpoint, digest size, etc. If yes, why is it not reused?

linea-besu/plugins/linea-sequencer/sequencer/src/test/java/net/consensys/linea/sequencer/liveness/LineaLivenessTxBuilderTest.java
Filter94: Is there a place that is polling the interruption status? I just don't understand how it becomes important for the first time in our codebase. Were we missing something important in the rest of the codebase?

linea-besu/plugins/linea-sequencer/build.gradle
Filter94: I see `lineth.kotlin-common-minimal-conventions` was imported instead

.github/workflows/maven-release.yml
Filter94: I think turning a freeform string in [this](https://github.com/LFDT-Lineth/lineth-monorepo/commit/72c65470c24cd4f6461b96c8fc8e780365055363#diff-1e1110e4b87ab2617e8052089ef9e5b403618114a3cf009ead9b3925d3bc89aaL14) commit into an options list was a mistake. We shouldn't need to modify this job when we add new gradle subproject. And any gradle subproject has to be publishable

linea-besu/plugins/besu-plugin-options-docgen/build.gradle
```diff
@@ -16,6 +16,11 @@ group = 'net.consensys.linea.besu.plugin'
 
 repositories {
   mavenCentral()
+  maven {
```
Filter94: Can't it onboard the minimal conventions too?

scripts/besu-plugin-options/lib.test.js
```diff
@@ -48,12 +48,12 @@ test("renderDefault uses em dash when unset", () => {
   assert.equal(renderDefault({ default: "1000" }), "`1000`");
 });
 
-test("rendered row count equals in-scope option count (63 = 51 + 12)", { skip }, async () => {
+test("rendered row count equals in-scope option count (65 = 53 + 12)", { skip }, async () => {
```
Filter94: Is there a way we could verify the same without modifying this file explicitly? Can we take some artifact as a source of truth? I think this test will guarantee fail on CI when a developer updates a besu plugin. And without CI we have no guardrails to prevent that

linea-besu/plugins/linea-sequencer/sequencer/src/test/kotlin/lineth/sequencer/liveness/Web3SignerDigestSignerTest.kt
```diff
@@ -0,0 +1,160 @@
...
+
+    assertThatThrownBy { signer.sign(ByteArray(32)).join() }
+      .hasRootCauseInstanceOf(IllegalStateException::class.java)
+      .hasRootCauseMessage("Web3Signer request failed with status 404")
+  }
+
+  private fun Sign.SignatureData.bytes(): ByteArray = r + s + v
+
+  private companion object {
+    const val SIGN_PATH = "/api/v1/eth1/sign/"
```
Filter94: Can we instead expose it as an internal static fields and import it here from the implementation?

linea-besu/plugins/sequencer-interfaces/src/main/kotlin/lineth/signing/NamedSignerProviderService.kt
```diff
@@ -0,0 +1,31 @@
...
+import org.hyperledger.besu.plugin.services.BesuService
+
+/**
+ * Provides transaction signers from a separately packaged Besu plugin.
+ *
+ * Implementations own provider-specific configuration and signer lifecycle. A returned signer's
+ * public key must be ready for synchronous access and encoded as the 64-byte unsigned secp256k1
+ * coordinates `x || y`.
+ */
+fun interface NamedSignerProviderService : BesuService {
```
Filter94: I don't follow what are we getting from `NamedSignerProviderService` turned into a service. This service is accessed exactly once when the plugin is registered and after that it's a dead weight.

My suggestion was to turn the Signer itself into a Service, because that way it can be injected into a Besu Plugin and reused by other plugins too

Filter94 (changes requested):

linea-besu/plugins/linea-sequencer/sequencer/src/main/java/net/consensys/linea/sequencer/liveness/Web3SignerDigestSigner.java
nadeemb53: The sign endpoint, the 32-byte digest requirement, the 65-byte signature format, and the request payload are all shared. Right now they live inside the coordinator's Vertx-based client, whereas the Besu side uses JDK HTTP/TLS and also has to resolve aliases to public keys. Depending on the coordinator module just to get at those constants would point the dependency the wrong way and drag Vertx into the Besu plugin.

I do agree the wire-protocol duplication should go. The right reusable piece is probably a transport-neutral `Web3SignerProtocol` helper in a shared JVM module that both HTTP implementations can use. I'd rather do that as its own change though, and leave their transport and config handling separate.

linea-besu/plugins/sequencer-interfaces/src/main/kotlin/lineth/signing/NamedSignerProviderService.kt
nadeemb53: I see your point. I'll make the required changes.

linea-besu/plugins/linea-sequencer/sequencer/src/test/java/net/consensys/linea/sequencer/liveness/LineaLivenessTxBuilderTest.java
nadeemb53: No explicit polling is required. This PR introduces `Future.get()` in `ECKeypairSignerAdapter` when bridging the async signer API to web3j’s sync signing API, which throws InterruptedException and clears the interrupt flag. Because we wrap that exception in IOException, we must restore the flag to avoid swallowing thread cancellation.

.github/workflows/maven-release.yml
```diff
@@ -10,15 +10,8 @@ on:
...
-        description: 'Library to release'
-        options:
-          - ':jvm-libs:linea:blob-compressor'
-          - ':jvm-libs:linea:blob-decompressor'
-          - ':jvm-libs:linea:blob-shnarf-calculator'
-          - ':jvm-libs:linea:linea-contracts:l1-rollup'
-          - ':jvm-libs:linea:linea-contracts:l2-message-service'
-          - ':linea-besu:plugins:sequencer-interfaces'
+        type: string
+        description: 'Library to release, e.g :jvm-libs:linea:blob-compressor'
```
Filter94, line 14: Just reverting the change exposes the vulnerability we initially had that was fixed in [this](https://github.com/LFDT-Lineth/lineth-monorepo/pull/3184) change. I believe Grade subproject coordinates string has to be validated, otherwise it's a regression

nadeemb53: I see

Filter94 (approved):

jonesho (approved):

PR #3669, opened 2026-07-29
fix(prover-ray): guard SSZ allocation sizes

gusiri:
This PR implements issue(s) #

### Checklist

  PR.

### DIFF
--- a/prover-ray/utils/ssz/ssz.go
+++ b/prover-ray/utils/ssz/ssz.go
@@ -59,6 +59,8 @@ const maxExtraDataBytes = 32
 
 const maxSSZSerializedBytes uint64 = 1<<32 - 1
 
+var maxInt = int(^uint(0) >> 1)
+
 // SSZ list/vector bounds mirrored from rollup_spec/stateless_input.py and
 // rollup_spec/canonical_ssz.py. The hand-written encoder must reject inputs
 // outside these bounds just like the reference remerkleable encoder does.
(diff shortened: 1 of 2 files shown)


prover-ray/utils/ssz/ssz.go
```diff
@@ -319,6 +328,20 @@ func addSSZLength(ctx string, total uint64, n int) (uint64, error) {
 	return total + add, nil
 }
 
+func checkedAllocCap(ctx string, n uint64) (int, error) {
+	if n > uint64(math.MaxInt) {
```
ivokub: Isn't it redundant check? `make` takes as input `int` anyway and `math.MaxInt` is the maximum value of int on platform.

In practice we cannot know ahead of time how much we can allocate - it depends how much memory there is left etc, so its a runtime decision. And we also know it won't be negative as we prepare the value ourselves. So I'm not sure these checks really add any value here.

ivokub (commented):
I'm not sure the PR is worth it -- it doesn't add any meaningful checks (only that length is less than MaxInt which is ensured by `int` type anyway).

ivokub:
> I'm not sure the PR is worth it -- it doesn't add any meaningful checks (only that length is less than MaxInt which is ensured by `int` type anyway).

And if any linter complains here, then imo its a linter bug and we should ignore it (//nolint etc)

gusiri:
@ivokub Yes, I agree. In practice we will not hit this overflow here. I initially avoided suppressing it because it is a CodeQL `security alert` rather than a lint warning, but I agree that it does not add much value.
I replaced it with a targeted suppression on that allocation line.

ivokub:
> @ivokub Yes, I agree. In practice we will not hit this overflow here. I initially avoided suppressing it because it is a CodeQL `security alert` rather than a lint warning, but I agree that it does not add much value. I replaced it with a targeted suppression on that allocation line.

Though which alert it is? I didn't see anything to fail in the CI. Somewhere in "code quality" tab?

ivokub (approved):

gusiri:
@ivokub Yes, it is in the Code Quality tab. 
The alert:
https://github.com/LFDT-Lineth/lineth-monorepo/pull/3600/checks?check_run_id=90734671419

PR #3679, opened 2026-07-31
feat(prover-ray): add the public-input-layout for the recursion

Soleimani193:
This PR implements the tailoring of the public inputs for a later recursion step. The messageBus public inputs are registered as the public inputs.

### Checklist

  PR.

---

> [!NOTE]
> **Medium Risk**
> This is a breaking API change to how public inputs are registered and ordered on the proof statement; incorrect layout or MessageBus ordering would break inter-shard recursion checks, though behavior is heavily tested.
> 
> **Overview**
> Replaces the flat `RegisterPublicInputs` list with a **`PublicInputLayout`** that groups statement cells by role (`MessageBus`, `ProgramVK`, `SharedRandomness`, `GuestPublicOutputs`, `IsLastShard`, etc.). Prove/verify still emit a flat `PublicInput`, but ordering is fixed by `PublicInputLayout.Cells()`; **`Unpack`** maps that wire format back to typed `PublicInputValues`. Registration is **`sys.PublicInputs.Register(field, cells...)`**, with duplicate and post–first-prove guards.
> 
> **`messagebus.Compile`** now registers each handle’s grand-product accumulator under **`MessageBusPI`** in alphabetical handle order so **`MessageBus[i]`** is the same logical handle on every shard for a future cross-shard/recursion check.
> 
> Tests cover layout unpack/duplicate rules, existing guest-output PI flows, and expanded cross-shard scenarios (including bidirectional traffic) asserting PI cells match accumulators and align by handle index.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aa85c158bb24985c5f04aae238bdba7da0eda795. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AlexandreBelling:
It is a bit overcomplicated for my tastes. I'll let you @ivokub make the cut on whether the abstraction is useful of not.

prover-ray/wiop/wiop_public_input_layout.go
```diff
@@ -0,0 +1,221 @@
...
+//	sys.PublicInputs.Register(GuestPublicOutputsPI, cellA, cellB)
+//
+// Struct-literal construction is also valid for simple cases:
+//
+//	sys.PublicInputs = &PublicInputLayout{GuestPublicOutputs: []*Cell{cell}}
+type PublicInputLayout struct {
+	// MessageBus holds one cell per registered message-bus handle, in
+	// alphabetical handle order (messagebus.Compile sorts handles before
+	// processing them, so the ordering is deterministic across all shards).
+	MessageBus []*Cell
```
AlexandreBelling: Better to have statically sized arrays here

prover-ray/wiop/wiop_public_input_layout.go
```diff
@@ -0,0 +1,221 @@
...
+	// SharedRandomnessContribution holds cells encoding this shard's additive
+	// contribution to the sharedRandomness.
+	SharedRandomnessContribution []*Cell
+	// GuestPublicOutputs holds cells encoding the guest program's public outputs.
+	GuestPublicOutputs []*Cell
+	// IsLastShard is a cell whose runtime value is field.One for the last
+	// shard and field.Zero otherwise. Nil if not used.
+	IsLastShard *Cell
+
+	// cachedCells and sealed are set once by the first publicInputCells call
```
AlexandreBelling: That does make sense to me. Why do we have a caching mechanism?

prover-ray/wiop/wiop_public_input_layout.go
```diff
@@ -0,0 +1,221 @@
...
+		cells = append(cells, l.IsLastShard)
+	}
+	return cells
+}
+
+// seal caches the result of Cells and marks the layout as sealed. Called once
+// by publicInputCells before the first Prove or Verify. Also validates nil
+// cells and duplicates as a safety net for layouts built via struct literals
+// (Register already prevents these when used). Idempotent.
+func (l *PublicInputLayout) seal() {
```
AlexandreBelling: Does not solve any issue

prover-ray/wiop/wiop_public_input_layout.go
```diff
@@ -0,0 +1,221 @@
...
+	"fmt"
+
+	"github.com/LFDT-Lineth/lineth-monorepo/prover-ray/maths/koalabear/field"
+)
+
+// PublicInputField identifies a semantic group within [PublicInputLayout].
+// Pass one to [PublicInputLayout.Register] to add cells to that group.
+type PublicInputField string
+
+const (
```
AlexandreBelling: This is completely opinionated by R5 while the wizard system is completely agnostic to it

prover-ray/wiop/wiop_public_input_layout.go
ivokub: I agree - we usually won't be calling more than once, so storing it in cache is unnecessary and adds complexity.

prover-ray/wiop/wiop_public_input_layout.go
```diff
@@ -0,0 +1,221 @@
...
+	l.sealed = true
+}
+
+// PublicInputValues carries the evaluated field-element values of
+// public inputs, organised by semantic role. It is produced by
+// [PublicInputLayout.Unpack] from the flat [PublicInput] returned by
+// [System.Prove].
+type PublicInputValues struct {
+	// MessageBus is indexed identically to [PublicInputLayout.MessageBus].
+	MessageBus                   []field.Gen
```
ivokub: Can we make it concrete to base or Ext for all fields?

I.e. I would think that ProgramVK would be `[]field.Element`, SharedRandomness as `field.Ext`, SharedRandomnessContribution as `field.Ext`, GuestPublicOutputs `[]field.Element` (but also as a fixed length, no?), IsLastShard `field.Element`.

Not sure about GrandProduct, but I'd think it would be extension element as well?

prover-ray/wiop/wiop_public_input_layout.go
```diff
@@ -0,0 +1,221 @@
...
+
+import (
+	"fmt"
+
+	"github.com/LFDT-Lineth/lineth-monorepo/prover-ray/maths/koalabear/field"
+)
+
+// PublicInputField identifies a semantic group within [PublicInputLayout].
+// Pass one to [PublicInputLayout.Register] to add cells to that group.
+type PublicInputField string
```
ivokub: We don't particularly need to use string descriptor here. It is fine if we just define
```
const(
    MessageBusPI uint = iota
    ProgramVkPI
    SharedRandomnessPI
    //
)

File: prover-ray/wiop/wiop_system.go [hunk 7]
```diff
@@ -44,15 +42,8 @@ type System struct {
...
-	// "statement". It is populated, in registration order, via
-	// [System.RegisterPublicInputs]. Public inputs are always cells (a
-	// column is exposed by opening positions into cells, see
-	// [ColumnPosition.Open]); a human-readable label, if desired, lives in the
-	// cell's own Context.Lable. Their values are carried separately from the
-	// [Proof] (in a [PublicInput], aligned to this order) and are received by
-	// [System.Verify] alongside the proof.
-	PublicInputs []*Cell
+	// PublicInputs defines the semantic structure of public inputs, mainly used for inter-shard consistency.
+	PublicInputs *PublicInputLayout
```
ivokub: Though one issue -- imo we shouldn't assume sharding always.

Imo it would be better to separate `PublicInputGuestOutputLayout` and the sharding related into `PublicInputShardLayout` etc

prover-ray/wiop/compilers/messagebus/messagebus.go
```diff
@@ -49,6 +49,17 @@ import (
...
+// PublicInputTag is the base tag under which this pass registers each handle's
+// accumulator cell as a public input. Handles are processed in alphabetical
+// order and the tag is suffixed by that index, so the handles of a shard get
+// MessageBus_0, MessageBus_1, … — retrieve one with
+// sys.LookupPublicInputByTag(PublicInputTag, i). The index is the handle's
+// position in [wiop.System.MessageBusHandles]; since every shard sorts its
+// handles the same way, index i denotes the same handle on every shard, which is
+// what lets the cross-shard layer match up by position. This numbering is why
+// [Compile] is single-invocation per system.
+const PublicInputTag wiop.PublicInputTag = "MessageBus"
```
AlexandreBelling, line 61: For the next PRs, we will also add the shared randomness and its partial contribution here

prover-ray/zkcdriver/risc5/public_inputs.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+// The base [wiop.PublicInputTag] values of the RISCV-ZKC public inputs. A role
+// spanning several cells is registered one cell at a time under a numeric
+// suffix, so its tags are Role_0 … Role_{n-1} (see
+// [wiop.System.RegisterPublicInputs]); a single-cell role is registered
+// unsuffixed.
+const (
+	MessageBusPI                   wiop.PublicInputTag = messagebus.PublicInputTag
+	ProgramVKPI                    wiop.PublicInputTag = "ProgramVerificationKey"
```
AlexandreBelling, line 15: One thing not figured out yet is that the FRI commitments are not cells. So there is a compatibility link to be written. Either we don't make them public-inputs and the 2-1 joiner guest program get them from the runtime directly or we write a small adapter creating a public-input cell equal to the commitment value.

prover-ray/zkcdriver/risc5/public_inputs.go
```diff
@@ -0,0 +1,61 @@
...
+	NumGuestPublicOutputs = 8
+
+	// NumSharedRandomness is the number of shared randomness cells. 8 because
+	// this is the state of a poseidon2 hasher.
+	NumSharedRandomness = 8
+
+	// NumSharedRandomnessContributions is not decided yet and is to be taken from
+	// another place. The negative value is an unset sentinel: it is deliberately
+	// unusable as a count or loop bound until it is given its real value.
+	NumSharedRandomnessContributions = -1
```
AlexandreBelling, line 35: It's really just a placeholder so that it can be updated by you @ThomasPiellard

prover-ray/wiop/wiop_public_input_layout.go
AlexandreBelling: Currently, in what I did public-inputs are mapped to cells. IMO, we could uniformize everything into basic field elements. We used to do that in the old version, but TBH I don't think this would simplify anything for the verifier, as you woud still need to repack everything into field extensions to do things with them anyway.

prover-ray/wiop/wiop_public_input_layout.go
AlexandreBelling: I kept a string descriptor for it as this is more printable. Also, they are sparsely defined (e.g. defined in several different places) so it would require a bit of lifting to make something consistent.

prover-ray/wiop/wiop_system.go
AlexandreBelling: I reverted to using an array of cells. From a side discussion, we will want to have homogeneity of the circuits so implicitly, all public-inputs will always be defined as of they were sharding but we will not use them for the 2-1 proof.

jonesho (approved):

PR #3689, opened 2026-08-03
feat(verifier-ray): PCS integration

Tabaie:
The PCS verifier works with recent versions of the prover, but the vanishing verifier cannot. Once the latter is brought up-to-date, we can have end-to-end tests and perhaps get rid of the separate PCS codegen fixture.


Tabaie:
Closing in favor of https://github.com/LFDT-Lineth/lineth-monorepo/pull/3692

PR #3692, opened 2026-08-03
feat(verifier-ray): PCS integration 2

YaoJGalteland:
This PR integrates full PCS/FRI verification into verifier-ray and removes the old raw-claim verification path.

The verifier now replays the shared Fiat–Shamir transcript, derives PCS challenges itself, rebuilds batch roots from transcript-bound commitments, authenticates committed openings through PCS/FRI, and routes those authenticated claims into vanishing verification. This closes the gap where PCS and vanishing could otherwise consume different values for the same committed columns.

The change also adds runtime layout reconstruction for committed dynamic columns, so a single baked PCS system can verify proofs across multiple dynamic module sizes.

The branch includes:

* supporting codegen changes
* dynamic-size transcript replay
* root and claim routing
* FRI/PCS verifier modules
* regenerated fixtures
* regression tests for dynamic-shift aliasing, no-fold `deep_alpha`, endpoint binding, malformed cell refs, and multi-size behavior

End-to-end testing is included. The generated `verify.zig` fixtures are built from real prover-ray proofs, and `verifier-ray/test/verifier_test.zig` runs those fixtures through the top-level `verifier.verify` path, covering:

* honest proof acceptance
* tampered proof rejection
* one baked system verifying multiple dynamic sizes


verifier-ray/src/crypto/fiat_shamir.zig
```diff
@@ -74,6 +74,14 @@ pub const Transcript = struct {
...
                 @compileError("fiat_shamir.randomManyIntegers: upper_bound must be a non-zero power of two");
         }
+        self.randomManyIntegersRuntime(out, upper_bound);
+    }
+
+    /// Runtime `upper_bound` variant, for the PCS layer whose codeword size is a
+    /// runtime function of the restricted FRI params. `upper_bound` must still be
+    /// a non-zero power of two (the FRI codeword size always is; the reconstructed
+    /// params guarantee it). Byte-identical squeeze order to the comptime form.
+    pub fn randomManyIntegersRuntime(self: *Transcript, out: []usize, upper_bound: usize) void {
```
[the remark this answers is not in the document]
YaoJGalteland: not a real issue, the  caller (verifier-ray/src/query/pcs.zig:392) passes codeword_size = 1 << params.log_codeword_size, which is always a non-zero power of two by construction.

verifier-ray/codegen/go.mod
```diff
@@ -17,7 +17,9 @@ require (
...
 	github.com/sirupsen/logrus v1.9.4 // indirect
 	github.com/x448/float16 v0.8.4 // indirect
-	golang.org/x/crypto v0.50.0 // indirect
+	golang.org/x/crypto v0.52.0 // indirect
 	golang.org/x/sync v0.20.0 // indirect
-	golang.org/x/sys v0.43.0 // indirect
+	golang.org/x/sys v0.45.0 // indirect
 )
+
+replace github.com/LFDT-Lineth/lineth-monorepo/prover-ray => ../../prover-ray
```
Tabaie: This is in `verifier-ray.yml`:
```yaml
on:
  pull_request:
    branches:
      - main
    paths:
      - 'verifier-ray/**'
      # Keep prover-ray off this trigger for now: verifier-ray uses a pinned
      # known-good prover-ray version and updates it manually to save CI runtime.
      # - 'prover-ray/**'
      # Keep arithmetization integration dependencies documented but non-blocking
      # while the RISC-V/zkc pipeline is still moving quickly.
      # - 'arithmetization/src/test/main.go'
      # - 'arithmetization/src/main/riscv/main.zkc'
      - '.github/workflows/verifier-ray.yml'
```
Should we add `prover-ray/**` as a trigger if we want to use a local copy rather than a pinned commit?

verifier-ray/src/crypto/fiat_shamir.zig
Tabaie: Better to do `element.value & (upper_bound-1)` instead

verifier-ray/codegen/pcs_zig_test.go
```diff
@@ -0,0 +1,298 @@
...
+// rejecting honest proofs. (P1b regression guard.)
+func TestDynamicModuleOrderFollowsSysModules(t *testing.T) {
+	sys := wiop.NewSystemf("dynorder")
+	r0 := sys.NewRound()
+	// Create modules in a known order: dynA, static, dynB.
+	dynA := sys.NewDynamicModule(sys.Context.Childf("dynA"), wiop.PaddingDirectionRight)
+	staticMod := sys.NewSizedModule(sys.Context.Childf("static"), 4, wiop.PaddingDirectionNone)
+	dynB := sys.NewDynamicModule(sys.Context.Childf("dynB"), wiop.PaddingDirectionRight)
+	_ = staticMod
+	_ = dynA.NewColumn(sys.Context.Childf("colA"), wiop.VisibilityOracle, r0)
```
Tabaie, line 125: Are the assignments necessary?

verifier-ray/codegen/actions.go
```diff
@@ -0,0 +1,111 @@
...
+	case *global.Verifier:
+		return true
+	case *logderivativesum.VerifierAction:
+		return true
+	case *lookuptologderivsum.ResultIsZeroVerifierAction:
+		return true
+	}
+	// The PCS opening action's concrete type is unexported, so it cannot be named
+	// in a type switch; match it by its fully-qualified type name instead.
+	return actionTypeName(action) == handledOpeningActionTypeName
```
ivokub: I think better would be to export, not handle using reflect.

verifier-ray/codegen/pcs_zig.go
```diff
@@ -0,0 +1,146 @@
...
+    .witness_map = &{{.WitnessName}},
+    .quotient_map = &{{.QuotientName}},
+    .batch_roots = &{{.RootsName}},
+    .zeta_coin_index = {{.System.ZetaCoinIndex}},
+};
+`
+
+// pcsIntArrayLiteral renders an int slice as a Zig array body `{ a, b, c }`
+// for one opened column's shift schedule.
+func pcsIntArrayLiteral(xs []int) string {
```
ivokub: Can use `intArray` from vanishing_zig.go instead

verifier-ray/src/query/pcs.zig
```diff
@@ -34,165 +35,401 @@ pub const Error = merkle.Error || fri.Error || error{
...
+            break :blk next;
+        };
+        routing.index_by_batch[batch_idx] = branch_idx;
+    }
+    return routing;
+}
+
+fn findRootIndex(roots: []const poseidon2.Digest, want: poseidon2.Digest) ?usize {
+    for (roots, 0..) |root, i| {
+        if (std.meta.eql(root, want)) return i;
```
ivokub: use poseidon2.eql.

verifier-ray/src/query/pcs.zig
```diff
@@ -34,165 +35,401 @@ pub const Error = merkle.Error || fri.Error || error{
...
     const final_alpha = transcript.randomExt();
-    if (comptime num_rounds > 0) challenges.fold_alphas[num_rounds - 1] = final_alpha;
+    challenges.deep_alpha = final_alpha;
+    if (num_rounds > 0) challenges.fold_alphas[num_rounds - 1] = final_alpha;
 
     transcript.updateExt(fri_proof.final_poly);
 
-    const codeword_size = comptime @as(usize, 1) << @intCast(params.log_codeword_size);
-    transcript.randomManyIntegers(&challenges.query_positions, codeword_size);
+    const codeword_size = @as(usize, 1) << @as(std.math.Log2Int(usize), @intCast(params.log_codeword_size));
```
ivokub: can just do `@as(usize, 1) << @intCast(params.log_codeword_size)`

verifier-ray/src/query/pcs.zig
```diff
@@ -34,165 +35,401 @@ pub const Error = merkle.Error || fri.Error || error{
...
+    /// opened size is 1 (num_rounds == 0), prover-ray still squeezes this final
+    /// alpha and uses it as alpha_DEEP even though there are no fold alphas.
+    deep_alpha: ext.Ext = ext.Ext.zero(),
     query_positions: []const usize,
     proof: OpeningProof,
+    /// Runtime dynamic-module sizes, in canonical dynamic-module order (the
+    /// same order `SizeSource.dynamic.index` indices into). Powers of two, and
+    /// each one must satisfy the emitted `min_size_log2` bound for every column
+    /// that references it.
+    module_sizes: []const usize = &.{},
```
ivokub, line 162: Perhaps we should use log2 of module sizes? Seems that down the line its more convenient that way. Right now there is imo a mismatch, for the codeword/plaintext size we use log2, but dynamic sizes are not.

verifier-ray/src/query/pcs.zig
```diff
@@ -34,165 +35,401 @@ pub const Error = merkle.Error || fri.Error || error{
...
+/// actual per-batch roots: distinct roots in input-opening order plus the
+/// input-tree branch index each batch maps to.
+pub fn InputRootRouting(comptime system: System) type {
+    return struct {
+        const Self = @This();
+        const batch_cap = @max(system.num_batches, 1);
+
+        distinct_count: usize = 0,
+        roots: [batch_cap]poseidon2.Digest = undefined,
+        index_by_batch: [batch_cap]usize = [_]usize{std.math.maxInt(usize)} ** batch_cap,
```
ivokub: So in this package we could get away with std import -- right now it seems we don't use anything which isn't compatible with zkvm, but having the import in general imo can lead to using some operations what we cannot do (i.e. IO etc.).

the maximum int value here we can possibly hardcode here, it is just a default value when not set.

For the log2 computation elsewhere in this package we can just store the dynamic module sizes in log2 (as we do anyway for FRI params the codeword size for example)

ivokub (changes requested):
In general it looks good. I haven't run full adversial review yet, but tests pass (both for honest and tampered cases), so I have confidence that its good. But we should definitely run a full agentic review as well.

Imo we can proceed with extracting the prover PR as well to export fields. In addition to what you have exported I think we can export the PCS opening verifier action struct so that we can just type assert and not use reflect+string comparison.

And I think we shoul also avoid the `std` package import. Right now it seems its fine here, but imo the problem generally is that this already imports a dependency which can be used for doing things which are incompatible in zkvm (i.e. reading files etc.). In most of our use cases imo we don't need it anyway (a la when we do shift, then we can just `@intCast` and the target type is inferred anyway).

Otherwise, looks good (though the note that we should separately run adversial review, but lets do it as a follow up after we have merged).

verifier-ray/src/query/pcs.zig
YaoJGalteland: module_sizes raw values are absorbed directly into the Fiat-Shamir transcript (byte-for-byte matching prover-ray's Runtime.AdvanceRound), and vanishing.zig uses the raw value directly as a field-exponent and root-of-unity order — only pcs.zig's internal consumption actually wants log2.

verifier-ray/src/query/pcs.zig
Tabaie: It's okay to use `std` in tests though right? Normally they would not be run in a zkVM.

Tabaie (dismissed):
I raised a few minor points. Looks great overall.

verifier-ray/codegen/actions.go
```diff
@@ -0,0 +1,108 @@
...
+// Any other action type — including new ones added to prover-ray later — trips
+// the error, forcing an explicit decision rather than a silent drop.
+func AssertAllVerifierActionsHandled(sys *wiop.System) error {
+	for _, round := range sys.Rounds {
+		for _, action := range round.VerifierActions {
+			if verifierActionIsHandled(action) {
+				continue
+			}
+			return &UnhandledVerifierActionError{
+				Type:  actionTypeName(action),
```
ivokub: I would just remove the reflection-based way and use `fmt.Sprintf("%T", action)` (it uses reflection internally still).

verifier-ray/testdata/generate/fri/go.mod
```diff
@@ -2,7 +2,7 @@ module github.com/consensys/linea-monorepo/verifier-ray/testdata/generate/fri
 
 go 1.25.7
 
-require github.com/LFDT-Lineth/lineth-monorepo/prover-ray v0.0.0-20260802202552-44452c8d0cdf
+require github.com/LFDT-Lineth/lineth-monorepo/prover-ray v0.0.0-20260805140308-d7137fe881bf
```
ivokub: I think now the prover side fixes in place we can just drop the `go.mod` for the `testdata/generate/fri` package and just use the one for `testdata/generate` (i.e. if we delete this go.mod then the one in the directory above will be used.)

verifier-ray/src/query/pcs.zig
ivokub: In tests std is fine (and kind of needed actually). It is only for the code which is eventually executed in zkvm.

The idea is to follow the principle of `no_std` in Rust, see the reasoning https://docs.rust-embedded.org/book/intro/no-std.html#bare-metal-environments. The same principle would apply here as well -- many parts of Zig std package is fine, but its just a precaution to avoid defining anything which makes running the verifier in zkvm impossible.

ivokub (dismissed):
Thanks for the changes. A few small points, but otherwise imo it is good.

verifier-ray/codegen/go.mod
ivokub: This is not relevant anymore, we don't replace prover-ray anymore (it was a placeholder).

verifier-ray/src/crypto/fiat_shamir.zig
YaoJGalteland: they’re equivalent, I used `% upper_bound` to match the same operation as in prover-ray. But yes, `& (upper_bound - 1)` is better.

ivokub (dismissed):

arijitdutta67 (dismissed):

ivokub (approved):

ivokub (approved):

PR #3694, opened 2026-08-03
fix(maru): declare jar task dependency on runtimeClasspath to fix manifest race

thedarkjester:
The Class-Path manifest attribute in maru/app/build.gradle is computed
  via a Provider that maps over configurations.runtimeClasspath.elements,
  deferring resolution to the jar task's execution phase. However, the
  jar task never declared a dependency on the tasks that build those
  classpath artifacts (e.g. :maru:linea-finalization-provider:jar), so
  under parallel execution Gradle could evaluate the manifest Provider
  before that task finished, failing with:

    Could not copy MANIFEST.MF ...
    Querying the mapped value of provider(java.util.Set) before task
    ':maru:linea-finalization-provider:jar' has completed is not supported

  Add dependsOn(configurations.runtimeClasspath) to the jar task to wire
  the correct task-graph edge. This only affects scheduling order when
  jar is actually requested — it does not resolve the configuration
  eagerly, so clean and other unrelated tasks are unaffected.

---

> [!NOTE]
> **Low Risk**
> Build-only Gradle task ordering change with no runtime or application logic impact.
> 
> **Overview**
> Fixes intermittent **`maru` `jar` failures** under parallel Gradle builds when writing **`MANIFEST.MF`**.
> 
> The **`jar`** task builds **`Class-Path`** from a lazy **`runtimeClasspath.elements`** provider. That provider was not tied to the tasks that produce project jars on that classpath (e.g. **`:maru:linea-finalization-provider:jar`**), so Gradle could evaluate the manifest before those tasks finished.
> 
> Adds **`dependsOn(configurations.runtimeClasspath)`** on **`jar`** so those producer tasks run first. Scheduling only applies when **`jar`** is requested; classpath resolution stays deferred so **`clean`** and unrelated tasks are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d6457f1beb04f1f5ac1ee9d89e606a07e0772ad8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/maru/app/build.gradle
+++ b/maru/app/build.gradle
@@ -104,6 +104,12 @@ jar {
   // jar task actually runs (execution phase), never during configuration. Resolving it eagerly
   // here (e.g. `.collect {}`) would force Besu dependency resolution on every invocation —
   // including `clean` — which is what previously dragged an on-demand Besu build into a clean.
+  // dependsOn(Configuration) wires the tasks that build runtimeClasspath's artifacts (e.g.
+  // :maru:linea-finalization-provider:jar) as task dependencies without resolving it here;
+  // without this, the manifest Provider below can be queried before those tasks finish under
+  // parallel execution, failing with "Querying the mapped value of provider(...) before task
+  // ':maru:linea-finalization-provider:jar' has completed is not supported".
+  dependsOn(configurations.runtimeClasspath)
   def classPathAttribute = configurations.runtimeClasspath.elements.map { locations ->
     locations
         .collect { it.asFile.name }


maru/app/build.gradle
```diff
@@ -104,6 +104,12 @@ jar {
   // jar task actually runs (execution phase), never during configuration. Resolving it eagerly
   // here (e.g. `.collect {}`) would force Besu dependency resolution on every invocation —
   // including `clean` — which is what previously dragged an on-demand Besu build into a clean.
+  // dependsOn(Configuration) wires the tasks that build runtimeClasspath's artifacts (e.g.
```
fluentcrafter, line 107: It feels to me that this just contradicts above comment no?

@jonesho it seems that the previous "fix" created another issue. Can you please follow up on this?

thedarkjester: @fluentcrafter I asked our friend that question before pushing.

```
No contradiction — the two are solving different problems, and I verified this empirically rather than just asserting it.

  What the comment warns against: calling .collect{} (or anything that iterates/materializes) directly on configurations.runtimeClasspath during configuration phase. That forces Gradle to fully resolve and build every producer project synchronously, right there in the script — including the Besu
  subproject (whose real artifacts are substituted in for the org.hyperledger.besu.internal:* coordinates via a project-dependency substitution). Since the jar { ... } block runs eagerly for every Gradle invocation on this project regardless of which task you actually asked for, that eager
  .collect{} was dragging a full Besu build into something as unrelated as clean.

  What dependsOn(configurations.runtimeClasspath) does: it only registers a task-graph edge — "if jar ends up in the execution plan, make sure these producer tasks run first." It doesn't touch the manifest Provider at all, and it doesn't resolve or build anything itself; the edge is only walked
  when Gradle assembles the execution plan for tasks that are actually requested. It's the same mechanism Gradle recommends for Test.classpath = configurations.runtimeClasspath — using a Configuration as a task input/dependency doesn't force premature resolution.

  I confirmed this isn't just theoretical:
  - Before my fix: ./gradlew :maru:app:jar --dry-run graph had no :maru:linea-finalization-provider:jar entry at all — the edge was genuinely missing (this is the actual bug).
  - After my fix: the edge appears, in the correct order, and ./gradlew :maru:app:clean --dry-run still completes in ~1s with zero Besu-related task activity — i.e., clean is unaffected.

  So the lazy-Provider fix (avoiding eager file materialization) and the dependsOn fix (correct task ordering) address two separate failure modes and don't conflict: the Provider stays lazy (only evaluated when jar executes), while dependsOn only affects scheduling when jar is actually going to
  run.
  ```

fluentcrafter (approved):

PR #3695, opened 2026-08-03
feat(arithmetization): weekly zkc metrics against the small reference block

### LINKED ISSUE #3693: feat: weekly zkc metrics against small reference block
Run tracer + constraints checker using zkc's main against the small reference stateless inputs. Produce week-over-week analysis.

OlivierBBB:
## Why

We had no week-over-week record of the RISC-V arithmetization, so drift stayed invisible until
someone tripped over it — max degree moving from d8 to d11, for instance. This adds a weekly job
that measures the arithmetization and appends a row to markdown tables in the repo.

**The markdown file is the ledger.** There is no separate data store.

## What it measures

Against `riscv-guests/l2-execution/test/testdata/stateless_input.ssz` — the small committed
reference block — with the guest built `KECCAK_ACCEL=true`:

1. `zkc compile --stats` on `main.zkc`
2. guest ELF → zkc JSON
3. `zkc trace --stats`
4. `zkc trace --stats --check`

Five tables in `arithmetization/docs/metrics/zkc-weekly-metrics.md`, newest row first: runs and
component versions (timestamp, weekly vs manual trigger, runner tier and size, arithmetization
commit, zkc ref and commit, link to the run); cost per task (outcome, wall clock, peak RSS, trace
cells); the constraint system (max degree, constraints by degree, lookups, complexity, static
cells); and the five most expensive modules, by complexity and by trace cells.

Runs Wednesday 03:00 UTC (05:00 Paris in summer) or on demand. The manual form takes a `zkc-ref`,
so "did a zkc change cost us?" can be answered without touching the arithmetization.

## First measurements

Seeded as the first row, measured locally with zkc `main` @ `49737ebb2275`:

| | |
|---|---|
| `compile --stats` | 2 s, 119 MiB |
| guest build | ~1 min |
| **`trace --stats`** | **3m42s, 7.70 GiB peak, 7 908 271 799 cells** |
| **`trace --stats --check`** | **still running when its 1 h cap fired, 12.14 GiB** |
| constraint system | max degree 8, 12 627 constraints, complexity 89 687, 6 387 lookups |

**Tracing the guest is cheap.** Three comments in the repo claimed that path needs ~15 GB or is
unimplemented; they are corrected here against measurement. Constraint *checking* is the expensive
half, and its true cost is still unknown — the first CI run, with a 240-minute cap, will tell us.

## Design notes

- **One job, one fixed runner tier**, deliberately: splitting across tiers would destroy the
  week-over-week comparability that is the entire point.
- **Trace and check are best-effort** with individual caps. A `TIMEOUT` is recorded as that week's
  data point rather than failing the job — knowing the ceiling *is* the measurement.
- **Failures name themselves**: `**FAIL** (rc 4)`, `**OOM/killed** (SIGKILL) … / 60.00 GiB`,
  `**CRASH** (SIGSEGV)`, `**FAIL** (3 failing)`, `skipped`. Peak RSS survives a kill.
- **A lost runner still records the week.** `if: always()` covers a failing step but not losing the
  runner, so a second small job appends a "run lost" placeholder, gated on the main job's
  `row_written` output so a red-but-recorded run cannot duplicate.
- **The timeout budget is validated before any measurement**, so raising `check-timeout-min` past
  the job cap fails fast instead of being silently truncated.
- Tables are located by an HTML marker, not by heading, so the surrounding prose stays editable.

## Plumbing worth knowing

- The tooling builds its own zkc into a private cache rather than calling `make install-zkc`, which
  overwrites `~/go/bin/zkc`.
- It calls `zkc` directly rather than through the guest Makefile's `exec` target, whose default
  `ZKC_EXEC_FLAGS ?= -q` no longer exists as a zkc flag. (Left for a follow-up, along with the
  execution-specs SSZ job that has been failing silently for three weeks.)
- A constraint failure does **not** change zkc's exit code, so pass/fail is grepped from the
  reported `failing …` lines.
- `time` wraps `timeout` and not the reverse, so a capped step still reports a real peak RSS.
- macOS's "peak memory footprint" is cumulative allocation, not a resident peak (154 GiB for a run
  whose true peak was 12 GiB), so the parser uses "maximum resident set size".

## Testing

Every failure rendering was reproduced against the renderer, not reasoned about: missing exit code,
rc 4, SIGKILL at 60 GiB, SIGSEGV, timeout, missing prerequisite, and the lost-run placeholder. The
first row was generated end-to-end from real measurements, and a second row confirmed newest-first
insertion. `go vet` clean; both workflows parse.

`weekly_zkc_metrics.sh` runs the same measurements locally, writing to a gitignored copy by default.


arithmetization/src/test/scripts/README-weekly-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+     JSON_EXT=/tmp/guest.json \
+     IN_BYTES="@$PWD/riscv-guests/l2-execution/test/testdata/stateless_input.ssz"
+
+zkc trace --stats         /tmp/guest.json arithmetization/src/main/riscv/main.zkc
+zkc trace --stats --check /tmp/guest.json arithmetization/src/main/riscv/main.zkc
+```
+
+## Gotchas this tooling encodes
+
+- **A constraint failure does not change zkc's exit code.** It reports the failing constraints and
```
DavePearce, line 126: This seems like something we should fix.

arithmetization/Makefile
```diff
@@ -56,13 +56,26 @@ install-zkc:
 	echo "Installing zkc from $(ZKC)..."; \
 	cd "$(ZKC)" && go install ./cmd/zkc
 
+# Build the zkc CLI to an explicit path instead of GOBIN, so a caller can pin one build without
```
DavePearce, line 59: I don't understand why this is necessary.

.github/workflows/arithmetization-weekly-zkc-metrics.yml
```diff
@@ -0,0 +1,481 @@
...
+          } >> "$GITHUB_ENV"
+          echo "zkc ref: requested '${requested}', using '${resolved}', sha '${sha_hint:-unknown}'"
+
+      # Build to an explicit path rather than GOBIN, so the measured binary is pinned and named
+      # rather than implicit on PATH, and pass it on with ZKC= (the override src/test/Makefile
+      # documents). Building from the checkout also gets vcs.revision stamped in for provenance.
+      - name: Install zkc
+        run: |
+          set -euo pipefail
+          make -C arithmetization install-zkc-to ZKC_BIN="${RUNNER_TEMP}/zkc-under-test"
```
DavePearce, line 146: I don't really understand why we are so concerned here with pinning the zkc version under test here.  I guess there is a concern this can interfere with other steps, which would then force them to reinstall zkc ??

DavePearce (approved):
I have approved since I cannot find anything specifically wrong.  However, this is introducing a lot of complex code which we will have to maintain.  It seems pretty clear to me that we cannot continue putting so much stuff into a combination of scripts, makefiles and workflow actions.  Instead, we need a single driver (e.g. written in `go`) which is structured so it can be understood and maintained.

PR #3697, opened 2026-08-03
fix(arithmetization): remove remaining -q flags from zkc workflows

### LINKED ISSUE #3696: fix(arithmetization): remove remaining -q flags from zkc workflows

OlivierBBB:
zkc dropped `-q`/`--quiet` upstream ([c8f15725](https://github.com/LFDT-Lineth/zkc/commit/c8f15725), first tagged in v1.2.23); verbosity now defaults to NONE, so quiet needs no flag. Three call sites still passed it and were broken against the pinned v1.2.25: the l2-execution guest Makefile's `ZKC_EXEC_FLAGS` default, the SSZ fixtures runner's competing `-zkc-flags` default, and `poseidon2` (where it made every *reject* case report PASS). All three now use an overridable `ZKC_EXEC_FLAGS` defaulting to `--fast`.

`verifier-ray` is deliberately untouched — its workflow pins v1.2.20, which predates the removal.

This is **one** cause of `arithmetization-execution-specs-ssz-fixtures.yml` being red, not the whole story: that job ran v1.2.20 on 07-13 and 07-20 where `-q` was valid, so only the 07-27 and 08-03 runs are attributable to it. Reproduced A/B — the fixed default exits 0, `-q` exits 2 via `unknown shorthand flag: 'q'`, matching those runs. `--gogen` is cleared as a suspect.

Why nobody could tell: `runGuest` discarded stderr and returned only `err == nil`, so a broken invocation looked identical to every fixture being invalid. It now prints the first failure's output, so the next run either passes or says what is wrong. Second commit adds the Slack notify job this weekly workflow never had — it has been red since its first run on 2026-07-09 with nobody informed.


DavePearce (dismissed):
I think it looks fine to me, though I noticed in one case `l2-execution/Makefile` it was turning of tracing by default and enabling fast mode instead.  I'm unsure whether this matters or not.

.github/workflows/arithmetization-execution-specs-ssz-fixtures.yml
```diff
@@ -36,3 +36,16 @@ jobs:
         env:
           ZKC_REF: ""
         run: make -C riscv-guests/l2-execution run-execution-specs-ssz-fixtures
+
+  # This job is scheduled and had no notification, so it stayed red for weeks without anyone
```
letypequividelespoubelles: I don't like most of claude's comment

arithmetization/src/test/zkc/poseidon2/Makefile
```diff
@@ -11,6 +11,11 @@
 
 SHELL := /bin/bash
 
+# Flags forwarded to `zkc execute`. Quiet needs no flag: zkc's verbosity defaults to NONE.
```
letypequividelespoubelles: rm claude's comment

arithmetization/src/test/zkc/poseidon2/Makefile
```diff
@@ -11,6 +11,11 @@
...
+# The field stays explicit on the commands below, so the test does not silently follow a change
```
letypequividelespoubelles: same

arithmetization/src/test/zkc/poseidon2/Makefile
```diff
@@ -11,6 +11,11 @@
...
+# of zkc's default.
+ZKC_EXEC_FLAGS ?= --fast
```
letypequividelespoubelles, line 14: I find the name weird, exec (in my mind ) is tracing, so !fast. Rename it ZKC_FAST_FLAGS or ZKC_PREFLIGHT_FLAGS ?

arithmetization/src/test/Makefile
```diff
@@ -40,8 +40,8 @@ ZKC             ?= zkc
 TEST ?=
 
 # Flags to use when invoking zkc exec via zkc-exec / elf-exec:
-# 	-q for quiet mode
 # 	--fast for fast mode (else tracing mode by default)
+# Quiet needs no flag: zkc's verbosity defaults to NONE, and guest printf appears only at -vvv.
```
letypequividelespoubelles: rm reference to 'quiet mode`it doesn't exist anymore

arithmetization/src/test/README.md
```diff
@@ -143,11 +143,11 @@ riscv-test elf-to-json BIN_EXT=asm/bin/test
 riscv-test elf-exec BIN_EXT=path/to/bin
 # Debug an already compiled ELF
 riscv-test elf-debug BIN_EXT=path/to/bin
-# Execute an already compiled ELF in quiet mode
-riscv-test elf-exec BIN_EXT=path/to/bin ZKC_EXEC_FLAGS=-q
+# Execute an already compiled ELF in fast mode (no tracing; quiet is the default)
```
letypequividelespoubelles: rm reference to 'quiet mode`it doesn't exist anymore

riscv-guests/l2-execution/Makefile
```diff
@@ -33,11 +33,13 @@ BIN      := $(MAKEFILE_DIR)zig-out/bin/$(NAME)
 # Sample fixture input used by exec/debug
 INPUT    ?= $(MAKEFILE_DIR)test/testdata/stateless_input.ssz
 
-# Flags forwarded to `zkc exec` by exec. Default -q (quiet: suppress the guest's per-instruction
-# printf). Override to pick the execution backend — e.g. "--quiet --gogen --fast" runs the guest via
-# zkc's generated-Go backend in fast mode, the runnability path used while the interpreter does not
-# implement tracing yet. "--quiet --check" runs the guest via the interpreter with tracing enabled.
-ZKC_EXEC_FLAGS ?=-q 
+# Flags forwarded to `zkc exec` by exec. Empty by default, which means the interpreter builds a
+# trace — the same behaviour as before, since the "-q" this used to carry only suppressed output and
```
letypequividelespoubelles: rm reference to 'quiet mode` it doesn't exist anymore

letypequividelespoubelles (changes requested):

arithmetization/src/test/zkc/poseidon2/Makefile
OlivierBBB: IIRC it's called ZKC_EXEC_FLAGS everywhere else ...

PR #3700, opened 2026-08-03
feat(maru): support custom validator signers

nadeemb53:
- add local/custom validator signer configuration and TOML parsing
- inject custom signer factories without exposing the P2P private key
- adapt asynchronous signers to the Besu `NodeKey` used by QBFT
- validate custom signer addresses against configured validator sets
- preserve signer lifecycle ownership and surface synchronous signing failures as failed futures

Stacked on #3666


maru/app/src/main/kotlin/maru/app/MaruAppFactory.kt
```diff
@@ -110,7 +119,9 @@ interface MaruAppFactoryCreator {
   ): LongRunningCloseable
 }
 
-class MaruAppFactory : MaruAppFactoryCreator {
+class MaruAppFactory(
+  private val validatorSignerFactory: ValidatorSignerFactory = DefaultValidatorSignerFactory,
```
Filter94: Why `validatorSignerFactory` goes into `MaruAppFactory`'s constructor and `p2pNetworkFactory` goes into `create` method?

maru/app/src/main/kotlin/maru/app/MaruApp.kt
```diff
@@ -49,14 +48,15 @@ interface LongRunningCloseable :
...
-class MaruApp(
+class MaruApp internal constructor(
   val config: MaruConfig,
   val beaconGenesisConfig: ForksSchedule,
   clock: Clock = Clock.systemUTC(),
   // This will only be used if config.p2pConfig is undefined
   val p2pNetwork: P2PNetwork,
   val beaconChain: BeaconChain,
-  private val privateKeyProvider: () -> ByteArray,
+  private val validatorNodeKey: NodeKey?,
```
Filter94: I think it's wrong to introduce a dependency for a third party lib in MaruApp's constructor.

For Vertx and Web3j, it's more of a shortcut, rather than a good architectural decision.

Why can't we use `Signer` here? The public key is available by `managedValidatorSigner!!.signer.publicKey()` and the private key is unavailable, only the signing capability. And that should be enough.

Also, I don't think we have to depend on a `ManagedValidatorSigner` class here. We should instead introduce `CloseableSigner` interface IMO

maru/app/src/main/kotlin/maru/app/MaruAppFactory.kt
Filter94: Do we really need to inject this factory or `MaruAppFactory::create` can accept the `CloseableSigner`, same as other overrides?

maru/app/src/main/kotlin/maru/app/MaruAppFactory.kt
```diff
@@ -139,17 +150,74 @@ class MaruAppFactory : MaruAppFactoryCreator {
...
     val l2EthWeb3j: Web3j? =
       config.forkTransition.l2EthApiEndpoint?.let {
         Web3j.build(HttpService(it.endpoint.toString()))
       }
 
     checkL2EthApiEndpointAndForks(clock, beaconGenesisConfig, l2EthWeb3j)
 
-    val blockHashing = ForkAwareBlockHashing(beaconGenesisConfig)
-
-    config.persistence.dataPath.createDirectories()
```
Filter94, line 151: 1. Why are we removing the `createDirectories` call from here?
2. Maybe it has to be a responsibility of the app itself?

maru/app/src/main/kotlin/maru/app/MaruAppFactory.kt
```diff
@@ -347,6 +416,79 @@ class MaruAppFactory : MaruAppFactoryCreator {
...
+      try {
+        managedValidatorSigner.close()
+      } catch (closeError: Throwable) {
+        error.addSuppressed(closeError)
+      }
+      throw error
+    }
+  }
+
+  private fun validateValidatorIdentity(
```
Filter94: This seems alien. Why are we doing this validation in the factory and not doing any other validations here? I don't think it's a responsibility of the factory

maru/app/src/main/kotlin/maru/app/QbftProtocolValidatorFactory.kt
```diff
@@ -28,13 +28,14 @@ import maru.p2p.SealedBeaconBlockBroadcaster
...
 import net.consensys.linea.metrics.MetricsFacade
+import org.hyperledger.besu.cryptoservices.NodeKey
 import org.hyperledger.besu.plugin.services.MetricsSystem
 import tech.pegasys.teku.ethereum.executionclient.web3j.Web3JClient
 import java.time.Clock
 
 class QbftProtocolValidatorFactory(
   private val qbftOptions: QbftConfig,
-  private val privateKeyBytes: ByteArray,
+  private val nodeKey: NodeKey,
```
Filter94: I think it has to be a signer here

maru/app/src/main/kotlin/maru/app/MaruApp.kt
```diff
@@ -305,22 +319,21 @@ class MaruApp(
...
-            is QbftConsensusConfig -> configuration.validatorSet
-            else -> throw IllegalArgumentException("")
-          }
-        }.toSet()
-    if (!validatorsFromAllForks.contains(localValidator)) {
-      log.warn(
-        "localValidator={} isn't found in any of validatorSet-s in any of the Forks in the Genesis file!",
-        localValidator,
-      )
+internal fun closeAll(vararg closeActions: () -> Unit) {
```
Filter94: I think this utility belongs to `Helpers`

maru/app/src/main/kotlin/maru/app/ValidatorSignerFactory.kt
```diff
@@ -0,0 +1,51 @@
...
+import maru.config.ValidatorSignerConfig
+import maru.config.ValidatorSignerType
+import org.hyperledger.besu.cryptoservices.NodeKey
+import java.util.concurrent.atomic.AtomicBoolean
+
+fun interface ValidatorSignerFactory {
+  fun create(config: ValidatorSignerConfig): ManagedValidatorSigner
+}
+
+class ManagedValidatorSigner(
```
Filter94: Is Maru the only place where we care about closing the signer?

maru/app/src/main/kotlin/maru/app/ValidatorSignerFactory.kt
```diff
@@ -0,0 +1,51 @@
...
+  private val closed = AtomicBoolean()
+
+  override fun close() {
+    if (closed.compareAndSet(false, true)) {
+      closeAction()
+    }
+  }
+}
+
+internal class ValidatorSignerResource(
```
Filter94: So it has both public `NodeKey` and `ManagedValidatorSigner` which could potentially not even relate to the same keypair? It's a very brittle design that partially defeats the point of `Signer` interface IMO

maru/consensus/src/main/kotlin/maru/consensus/qbft/adapters/SignerSecurityModule.kt
```diff
@@ -0,0 +1,112 @@
...
+  override fun sign(bytes: ByteArray): SafeFuture<Secp256k1Signature> {
+    require(bytes.size == Bytes32.SIZE) {
+      "Validator signer input must be ${Bytes32.SIZE} bytes, got ${bytes.size}"
+    }
+    val signature = SecpCrypto.signatureAlgorithm.sign(Bytes32.wrap(bytes.copyOf()), keyPair)
+    return SafeFuture.completedFuture(Secp256k1Signature(signature.r, signature.s))
+  }
+}
+
+internal class SignerSecurityModule(
```
Filter94, line 26: We are making an adapter to adapt `Signer` to a third party interface, which we use instead of our interface. I believe `QbftFinalStateAdapter` is the only place where it's necessary. In the rest of the places we can use `Signer` and `CloseableSigner`

maru/consensus/src/main/kotlin/maru/consensus/qbft/adapters/SignerSecurityModule.kt
Filter94, line 1: Can we match the file name to the main class in this file?

maru/consensus/src/test/kotlin/maru/consensus/qbft/adapters/SignerSecurityModuleTest.kt
```diff
@@ -0,0 +1,136 @@
...
+      }
+
+    assertThatThrownBy { SignerSecurityModule(signer).sign(Bytes32.wrap(digest)) }
+      .isInstanceOf(SecurityModuleException::class.java)
+      .hasMessageContaining("Validator signing failed")
+      .hasCause(providerError)
+  }
+
+  @Test
+  fun `adapter preserves interruption while waiting for a signature`() {
```
Filter94, line 100: What is the point of this test? Why is it important to "preserve interruption"?

maru/app/src/main/kotlin/maru/app/MaruAppFactory.kt
nadeemb53: ValidatorSignerFactory is fixed by the embedding distribution, while p2pNetworkFactory is an existing per-creation test override. That explains the difference, although I agree the API is inconsistent.
I'll fix this in an upcoming commit.

nadeemb53: The complication is that `MaruAppCli` parses `signer-type` and `signer-name` inside the call path, so the embedding distribution cannot resolve a signer beforehand without duplicating. I’ll retain a resolver/factory at this boundary, but pass only the resulting closeable signer through app construction.

maru/app/src/main/kotlin/maru/app/MaruAppFactory.kt
nadeemb53: The call was moved, not removed. It now runs in the outer create immediately before getOrGeneratePrivateKey, because signer composition happens earlier and may require the local key.

The factory already owns key and database setup, so I’d keep that responsibility there.

maru/app/src/main/kotlin/maru/app/MaruAppFactory.kt
nadeemb53: Hmm I'll have the factory delegate the logic to a focused validator/helper

maru/app/src/main/kotlin/maru/app/ValidatorSignerFactory.kt
nadeemb53: Besu also closes its provider registry during plugin shutdown. In Maru, the enterprise composition creates the registry per app, so the app owns closing it.

maru/consensus/src/test/kotlin/maru/consensus/qbft/adapters/SignerSecurityModuleTest.kt
nadeemb53: Besu’s signing API is synchronous, so the adapter waits on Future.get(). When get() throws InterruptedException, it clears the thread’s interrupt flag. The adapter must restore it before wrapping the exception so cancellation and shutdown can still observe the interrupt.

maru/app/src/main/kotlin/maru/app/ValidatorSetMembership.kt
```diff
@@ -0,0 +1,79 @@
...
+            else ->
+              throw IllegalArgumentException(
+                "Unsupported consensus configuration: ${configuration::class.qualifiedName}",
+              )
+          }
+        }.toSet()
+
+    if (validator !in validatorsFromAllForks) {
+      if (qbftConfig.validatorSigner.type == ValidatorSignerType.CUSTOM) {
+        throw IllegalArgumentException(
```
Filter94: Why does it throw only when validator signer type is custom?

maru/app/src/main/kotlin/maru/app/ValidatorIdentityValidator.kt
```diff
@@ -0,0 +1,79 @@
...
+import maru.consensus.ForkSpec
+import maru.consensus.ForksSchedule
+import maru.consensus.QbftConsensusConfig
+import maru.core.Validator
+import maru.crypto.SecpCrypto
+import org.apache.logging.log4j.LogManager
+import org.apache.tuweni.bytes.Bytes
+import org.hyperledger.besu.ethereum.core.Util
+
+internal object ValidatorIdentityValidator {
```
Filter94: `ValidatorIdentityValidator` has a function `validate` which validates the signer and `validatorFor` which returns a `Validator` object, which doesn't have `validate` method, but rather represents Protocol's Validator role.

Can you please work on some better naming?

maru/app/src/main/kotlin/maru/app/QbftProtocolValidatorFactory.kt
```diff
@@ -34,7 +36,7 @@ import java.time.Clock
 
 class QbftProtocolValidatorFactory(
   private val qbftOptions: QbftConfig,
-  private val privateKeyBytes: ByteArray,
+  private val validatorSigner: Signer<Secp256k1Signature>,
```
Filter94: Everything here is related to the validator, so `validatorSigner` feels like a poor name for me. IMO, `signer` is enough

maru/app/src/main/kotlin/maru/app/MaruAppFactory.kt
```diff
@@ -139,17 +148,74 @@ class MaruAppFactory : MaruAppFactoryCreator {
     log.info("configs={}", config)
     log.info("beaconGenesisConfig={}", beaconGenesisConfig)
 
+    val blockHashing = ForkAwareBlockHashing(beaconGenesisConfig)
+
+    config.persistence.dataPath.createDirectories()
+    val privateKey = getOrGeneratePrivateKey(config.persistence.privateKeyPath)
+    val validatorSigner =
+      createValidatorSigner(
```
Filter94: 1. Is it possible to instantiate the signer inside the app, after we determine, whether a node is a validator or a follower?
2. Why the logic to instantiate the signer isn't just a single call to `validatorSignerFactory`? Why do we have 2 levels of branching for creating the signer?

maru/app/src/test/kotlin/maru/app/ValidatorSignerFactoryTest.kt
```diff
@@ -0,0 +1,209 @@
...
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.io.TempDir
+import java.net.URI
+import java.nio.file.Path
+import java.time.Clock
+import java.time.Instant
+import java.time.ZoneOffset
+import kotlin.time.Duration.Companion.seconds
+
+class ValidatorSignerFactoryTest {
```
Filter94: Do we really need `MaruAppFactory` to be involved here? If we move the logic related to the signer creation into the `SignerFactory` out of `MaruAppFactory` I believe the testing could be simplified and more focused

maru/app/src/main/kotlin/maru/app/ValidatorIdentityValidator.kt
```diff
@@ -0,0 +1,79 @@
...
+    require(encodedPublicKey.size == PUBLIC_KEY_SIZE) {
+      "Validator signer public key must be $PUBLIC_KEY_SIZE bytes, got ${encodedPublicKey.size}"
+    }
+    val publicKey =
+      try {
+        SecpCrypto.signatureAlgorithm.createPublicKey(Bytes.wrap(encodedPublicKey))
+      } catch (error: Exception) {
+        throw IllegalArgumentException("Invalid validator signer public key", error)
+      }
+    require(SecpCrypto.signatureAlgorithm.isValidPublicKey(publicKey)) {
```
Filter94: This is a partial copypaste from `SignerSecurityModule`. It it possible to reuse this code?

maru/consensus/src/main/kotlin/maru/consensus/qbft/adapters/SignerSecurityModule.kt
```diff
@@ -0,0 +1,92 @@
...
+    val signature =
+      try {
+        signer.sign(dataHash.toArray()).get()
+      } catch (error: InterruptedException) {
+        Thread.currentThread().interrupt()
+        throw SecurityModuleException("Interrupted while waiting for validator signature", error)
+      } catch (error: Exception) {
+        throw SecurityModuleException("Validator signing failed", unwrapCompletionError(error))
+      }
+    return object : Signature {
```
Filter94: Why don't we return a named adapter class here?

maru/app/src/main/kotlin/maru/app/ValidatorSetMembership.kt
nadeemb53: This was done for backwards compatibility. Local signer mismatches previously logged a warning, while custom signer support is new, so a mismatched external key fails startup.

maru/app/src/main/kotlin/maru/app/MaruAppFactory.kt
nadeemb53: Signer construction belongs in `MaruAppFactory`, not MaruApp. Followers already skip signer creation. I agree the LOCAL/CUSTOM branching should move into the signer factory.

maru/app/src/main/kotlin/maru/app/ValidatorSetMembership.kt
Filter94: If we want to make this stricter, I don't really see the reason not to do it for the regular validators, because the inconsistency here is worse than backwards incompatibility. Also I don't see why would anyone rely on this behavior and have a validator configured that doesn't actually participate in the protocol as a validator

maru/app/src/main/kotlin/maru/app/ValidatorSignerFactory.kt
```diff
@@ -0,0 +1,71 @@
...
+    require(config.type == ValidatorSignerType.CUSTOM) {
+      "CustomValidatorSignerFactory is only used for custom validator signers"
+    }
+    throw IllegalArgumentException(
+      "Custom validator signer '${config.name}' requires an external CustomValidatorSignerFactory",
+    )
+  }
+}
+
+internal class ValidatorSignerFactory(
```
Filter94: Sorry for nitpicking, but I feel like it's important to nail it here.

I think we're missing the distinction between the `ValidatorSignerFactory` and `CustomValidatorSignerFactory`. At the first glance it looks like delegation, but these factories have slightly different focus. Can we reflect this difference in the naming?

maru/app/src/main/kotlin/maru/app/ValidatorSignerIdentity.kt
```diff
@@ -0,0 +1,20 @@
...
+ */
+package maru.app
+
+import linea.crypto.Secp256k1Signature
+import linea.crypto.Signer
+import maru.core.Validator
+import maru.crypto.decodeSecp256k1PublicKey
+import org.hyperledger.besu.ethereum.core.Util
+
+internal fun Signer<Secp256k1Signature>.toValidator(): Validator {
```
Filter94: Does it belong to the `app` or `crypto`?

nadeemb53: I'll move it to crypto

maru/app/src/main/kotlin/maru/app/ValidatorSignerFactory.kt
nadeemb53: rename `ValidatorSignerFactory` to `ValidatorSignerInitializer` then?

Filter94 (approved):

fluentcrafter (approved):

PR #3716, opened 2026-08-05
fix(coordinator): make runtime invariants explicit

nadeemb53:
- report the exact proof index when a required L2 execution or rollup proof response is missing
- validate blob compression proofs at submission and finalization boundaries
- replace the gas-cap non-null assertion with a named invariant
- add regression coverage for missing prover responses

Refs #3686


.github/workflows/coordinator-testing.yml
```diff
@@ -45,7 +45,9 @@ jobs:
           pnpm-install-options: '--frozen-lockfile --prefer-offline --filter contracts --ignore-scripts'
       - name: Build coordinator and Unit tests
         run: |
-          ./gradlew -V coordinator:app:buildNeeded
+          ./gradlew -V \
+            coordinator:app:buildNeeded \
+            :coordinator:clients:prover-client:riscv-client:test
```
jonesho, line 50: why do we need to call this `:coordinator:clients:prover-client:riscv-client:test` explicitly, it should run in `coordinator:app:buildNeeded`, no?

nadeemb53: `coordinator:app` does not currently depend on riscv-client
Both depend on file-based-client
Therefore, app:buildNeeded never schedules `riscv-client:test`. The explicit task ensures the regression tests added here run in CI.

fluentcrafter: I would remove this because it will be added to the app later. 
Managing module dependencies shall stay is Gradle's realm scope responsibility, not CI. 

If we want to include the build, lets then add the `:coordinator:clients:prover-client:riscv-client` to the APP, maybe as runtimeOnly or something?

coordinator/clients/prover-client/riscv-client/src/main/kotlin/linea/coordinator/clients/prover/riscv/RollupProverClient.kt
```diff
@@ -34,14 +34,18 @@ internal class FileBasedRollupProofRequestDtoMapper(
             chainId = chainId,
             blobs = request.blobs.map { it.fromDomainObject() },
             parentShnarf = request.parentShnarf.encodeHex(),
-            l2ExecutionProofs = l2ExecutionProofResponseDtos.map { it ->
+            l2ExecutionProofs = l2ExecutionProofResponseDtos.mapIndexed { index, response ->
+              val proofIndex = request.l2Executions[index]
+              val proofResponse = requireNotNull(response) {
+                "L2 execution proof response was not found for proofIndex=$proofIndex"
+              }
```
fluentcrafter: efficiency: lazy lookup 
```suggestion
              val proofResponse = requireNotNull(response) {
                "L2 execution proof response was not found for proofIndex=${request.l2Executions[index]}"
              }
```

Please apply in other places

coordinator/clients/smart-contract-client/src/main/kotlin/net/consensys/linea/contract/l1/Web3JLinethRollupSmartContractClient.kt
```diff
@@ -275,3 +277,10 @@ class Web3JLinethRollupSmartContractClient internal constructor(
       }
   }
 }
+
+private fun List<BlobRecord>.requireCompressionProofs(operation: String) =
```
fluentcrafter: this name is misleading: 
```suggestion
private fun List<BlobRecord>.getCompressionProofs(operation: String) =
```

coordinator/clients/smart-contract-client/src/test/kotlin/net/consensys/linea/contract/l1/Web3JLinethRollupSmartContractClientTest.kt
```diff
@@ -0,0 +1,60 @@
...
+import org.mockito.kotlin.any
+import org.mockito.kotlin.anyOrNull
+import org.mockito.kotlin.doReturn
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.spy
+import org.mockito.kotlin.whenever
+import org.web3j.protocol.Web3j
+import tech.pegasys.teku.infrastructure.async.SafeFuture
+
+class Web3JLinethRollupSmartContractClientTest {
```
fluentcrafter: Did you write this test, or was it AI-generated?

IMO, this test adds maintenance overhead without real value...

Why, they test more "concrete implementation" than "soundness/correctness". Unfortunately, for soundness/correctness, we need an integration test where true calls are sent to a real smart contract on an execution client. 

@Filter94 WDYT?

nadeemb53: All 3 PRs are AI-generated completely. I have driven them to make sure they follow the general guidelines and reviewed them.

coordinator/clients/smart-contract-client/src/main/kotlin/lineth/contract/l1/Web3JLinethRollupSmartContractClient.kt
```diff
@@ -125,26 +125,34 @@ class Web3JLinethRollupSmartContractClient internal constructor(
...
   override fun submitBlobs(blobs: List<BlobRecord>, gasPriceCaps: GasPriceCaps?): SafeFuture<String> {
-    return getVersion()
-      .thenCompose { version ->
-        val function = Web3JLinethRollupFunctionBuilders.buildSubmitBlobsFunction(version, blobs)
-        web3jContractHelper.sendBlobCarryingTransactionAndGetTxHash(
-          function = function,
-          blobs = blobs.map { it.blobCompressionProof!!.compressedData },
-          gasPriceCaps = gasPriceCaps,
-        )
+    return SafeFuture.completedFuture(Unit)
```
fluentcrafter, line 128: IMO this makes the code unnecessarily more complex (and slightly inefficient) just to avoid a `!!`.

We shall be more critical of taking AI feedback. 

In this context, we cannot have nulls. 

1. The submission controller shall only fetch proven records
2. integration/e2e tests will catch this, so it won't make it to prod.
3. This change does not add extra safety/soundness; it just adds a better error message if it happens...

fluentcrafter (approved):

PR #3724, opened 2026-08-05
refactor(prover-ray): export PCS batch/layout helpers and OpeningVeri…

YaoJGalteland:
Exports FRILogInverseRate, FRINumQueries, FRIMaxCommittableSizeLog2, FRIStaticParams, BatchRef, CommittedBatches, GetLayout, RecoverBatchClaims, and OpeningVerifierAction from wiop/compilers/pcs. These are pure renames (no behavior change) so the verifier-ray codegen can consume the prover's own FRI/PCS parameters and batch layout directly, and can type-assert on OpeningVerifierAction instead of matching its type name via reflection.

### DIFF
--- a/prover-ray/wiop/compilers/pcs/pcs.go
+++ b/prover-ray/wiop/compilers/pcs/pcs.go
@@ -43,9 +43,9 @@ import (
 )
 
 const (
-	// friLogInverseRate is the log2 of the FRI blow-up factor (codeword size /
+	// FRILogInverseRate is the log2 of the FRI blow-up factor (codeword size /
 	// plaintext size).
-	friLogInverseRate = 1
+	FRILogInverseRate = 1
 )
 
 // friNumQueries is the number of FRI query openings. This is obtained from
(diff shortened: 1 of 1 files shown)


ivokub (dismissed):

prover-ray/wiop/compilers/pcs/pcs.go
```diff
@@ -89,6 +89,36 @@ func staticFRI() (fri.Params, []*fri.RSEncoder) {
 	return staticFRIParams, staticFRIEncoders
 }
 
+// FRILogInverseRate is the log2 of the FRI blow-up factor (codeword size /
+// plaintext size). Exported so out-of-package consumers — notably the
+// verifier-ray codegen — can reconstruct the per-proof FRI params from a
+// protocol's largest opened size (log_codeword = maxSize + FRILogInverseRate)
+// without duplicating the constant.
+const FRILogInverseRate = friLogInverseRate
```
ivokub: We can just make the `friLogInverseRate` public, no need to assign to new const?

prover-ray/wiop/compilers/pcs/pcs.go
```diff
@@ -89,6 +89,36 @@ func staticFRI() (fri.Params, []*fri.RSEncoder) {
 	return staticFRIParams, staticFRIEncoders
 }
 
+// FRILogInverseRate is the log2 of the FRI blow-up factor (codeword size /
```
ivokub: And also the comment is way to explicit -- we don't need here to describe that verifier-ray uses it. Also for the comments below

ivokub (approved):

PR #3738, opened 2026-08-06
feat(misc): Added flexible blobs spec

Filter94:
This PR implements issue(s) #

### Checklist

  PR.


rollup_spec/src/rollup_spec/prover_io/schemas/getZkRollupAggregationProofV1.request.schema.json
```diff
@@ -56,8 +56,12 @@
         "endFtxRollingHash",
         "endProcessedFtxNumber",
         "filteredAddressesHash",
-        "parentShnarf",
-        "endShnarf",
+        "parentDrh",
```
AlexandreBelling: Drh -> dataRollingHash ?

rollup_spec/src/rollup_spec/l1_rollup.py
```diff
@@ -95,26 +112,50 @@ class FinalizationSubmission:
...
+    """
     pi = submission.public_inputs
 
     if not verify_rollup_aggregation_snark(submission.proof, pi):
         raise Exception("invalid rollup-aggregation proof")
-    if pi.parent_shnarf != state.current_finalized_shnarf:
-        raise Exception("parentShnarf does not match current finalized shnarf")
-    if pi.end_shnarf not in state.submitted_shnarf_last_block_hashes:
-        raise Exception("endShnarf was not anchored by a blob submission")
+    if keccak256(prev_drh + _encode_offset(prev_offset)) != state.current_finalized_position_commitment:
```
AlexandreBelling: Just to be sure. Does that mean we can finalize in the middle of a blob?

rollup_spec/src/rollup_spec/l1_rollup.py
```diff
@@ -168,8 +209,10 @@ def finalize_rollup(state: LinethRollupState, submission: FinalizationSubmission
         if vk not in state.approved_vks:
             raise Exception("program VK is not approved")
 
-    state.current_finalized_shnarf = pi.end_shnarf
-    state.current_finalized_last_block_hash = state.submitted_shnarf_last_block_hashes[pi.end_shnarf]
+    state.current_finalized_position_commitment = keccak256(
+        pi.end_drh + _encode_offset(pi.end_offset)
```
AlexandreBelling: Just to be sure. We are doing the keccak of a concatenation, right?

rollup_spec/src/rollup_spec/l1_rollup.py
Filter94: Yes we can

rollup_spec/src/rollup_spec/l1_rollup.py
Filter94: Yes, `+` is plain concatenation

jonesho (dismissed):

AlexandreBelling (dismissed):

rollup_spec/src/rollup_spec/rollup.py
```diff
@@ -94,179 +99,209 @@ class TruncatedEthereumBlock:
...
+    by this proof — relevant only at the two ends of the touched range, never
+    per-chunk: `opaque_prefix_bytes` fills the start of the FIRST touched
+    chunk when `start_offset > 0`; `opaque_suffix_bytes` fills the end of the
+    LAST touched chunk when this proof's data doesn't reach `chunkSize`. Both
+    default to empty. The guest witnesses them as opaque bytes purely to
+    reconstruct each boundary chunk's full published content for KZG
+    verification — their content is never interpreted, only reproduced
+    (§2.2: "opaque boundary bytes").
+
+    Mid-chunk starts (`start_offset > 0`): `parent_data_rolling_hash` is already the dataRollingHash
```
jonesho, line 215: One thing a bit confused to me is the input `parent_data_rolling_hash` vs `boundary_prev_data_rolling_hash`: I understand that the `boundary_prev_data_rolling_hash` is `the dataRollingHash value before the first touched chunk` but why it's optional and only requires when `start_offset` > 0? And why `parent_data_rolling_hash` has a prefix `parent_` but not sth like `first_chunk_`, if it actually is the drh value *after* folding the first touched chunk?

Filter94: >but why it's optional

In case blob boundary exactly matches conflation boundary, the previous blob is simply not needed. Another case is when we intentionally ignore the parent blob's suffix (because it's invalid by mistake, for example) That's why it's optional

> only requires when start_offset > 0?

If parent blob is non-empty, then we must read it starting from some `start_offset != 0`. If we must read the whole blob starting from offset 0, it shouldn't be a parent blob, it should be the "owned" first blob.

> And why parent_data_rolling_hash has a prefix parent_ but not sth like first_chunk_

Because it really means the parent and in some cases it doesn't match the first chunk. In a sense that "hash of a blob, preceeding the bytes we're interested in in this Rollup execution". If blob boundary exactly matches the conflation boundary, parent hash becomes a blob that doesn't have any bytes that are related to the execution proofs verified in a Rollup execution

rollup_spec/src/rollup_spec/prover_io/schemas/getZkRollupAggregationProofV1.request.schema.json
Filter94: I made the expansion everywhere. I think it's better than an abbreviation

rollup_spec/src/rollup_spec/rollup.py
jonesho: Thanks for the explanation, now it is clear to me, one small suggestion: can you add description for `parent_data_rolling_hash` on the rollup request schema? maybe sth like:
```
"description": "The dataRollingHash value after the last touched chunk of the previous rollup proof, which is essentially the dataRollingHash value after the first touched chunk when startOffset > 0"
```

jonesho (dismissed):

rollup_spec/src/rollup_spec/rollup.py
Filter94: Can you please check if this is clear? https://github.com/LFDT-Lineth/lineth-monorepo/pull/3738/commits/0b9b6973d8a378044e1d7b77855441423e23c7b2

rollup_spec/src/rollup_spec/prover_io/schemas/getZkRollupProofV1.response.schema.json
```diff
@@ -56,7 +56,10 @@
         "endFtxRollingHash": { "$ref": "#/$defs/hash32" },
         "endProcessedFtxNumber": { "$ref": "#/$defs/uint" },
         "filteredAddressesHash": { "$ref": "#/$defs/hash32" },
-        "parentDataRollingHash": { "$ref": "#/$defs/hash32" },
+        "parentDataRollingHash": {
+          "description": "The dataRollingHash accumulator value at this proof's start stream position, paired with startOffset (§2.4, §3.1): a pure DA accumulator (Hash(prevDataRollingHash, chunkHash)) carrying no execution continuity. Chained proof-to-proof by checking it against the preceding proof's endDataRollingHash (§2.3); at the start of the aggregated range it is simply forwarded from the first proof, unchecked internally, for L1 to verify at finalization (§5).",
+          "$ref": "#/$defs/hash32"
```
jonesho: tbh, I found only the description in `rollup_spec/src/rollup_spec/prover_io/schemas/getZkRollupProofV1.request.schema.json` is useful, imo other places can be removed, wdyt?

rollup_spec/src/rollup_spec/prover_io/schemas/getZkRollupProofV1.request.schema.json
```diff
@@ -45,7 +45,10 @@
           "description": "Foreign bytes after endOffset in the LAST touched chunk only (§3.1); empty (0x) or absent otherwise.",
           "$ref": "#/$defs/hex"
         },
-        "parentDataRollingHash": { "$ref": "#/$defs/hash32" },
+        "parentDataRollingHash": {
+          "description": "The dataRollingHash accumulator value this proof starts from, paired with startOffset (§2.2, §3.1). If startOffset = 0, this is R_0 (the accumulator before the first touched chunk, folded forward normally from chunk 1). If startOffset > 0 (mid-chunk start), this is already R_1 (the accumulator after the first touched chunk), whose preimage is opened via boundaryPrevDataRollingHash before folding continues from chunk 2.",
```
jonesho: nit: `before folding continues from chunk 2 (if any).`

jonesho (dismissed):

AlexandreBelling (dismissed):

rollup_spec/src/rollup_spec/prover_io/schemas/getZkRollupProofV1.response.schema.json
Filter94: Ok

gauravahuja (approved):

fluentcrafter (approved):

PR #3743, opened 2026-08-07
feat(prover-ray): preflight helpers

ThomasPiellard:
Add a preflight hook to derive the shared randomness in the message bus compiler. The expected caller's sequence is now:

```
  // 1. declare MessageBus entries as part of normal constraint setup
  sys.NewMessageBusSend(ctx, originShard, handle, tab)
  sys.NewMessageBusReceive(ctx, originShard, handle, tab)

  // 2. wire the shared randomness before compiling
  messagebus.RegisterPreflightSeed(sys, sets, hasher)

  // 3. compile as usual
  messagebus.Compile(sys)
  grandproduct.Compile(sys)
```

  latestUnreducedParticipantRound mirrors the logic of the existing
  latestParticipantRound but works directly on sys.MessageBuses without
  needing the pre-built byHandle map, so it can run before Compile has
  grouped entries.
  
  For the MSHash, the old prover's MSHash is used, and implements the following additive hash interface:
  ```
type MSetHash [MSetHashSize]field.Element

// Insert adds msg to the accumulator. Panics on an empty msg.
func (m *MSetHash) Insert(msg ...field.Element)

// Remove removes msg from the accumulator. Panics on an empty msg.
func (m *MSetHash) Remove(msg ...field.Element)

// Add combines two accumulators in place: m += other.
func (m *MSetHash) Add(other MSetHash)
```

### CODE REVIEW

File: prover-ray/preflight/preflight.go [hunk 1]
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"github.com/LFDT-Lineth/lineth-monorepo/prover-ray/crypto/koalabear/fri"
+	"github.com/LFDT-Lineth/lineth-monorepo/prover-ray/maths/koalabear/field"
+)
+
+// CrossShardSet bundles one logical cross-shard column fragment with the RS
+// encoders needed to commit to it. Build one CrossShardSet per shard fragment
+// before calling [Run].
+type CrossShardSet struct {
```
bogdanbear: I would call this something along the lines of BusInputSet

prover-ray/wiop/compilers/grandproduct/crossshard.go
```diff
@@ -0,0 +1,105 @@
...
+// skipInShardCheck controls whether the message-bus compiler emits a per-shard
+// "product == 1" assertion. Set it to false when the permutation is balanced
+// within each shard (each shard's A multiset equals its own B multiset). Set
+// it to true when rows cross shard boundaries and the balance check must be
+// deferred to a cross-shard join layer.
+func CompileCrossShardPermutations[P any](
+	sys *wiop.System,
+	originShard string,
+	handle string,
+	perms []*wiop.TableRelationQuery,
```
bogdanbear: As discussed in the daily, the crossshard perms do not exist in our design

prover-ray/preflight/preflight.go
```diff
@@ -0,0 +1,54 @@
...
+func Run[P any](sets []CrossShardSet, hasher AdditiveHasher[P]) field.Octuplet {
+	acc := hasher.Identity()
+	for _, s := range sets {
+		cs := fri.Commit(s.Encoders, s.Table)
+		root := cs.Tree.Nodes[0]
+		a := hasher.Hash(root)
+		acc = hasher.Combine(acc, a)
+	}
+	return hasher.ToSeed(acc)
+}
```
bogdanbear, line 54: so this is where the preflight set is processed and turned into a hash. Makes sense!

prover-ray/wiop/compilers/grandproduct/crossshard.go
```diff
@@ -0,0 +1,105 @@
...
+	coinRound := maxRound.EnsureNext()
+	coinRound.RegisterPreSamplingHook(&preflightSeedHook[P]{sets: sets, hasher: hasher})
+
+	// Declare one Send entry per A-side table and one Receive entry per B-side
+	// table, all on the same handle so the message-bus compiler aggregates
+	// their contributions into a single GrandProduct per shard.
+	for qi, q := range perms {
+		qCtx := compCtx.Childf("perm-%d", qi)
+		for ti, tab := range q.A {
+			mb := sys.NewMessageBusSend(qCtx.Childf("send-%d", ti), originShard, handle, tab)
```
bogdanbear: We need something that registers the message bus, plugs in your shared coins, and then calls the messagebus compiler. We should not call send and receive in this PR.

bogdanbear (dismissed):

prover-ray/wiop/compilers/messagebus/messagebus_preflight.go
```diff
@@ -0,0 +1,68 @@
...
+	return best
+}
+
+// preflightSeedHook is a [wiop.ProverAction] that fires as a PreSamplingHook
+// on the coin round. It recomputes the shared Fiat-Shamir seed from the
+// pre-distributed cross-shard column sets and injects it via SetFSState so
+// that the subsequent α and β coins are identical on every participating shard.
+// Because [wiop.Runtime.AdvanceRound] runs PreSamplingHooks on both the prover
+// and the verifier, no separate verifier action is needed.
+type preflightSeedHook[P any] struct {
```
AlexandreBelling: Not a blocker but this will need to be exported by the verifier-ray

prover-ray/preflight/preflight.go
```diff
@@ -0,0 +1,54 @@
...
+// accumulates the sum A = Σ AdditiveHash(R_i), and converts A to a
+// [field.Octuplet] via [AdditiveHasher.ToSeed]. Every shard that holds the
+// same S_i data produces the same octuplet regardless of processing order,
+// because the group operation is commutative.
+//
+// The octuplet is used as the Fiat-Shamir seed: each shard's prover and
+// verifier call [wiop.Runtime.SetFSState] with it inside a
+// [wiop.Round.RegisterPreSamplingHook] so the shared challenges α and β are
+// derived from an identical state on every participating shard.
+package preflight
```
AlexandreBelling, line 16: Not a blocker, but it would be better grouping it with zkcdriver/risc5. Reason is to avoid too many top-level packages.

prover-ray/wiop/compilers/messagebus/messagebus_preflight.go
```diff
@@ -0,0 +1,68 @@
...
+	sets   []preflight.BusInputSet
```
AlexandreBelling, line 61: For the preflight itself you will want to use the full commitment armada. But here, you can get the value of the commitment by getting it from the runtime directly. Otherwise, it would be computed twice which is a no-no.

AlexandreBelling (dismissed):

prover-ray/preflight/preflight.go
ThomasPiellard: maybe we can move it in another PR, moving it to risc5 creates an import cycle

AlexandreBelling (approved):

bogdanbear (approved):

PR #3749, opened 2026-08-07
perf(prover-ray): 2x faster FRI commit

gbotrel:
FRI commit is now faster than native Plonky3 (AVX-512, Rayon, fat LTO) in every measured configuration. Roots bit-identical.

| shape, rate 2 | before | after | plonky3 |
|---|---:|---:|---:|
| 2^20 x 16 | 77.9 ms | **31.8 ms** | 65.7 ms |
| 2^16 x 256 | 17.1 ms | **12.0 ms** | 25.0 ms |
| 2^12 x 4096 | 17.9 ms | **10.9 ms** | 18.5 ms |
| 10 GiB (2^22 x 640) | 2.67 s | **1.40 s** | 1.76 s |

96 cores, cold one-shot processes, medians of 5.

What:
- Parallel Merkle levels + gnark-crypto's 16-lane state-seeded Poseidon2 compression ([gnark-crypto#871](https://github.com/Consensys/gnark-crypto/pull/871), pinned commit `000c30b45f17`).
- Field-specific COBRA bit-reversal selection over gnark-crypto's generic helpers.
- Rate-2 LDE: first half of the bit-reversed codeword is bitrev(input); second half is one half-size coset FFT.
- Leaves hash directly into tree node storage; codewords slab-allocated per size.
- Opt-in `WithConsumeWitness` + `GOMEMLIMIT`: 10 GiB peak RSS 31.6 -> ~23 GiB.

### Checklist

* [x] Focused tests, purego tests, formatting, and golangci-lint pass locally.
* [ ] If this change is deployed to any environment (including Devnet), E2E test coverage exists or is included in this PR.
* [x] No breaking API changes.


prover-ray/crypto/koalabear/fri/commitment.go
```diff
@@ -36,9 +36,32 @@ type CommitterState struct {
...
+}
+
+// WithConsumeWitness transfers ownership of the witness columns to Commit:
+// each plaintext column is released as soon as its codeword is computed, so
+// the witness and its full encoding never coexist and peak memory drops by
+// roughly the witness size on large batches. The caller must not use the
+// witness table after the call. The plaintext values remain available: for a
+// rate-two code the first half of every codeword is exactly the bit-reversed
+// plaintext column.
+func WithConsumeWitness() CommitOption {
```
gbotrel: maybe a bit premature, might end up trimming that

prover-ray/crypto/koalabear/fri/fri.go
```diff
@@ -655,19 +665,6 @@ func extLimbs(e field.Ext) [6]field.Element {
 	}
 }
 
-// mapExtToOctuplet converts a slice of field extensions into a slice of
-// octuplets, packing each extension's six coordinates into the first six
-// octuplet entries and leaving coordinates 6 and 7 zero. It is the slice-wise
-// inverse of octupletToExt.
-func mapExtToOctuplet(exts []field.Ext) []field.Octuplet {
```
YaoJGalteland, line 662: mapExtToOctuplet (inlined into buildTreeExt) is deleted, but the test file still calls it directly in TestOctupletExtRoundTrip/TestBuildTreeExtOpenRecover.

prover-ray/crypto/koalabear/fri/fri.go
```diff
@@ -655,19 +665,6 @@ func extLimbs(e field.Ext) [6]field.Element {
...
-	for i := range exts {
-		limbs := extLimbs(exts[i])
-		copy(res[i][:], limbs[:])
-	}
-	return res
-}
-
 // openRunningQueryExt builds the Merkle opening data for query index s across
 // running extension folding levels. Input-tree openings are carried separately.
 func (st *ProverState) openRunningQueryExt(s int) RunningQuery {
```
YaoJGalteland, line 683: this line still calls `gutils.BitReverse` directly instead of the local `bitReverse()` wrapper.

For `invTwiddles`, which is `[]field.Element`, the local wrapper should be faster for large inputs because it can use the COBRA path:

* `gutils.BitReverse[T]` falls back to the naive path when `arm64 || len(v) < 2^21 || sizeof(T) < 8`.
Since `field.Element` is 4 bytes, `sizeof(T) < 8` is always true, so calling `gutils.BitReverse` directly on `[]field.Element` always takes the naive path.
* For large inputs (`>= 2^21`), the local `bitReverse()` wrapper can take the faster COBRA path.

So I think this call should use the local wrapper as well.

prover-ray/crypto/koalabear/fri/bitreverse.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/LFDT-Lineth/lineth-monorepo/prover-ray/maths/koalabear/field"
+	gutils "github.com/consensys/gnark-crypto/utils"
+)
+
+// bitRevCobraMinLog2 is the naive/COBRA crossover measured on a 96-core
+// EPYC 9R45 with all cores bit-reversing concurrently (the Encode regime):
+// naive wins through 2^20, COBRA wins 1.4x at 2^21 and 2.1x at 2^22.
+const bitRevCobraMinLog2 = 21
+
+func bitReverse(v []field.Element) {
```
YaoJGalteland: would it be clearer to rename it to `bitReverseBase`? It only works on base-field elements, so the current name sounds more generic than it actually is. It would also avoid confusion with `gutils.BitReverse`.

YaoJGalteland: have you compared the performance with `gutils.BitReverse`?

On my machine (2× Intel Xeon Platinum 8488C / Sapphire Rapids, 192 vCPUs, 1.5 TiB RAM, Ubuntu 24.04), the local wrapper doesn’t seem to improve performance.

At `2^21`—the one size where the wrapper actually switches to COBRA while `gutils.BitReverse` stays on the naive path because of the 4-byte guard, the wrapper is actually around 20% slower.

So on this machine, it seems to be the opposite of the intended optimization.

YaoJGalteland (commented):
have you compared the performance across different FRI rates, e.g. `2, 4, 8, 16`?

I remember we used rate `16` for Vortex before. Is that also a plausible parameter for FRI, or is the practical range more constrained here?

prover-ray/crypto/koalabear/fri/bitreverse.go
gbotrel: yes let's keep simple for now (gnark-crypto exposes both now) and tune when we have real production benchmark / target

AlexandreBelling (dismissed):

prover-ray/crypto/koalabear/fri/bitreverse.go
```diff
@@ -0,0 +1,13 @@
+package fri
+
+import (
+	gutils "github.com/consensys/gnark-crypto/utils"
+)
+
+func bitReverse[T any](v []T) {
```
AlexandreBelling, line 7: Nit: seems a bit superfluous to have private wrappers for theses.

gbotrel: so this is the result of a round trip of experiments ; following @YaoJGalteland review I reverted some optimizations, I will need these when we tune them depending on the size and target architecture 👍

YaoJGalteland (approved):

AlexandreBelling (approved):

PR #3766, opened 2026-08-11
feat(prover-ray): shared randomness public input

AlexandreBelling:
## What

Makes the message-bus permutation challenges α and β derivable from a γ supplied
from outside the proof, so that several shards can agree on them, instead of each
shard drawing them from its own Fiat-Shamir transcript.

Opt-in, off by default:

    messagebus.Compile(sys, messagebus.CompileOptions{SharedRandomness: true})

When set, `Compile` declares γ (8 cells, round 0) and this shard's contribution to
it (328 cells, coin round) as public inputs, and registers a pre-sampling hook that
installs γ as the FS state immediately before α and β are drawn. `AdvanceRound` runs
pre-sampling hooks on prover and verifier alike, so both sides derive the same
challenges with no extra verifier action.

## Key changes

- `messagebus`: `CompileOptions.SharedRandomness`; γ + contribution public inputs;
  `SharedRandomnessSeedHook`; contribution prover action and its verifier check;
  `HasSharedRandomness` so assignment paths can ask whether γ applies.
- Registration lives inside `Compile` rather than in a separate exported call: it is
  the same call that fixes the coin round, so the hook and the coins can no longer
  land on different rounds — a divergence that was silent, not an error.
- `multiset_hashing`: drops the stateless `Hasher` struct for package-level
  `Hash`/`Combine`/`Identity`/`ToSeed`.
- `preflight`: `Run` loses the `AdditiveHasher[P]` interface and its type parameter
  and calls the multiset hash directly.
- `wiop`: `Runtime.CurrentRound` now means the same thing on both sides — the round
  the running action was registered on — by tracking the round during the
  verifier-action pass. Without it a verifier action mirroring a prover action reads
  the last round instead of its own and fails on any protocol whose coin round is not
  the final round (i.e. every PCS-compiled one).
- `zkcdriver`: `AssignFromTrace` / `AssignWithPreRead` take γ, assigned only when the
  system was compiled to expect it.

## Supersedes #3743

That PR landed an earlier take on the same feature (`preflight.AdditiveHasher`,
`messagebus.RegisterPreflightSeed`, `PreflightSeedHook`). Both are removed here.
Its hook recomputed γ *inside* the proof from every shard's column sets, which a
shard's prover cannot do — it holds only its own data — and a verifier can do even
less, holding none. γ arrives as a public input instead. Nothing else referenced
that API.

## Not yet wired

- γ is a zero placeholder. Producing a real one needs `preflight.Run` reworked to
  take per-shard inputs and be pipelined; the current signature is a stand-in.
- The zkc driver declares no message-bus entry yet, so `SharedRandomness: true` in
  its pipeline registers nothing today. Left on deliberately, so the seeded path
  engages by itself once the arithmetization emits bus entries.
- The contribution hashes the PCS commitments of the rounds preceding the coin
  round; reconciling that preimage with what `preflight.Run` hashes is follow-up.

## Testing

`go vet ./...` clean. All `wiop` packages pass, including five new tests covering
seeded/unseeded coin agreement, γ→coin sensitivity, the public-input surface, and
the round placement of both cell families. `zkcdriver` still exceeds a 45s test
timeout on its synced integration suite, unchanged by this PR and with no assertion
failures.


prover-ray/preflight/preflight.go
```diff
@@ -1,22 +1,24 @@
-// Package preflight implements the pre-phase that establishes a shared
```
ThomasPiellard, line 1: this package is never used

AlexandreBelling: Yes, but it will eventually be used once we wire it with the arithmetization to concretely compute the shared randomness.

prover-ray/crypto/koalabear/multiset_hashing/msethash.go
```diff
@@ -91,37 +92,35 @@ func (m *MSetHash) update(rem bool, msg ...field.Element) {
...
+// Hash maps a Merkle root (8 field elements) into a fresh accumulator by
+// inserting the root elements.
+func Hash(root field.Octuplet) MSetHash {
 	var m MSetHash
 	m.Insert(root[:]...)
 	return m
 }
 
-// Combine implements [preflight.AdditiveHasher].
-func (Hasher) Combine(a, b MSetHash) MSetHash {
```
ThomasPiellard, line 111: Combine is never used, yet it is this feature that guarantees that the randomness is the same between the shards; what are the shared randomness tests actually testing ?

prover-ray/wiop/compilers/messagebus/shared_randomness.go
```diff
@@ -0,0 +1,219 @@
...
+	for i := range rt.CurrentRound().ID {
+		if !rt.System.Rounds[i].HasCommitment {
+			logrus.Warnf(
+				"No commitment found for round: %v. Did you use a message bus? "+
+					"And did you reduce the current system using a PCS?", i)
+			continue
+		}
+
+		com := rt.Commitments[i]
+		hasher.WriteElements(com[:]...)
```
YaoJGalteland, line 165: The new shared-randomness contribution public input does not match the accumulation that preflight.Run uses to derive γ.

The new shared-randomness contribution public input computes: multisethashing.Hash(Poseidon2(R_0 || ... || R_{n-1})). So all roots are absorbed sequentially into one hash state. The result depends on concatenation order, and each root’s contribution can’t be isolated or combined independently across shards.

But preflight computes: γ_preflight = ToSeed( **Σᵢ** multisethashing.Hash(Rᵢ) ), Here each root is hashed independently and the hashes are then added, so shard roots can arrive in any order.

AlexandreBelling: Ah no, this one hash the commitments of round 0 (program commitment) and round 1 (message bus) together. This is not "one commitment per shard", a local-shard prover does not get the message bus data of the other shards in its witness. This implementation is not a concurent implementation of preflight.Run.

What is still under the hook is that preflight.Run assume a unique commitment per shard but that is inaccurrate. But we can see that later.

bogdanbear (approved):

prover-ray/wiop/compilers/messagebus/shared_randomness.go
bogdanbear: So, unlike in the old prover where round 0 was the only special one, now both round 0 and 1 are reserved for special purposes (program commitment and bus), right?

ThomasPiellard (approved):

PR #3768, opened 2026-08-11
feat(verifier-ray): add bound check

YaoJGalteland:
Mirror the bound check https://github.com/LFDT-Lineth/lineth-monorepo/pull/3767 on verifier-ray end

### DIFF
--- a/verifier-ray/codegen/coin_routing.go
+++ b/verifier-ray/codegen/coin_routing.go
@@ -27,6 +27,10 @@ type CoinRouting struct {
 	// byte-exact with the prover. Counted in the same order as the proof's
 	// module_sizes (VanishingSystem dynamic-module order).
 	DynamicModuleCount int
+	// ColumnSizeMaxSupported mirrors wiop.ColumnSizeMaxSupported: the maximum
+	// dynamic-module size a proof may declare. The Zig verifier rejects any
+	// module_sizes entry above this before it reaches downstream arithmetic.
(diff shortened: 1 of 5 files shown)


verifier-ray/codegen/coin_routing.go
```diff
@@ -27,6 +27,10 @@ type CoinRouting struct {
 	// byte-exact with the prover. Counted in the same order as the proof's
 	// module_sizes (VanishingSystem dynamic-module order).
 	DynamicModuleCount int
+	// ColumnSizeMaxSupported mirrors wiop.ColumnSizeMaxSupported: the maximum
+	// dynamic-module size a proof may declare. The Zig verifier rejects any
+	// module_sizes entry above this before it reaches downstream arithmetic.
+	ColumnSizeMaxSupported int
```
AlexandreBelling: This is a constant in prover-ray, so in practice you could also make it a constant in verifier-ray.

AlexandreBelling: Also, this would make the blast radius of the PR much narrower as you would not have to modify the test cases.

YaoJGalteland: Since this value is constant, I'll hardcode it. that keeps the few lines change isolated to root.zig.

AlexandreBelling (approved):

arijitdutta67 (approved):

Issue #3086, opened 2026-05-13
[RISCV] - coordinator - conflation support


fluentcrafter:
# Description

Adapt coordinator conflation flow for new RISC-V flow.

PR #3769, opened 2026-08-12
feat(coordinator): add StartingPoint sealed class to BlockCreationMonitor

gauravahuja:
#3086 

## Summary

- Replaces the `startingBlockNumberExclusive: Long` constructor parameter on `BlockCreationMonitor` with a `StartingPoint` sealed class (`ByBlockNumber` | `ByTimestamp`), allowing the coordinator to self-coordinate from a cutover timestamp without requiring external block-number resolution.
- `ByTimestamp` flow: polls the latest L2 block until its timestamp reaches the target, then binary-searches `[0, latestBlock]` in O(log n) RPC calls to find the exact first block at or after the timestamp.
- Updates `ConflationAppV1` and `ConflationBacktestingApp` call sites to use `ByBlockNumber`.
- Adds three new tests covering `ByTimestamp`: happy path, wait-for-cutover, and binary-search with non-trivial block times.

## Test plan

- [ ] `ByTimestamp should start monitoring from first block at or after cutover timestamp`
- [ ] `ByTimestamp should wait for L2 to reach cutover timestamp before starting`
- [ ] `ByTimestamp should walk forward to find exact first block when estimate undershoots`
- [ ] All existing `BlockCreationMonitorTest` tests pass

---

> [!NOTE]
> **Medium Risk**
> Touches core conflation block-monitor startup, where an incorrect start block could skip or mis-chain L2 blocks. Existing production call sites still use block-number start, which limits immediate blast radius.
> 
> **Overview**
> Lets `BlockCreationMonitor` start from a **cutover timestamp** instead of only a known block number, so the coordinator can self-resolve the first L2 block without external block-number lookup.
> 
> Replaces `startingBlockNumberExclusive` with a `StartingPoint` sealed class (`ByBlockNumberExclusive` | `ByTimestampInclusive`). The timestamp path waits until the latest L2 block reaches the target, then binary-searches for the first block at or after that timestamp and seeds parent-hash tracking from there.
> 
> Existing `ConflationAppV1` and `ConflationBacktestingApp` call sites keep block-number startup via `ByBlockNumberExclusive`. Adds tests for happy path, wait-for-cutover, and binary-search resolution.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2112e07771bc5a74b0b4deec6da2a6965407047d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/coordinator/app/src/main/kotlin/lineth/coordinator/blockcreation/BlockCreationMonitor.kt
+++ b/coordinator/app/src/main/kotlin/lineth/coordinator/blockcreation/BlockCreationMonitor.kt
@@ -23,7 +23,7 @@ import kotlin.time.Instant
 class BlockCreationMonitor(
   private val vertx: Vertx,
   private val ethApi: EthApiBlockClient,
-  private val startingBlockNumberExclusive: Long,
+  private val startingPoint: StartingPoint,
   private val blockCreationListener: BlockCreationListener,
   private val lastProvenBlockNumberProviderSync: LastProvenBlockNumberProviderSync,
   private val config: Config,
@@ -36,6 +36,11 @@ class BlockCreationMonitor(
   name = "BlockCreationMonitor",
   timerSchedule = TimerSchedule.FIXED_DELAY,
 ) {
+  sealed class StartingPoint {
+    data class ByBlockNumberExclusive(val blockNumberExclusive: Long) : StartingPoint()
+    data class ByTimestampInclusive(val timestampInclusive: Instant) : StartingPoint()
+  }
+
   data class Config(
     val pollingInterval: Duration,
     val blocksToFinalization: Long,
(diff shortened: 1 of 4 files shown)


coordinator/app/src/main/kotlin/lineth/coordinator/blockcreation/BlockCreationMonitor.kt
```diff
@@ -72,32 +82,103 @@ class BlockCreationMonitor(
...
+      log.info(
+        "Cutover timestamp reached. Binary searching for exact first block in 0..{}. targetTimestamp={}, " +
+          "latestBlockNumber={}, " +
+          "latestBlockTimestamp={}",
+        latestBlock.number,
+        targetTimestamp,
+        latestBlock.number,
+        Instant.fromEpochSeconds(latestBlock.timestamp.toLong()),
+      )
+      binarySearchFirstBlockAtOrAfterTimestamp(0L, latestBlock.number.toLong(), targetTimestamp)
```
Filter94, line 148: Shouldn't we pass the `_nexBlockNumberToFetch` instead of 0 here to speed up the retries? I don't think we want to run the binary search over the whole chain history on every retry

Filter94 (approved):

coordinator/app/src/main/kotlin/lineth/coordinator/blockcreation/BlockCreationMonitor.kt
gauravahuja: _nexBlockNumberToFetch is not set in the timestamp case.

> I don't think we want to run the binary search over the whole chain history on every retry

It happens only once. When the latest block crosses the cutover timestamp, we do the binary search to find the first block before the cutover timestamp.

Filter94: I see. I misinterpreted the code, I thought the binary search call is running inside the retry future, but it's composed with a completed retry

PR #3792, opened 2026-08-14
fix(prover-ray): range check bug resolved

bogdanbear:
Fix vacuous RangeCheck.Check on dynamic modules

RangeCheck.Check sized its row loop with Module.Size(), which is 0 for a dynamic module, so it accepted any value outside [0, B) — and zkcdriver puts essentially every real range constraint on a dynamic module. Now resolves the height via RuntimeSize(rt) and indexes with ElementAtN(m.Padding, n, row), keeping the padding split correct too. Reference checker only: the compiled reduction already enforced the bound, so no proof was forgeable. 

### Checklist

### DIFF
--- a/prover-ray/wiop/query_range_check.go
+++ b/prover-ray/wiop/query_range_check.go
@@ -23,10 +23,12 @@ func (rc *RangeCheck) Round() *Round { return rc.Handle.Round() }
 // Check implements [Query]. Verifies that every row of Handle lies in [0, B).
 func (rc *RangeCheck) Check(rt *Runtime) error {
 	m := rc.Handle.Module
-	n := m.Size()
+	// RuntimeSize, not Size: Size() is 0 for a dynamic module, which would empty
+	// the loop and make the check vacuous.
+	n := m.RuntimeSize(rt)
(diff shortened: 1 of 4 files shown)


prover-ray/wiop/query_range_check.go
```diff
@@ -23,10 +23,12 @@ func (rc *RangeCheck) Round() *Round { return rc.Handle.Round() }
 // Check implements [Query]. Verifies that every row of Handle lies in [0, B).
 func (rc *RangeCheck) Check(rt *Runtime) error {
 	m := rc.Handle.Module
-	n := m.Size()
+	// RuntimeSize, not Size: Size() is 0 for a dynamic module, which would empty
```
AlexandreBelling, line 26: Can you add a mechanism that panics when calling Size on a runtime-sized module. To me this is the core of the problem that you PR attempt to fix.

bogdanbear: Good call, adding it.

AlexandreBelling (approved):

ThomasPiellard (approved):

PR #3794, opened 2026-08-14
feat(prover-ray): proof serde toward verifier-ray

AlexandreBelling:
## Proof image serialization

Serializes a `wiop.Proof` into the byte image `verifier-ray` consumes with **zero decode work**: one pointer cast, no parsing, no allocation, no fix-up pass. The image doubles as the guest's RAM witness.

The format is not a free choice. `verifier-ray/src/main.zig` already casts its input region directly:

```zig
fn loadR5Input() *const verifier.VerifyInput {
    return @ptrCast(@alignCast(&_in_start));
}
```

That cast is an existing contract with no producer behind it. This PR supplies the producer, so the format *is* the Zig ABI of `verifier.VerifyInput`, measured from the compiler rather than designed.

### What this adds

**`prover-ray/wiop/proofserialization/`**

| file | role |
|---|---|
| `projection.go` | `wiop.Proof` → the verifier's round-major shape |
| `encode.go` | depth-first inline layout, absolute pointers baked in at encode time |
| `decode.go` | `Decode` / `Validate`, strict host-side reader |
| `layout.go` | the ABI mirror: sizes, field offsets, discriminants |
| `measure.go` | image size and composition without building one |
| `README.md` | the format and the reasoning behind it |

**`verifier-ray/`** — `src/proof_abi.zig` pins the proof types' layout with comptime assertions; `test/proof_abi_test.zig` pins the discriminant offsets `@offsetOf` cannot express; `test/proof_image_test.zig` mmaps a Go-produced image and reads it as a real `verifier.VerifyInput`. The Zig changes are assertion-only, apart from one wire-neutral field reorder in `merkle.zig` (see below).

24 files, +4071/−5.

### The contract, in brief

- **Root is `verifier.VerifyInput`** (`proof` @0, `public_inputs` @112), because that is what the loaders cast — not `verifier.Proof`.
- **`rounds[*].cells` omits public-input cells.** The flat `public_inputs` statement carries them in registration order. Carrying them in both places would absorb them twice and desynchronise the transcript replay.
- **Pointers are absolute**, computed as `base + offset` at encode time, so the guest dereferences them with no arithmetic. An image is therefore only valid at the base it was relocated for.
- **Empty slices carry a non-null pointer.** Zig's `[]const T` holds a non-optional pointer, so null is UB even at length zero.

### Layout is asserted, not guaranteed

Zig's `auto` layout gives no ABI guarantee, and `extern struct` is unavailable: it rejects slices outright — *"slices have no guaranteed in-memory representation"* — which is exactly the representation this format depends on. So the layout is pinned by assertion and checked from both sides.

This matters concretely: **`merkle.Branch` did not lay out in declaration order.** It declared `leaf` before `siblings`, but `siblings` sits at offset 0, because Zig stable-sorts fields by alignment descending. An encoder written from the declarations would be wrong, and a wrong image still casts cleanly — it surfaces as an unrelated failure deep in verification rather than as a layout error. `Branch` is reordered to match (no ABI change), and the align-descending convention is documented so declaration order and memory order stay equal.

Three guards, covering different failure modes:

| guard | catches |
|---|---|
| `proof_abi.zig` comptime assertions | Zig's layout moving under the pinned numbers |
| `abi_agreement_test.go` | the Go constants and the Zig pins diverging |
| `proof_image_test.zig` | the two sides disagreeing on actual bytes |

The third is not redundant: a round-trip test cannot see a *symmetric* error, because encoder and decoder share `layout.go`. Changing an offset keeps them consistent with each other while the image disagrees with the verifier.

### Proof size

The `modexp_u256` proof is **43 MiB**, and essentially all of it is one thing: **opened row values account for 41.3 MiB**. Encoding overhead, Merkle digests, cells and every other component are rounding errors next to that.

The driver is structural. A query does not open one row per input tree — the multi-size commitment opens **one conjugate row pair per distinct committed size**, per tree. Each query therefore reveals roughly twice the entire committed column count:

```
proof ≈ queries × 2 × (total committed columns) × element size
```

For `modexp_u256`: 229 queries (`FRILogInverseRate = 1`, blow-up 2) over ~9,150 committed columns across ~15 sizes, with 29% of values being 24-byte extension elements.

**Expect the production system to be several times this**, once precompiles and the modules not yet in the proof system are added, since size scales with committed column count. The levers, in order of effect: query count (a higher blow-up rate cuts it several-fold), total committed columns, the proportion of extension columns (6× a base column), and the number of distinct sizes. All prover-side; none of them are serialization. Documented in README §11.

### Testing

- **Round trip on real proofs** — `Prove` → `Project` → `Encode` → `Decode` → re-`Encode`, over 29 PCS-compiled scenarios.
- **Cross-language** — `proof_image_test.zig` maps a Go-produced image and reads it as the actual Zig type, with no Zig-side parsing. Verified to fail when the image is corrupted; skips rather than passes where `MAP_FIXED` is unavailable.
- **Size model validated against the encoder** — `Measure`'s arithmetic and `len(Encode(...))` agree byte for byte.
- **Mutation-tested** — 52 injected defects (wrong offsets, inverted tags, dropped coordinates, removed bounds checks, perturbed constants), 51 caught; the survivor is an unreachable invariant assertion.
- **Validator** — a zero-decode format has no parse step, so `Validate` is the only thing between a malformed image and out-of-bounds guest reads. Bounds, alignment, null and discriminant checks are each covered.

Also included: a one-line CI fix. `verifier-ray.yml` pinned `ZKC_VERSION: 'v1.2.26'`, which cannot parse the `/%` divmod operator that `arithmetization/src/main/lib/keccak/impl.zkc` already uses, so `make zkc-verify-passing` fails with `unexpected token`. Bumped to `v1.2.27`. Pre-existing on `main` and unrelated to serialization — it goes unnoticed because `verifier-ray.yml` excludes `arithmetization/**` from its triggers, so a change there cannot fail the workflow that consumes it.

### Future work

Deliberately out of scope, to keep this PR from reshaping verifier-ray:

- **Consuming the image in verifier-ray's runtime.** `testdata/generate` would emit an image from a real proof and a Zig test would call `verifier.verify()` on it, giving an end-to-end round trip on every `make test`. Needs a prover-ray export or a `wiop.CaptureProof` refactor, a module pin bump, and a Makefile change to keep the emitted image from going stale.
- **`entry_claims` is caller-supplied.** It is a re-slicing of the proof's claimed values into the verifier's compiled layout order, and that order lives in verifier-ray's `PcsSystem`. It is not derivable from a `wiop.Proof`, so it is a seam between the two systems rather than a missing feature. A projected proof cannot verify until the call site supplies it.
- **Native loader base.** Absolute pointers break `loadNativeInput`'s `mmap(null, ...)`; README §5.2 recommends `MAP_FIXED`. A verifier-ray decision, and it blocks nothing here — the cross-language test maps at a test-only address.

### Checklist

* [ ] If this change is deployed to any environment (including Devnet), E2E test coverage exists or is included in this PR. — *not deployed; no encoder call sites yet*
* [x] I have informed the team of any breaking changes if there are any. — *none; verifier-ray changes are assertion-only plus one wire-neutral field reorder*

**Local runs:** `go test ./wiop/proofserialization/` green, `zig build test` green, `zig build -Dr5=true` green, `make zkc-verify-passing` green, `gofmt` / `vet` / `zig fmt` clean.

Two unrelated failures in `wiop/compilers` (`TestFullPipeline_LookupRowLimit_TamperedDynamicSize`, `TestCompilePermutation_RowLimit_VerifierRejects`) reproduce on `main` and are being fixed separately — both tamper to sizes above `ColumnSizeMaxSupported`, which trips the generic bound check before the row-limit action they assert on.


prover-ray/wiop/proofserialization/README.md
```diff
@@ -0,0 +1,681 @@
...
+  keeping both, and for the cross-language golden test still outstanding.
+
+  `abi_agreement_test.go` closes the drift direction nothing else covered:
+  `proof_abi.zig` catches Zig's layout moving, and the encoder's own tests catch
+  Go bugs against Go's constants, but neither notices the two sides' *numbers*
+  diverging. It parses `proof_abi.zig` and compares every pinned size, offset and
+  discriminant against `layout.go`, reporting which constant disagrees and what it
+  would corrupt. Verified to fire by perturbing a pin.
+
+- **Done — the projection.** `project.go`: `Project(sys, proof, pub, entryClaims)`
```
YaoJGalteland: entryClaims is stale

prover-ray/wiop/proofserialization/README.md
```diff
@@ -0,0 +1,681 @@
...
+  turns a `wiop.Proof` into the verifier's shape. The Go maps disappear here,
+  which is why they were never an obstacle to the dump. Cells go out round-major
+  in declaration order *including public inputs*, since the verifier absorbs them
+  in that order and omitting them would desynchronise the transcript replay.
+
+  `TestProjectEncodeDecode_EndToEnd` closes the loop on real proofs:
+  prove → project → encode → decode → re-encode, over all 29 scenarios.
+
+  **`entry_claims` is still a parameter, not derived.** Its canonical ordering is
```
YaoJGalteland: stale doc

prover-ray/wiop/proofserialization/decode.go
```diff
@@ -0,0 +1,395 @@
...
+
+// Validate reports whether image is a well-formed proof image at base: every
+// pointer in range and aligned, every length consistent with the image size.
+//
+// It validates structure, not values — the verifier's own checks
+// (fri.checkOpeningProofShape and friends) remain responsible for those. It
+// allocates, since it decodes to walk; that is acceptable because it runs
+// host-side, never in the guest.
+func Validate(image []byte, base uint64) error {
+	_, err := Decode(image, base)
```
YaoJGalteland, line 74: `Validate` currently just calls `Decode`.

`Decode` checks that pointers and lengths are internally consistent, but it does not enforce the global `MaxImageSize` limit that `Encode` applies to the guest input region.

That means an externally supplied image can be:

* larger than `MaxImageSize`
* internally self-consistent
* still accepted by `Validate`

I think this is a correctness gap. “Valid for this format” should also mean “loadable into verifier-ray’s guest memory layout,” not just that the pointer arithmetic is valid on the host.

Without the size check, `Validate` can approve an image that the guest cannot actually accept.

prover-ray/wiop/proofserialization/decode.go
```diff
@@ -0,0 +1,395 @@
...
+	}
+	if payload%uint64(elemAlign) != 0 {
+		return 0, 0, fmt.Errorf("proofserialization: %s: payload offset %d is not %d-byte aligned",
+			what, payload, elemAlign)
+	}
+	if payload+count*uint64(elemSize) > uint64(len(d.buf)) {
+		return 0, 0, fmt.Errorf("proofserialization: %s: %d elements of %d bytes at offset %d "+
+			"run past the end of the %d-byte image", what, count, elemSize, payload, len(d.buf))
+	}
+	return int(payload), int(count), nil
```
YaoJGalteland, line 122: iiuc, the rule should be:

* the root `VerifyInput` lives at offsets `[0, SizeVerifyInput)`
* any non-empty slice backing storage must live at `[SizeVerifyInput, len(image))`

So for any slice with `len > 0`, the validator should enforce:

```text
payload >= SizeVerifyInput
```

Without that check, a malformed image can point a slice payload back into the root object itself, causing the guest to interpret `VerifyInput` header bytes as slice elements.

YaoJGalteland (commented):
It looks good overall. I only found two minor issues around the size checks.

prover-ray/wiop/proofserialization/project.go
```diff
@@ -0,0 +1,207 @@
...
+
+// projectRounds builds one RoundMessage per round, in round order.
+func projectRounds(
+	sys *wiop.System,
+	proof wiop.Proof,
+	isPublicInput map[wiop.ObjectID]bool,
+) ([]RoundMessage, error) {
+	rounds := make([]RoundMessage, len(sys.Rounds))
+
+	for _, r := range sys.Rounds {
```
YaoJGalteland, line 79: change 
```
rounds := make([]RoundMessage, len(sys.Rounds))
for _, r := range sys.Rounds {
```
to 

```
replayedRoundCount := max(len(sys.Rounds)-1, 0)
rounds := make([]RoundMessage, replayedRoundCount)
for _, r := range sys.Rounds[:replayedRoundCount] { ... }

```

this change makes the serialized proof's round list match exactly what verifier-ray's replay loop expects

YaoJGalteland (dismissed):

YaoJGalteland (approved):

PR #3807, opened 2026-08-17
fix(sequencer): deduplicate forced transactions in pool

### LINKED ISSUE #3776: Sequencer includes FTX in non-monotonic order, halting coordinator finalization
## Impact
__environment Local and CI__

__an user is impacted [Y], if yes type of user: Currently only local as it's not code running in production, but if live could halt the coordinator until it's restarted__

## Describe the bug

This bug is related to occasional failure of `forced_transaction.spec.ts` tests when ran concurrently in the CI. 
When two or more FTX are in flight to the Sequencer, the Sequencer can include the higher nonced FTX on smaller block number and halt the coordinator.

`ForcedTransactionsSafeBlockNumberManager.kt:66` assumes FTX results from the sequencer arrive in non-decreasing block order. 

The e2e tests are ran concurrently, and with two FTXs in flight at once, the sequencer can place the higher-numbered one in an earlier block. For example, one of the workflows linked below has the next values:
Aug 4 run: 
- the higher nonce FTX
  - ftx=5
  - block 25
- the lover nonce FTX
  - ftx=4
  - block 26 

If the higher-block result lands first, `safeBlockNumber` advances past the other, and when the second result arrives the invariant throws `IllegalStateException`. The service then retries the same message every 5s forever. Finalization halts and every test waiting on FinalizedStateUpdated times out.

### Steps to Reproduce
_A bug that we can't reproduce is hard to resolve, to ensure the bug is fixed in a timely manner, it is crucial to indicate how to reproduce it. If the bug can't be reproduced, give specific details on how it happened._

Steps to reproduce the bug:
1. Checkout to `test/ftx-safeblock-out-of-order`
2. Run `./gradlew :coordinator:ethereum:forced-transactions:test \
  --tests "lineth.ftx.conflation.ForcedTransactionsSafeBlockNumberOutOfOrderTest"`

PR with the reproduction test: https://github.com/LFDT-Lineth/lineth-monorepo/pull/3777

### Actual result
Related workflow failures:
[July 28th](https://consensys.slack.com/archives/C09GY4Q5PME/p1785216981282769), [Aug 3rd](https://consensys.slack.com/archives/C09GY4Q5PME/p1785768234660769), [Aug 4th](https://consensys.slack.com/archives/C09GY4Q5PME/p1785853233880729)

### Expected result
To be discussed by the protocol team.
- Is this an issue with the Sequencer, is it violating ordering, should it behave this way?
- Is this an issue with the Coordinator, is sequencer behaving as expected and the Coordinator is the one that should be tolerating this scenario?

nadeemb53:
Fixes #3776. 
Deduplicates pending and completed forced transactions so coordinator redelivery cannot leave stale queue entries that overwrite inclusion status.


linea-besu/plugins/linea-sequencer/sequencer/src/main/java/lineth/sequencer/forced/LineaForcedTransactionPool.java
```diff
@@ -69,6 +69,7 @@ public class LineaForcedTransactionPool
       7_200;
 
   private final Deque<ForcedTransaction> pendingQueue = new ConcurrentLinkedDeque<>();
+  private final Set<Long> pendingTransactionNumbers = ConcurrentHashMap.newKeySet();
```
Filter94: I feel like the this becomes bloated. What if we use an `ConcurrentSkipListMap` here instead of the `pendingQueue`?

nadeemb53: my understanding is that the set keeps the transaction number reserved while the transaction moves from the queue into the status cache. With only a map, removing the entry before recording the status creates a window where a retry can add it again, so we’d need extra locking. That doesn’t end up being simpler.

Filter94: What if we avoid this window by reordering the operations? Same as you've done with the `pendingTransactionNumbers`, removal from the queue after the status is updated

Also, a validation whether a transaction is within the `statusCache` already with `INCLUDED` status could also help with the duplicates. And I think we don't care about the resubmission with other, initially rejected statuses

nadeemb53: Hmm let me try

Filter94 (approved):

PR #3808, opened 2026-08-17
fix(guest): `install-native-crypto-deps` target and `ZKC_EXEC_FLAGS` propagation

### LINKED ISSUE #3728: fix(ci): execution specs tests
Tests are failing, see https://github.com/LFDT-Lineth/lineth-monorepo/actions/workflows/arithmetization-execution-specs-ssz-fixtures.yml

lorenzogentile404:


riscv-guests/l2-execution/scripts/run_execution_specs_ssz_fixtures.go
```diff
@@ -0,0 +1,422 @@
+package main
```
lorenzogentile404: This has been deleted in `main`. Understand if the logic has been moved somewhere else.

Filter94 (dismissed):

riscv-guests/Makefile
```diff
@@ -62,7 +62,7 @@ install-native-crypto-deps:
 		$$sudo cp /tmp/blst-src/libblst.a "$$prefix/lib/"; \
 		$$sudo cp /tmp/blst-src/bindings/blst.h /tmp/blst-src/bindings/blst_aux.h "$$prefix/include/"; \
 	fi; \
-	if [ -f "$$prefix/lib/libmcl.so" ] || [ -f "$$prefix/lib/libmcl.dylib" ]; then \
+	if { [ -f "$$prefix/lib/libmcl.so" ] || [ -f "$$prefix/lib/libmcl.dylib" ]; } && { [ "$$uname_s" != "Darwin" ] || [ -f "$$prefix/lib/libmcl.a" ]; }; then \
```
amkCha, line 65: I don't know - what is this doing ?

arithmetization/src/test/Makefile
```diff
@@ -133,15 +133,15 @@ compile: require-src
 
 # Convert and execute an already compiled external ELF
 elf-exec: elf-to-json install-zkc
-	$(MAKE) -C $(MAKEFILE_DIR) --no-print-directory zkc-exec JSON=$(JSON_EXT)
+	$(MAKE) -C $(MAKEFILE_DIR) --no-print-directory zkc-exec JSON=$(JSON_EXT) ZKC_EXEC_FLAGS="$(ZKC_EXEC_FLAGS)"
```
amkCha, line 135: are you calling elf-exec directly ? Through make -C risv-guests/l2-execution exec it's passed already

lorenzogentile404: I am not sure I understand. Here I added the `ZKC_EXEC_FLAGS` otherwise at the internal invokation the default values was used instead.

riscv-guests/Makefile
lorenzogentile404: Basically there is a difference between Linux and MacOS in the libraries we install. In the case of MacOS (when `uname_s` is equal to `Darwin`) the file `libmcl.a` needs to be moved from the `tmp` folder under `lib`. In the case of Linux that file is already under `lib`.
So, when we are on MacOS and `libmcl.a` is not under `lib`, then we know that the library/target still needs to be executed fully, so we do not skip the body of this if.

amkCha: thanks for the explanation - if the script adapts to both OS then, looks good :+1:

arithmetization/src/test/Makefile
amkCha: what's the command line for an internal invocation ? you mean when you type `make elf-exec` directly ?

amkCha (dismissed):

arithmetization/src/test/Makefile
lorenzogentile404: For example, when a target is included as a dependency (in this case `elf-to-json` `install-zkc`) I call those internal, as the values of the flags are "inherited", while if you invoke `make` again, then it is like a fresh new call.

amkCha (approved):

DavePearce (approved):
LGTM

PR #3817, opened 2026-08-18
feat(prover-ray): added guest output to the public inputs

Soleimani193:
This PR integrates guest output into the prover's public inputs.

### Checklist

  PR.

---

> [!NOTE]
> **High Risk**
> Changes how guest outputs enter the public-input vector and adds length pinning constraints; mistakes here directly affect proof soundness for attested program results.
> 
> **Overview**
> Guest program results are now exposed as **public inputs** the verifier checks, instead of staying only in the trace.
> 
> **zkcdriver** discovers the arithmetization’s `pub output` memory during schema import (`PublicOutput` / `PublicOutputs`) and records its address and data columns. **risc5** uses that metadata to pin the dynamic output memory to **`NumGuestPublicOutputs`** via a length constraint on the last address, then registers each output cell as a public input. **`GetGuestPublicOutputs`** reads those values from the constrained trace so they match what the proof attests.
> 
> **`Core.New`** always calls **`RegisterGuestPublicOutputs`** after the circuit is loaded; backend tests compile a small **`guest_output.zkc`** fixture that declares the expected public output shape. New integration tests prove/verify the happy path and assert wrong-length guests fail verification and panic in the prover.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d8b83182a269c3e4d99e77d647cef07cdd76c95e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


prover-ray/backend/core.go
```diff
@@ -56,6 +57,11 @@ func New(cfg Config) (*Core, error) {
 	sys.NewRound()
 	driver := zkcdriver.NewZkCDriver(sys, zkcdriver.Settings{}, binFile)
 
+	// Must run after the arithmetization is defined, so the guest_output columns
+	// exist, and before the compiler passes, which discharge the openings it
+	// registers.
+	risc5.RegisterGuestPublicOutputs(sys)
```
AlexandreBelling, line 63: Not a blocking comment but we should envision this function call to also register all the things (just to avoid having to call 10 different functions). Or maybe we should fold that into a function wrapping the public input registration with the zkc driver

AlexandreBelling (approved):
LGTM. You can ignore my comment

ThomasPiellard (approved):

PR #3821, opened 2026-08-19
fix(prover-ray): entry_claims is removed and built from ctx.rounds

YaoJGalteland:
This looks correct and the explanation is strong. I’d tighten it slightly around the “single source of truth” point:

**Before this PR:** `entry_claims` was stored directly on `Proof.pcs_opening`, so the prover supplied it independently to the verifier. The same underlying `LagrangeEval.EvaluationClaims` values also existed in `rounds[*].cells`, which were separately absorbed into the Fiat–Shamir transcript. That created two sources of truth for the same claims, with no check that they agreed.

**After this PR:** `entry_claims` is removed from `PcsOpening` entirely. Instead, at the beginning of `verify()`:

1. `pcs.reconstruct(...)` computes the canonical PCS entry order.
2. `pcs.buildEntryClaims(...)` reads the claim values directly from `ctx.rounds[*].cells` using the codegen-baked `claim_cells: []const CellRef` coordinates and writes them into a local stack buffer in that canonical order.
3. The resulting `entry_claims` slice is then passed to both `pcs.verify(...)` and `routeClaims(...)`.

So `entry_claims` is no longer an independent proof input. It is derived inside the verifier from the already transcript-bound round cells, making `rounds[*].cells` the single source of truth for those values.

### DIFF
--- a/prover-ray/wiop/proofserialization/types.go
+++ b/prover-ray/wiop/proofserialization/types.go
@@ -93,18 +93,6 @@ type OpeningProof struct {
 	FriProof     FriProof
 }
 
-// PcsOpening is the prover-supplied half of a PCS opening. Mirrors Zig's
-// verifier.PcsOpening.
-//
-// The claimed evaluations are not carried here: they are ordinary
-// `LagrangeEval.EvaluationClaims` cells, already present in
-// `Proof.Rounds[*].Cells`. The verifier reconstructs them at verify time by
-// reading the compiled system's per-column claim-cell table against the
-// rounds it already has.
-type PcsOpening struct {
-	Proof OpeningProof
-}
-
 // Proof mirrors Zig's verifier.Proof: the verifier-visible transcript.
 //
 // Note that Rounds[*].Cells OMITS public-input cells; those travel in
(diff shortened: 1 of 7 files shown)


AlexandreBelling (dismissed):

prover-ray/wiop/proofserialization/types.go
```diff
@@ -94,11 +94,15 @@ type OpeningProof struct {
...
+//
+// The claimed evaluations are not carried here: they are ordinary
+// `LagrangeEval.EvaluationClaims` cells, already present in
+// `Proof.Rounds[*].Cells`. The verifier reconstructs them at verify time by
+// reading the compiled system's per-column claim-cell table against the
+// rounds it already has.
 type PcsOpening struct {
-	EntryClaims [][]Ext
-	Proof       OpeningProof
+	Proof OpeningProof
```
AlexandreBelling: At this point, do we even need a separate type for the OpeningProof? We should use the opening proof directly

YaoJGalteland: Good finding! Yes, I flattened the structure, so Proof now holds PcsOpening.OpeningProof directly.

AlexandreBelling (approved):

PR #3828, opened 2026-08-19
perf(prover-ray): Merkle capping

### LINKED ISSUE #3616: perf(prover-ray/verifier-ray): FRI Merkle Capping
For $q$ FRI queries, if we remove the top $l$ steps of every query and instead open the $l$ upper levels of the tree, the verifier saves $ql$ hashes and adds $2^{l-1}$ ones. To optimize the savings $d := ql-2^{l-1}$, we look to vanish $\frac \partial {\partial l} d = q - 2^{l-1}/\log_2 e$. This occurs at $l = \log_2 q + \log_2\log_2 e$. The possible presence of auxiliary leaves complicates this matter, diminishing the optimal $l$. So the verifier can let the prover chose $l$, with a guard $l < 1 + \log_2 q + \log_2\log_2 e$ against malicious behavior (or simply refuse to consume proofs that are too large.)

Tabaie:
Implements Merkle capping. Benchmarked on `bench_pcs` that verifies one opening on one column of size $2^{19}$.
```
                |   main   |  This PR   |  relative delta
--------------------------------------------------------------
Native  runtime | 12.88 ms |  10.45 ms  |    -18.8%
 RISC-V cycles  |  54.71 M |   36.82 M  |    -32.7%
```
"Native" performance was measured on an M5 laptop. Prover gains were modest and within noise bounds.


AlexandreBelling (dismissed):

verifier-ray/bench/bench_pcs/bench_main.zkc
```diff
@@ -0,0 +1,46 @@
...
+include "../../../arithmetization/src/main/riscv/utils/bit_flip.zkc"
+include "../../../arithmetization/src/main/riscv/utils/negative.zkc"
+include "../../../arithmetization/src/main/riscv/utils/register_utils.zkc"
+include "../../../arithmetization/src/main/riscv/memory.zkc"
+
+// The repository-wide main.zkc prints one line per RISC-V instruction. That is
+// useful for debugging but makes a large proof benchmark I/O-bound. This entry
+// point keeps the same interpreter and cycle counter while leaving only guest
+// write syscalls (the phase markers) visible.
+fn main<registers, ram, keccak_state, poseidon2_state>() {
```
ivokub, line 17: I'm not sure it would be good to have a copy of the interpreter loop. This means that whenever the arithmetization changes then we get off-sync. Particularly -- recently the R5 interpreter was completely refactored to use the "predecoded" R5 ELF.

verifier-ray/src/query/pcs.zig
```diff
@@ -502,6 +532,178 @@ pub fn deriveChallenges(
...
 }
 
+fn inputAuxDepth(rate_log: u8, size_log2: u8, bottom_size_log2: u8) ?usize {
+    if (size_log2 >= bottom_size_log2) return null;
+    const encoded_log = @as(usize, rate_log) + size_log2;
+    if (encoded_log == 0) return null;
+    return encoded_log - 1;
+}
+
+test "input auxiliary table includes encoded size two" {
```
ivokub, line 542: I would keep the tests in separate file.

verifier-ray/testdata/generate/go.mod
```diff
@@ -26,3 +26,5 @@ require (
 )
 
 replace github.com/consensys/linea-monorepo/verifier-ray/codegen => ../../codegen
+
+replace github.com/LFDT-Lineth/lineth-monorepo/prover-ray => ../../../prover-ray
```
ivokub, line 30: Instead of using `replace` directive (and thus not allowing us to pin the prover version we support), it is better to split this PR into two parts -- the prover part and the verifier part. After the prover part is merged then we can pin the concrete prover-ray version in this same go.mod file.

prover-ray/wiop/proofserialization/README.md
```diff
@@ -456,13 +461,14 @@ memory image is fine as a RAM witness and wrong as a network payload.
 Identical across all four programs, since they follow from the FRI parameters
 rather than the circuit: 229 queries, 4 input trees per query, opening depth 17,
 16 FRI rounds (15 round roots), 15 layers per running query, 3435 branches,
-30,915 branch sibling digests, 1 final-poly coefficient. Only `row data` and
-`cells` scale with the program.
+30,915 branch sibling digests, 1 final-poly coefficient. The cap structure is
+fixed by the FRI/input-tree parameters; `row data`, cap row data, and `cells`
```
ivokub, line 465: We should update the number of digests here after capping was implemented.

prover-ray/crypto/koalabear/fri/tree.go
```diff
@@ -71,6 +73,24 @@ type Branch struct {
...
+// MerkleCap authenticates the shared upper part of a Merkle tree. Nodes are
+// the frontier, from left to right; Aux is the heap-order prefix above it.
+// An empty cap represents depth zero and carries no copy of the known root.
+type MerkleCap struct {
+	Nodes []field.Octuplet
+	Aux   []*field.Octuplet
+}
+
+// merkleCapDepth selects the shared prefix for a plain binary tree. The
+// frontier has at most one node per query, and at least one branch step stays
```
ivokub, line 85: Comment correct? Should have "between q to 2q" nodes?

ivokub (changes requested):
In general looks good, I didn't find any implementation issues. It makes sense, looks good and is well tested everywhere.

I would do a few changes though:
* separate the prover and verifier part into separate PRs so that we can pin concrete prover version in `go.mod` instead of using the replace directive (in which case every time something is changed in the prover it also changes how the verifier accesses the APIs)
* lets not reimplement the main.zkc in the bench. We should be better in sync with arithmetization in case anything changes there (probably best approach is to just use arithmetization directly).
* small comment about inline test.
* it seems the Merkle capping is not documented in the prover dir? I think we should have some documentation, even as `crypto/koalabear/fri/doc.go` (or in `prover-ray/docs/`) . It doesn't have to be very extensive, but rather a high-level description (so that the people looking at the package would understand what is going on, right now the whole FRI/PCS implementation is complicated enough that without LLMs it is difficult to figure out what is going on). If anyone wants more specific details about the implementation, then they can just ask LLMs.

prover-ray/wiop/proofserialization/README.md
Tabaie: A lot of the figures in this doc are stale. Seems like the harness that measured them is no longer there. Imo we should either bring it back or get rid of this file entirely. @AlexandreBelling what do you think?

ivokub: I guess we can get rid of the section, I agree it will get stale over time. If we want to have some nice up-to-date numbers, then we could have a `Benchmark` test where we report custom metrics https://pkg.go.dev/testing#B.ReportMetric. Then when running `go test -bench` then we get a nice output which we can later run into https://pkg.go.dev/golang.org/x/perf/cmd/benchstat so that we can have nice diffs over time.

ivokub (dismissed):
Conditional approve -- the numbers in the README are stale, but I think we can keep it for now for reference and come up with a better solution later. Or if you want to remove the section then its also fine (but I think the historical reference is also nice to have)

ivokub (approved):

PR #3844, opened 2026-08-24
fix(prover-ray): fix shared randomness and fiat-shamir

Soleimani193:
This PR fixes several issues and bugs in the shared randomness computation and the Fiat-Shamir process

**Problem**
messagebus.Compile(..., {SharedRandomness: true}) derives the cross-shard challenges α, β from a seed γ handed to the shard by the orchestrator. It did so with a PreSamplingHook on the coin round whose Run calls rt.SetFSState(γ) — which replaces the entire Fiat-Shamir state. AdvanceRound then sampled every coin on that round from the replacement and never restored the original. Two bugs follow.

**1. The transcript was discarded, so Fiat-Shamir stopped binding it.**

AdvanceRound absorbs the closing round's commitment, its cells, and the dynamic module sizes — then runs the hook, which throws all of it away:
AdvanceRound(1 → 2):
  absorb module sizes, Commitments[1], round-1 cells   ← all discarded at the next step
  SetFSState(γ)                                        ← transcript replaced, permanently
  sample α, β
round 2, 3, … : every later coin derives from γ + whatever came after
From that point on, nothing absorbed before the coin round influenced any challenge 

**2. Per-round seeding captured other passes' coins.**

The coin round is not the message bus's private property. Three passes anchor their coin round at "one past the last witness round" and reuse whatever sits in that slot (EnsureNext appends only when the anchor is the tail):
In the zkc pipeline the whole RV witness sits in one round — scanColumns pins every column to Rounds[0] — so all three resolve to round 1 regardless of pass order.  A hook seeding the whole round therefore also seeded γ_lookup, making it a function of γ alone with no dependence on the multiplicity columns lookuptologderivsum explicitly requires it to be bound to.

**Fix**
Seeding becomes per-coin. CoinField.MarkSeeded() opts a coin in; AdvanceRound snapshots the transcript before the hook, draws the marked coins from the seed, restores the snapshot, then draws the rest from it. messagebus marks α and β only.

α and β keep their exact values: they are still drawn consecutively from the one γ-seeded sponge (α = digest(γ), β = digest(γ‖0)), staying distinct through the usual safeguard update.

Also collapsed Round.PreSamplingHooks []ProverAction to a single PreSamplingHook, panicking on a second registration. A round holds one Fiat-Shamir state, so stacking could never give two coins two seeds; the old code warned about this and defined last-hook-wins semantics that were never used.

**Problem with Contribution Computation:**

The computation of shared randomness contribution is not consistent with the WIOP description, especially with the shared randomness computation from preflight. It uses the commitments from all the previous rounds.
This can be seen as
contribution =  MultSetHash(setup, ProgramVK, busColumnsOfShard)
while the shardRandomness is MultiSetHash(AllBusColumns) = MultiSetHash(busColumnsOfShard_0, ....)

**Missing Tests:**
 The PR also adds the tests for the message bus when randomness is seeded via the shared randomness implementation, thereby testing the integration of shared randomness and the message bus.

### Checklist

  PR.

---

> [!NOTE]
> **High Risk**
> Changes Fiat–Shamir transcript handling and cross-shard γ contribution hashing; mistakes here would affect soundness of sharded message-bus proofs rather than cosmetic behavior.
> 
> **Overview**
> Fixes **Fiat–Shamir shared-randomness seeding** so cross-shard γ reaches only the coins that need it, and aligns **shard contributions** with preflight/orchestrator semantics.
> 
> **Per-coin seeding (`CoinField.MarkSeeded`)** — `AdvanceRound` now snapshots the transcript before the pre-sampling hook, draws **marked** coins from the hook’s state (e.g. γ), **restores** the snapshot, then draws unmarked coins from the local transcript. Message bus marks **α and β only**, so later challenges (including other compiler passes on the shared coin round) stay bound to prior commitments instead of living entirely on γ.
> 
> **Single pre-sampling hook per round** — `PreSamplingHooks` becomes one `PreSamplingHook`; a second registration **panics** instead of last-hook-wins with a warning.
> 
> **Contribution preimage** — Shared randomness contribution is no longer a multiset hash over all earlier round commitments; it is **`multisethashing.Hash` of the bus round’s PCS commitment** (the round immediately before the coin round), matching what preflight accumulates per shard.
> 
> **Tests** — Adds coverage for scoped seeding (foreign coins / FS state), γ-as-public-input, and the full **γ-from-contributions** loop with PCS plus cross-shard grand-product balance under derived α/β.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7957cdd5dc36a1fe9c114b1fcde425c0d73a3e57. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


prover-ray/wiop/compilers/messagebus/shared_randomness.go
```diff
@@ -142,30 +151,37 @@ type SharedRandomnessContributionAssigner struct{}
...
-	hasher := poseidon2.NewMDHasher()
-	for i := range rt.CurrentRound().ID {
-		if !rt.System.Rounds[i].HasCommitment {
-			logrus.Warnf(
-				"No commitment found for round: %v. Did you use a message bus? "+
-					"And did you reduce the current system using a PCS?", i)
-			continue
-		}
+	// In range: registerSharedRandomness refuses a coin round of 0.
+	busRound := rt.System.Rounds[rt.CurrentRound().ID-1]
```
YaoJGalteland: spec (preflight.Run):   Hash( commit( all bus columns, and only bus columns ) )
  code:                   Hash( Commitments[coinRound.ID - 1] )
                                └── the commitment of one round, whole

they only match for the single bus round case

prover-ray/wiop/compilers/messagebus/shared_randomness.go
```diff
@@ -142,30 +151,37 @@ type SharedRandomnessContributionAssigner struct{}
...
+// through here, so neither can drift from the other's preimage.
+//
+// The value returned here is exactly the Hash(root) that
+// [github.com/LFDT-Lineth/lineth-monorepo/prover-ray/preflight.Run] accumulates
+// for that shard, which is what lets an aggregator recompute γ from the shards'
+// contributions.
+//
+// This assumes the bus round holds the bus input set and nothing else: the
+// contribution hashes the round's commitment whole, while preflight commits the
+// bus columns alone. The assumption is not checked here. Breaking it does not
```
YaoJGalteland, line 171: `sharedRandomnessContribution` documents that it assumes the bus round contains exactly the bus input set, but that assumption doesn’t seem to be enforced.

If either half is violated — a bus entry comes from an earlier round, or a non-bus column shares the bus round — the failure only shows up later at the aggregation layer, while all shard-local checks still pass.

Could we at least enforce the cheap half (all bus columns on one round) in `registerSharedRandomness` by panicking if:

```go
entry.Round().ID != coinRound.ID-1
```

The extra-column half ("and nothing else") could stay documented for now if enforcing it would be too restrictive for planned layouts.

Soleimani193: Yes, later in a separate PR, we would put all the  (sharded) preflight data in the bus round,  that is, one round before the coin round. And with the following code, we are sure it takes them from the correct round.

PR #3850, opened 2026-08-25
feat(arithmetization): support sharding of interpreter() function

### LINKED ISSUE #3849: feat: update arithmetisation to support sharding
# Summary

The arithmetisation requires a minor update as a preparatory step for sharding.  Specifically, the loop wrapping the `interpreter()` function must be removed and replaced with a (recursive) no-return call.

DavePearce:
This puts in place support for sharding on the interpreter() function. This is done by converting the interpreter loop into a recursive "no return" function call.  The key about this approach, is that it makes it relatively simple to then shard the interpreter() function.

### DIFF
--- a/arithmetization/src/main/riscv/interpreter.zkc
+++ b/arithmetization/src/main/riscv/interpreter.zkc
@@ -188,29 +188,30 @@ include "../lib/io/write_output.zkc"
 // The tiers above list explicit semantic opcodes (`*_WB`, branches, stores,
 // precompiles, …); anything not matched there eventually reaches the
 // `interpreter_c` default and behaves as a no-op.
-fn interpreter<registers, ram, keccak_state, poseidon2_state>(clk:u32, base:Address, pc:Address, output_write_address:OutputAddress) -> ! {
+
+fn interpreter<registers, ram, keccak_state, poseidon2_state>(clock_cycle:u32, base:Address, pc:Address, output_write_address:OutputAddress) -> ! {
+
     var new_output_write_address:OutputAddress
     var new_pc:Address
     var index:u62
     var rem:u2
     // determine index of predecoded program
     index::rem = pc - base
-    // NOTE: This section pertains to the predecoding justification
-    // It is pending decision and is commented for now.
+    // -- PREDECODING PROOF SECTION -- 
     // Execution must stay inside the span justified by predecoding, at 4-byte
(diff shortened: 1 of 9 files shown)


letypequividelespoubelles (dismissed):

arithmetization/src/main/riscv/main.zkc
amkCha: Should we update this as well so we have all the uncomment ready at hand ?

arithmetization/src/main/riscv/interpreter.zkc
```diff
@@ -188,9 +188,27 @@ include "../lib/io/write_output.zkc"
 // The tiers above list explicit semantic opcodes (`*_WB`, branches, stores,
 // precompiles, …); anything not matched there eventually reaches the
 // `interpreter_c` default and behaves as a no-op.
-//
-fn interpreter<registers, ram, keccak_state, poseidon2_state>(index:u64, pc:Address, output_write_address:OutputAddress) -> (new_pc:Address, new_output_write_address:OutputAddress) {
+fn interpreter<registers, ram, keccak_state, poseidon2_state>(clk:u32, base:Address, pc:Address, output_write_address:OutputAddress) -> ! {
```
amkCha: for readability, I had a preference for `clock_cycle`, more explicit

arithmetization/src/main/riscv/main.zkc
```diff
@@ -76,43 +76,6 @@ fn main<registers, ram, keccak_state, poseidon2_state>() {
...
-        var index:u62
-        var rem:u2
-        // rem is the remainder of pc - base divided by 4
-        index::rem = pc - base
-
-        // NOTE: This section pertains to the predecoding justification
-        // It is pending decision and is commented for now.
-        // Execution must stay inside the span justified by predecoding, at 4-byte
-        // aligned PCs, so decoded[(pc - base) / 4] matches the row checked at pc.
-        // if pc<base || pc >= exec_end || rem != 0 {
```
amkCha, line 105: In the new interpreter function, were you able to port this check ? Even if commented, so we can uncomment easily if we want to enable predecoding justification 🙏

DavePearce: yes, I put that check back now.

letypequividelespoubelles (approved):

DavePearce (commented):
LGTM

Issue #3825, opened 2026-08-19
[RV] - Rollup guest program


Filter94:
## Description
We need to add an implementation of the Rollup guest program

## Motivation
Execution of Rollup guest program will be proven and verification of these proofs on L1 proves the state transition on L2

## Tasks
- [ ] Implement rollup guest program

## Acceptance criteria
- [ ] Rollup guest program is implemented
- [ ] It is covered with the tests running on CI
- [ ] There is CI job that test the integration between the compiled guest program and the ZkC interpreter

## Risks
- [ ] Bugs

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #3854, opened 2026-08-25
chore(misc): rollup zstd kzg clarifications

Filter94:
This PR implements issue(s) #3825

### Checklist

  PR.


rollup_spec/Readme.md
```diff
@@ -228,15 +228,15 @@ Each l2-execution proof `Eₑ` (`e ∈ [1, N]`) has:
...
+2. **Verify and fold the touched chunks.** Given `startOffset` (public input) and `ownStreamBytes` (step 1), derive `T` — the number of chunks `ownStreamBytes` spans — and `endOffset = len(ownStreamBytes) - (T-1)·chunkSize + startOffset`, asserting `0 < endOffset ≤ chunkSize`. For each chunk `k ∈ [1, T]`: slice the corresponding portion of `ownStreamBytes` (all of it for an interior chunk; `chunkSize - startOffset` bytes for the first chunk if `k = 1`; up to `endOffset` bytes for the last chunk if `k = T`), combine it with the witnessed `opaquePrefixBytes` / `opaqueSuffixBytes` — relevant only for `k = 1` / `k = T` respectively, never for interior chunks — to reconstruct the chunk's full `chunkSize` bytes, and run the chunk-kind-dispatched commitment check above (§2.2, §3.1). A calldata chunk instead resolves its own extent — a whole number of segments — by matching its anchored hash at the candidate segment-end offsets (§3.1), so `T` and `endOffset` follow from the hash-matched walk across the heterogeneous chunks. Fold the dataRollingHash:
```
Tabaie: `endOffset = len(ownStreamBytes) - (T-1)·chunkSize + startOffset` should become something like $\text{endOffset} =\text{startOffset}+ \text{len(ownStream)}-\sum_{k=1}^{T-1} c_k$ where $c_k$ can vary based on whether the chunk is a blob or calldata.

Tabaie (changes requested):
Just in one spot the documentation didn't reflect the mixed source of chunks.
More seriously, the following check doesn't take the same into account. It only considers the fixed blob length as a correct boundary regardless of whether the last chunk was a blob or calldata.
```python
    offsets_at_chunk_boundary = (
        left.public_inputs.end_offset == BLOB_BYTES_LENGTH and right.public_inputs.start_offset == 0
    )
```

srinathln7 (commented):
Since this PR enables the calldata path, I think we need to update the aggregation logic in `rollup_spec/src/rollup_spec/rollup_aggregation.py`.

Currently, around line 153 `rollup_spec/src/rollup_spec/rollup_aggregation.py`, the code checks:

```
offsets_continuous      = left.end_offset == right.start_offset
offsets_at_chunk_boundary = (left.end_offset == BLOB_BYTES_LENGTH and right.start_offset == 0)
```

But with calldata support, this breaks. A proof ending with a calldata chunk (e.g., 40000 bytes) sets end_offset = 40000. Since calldata chunks aren't shared between proofs, the next proof starts fresh at `start_offset = 0`.

Neither condition above is satisfied, so the aggregator incorrectly rejects valid proof pairs. I think this half is missing. WDYT @Filter94

Filter94:
> Just in one spot the documentation didn't reflect the mixed source of chunks. More seriously, the following check doesn't take the same into account. It only considers the fixed blob length as a correct boundary regardless of whether the last chunk was a blob or calldata.
> 
> ```python
>     offsets_at_chunk_boundary = (
>         left.public_inputs.end_offset == BLOB_BYTES_LENGTH and right.public_inputs.start_offset == 0
>     )
> ```

Now the check is just `left.public_inputs.end_offset != right.public_inputs.start_offset` because the start and end offset encoding is consistent and it matches between blobs and calldata and 0 means "complete chunk" was consumed or should be ignored completely (depending on the logic).

rollup_spec/Readme.md
Filter94: Now calldataLength is introduced to the ChunkWitness to compute this sum

rollup_spec/src/rollup_spec/rollup.py
```diff
@@ -803,21 +911,24 @@ def rlp_encode_truncated_blocks(blocks: Sequence[TruncatedEthereumBlock]) -> byt
...
 
 
-def compress_lz4(data: bytes) -> bytes:
+def compress_zstd(data: bytes) -> bytes:
     """
-    LZ4-compress the canonical RLP-encoded truncated-block payload using
-    the raw LZ4 block format (no 4-byte uncompressed-size header). The
-    rollup guest zero-pads this output to `BLOB_BYTES_LENGTH` and
+    zstd-compress the canonical RLP-encoded truncated-block payload (§3.1).
+    The rollup guest zero-pads this output to `BLOB_BYTES_LENGTH` and
```
Tabaie: Assumes blob-only chunks

rollup_spec/src/rollup_spec/proof_io_v1.py
Tabaie, line 453: Not sure about this but would it be more accurate to say it's a `{chunkHash, isCalldata}` tuple?

srinathln7 (dismissed):

Tabaie:
I had AI look for stale references to the blob-only chunk design and it came up with the following:

`Readme.md:61`	"Blobs (or calldata submissions) are fixed-size chunks" — names calldata, then calls it fixed-size. Same line: "a chunk may be shared between adjacent rollup proofs" — blob-only now.
`Readme.md:84`	"Chunks (blobs)" equates the two. Same line: "several rollup proofs' conflations may share one chunk".
`Readme.md:99`	"l1_rollup.py models the contract-facing blob anchoring" — `src/rollup_spec/l1_rollup.py:115` is kind-agnostic.
`Readme.md:439`	the duplicate paragraph (finding 4).
`Readme.md:451`	§3.2 heading "Blob Payload", and its opening "The DA blob must contain…" — the payload is chunk-kind-agnostic.
`Readme.md:602`	§5.2 heading "Calldata Blob Submission Interface".
`Readme.md:757`	"implicitly across blob boundaries within a multi-blob rollup proof".
`Readme.md:827`	"one rollup proof can cover K ≥ 1 blobs", with K = 1/K > 1. The interconnection row two lines up was changed to "T ≥ 1 mixed chunks" in this same commit; this one was missed.

rollup_spec/src/rollup_spec/rollup.py
```diff
@@ -178,114 +210,178 @@ def _truncate_conflation(
...
+            except Exception as exc:
+                raise Exception("invalid chunk KZG commitment computation") from exc
+            computed_chunk_hash = Hash32(kzg_commitment_to_versioned_hash(chunk_kzg_commitment))
+            if computed_chunk_hash != chunk.chunk_hash:
+                raise Exception(f"chunk {i} computed KZG commitment does not match chunkHash")
+        else:
+            # Range-aligned calldata chunk: it packs a whole number of
+            # segments, so it carries no opaque boundary bytes and its start
+            # sits at a fresh stream/segment boundary.
+            if len(prefix) != 0 or len(suffix) != 0:
```
Tabaie, line 344: I think this will never fire.

rollup_spec/src/rollup_spec/rollup.py
```diff
@@ -178,114 +210,178 @@ def _truncate_conflation(
...
+                raise Exception(f"chunk {i} is a calldata chunk and carries no opaque bytes")
+            if is_first and start_offset != 0:
+                raise Exception(f"chunk {i} is a calldata chunk and starts at offset 0")
```
Tabaie: Same here. Unreachable.

rollup_spec/src/rollup_spec/rollup.py
```diff
@@ -178,114 +210,178 @@ def _truncate_conflation(
...
+            # bytes don't, reconstructing exactly BLOB_BYTES_LENGTH.
+            own_slice_len = BLOB_BYTES_LENGTH - len(prefix) - len(suffix)
+            if own_slice_len < 0:
+                raise Exception(f"chunk {i} opaque bytes exceed the blob chunk size")
+            last_chunk_len = BLOB_BYTES_LENGTH - len(suffix)
+            if own_slice_len <= 0:
+                raise Exception(f"chunk {i} must contain owned bytes")
+            own_slice = own_stream_bytes[cursor:cursor + own_slice_len]
+            cursor += own_slice_len
+            full_chunk_bytes = prefix + own_slice + suffix
```
Tabaie, line 316: Should this take field element packing into account? I think a BLS12-381 scalar field element has only room for 254 data bits, so this will be slightly inflated.

Tabaie (changes requested):
A few more inconsistencies but I think it's close.

rollup_spec/src/rollup_spec/rollup.py
Filter94: So we're missing the packing step as well? Whatever the raw byte output we get from ZSTD we have to pack it by 31 byte?

Filter94: If so, the packing issue wasn't introduced in this PR, so I suggest to address the smaller issues and handle the packing in another PR

Tabaie: 31 bytes works. In the current scheme we do 254 bits and get more space in the blob.

Tabaie (approved):

PR #3859, opened 2026-08-26
fix(verifier-ray): bug fixes

YaoJGalteland:
## What

Three independent bugfixes surfaced while running a real, full-scale RISC-V proof through the Zig verifier (smaller synthetic fixtures didn't exercise any of these paths):

1. **PCS aliasing-shift false rejection** (`src/query/pcs.zig`, `test/pcs_test.zig`)
   `reconstruct()` rejected honest dynamic-module sizes when two raw shifts aliased to the same domain point, while prover-ray's `RecoverBatchClaims` legitimately dedupes them before batching. Applies the same dedup in `reconstructQueryValueAt` instead of enforcing `dyn.min_size_log2` as a floor.

2. **Comptime monomorphization blowup in vanishing evaluation** (`src/query/vanishing.zig`, `src/query/logderivativesum.zig`, `codegen/vanishing_zig.go`)
   `evalExpr`/`evalOp` took the expression-tree index/op as `comptime` parameters, so Zig generated a distinct function instantiation per unique node ever evaluated. Real arithmetization modules carry thousands of expression nodes, which blew up into a stack overflow well before any actual recursion-depth problem. Makes `expr_index`/`op` ordinary runtime parameters while keeping `module`/`static_n` comptime, bounding recursion by each module's actual (shallow) expression-tree depth. Also raises the comptime branch-eval quota in `logderivativesum.zig`, and switches codegen's constant-literal emission from `field.Element.init(...)` to a direct struct literal — same "large real module exceeds default comptime limits" family of fix. Includes regenerated `testdata/generated/{vanishing,verify}.zig` fixtures reflecting that literal-syntax change (otherwise `make verify-testdata` would fail).

3. **Stale `prover-ray` module pin breaks the codegen/testdata-generate build** (`codegen/go.mod`, `testdata/generate/go.mod`)
   Both modules required a `prover-ray` pseudo-version pinned to an older commit whose own `go.mod` declares `github.com/LFDT-Lineth/zkc v1.2.25-...`, while this branch's own base (`fixup/bug-in-proofserialization-roundcount`) is already at `zkc v1.2.27`. Any `prover-ray` package that calls a zkc API only present in the newer version breaks `go build`/`make generate-testdata` against the stale pin. Bumps both requirements to the pseudo-version for that base commit (`v0.0.0-20260824082906-f6a0ec502d4d`), so both modules resolve through the real module proxy like every other dependency, rather than pointing at a stale point-in-time snapshot.

Part 1 of a split of #3832.

## Testing

`zig build test`: 76/76 passing. `go test ./codegen/...`: passing. `make generate-testdata` / `make verify-testdata`: passing (no drift) with `prover-ray` resolved via the real module proxy.

### DIFF
--- a/verifier-ray/src/query/pcs.zig
+++ b/verifier-ray/src/query/pcs.zig
@@ -31,6 +31,7 @@ pub const Error = merkle.Error || fri.Error || error{
     RowShapeMismatch,
     ConjugateRowShapeMismatch,
     ClaimPointOnQueryPoint,
+    InconsistentAliasedClaim,
     MissingTopLevelAux,
     BoundaryFinalSelfMismatch,
     BoundaryFinalSiblingMismatch,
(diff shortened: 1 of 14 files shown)


AlexandreBelling (dismissed):
Approved on the basis that these changes are on import and testdata. Except for a genuine fix in pcs.go 

EDIT: Ah but the CI still does not pass...

verifier-ray/src/query/pcs.zig
```diff
@@ -761,6 +774,28 @@ fn reconstructQueryValueAt(
...
+            // already-seen point here reproduces that same single-term
+            // contribution instead of double-counting it.
+            var already_seen = false;
+            for (shifts[0..k]) |prior_shift| {
+                if (shiftedPoint(size_log2, prior_shift, zeta).eql(point)) {
+                    already_seen = true;
+                    break;
+                }
+            }
+            if (already_seen) continue;
```
ivokub: **Blocking — aliased claims must be equality-bound before deduplication.**

These entries come from distinct transcript cells. Skipping the duplicate authenticates only the first value through PCS, while `routeClaims` can later route both raw-shift slots independently into vanishing. A malicious prover could keep the first claim consistent with the commitment and choose a different skipped claim to satisfy the vanishing identity.

`prover-ray`'s `RecoverBatchClaims` explicitly rejects inconsistent duplicate values. A localized fix would retain the index of the earlier alias and compare the two claims before continuing:

```zig
for (shifts[0..k], 0..) |prior_shift, prior_k| {
    if (shiftedPoint(size_log2, prior_shift, zeta).eql(point)) {
        if (!entry_claims[i][prior_k].eql(entry_claims[i][k])) {
            return Error.InconsistentAliasedClaim;
        }
        already_seen = true;
        break;
    }
}
```

Please add `InconsistentAliasedClaim` to the PCS error set and an adversarial regression that changes one of two aliased claim cells. Alternatively, retain the `min_size_log2` rejection until equality binding is implemented.

verifier-ray/src/query/vanishing.zig
```diff
@@ -297,12 +347,12 @@ fn cancellationAtPoint(
...
 
-fn staticRootPower(comptime n: usize, comptime k: usize) field.Element {
+fn staticRootPower(comptime n: usize, k: usize) field.Element {
     const omega = field.rootOfUnityBy(n) catch unreachable;
-    return omega.powComptime(k);
+    return omega.pow(@as(u64, k));
 }
 
-fn normalizePosition(comptime position: i32, comptime static_n: usize, dynamic_n: usize) usize {
+fn normalizePosition(position: i32, comptime static_n: usize, dynamic_n: usize) usize {
```
ivokub, line 391: **Dynamic selector positions need runtime bounds validation.**

`module_sizes` is proof-supplied. If `position < -n`, the subtraction below underflows; positive positions `>= n` are also accepted and effectively reduced by root-of-unity exponentiation. The Go reference's `LagrangeSelector.resolvedRow` rejects positions outside `[-n, n)`.

This can be fixed cheaply by validating before every normalization:

```zig
fn validPosition(position: i32, n: usize) bool {
    if (position >= 0) {
        return @as(usize, @intCast(position)) < n;
    }
    const magnitude: usize = @intCast(-@as(i64, position));
    return magnitude <= n;
}
```

Then, in `evalLagrangeSelector`—and preferably `cancellationAtPoint`, which uses the same helper:

```zig
const n = if (static_n != 0) static_n else ctx.dynamic_n;
if (!validPosition(position, n)) {
    return error.LagrangeSelectorPositionOutOfRange;
}
```

Please add the error to `vanishing.Error` and tests for `position == n`, `position == -n-1`, and the valid boundary `position == -n`.

ivokub (changes requested):
Otherwise looks good, only two small issues my review found.

AlexandreBelling (dismissed):

ivokub (approved):

PR #3861, opened 2026-08-26
feat(verifier-ray): replay shared-randomness pre-sampling hooks in transcript

YaoJGalteland:
## What

This PR mirrors prover-ray #3766 on the verify side. Concretely:

**#3766 (prover-ray)** introduced `messagebus.CompileOptions.SharedRandomness`: an opt-in flag that makes a shard's message-bus permutation challenges α/β derivable from a cross-shard γ seed, instead of each shard drawing them from its own Fiat-Shamir transcript. When enabled, `Compile` declares γ (an 8-limb Poseidon2 octuplet) plus this shard's contribution as public inputs, and installs a pre-sampling hook that overwrites the FS transcript state with γ immediately before α/β are drawn, inside `Runtime.AdvanceRound`.

**This PR (verifier-ray)** replays that exact mechanic on the verify side:

* **`codegen/coin_routing.go`** — structurally detects `messagebus.HasSharedRandomness(sys)` on the compiled prover-ray system, finds which round carries the pre-sampling hook, and locates γ's 8 public-input cells via `sys.LookupPublicInputByTag(messagebus.SharedRandomnessSeedPI, i)` — reading the exact tag/API #3766 added.
* **`codegen/spec_zig.go`** — emits two new fields into the generated Zig `Spec`: `shared_randomness_coin_round` and `shared_randomness_gamma_refs`.
* **`src/protocol/root.zig`** — `replayWithTranscript` checks, at each round, whether `round_index == spec.shared_randomness_coin_round`; if so it calls `gammaDigest()` to rebuild γ from the bound round cells and calls the new `transcript.setState(...)` to overwrite the FS state — right before squeezing that round's coins. This is a byte-for-byte mirror of prover-ray's `Runtime.AdvanceRound` running its `PreSamplingHooks` after absorbing the round's commitment/cells but before squeezing coins.
* **`src/crypto/fiat_shamir.zig`** — adds `Transcript.setState`, mirroring prover-ray's `FiatShamir.SetState`/`MDHasher.SetStateOctuplet`.
* **`src/crypto/multiset_hashing.zig` + `src/query/shared_randomness.zig`** — ports the SIS-based multiset hash and adds the `shared_randomness` sub-verifier: it recomputes the Poseidon2 sponge over the proof's own committed pre-coin rounds and compares the multiset-hash expansion limb-by-limb against the claimed contribution public-input cells. `verifier.verify` wires it into `verifier.Systems` (`.shared_randomness`) and runs it after the row-limit checks.
* **`codegen/shared_randomness*.go`** — extracts the `SharedRandomnessContributionChecker` action and emits the `SharedRandomnessSystem` metadata (ordered rounds + contribution cell refs) consumed by the Zig sub-verifier.

The net effect: two independent shards with divergent local transcript state still derive the same α/β, because both override their transcript with the shared γ at the same point in the round sequence — which is exactly what #3766 built and this PR verifies.

**Bugs found/fixed along the way:**

* `gammaDigest` compared `spec.shared_randomness_gamma_refs.len` against `poseidon2.Digest.len`, but `Digest` is an array *type*, not a value, so `.len` doesn't resolve on it. This path was never compiled by any existing test. Fixed via `@typeInfo(poseidon2.Digest).array.len`.
* **Base-field enforcement on contribution limbs.** The contribution check previously lifted each claimed cell with `toExt()`, which erases the base/extension distinction and would accept an ext-encoded limb equal to `lift(contribution[i])`. The prover-side contract (`messagebus.contributionCell`) requires these cells to be base-field only, and the sibling `gammaDigest` path already rejects ext-encoded cells. The check now pattern-matches the cell and returns `ContributionNotBaseField` for an ext-encoded limb, comparing base-to-base. A regression test re-tags an honest base limb as ext (same numeric value) and asserts it is rejected.

Part 2 of a split of #3832.

## Testing

`zig build test`: passing (includes the new SharedRandomnessContribution E2E fixture and the base-field regression test). `go vet` / `go test ./codegen/...`: passing.

The E2E fixture's negative case flips one of the 328 shared-randomness contribution limbs in the public-input statement and asserts `shared_randomness.verify` rejects it; hand-built unit tests additionally pin that a tampered committed-round octuplet, a tampered contribution limb (including past the first chunk), and an ext-encoded limb are each rejected.

## TODO

- Update when https://github.com/LFDT-Lineth/lineth-monorepo/pull/3844 is merged.

~~- After #3884 merges: restack the chain onto the new main, apply the verifier-ray zkc version update, regenerate fixtures, and zkc-verify-aggregator should go green.~~ (done — #3884 merged, branch rebased onto main)
~~- #3917 fixes the rowlimit action missing gap~~ (done — #3917 merged; the rowlimit codegen now enforces the action-specific `Limit`, and the branch base includes it)

---

> [!NOTE]
> **High Risk**
> Touches Fiat-Shamir replay and cross-shard randomness binding—any mismatch with prover-ray breaks α/β and downstream checks; the new contribution path is security-critical for shard honesty.
> 
> **Overview**
> Adds **verifier-ray** support for prover-ray’s opt-in `messagebus.CompileOptions.SharedRandomness`, so cross-shard message-bus coins (α/β) replay the same way the prover does.
> 
> **Transcript replay:** Go `BuildCoinRouting` emits `shared_randomness_coin_round` and γ limb refs into `protocol.Spec`. `replayWithTranscript` rebuilds γ from bound public-input cells and calls new `Transcript.setState` immediately before squeezing that round’s coins, matching `Runtime.AdvanceRound` pre-sampling hooks.
> 
> **Contribution check:** New `shared_randomness` sub-verifier (plus `multiset_hashing`) recomputes the Poseidon2 sponge over pre–coin-round commitments and compares the expanded digest to the 328 contribution public-input limbs; codegen adds `BuildSharedRandomnessSystem`, Zig emission, and allowlists `SharedRandomnessContributionChecker`. `verifier.verify` runs the check after row limits.
> 
> **Tests/fixtures:** Unit tests for γ-insensitive coins, contribution tampering, and base-field-only limbs; fixture generator adds a `SharedRandomnessContribution` E2E case with a flipped contribution limb negative test. `public_input` spec validation gets a higher comptime branch quota for large PI layouts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e9071a53cd74778f800ec72b78be1ff145962b74. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ivokub (commented):
**Blocking: shared-randomness systems still cannot be emitted by verifier-ray**

The transcript reset implemented here appears to match prover-ray’s `AdvanceRound` ordering, but the shared-randomness path is not usable end to end.

`messagebus.Compile(...SharedRandomness: true)` registers a `*messagebus.SharedRandomnessContributionChecker`. Verifier-ray does not implement or recognize this action in `verifierActionIsHandled`, so `AssertAllVerifierActionsHandled` rejects any system containing an actual shared-randomness message bus.

I reproduced this with a minimal message-bus system: `BuildCoinRouting` correctly found the shared round and eight γ references, but the action audit failed with:

```text
codegen: round 1 registers verifier action
*messagebus.SharedRandomnessContributionChecker,
which the verifier-ray codegen does not emit
```

Neither stacked follow-up closes this gap:

- #3862 only adds guest ELF coverage.
- #3863 passes `SharedRandomness: true`, but `main.zkc` currently emits no message-bus entries, so `messagebus.Compile` returns early. Its generated Zig spec has no `shared_randomness_coin_round` or γ references, and the action audit never encounters the contribution checker.

This fails closed, so it is not presently an acceptance/soundness vulnerability. It is a functional blocker for the feature described by this PR.

Please either:

1. implement the contribution check in Zig/codegen and register `SharedRandomnessContributionChecker` as handled, with a real message-bus integration test and a tampered-contribution rejection test; or
2. narrow the PR description explicitly to the transcript-replay primitive and track the missing verifier action as a blocking follow-up before shared randomness is enabled in a real system.

ivokub (changes requested):

verifier-ray/src/query/shared_randomness.zig
```diff
@@ -0,0 +1,72 @@
...
+    if (system.contribution_refs.len == 0) return;
+
+    var hasher = poseidon2.MDHasher.init();
+    inline for (system.rounds) |round| {
+        if (!round.has_commitment) continue;
+        if (round.round >= ctx.rounds.len) return error.MissingRoundCommitment;
+        const commitment = ctx.rounds[round.round].commitment orelse return error.MissingRoundCommitment;
+        hasher.writeElements(&commitment);
+    }
+    const digest = hasher.sumDigest();
```
ivokub, line 70: **Blocking: this stops at the 8-limb intermediate digest, but the contribution is a 328-limb `MSetHash`**

The Go reference performs two distinct hashing steps:

1. `sharedRandomnessContribution` absorbs every commitment preceding the coin round and calls `hasher.SumDigest()`, producing an 8-element root (`prover-ray/wiop/compilers/messagebus/shared_randomness.go:154-168`).
2. It then returns `multisethashing.Hash(root)`, not the root itself. `Hash` inserts that root into an `MSetHash` (`prover-ray/crypto/koalabear/multiset_hashing/msethash.go:100-105`). `MSetHash` contains `41 * 8 = 328` field elements, and `update` derives all 41 Poseidon2 chunks by repeatedly appending a zero octuplet (`msethash.go:23-39,68-92`).

That distinction is reflected in the protocol layout: `NumSharedRandomnessContribution` is `len(multisethashing.MSetHash{})`, and `BuildSharedRandomnessSystem` consequently emits 328 `contribution_refs`.

The Zig implementation currently computes only the first 8-element `sumDigest()` and then indexes it once for every emitted contribution reference. A real shared-randomness system therefore reaches `digest[8]` and beyond. The direct Zig tests do not expose this because their synthetic `two_round_system_refs` contains only eight entries.

Please port the equivalent of `multisethashing.Hash` to Zig and compare all 328 generated contribution limbs:

```text
commitments -> MDHasher.sumDigest() -> multiset_hashing.Hash(root) -> compare 328 limbs
```

This must also be wired into the full generated verifier. `verifier.verify` already calls `shared_randomness.verify(systems.shared_randomness, ctx)`, but `writeVerifyCase` currently omits `.shared_randomness = system_N_shared_randomness` from the generated `verifier.Systems` initializer, so it defaults to the empty no-op system.

We need this because these public-input cells are the shard contribution combined by the cross-shard shared-randomness protocol. Once `SharedRandomnessContributionChecker` is marked as handled, verifier-ray must enforce the same 328-limb relation as the Go verifier; otherwise the previous fail-closed action audit has been removed without equivalent verification.

Please add one generated message-bus integration case that goes through `BuildCompiledSystem` and `verifier.verify`, accepts the honest proof, and rejects a proof/public-input statement with one of the 328 contribution limbs changed. That would exercise the Go metadata extraction, generated Zig wiring, multiset-hash expansion, public-input binding, and final verifier dispatch together.

ivokub (approved):
Thanks for addressing the comments -- the PR looks good now.

PR #3862, opened 2026-08-26
feat(prover-ray): add guest ELFs to cover testing the full RISC-V instruction set

YaoJGalteland:
## What

`main.zkc` compiles once into a generic RISC-V interpreter circuit — it never encodes any specific program. The only way to test that the circuit correctly constrains each instruction's semantics is to feed it a real execution trace, which means feeding it a real guest ELF. Without one, the circuit can only be checked against synthetic hand-built fixtures, never against an actual RISC-V binary running through the real pipeline.

`internal/minimal-elf` is a small ELF-builder (not a compiler toolchain) that hand-assembles a minimal, valid RISC-V ELF exercising the full RV64I + M-extension + custom-precompile surface in a single guest. The guest self-checks every result and exits 0 (pass) or 1 (fail) through the guest exit syscall, so a wrong circuit constraint shows up as either a rejected honest proof or an accepted dishonest one.

`backend/zkc-r5/minimal_guest.go` exposes it as a ready-made witness for driver tests:

* **`AllInOneGuestELF`** — one 624-byte guest covering: memory round-trip (SW/LW); LUI, JAL, JALR, an R-type base op, and an M-extension op; all six B-type branch variants (BEQ, BNE, BLT, BGE, BLTU, BGEU), both taken and not-taken; SB/LB, SH/LH, SD/LD including sign-extension correctness; the `R_POSEIDON2`/`R_KECCAK` custom precompiles checked against known-good vectors; `R_WRITE_OUTPUT` copying known bytes into the public `guest_output` buffer; every I-type ALU immediate (SLTI, SLTIU, XORI, ORI, ANDI, SLLI, SRLI, SRAI, ADDIW, SLLIW, SRLIW, SRAIW); and the RV64 R-type word-width ops (ADDW, SUBW, SLLW, SRLW, SRAW) — each cross-checked against an independently computed expected value.

`zkcdriver/r5_test.go`'s `TestRisc5InstructionCoverageGuest` exercises it through the real zkc driver pipeline, including the write-output assertion.

Since `main.zkc`'s circuit doesn't depend on the guest witness, proving 10 separate per-instruction-family guests produced 10 proofs of identical size for no extra coverage over one guest exercising every instruction in a single trace — see [downstream discussion](#3863) for the proof-size confirmation. This PR originally introduced those 10 guests (`ExitZeroGuestELF`, `ExitOneGuestELF`, `MemoryRoundTripGuestELF`, `ArithmeticGuestELF`, `BranchesGuestELF`, `LoadStoreWidthsGuestELF`, `ImmediateALUGuestELF`, `WordWidthGuestELF`, `Poseidon2GuestELF`, `KeccakGuestELF`, `WriteOutputGuestELF`) individually; they've since been consolidated into `AllInOneGuestELF`. `ExitOneGuestELF` (which `main.zkc` must *reject*, testing the negative path) is not part of the merge and has no replacement here — nothing in this stack currently exercises that rejection path.

The merged guest was assembled with the real `riscv64-linux-gnu-as` toolchain rather than hand-encoded, which caught a real bug in transit: `SRLIW`/`SRLW` zero-extend their 32-bit result (unlike `ADDIW`/`SLLIW`/`SRAIW`/`SRAW`, which sign-extend), so their comparison constants need an extra mask that a naive port of the original per-guest logic was missing.

Part 3 of a split of #3832.

## Testing

`go vet ./...`: clean. `go test ./internal/minimal-elf/... ./backend/zkc-r5/...`: passing. `go test ./zkcdriver/... -run TestRisc5`: passing (~12s, down from ~100s across the original 10 guests).

---

> [!NOTE]
> **Low Risk**
> Changes are limited to test fixtures and a new zkcdriver integration test; no production proving, auth, or data-path behavior is altered.
> 
> **Overview**
> Adds a **single consolidated test guest** for the RISC-V arithmetization instead of relying only on minimal `auipc`-only ELFs or many per-family binaries.
> 
> **`internal/minimal-elf`** now exposes `AllInOneSectionData` (hand-assembled ~624-byte RV64I + M + custom-precompile program with inline pseudocode docs) and `AllInOneElfProgram`, built via the existing `Make` helper. The guest self-checks memory, branches, sub-word loads/stores, Poseidon2/Keccak precompiles, immediate and word-width ALU ops, and `R_WRITE_OUTPUT`, failing through the Linux-style exit syscall if any check mismatches.
> 
> **`zkcdriver/r5_test.go`** adds `TestRisc5InstructionCoverageGuest`, which feeds `AllInOneElfProgram` through `predecoding` and the real embedded `main.zkc` interpreter with **trace constraint checking enabled** (not full prove/verify), and asserts **`guest_output` is `"ABC"`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9f1424da2f6e751501150ee2f643414e0a85a725. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ivokub (dismissed):
Thanks! Really nice feature!

Though lets wait merging it before the ones before are (i.e. lets not merge this one into previous PR).

AlexandreBelling (dismissed):

AlexandreBelling (dismissed):

ivokub (dismissed):

prover-ray/internal/minimal-elf/minimal_elf.go
```diff
@@ -13,6 +13,227 @@ const (
 // ValidSectionData is a valid RISC-V auipc x5, 0 instruction encoding.
 var ValidSectionData = []byte{0x97, 0x02, 0x00, 0x00}
 
+// AllInOneSectionData is a tiny valid RISC-V program that exercises the full
```
AlexandreBelling, line 16: Wouldn't that be better to directly write that in a proper elf-file and load it from disc? I also do not see a point of embedding it in the binary.

AlexandreBelling: Not a blocking comment. Just pointing out this is not a practice we should generalize to everytime our code interacts with files.

ivokub: The MinimalElf/AllInOneElf are not necessary for proving -- it is just some test R5 ELF which can be used in tests. Right now in the tests we need to build the `verifier-ray` executable (which requires zig pipeline), but could only use the synthetic R5 ELF as well. See also comment above for generating a synthetic ELF with some particular desired features (i.e. nb on instructions, expected number of R5 cycles per instruction etc.).

And imo -- for the final binaries the synthetic ELF will not be included as its only used in the tests, so the linker should ignore it.

prover-ray/zkcdriver/r5_test.go
```diff
@@ -6,10 +6,47 @@ import (
...
+// every compiled constraint. The single guest exercises the full RV64I +
+// M-extension + custom-precompile surface (memory round-trip, LUI/JAL/JALR,
+// ADD/MUL, all six branch variants, sub-word loads/stores with sign extension,
+// the Poseidon2/Keccak/write-output precompiles, every immediate ALU op, and
+// RV64 word-width ops) in one trace, and asserts the write-output bytes. This
+// only traces and checks constraints — it does not run the (expensive)
+// prove/verify pipeline, which is already covered by
+// verifier-ray/codegen/riscv_bootstrap.go.
+func TestRisc5InstructionCoverageGuest(t *testing.T) {
+	binf, err := compileBinaryConstraints(riscvMainZkcPath)
```
ivokub: Can you use
```go
	binf, err := embedded.CompiledBinaryFile()
	if err != nil {
		b.Fatalf("failed to compile embedded R5 arithmetization: %v", err)
	}
```
instead? This gets the pinned arithmetization version (see the benchmark below).

ivokub (approved):

PR #3863, opened 2026-08-26
feat(verifier-ray): verify a real R5 proof image end-to-end against main.zkc

YaoJGalteland:
## What

Wires verifier-ray up to verify a real, honest RISC-V proof end-to-end: compiling `arithmetization/src/main/riscv/main.zkc` into a `wiop.System`, proving `zkc_r5.AllInOneGuestELF` (a single guest that exercises the full RV64I + M-extension + custom-precompile surface in one witness, from #3862) through the actual prove/PCS/verify pipeline, and verifying that proof through the real Zig verifier instead of relying only on synthetic scenario fixtures.

* `codegen/riscv_bootstrap.go` compiles `main.zkc` through the full compiler pipeline (`nonnative -> rangecheck -> lookuptologderivsum -> messagebus -> grandproduct -> logderivativesum -> localvanishing -> global -> pcs`) and exposes `BuildAllInOneHonestRiscvArtifacts()`, which proves `AllInOneGuestELF` against it and returns the verifier-facing artifacts.
* `codegen/generate-riscv-system` drives `BuildAllInOneHonestRiscvArtifacts()` to emit `testdata/generated/riscv_system.zig` and `testdata/riscv_proof_image.bin` (the single committed cross-language fixture).
* `src/main.zig`'s native and R5-zkVM entry points now load and verify this real proof image (mmap on native, linked memory on R5) instead of only embedded synthetic fixtures.
* `test/riscv_proof_image_test.zig` replaces the old hand-built-fixture byte test with a real end-to-end `verifier.verify()` call against the honest proof.
* `test/riscv_system_test.zig` smoke-tests the generated system type-checks as `verifier.Systems`.

## Architecture: how main.zkc and the guest ELF become a verified proof

`main.zkc` compiles once into a generic RISC-V interpreter circuit (every instruction's AIR logic — ADD, SUB, branches, memory, ...); it never encodes any specific program. The guest ELF supplies the specific program as public input (raw instruction/data bytes), and only running the interpreter against those bytes produces the witness (the actual execution trace) that gets proven.

```
End-to-end:

zkc source                                        guest ELF bytes (+ input data)
   │                                                        │
   │ compileBinaryConstraints                               │ zkcr5.PrepareInput
   v                                                        v
constraints.BinaryFile                            public inputs (entry point, memory blobs)
   │                                                        │
   │ NewZkCDriver + wiop.NewSystemf                         │
   v                                                        │
wiop.System                                                 │
   │                                                        │
   │ runCompilePipeline                                     │
   v                                                        │
compiled wiop.System ───────────────────────────────────────┤
   │                                                        │ binFile.Trace
   │ codegen.BuildCompiledSystem                            v
   v                                            compiled wiop.System + witness
codegen.CompiledSystem                                       │
   │                                                         │ sys.Prove
   │ codegen.WriteCompiledSystemZig                          v
   v                                            wiop.Proof + wiop.PublicInput
riscv_system.zig                                             │
   │                                                         │ proofserialization.Project / Encode
   │ Zig import + smoke test                                 v
   v                                                  riscv_proof_image.bin
verifier.Systems                                            │
   │                                                        │ mmap + @ptrCast
   │                                                        v
   │                                              verifier.VerifyInput
   │                                                         │
   └─────────────────────-─┬─────────────────────────────────┘
                           │ verifier.verify(spec, systems, proof, public_inputs)
                           v
                    accept / reject
```
So the verifier needs exactly two things, from two independent supply chains that only meet at verify-time:

1. **`verifier.Systems`** — "What is the protocol?" Compiled once from `main.zkc`; identical for every RISC-V guest ELF ever run through it.
2. **`verifier.VerifyInput`** — "What proof/instance should I verify?" Specific to one guest ELF's honest execution.

## Bugs found and fixed along the way

**Stale `prover-ray` pin blocked the build.** `codegen/go.mod` and `testdata/generate/go.mod` pinned a `prover-ray` pseudo-version predating `AllInOneGuestELF` (#3862) and the `zkc v1.2.27` bump this stack's base already carries. Both were repinned to #3862's actual pushed commit via real proxy resolution, dropping an interim `replace` directive. `make generate-testdata` / `make verify-testdata` now pass cleanly with zero drift.

**Two different fixtures silently collided on the same filename.** `codegen/generate-riscv-system` wrote the real `main.zkc` honest-proof image to `testdata/proof_image.bin` — the exact path prover-ray's own `TestVerifierRayImageIsUpToDate` (`wiop/proofserialization/abi_agreement_test.go`) uses for an unrelated, small synthetic `VerifyInput` fixture at a different base address (`0x400000000`, for macOS mmap compatibility, vs. this real proof's `0x08800000`/`GuestBase`). Whichever generator ran last clobbered the other's fixture with content its own reader can't decode. Renamed the real-proof fixture to `testdata/riscv_proof_image.bin` everywhere, so the two fixtures stop sharing a path. Also hardened the proof-image test's mmap from `MAP_FIXED` to `MAP_FIXED_NOREPLACE`, so a future collision fails loudly instead of silently overlapping.

**A binary proof fixture was silently subject to git's line-ending normalization.** `.gitattributes`' `* text eol=lf` catch-all had no exclusion for `verifier-ray/**/*.bin`, so git treated these ~52MB proof images as text and applied CRLF/LF normalization on checkout/checkin — capable of silently corrupting bytes that happen to look like line endings. The previously-committed `testdata/riscv_proof_image.bin` had in fact been corrupted this way and failed `proofserialization.Decode` outright (`presence flag 8 is neither null nor present`) despite `git diff` reporting no change, because the stale index entry had itself been normalized to match. Added `verifier-ray/**/*.bin binary` (mirroring the existing `prover/**/*.bin` entries) and re-added the fixture via `git add --renormalize`.

None of the first two issues were caught before because **the verifier-ray CI workflow has never actually run a check on this PR** (confirmed via `gh api .../check-runs` on past head commits — no `verifier-ray` job appears), despite the PR touching `verifier-ray/**`.

## Why split out

Final part of a split of #3832, stacked on the guest-ELF generator PR (#3862). This is the final wiring plus the generated production system — necessarily last since it depends on every prior part (bugfixes, shared-randomness plumbing, guest-ELF fixtures).

## Testing

* `zig build test`: 80/80 passing (~22s).
* `zig build test -Dverifier-profiling=true`: passing.
* `make fmt` / `make test` / `make generate-testdata` / `make verify-testdata`: passing, no drift.
* `make build-release`, `make run-passing`, `make run-failing-expected`: passing.
* `go test ./codegen/...` (including `TestBuildAllInOneHonestRiscvArtifacts`, ~9s): passing.

---

> [!NOTE]
> **Medium Risk**
> Touches proof loading, in-memory layout relocation, and committed verifier metadata for the real RISC-V circuit; incorrect pointer rebasing or fixture drift could break or mis-run verification smoke tests.
> 
> **Overview**
> **End-to-end RISC-V verifier bootstrap:** verifier-ray now smoke-tests and CI-checks against a **real** honest proof from `arithmetization/.../main.zkc` and `zkc_r5.AllInOneGuestELF`, not only synthetic embedded scenarios.
> 
> Go adds **`BuildAllInOneHonestRiscvArtifacts()`** and **`generate-riscv-system`**, which prove the guest through the full prover pipeline and commit **`testdata/generated/riscv_system.zig`** plus **`testdata/riscv_proof_image.bin`** (kept separate from prover-ray’s synthetic `proof_image.bin`). **`make generate-testdata` / `verify-testdata`** now guard both the Zig system and the proof image.
> 
> **Runtime loading:** default **`run`** / **`zkc-verify`** feed the proof bytes at `_in_start` via **`@testdata/riscv_proof_image.bin`**; **`main.zig`** verifies with the generated **`riscv_system`** spec/systems. Native load tries **`MAP_FIXED`** at guest base **`0x08800000`**, otherwise reads the file and calls new **`image_relocation.rebase`** so macOS can mmap at an arbitrary address. Embedded invalid fixtures remain for negative tests only.
> 
> **Tests & hygiene:** **`riscv_proof_image_test.zig`** runs full **`verifier.verify`** on the committed image; **`proof_image_test.zig`** (byte-level synthetic fixture) is removed. **`.gitattributes`** marks **`verifier-ray/**/*.bin`** as binary to avoid CRLF corruption of large proof blobs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 85f0d076b952e8646495f0d45cf8a546ef409c39. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


verifier-ray/src/main.zig
```diff
@@ -11,7 +12,8 @@ const is_native_os = builtin.target.os.tag == .linux or builtin.target.os.tag ==
 const is_native_arch = builtin.target.cpu.arch == .x86_64 or builtin.target.cpu.arch == .aarch64;
 const is_supported_native = is_native_os and is_native_arch;
 
-const native_input_path: [:0]const u8 = "zig-out/input.bin";
+const native_input_path: [:0]const u8 = "testdata/riscv_proof_image.bin";
+const input_guest_base: usize = 0x08800000;
```
YaoJGalteland, line 17: Could we update the production `GuestBase` and the zkC linker script (0x08800000) to an address that works on both macOS and Linux?

On arm64 macOS, the low 4 GiB of the address space (0x00000000 – 0x100000000) is reserved/restricted by the kernel and dyld, so `MAP_FIXED` at the current `0x08800000` fails with `EPERM` or `ENOMEM`.

On Linux/x86_64, only the very low null-pointer guard region (0x0000 – 0xFFFF) is reserved, so `0x08800000` works there.

An address above `0x100000000` should work on both platforms. For example, I tested `0x400000000` (16 GiB) successfully.

ivokub: Resolved -- we now mmap without specifying the target and just rewrite the pointers with your relocation library.

ivokub (approved):

arijitdutta67 (approved):

PR #3872, opened 2026-08-27
fix(ci): derive e2e image configuration from artifacts

nadeemb53:
- make coordinator and Maru images expose their launcher JAR through `LINETH_APP_JAR`
- make Compose consume the image-owned launcher path and canonical image repositories
- discover and load Docker artifacts from their manifests instead of service-specific filenames
- select E2E tags from one canonical image table, preserving explicit caller tags and falling back to `develop`

### DIFF
--- a/docker/compose-spec-l2-services.yml
+++ b/docker/compose-spec-l2-services.yml
@@ -13,7 +13,7 @@ services:
       - ./config/coordinator/coordinator-config-v2.toml:/coordinator/coordinator-config-v2.toml:ro
 
   maru:
-    image: ${MARU_IMAGE_NAME:-consensys/maru}:${MARU_TAG:-1.3.0-20260714-ced387a}
+    image: consensys/maru:${MARU_TAG:-1.3.0-20260714-ced387a}
     hostname: maru
     container_name: maru
     profiles: [ "l2", "l2-bc", "debug" ]
@@ -22,17 +22,18 @@ services:
         condition: service_healthy
     environment:
       LINETH_USE_MARU_OVERRIDE_CONFIG: ${LINETH_USE_MARU_OVERRIDE_CONFIG:-false}
-      # Which jar to launch. Override to point at a different entrypoint jar.
-      MARU_COMMAND: exec java -Xmx384m -Dlog4j2.configurationFile=configs/log4j.xml -jar ${MARU_JAR:-maru.jar} --genesis-file /initialization/genesis-maru.json --config configs/config.toml
     command:
       - /bin/sh
       - -c
       - |
+        set -- java -Xmx384m -Dlog4j2.configurationFile=configs/log4j.xml \
+          -jar "$$LINETH_APP_JAR" \
+          --genesis-file /initialization/genesis-maru.json \
+          --config configs/config.toml
         if [ "$$LINETH_USE_MARU_OVERRIDE_CONFIG" = "true" ]; then
-          $$MARU_COMMAND --config configs/config.overrides.toml
-        else
-          $$MARU_COMMAND
+          set -- "$$@" --config configs/config.overrides.toml
         fi
+        exec "$$@"
 #    command: [ 'echo' , 'Forced exit to run in IntellIJ' ]
     volumes:
       - ./config/maru/sequencer.config.toml:/opt/consensys/maru/configs/config.toml:ro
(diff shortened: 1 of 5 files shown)


docker/compose-spec-l2-services.yml
```diff
@@ -22,17 +22,18 @@ services:
...
     environment:
       LINETH_USE_MARU_OVERRIDE_CONFIG: ${LINETH_USE_MARU_OVERRIDE_CONFIG:-false}
-      # Which jar to launch. Override to point at a different entrypoint jar.
-      MARU_COMMAND: exec java -Xmx384m -Dlog4j2.configurationFile=configs/log4j.xml -jar ${MARU_JAR:-maru.jar} --genesis-file /initialization/genesis-maru.json --config configs/config.toml
     command:
       - /bin/sh
       - -c
       - |
+        set -- java -Xmx384m -Dlog4j2.configurationFile=configs/log4j.xml \
+          -jar "$$LINETH_APP_JAR" \
```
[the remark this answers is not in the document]
nadeemb53: Addressed in 932ed2556c. Maru and coordinator now fall back to their legacy OSS JAR paths when LINETH_APP_JAR is absent, while images implementing the new contract continue to supply their own full path.

docker/compose-spec-l2-services.yml
```diff
@@ -13,7 +13,7 @@ services:
       - ./config/coordinator/coordinator-config-v2.toml:/coordinator/coordinator-config-v2.toml:ro
 
   maru:
-    image: ${MARU_IMAGE_NAME:-consensys/maru}:${MARU_TAG:-1.3.0-20260714-ced387a}
+    image: consensys/maru:${MARU_TAG:-1.3.0-20260714-ced387a}
```
jonesho: @nadeemb53 is it possible to keep the injectable env image name `MARU_IMAGE_NAME` (i.e. which can be either `consensys/maru` or `consensys/lineth-enterprise-maru`)? same for `BESU_PACKAGE_IMAGE_NAME` and `LINEA_COORDINATOR_IMAGE_NAME`? as I found it useful to have a way to inject the image names, especially when running these stack locally

We could perhaps modify the logic [here](https://github.com/LFDT-Lineth/lineth-monorepo/pull/3872/changes#diff-d2c333addaa2baa2ab985deb9be7698273e02e5332ada2bf2855cddc3f04c33fR108) to extract the actual image name from the downloaded artifacts, and then properly set those `*-IMAGE_NAME` env variables, wdyt?

nadeemb53: I'll add the injectable Compose image-name variables back for local use.

`manifest.json` only tells us the repository tags. It doesn't say which service or environment variable a given image belongs to, so we can't map them automatically without adding extra image metadata.

jonesho (approved):

docker/compose-spec-l2-services.yml
jonesho: @nadeemb53 but [repo_tags](https://github.com/LFDT-Lineth/lineth-monorepo/pull/3872/changes#diff-d2c333addaa2baa2ab985deb9be7698273e02e5332ada2bf2855cddc3f04c33fR114) should give us the "image_name:image_tag" for the loaded docker image, no? I was thinking about extracting the image name and image tag from `repo_tags` and assign the values of `*_IMAGE_NAME` and `*_TAG` used in the [docker/compose-spec-l2-services.yml](https://github.com/LFDT-Lineth/lineth-monorepo/pull/3872/files/932ed2556c27062dbf35ab6c44d813cc89bc3606#diff-e538882b59032d8ac081f9c6cf4f390597b0563f303974062f5b7f318e3b0507) file properly

nadeemb53: @jonesho yes pulling `image_name` and `image_tag` out of `repo_tags` is fine. I am not sure about deciding which `*_IMAGE_NAME` and `*_TAG` variables to assign without falling back on repository-name patterns.

We could treat suffixes like `-maru` and `-coordinator` as the convention, but that just trades one contract for another. The canonical alias gives us an explicit identity instead. Does that distinction make sense?

jonesho: yes, we could leave it like this for now

fluentcrafter (approved):

PR #3874, opened 2026-08-28
refactor(arithmetization): extract elfmapping and predecoding into Go pkg, extract elf_to_json into reusable cmd

ivokub:
This PR refactors the tools in the arithmetization so that they can be reused downstream in prover and verifier as Go packages instead of needing to run through elf_to_json tool. This helps with running Go tests in prover-ray, implementing the prover backend and code generation of the spec in verifier-ray

### Checklist

  PR.


Soleimani193 (approved):

AlexandreBelling (approved):

arithmetization/gopkg/predecoding/doc.go
```diff
@@ -0,0 +1,216 @@
+// Package predecoding converts executable R5 ELF blobs into the decoded
```
amkCha, line 1: Isn't it better as a .md ? Or having it as .go enable certain functionalities ?

ivokub: Yes, then it becomes package documentation, available through `go doc` (and also the Go language server can access it). We don't have to open the repository and look for the `.md` files to find it. The point is that downstream we want to consume Go package APIs, not look into the implementation in the corresponding git repo.

I.e. look at the example below:
```bash
$ go doc github.com/LFDT-Lineth/lineth-monorepo/arithmetization/gopkg/predecoding
package predecoding // import "github.com/LFDT-Lineth/lineth-monorepo/arithmetization/gopkg/predecoding"

Package predecoding converts executable R5 ELF blobs into the decoded
instruction table consumed by the R5 arithmetization.

# Overview

The R5 interpreter dispatches on a predecoded instruction table instead of
decoding opcode fields during every execution step. Predecode consumes an
elfmapping.Program and returns instruction_base plus the packed decoded table.
...
const InstructionBaseInput = "instruction_base" ...
const DefaultMaxDecodedRecords uint64 = 2_000_000
func PrepareInputs(elfBytes []byte, inputData []byte, options ...Option) (map[string][]byte, error)
type DecodedProgram struct{ ... }
    func Predecode(program elfmapping.Program, options ...Option) (DecodedProgram, error)
type Option func(*config) error
    func WithIncludeExecutable() Option
    func WithMaxDecodedRecords(maximum uint64) Option
    func WithSectionsWriter(writer io.Writer) Option

$ go doc github.com/LFDT-Lineth/lineth-monorepo/arithmetization/gopkg/predecoding.Predecode
package predecoding // import "github.com/LFDT-Lineth/lineth-monorepo/arithmetization/gopkg/predecoding"

func Predecode(program elfmapping.Program, options ...Option) (DecodedProgram, error)
    Predecode creates one packed instruction row for every four-byte word in the
    aligned span containing the program's executable blobs.
```
(it also looks good in IDEs, example below)
image

PR #3893, opened 2026-08-31
feat(verifier-ray): verify two r5 proofs 

YaoJGalteland:
## Summary

Adds a two-proof aggregator to verifier-ray: `verifier.verifyPair` verifies two proofs against the same compiled RISC-V system and checks their public-input statements agree, so a caller (e.g. a recursive R5 guest) can attest "both of these proofs hold, about the same statement" as a single combined claim.

- **`verifier.verifyPair`/`AggregatorInput`** (`src/verifier.zig`) — validates both proofs' public-input lengths against the spec, checks the two statements are element-wise equal (`checkPublicInputConsistency`), then runs `verify` on each. The length check runs before the consistency comparison so an untrusted pair image can't force wasted work or an out-of-bounds read with a matching-but-invalid length.
- **Aggregator entry point** (`src/main.zig`, `build.zig`) — a `-Daggregator` build flag now actually dispatches `main()`/`r5_main()` to `verifyPair` via new `loadNativePairInput`/`loadR5PairInput` loaders, instead of silently falling through to the single-proof path.
- **Pair image format** (`codegen/aggregator_image.go`) — `EncodeAggregatorPair` packs two proofs into one image: a 16-byte header of two absolute pointers followed by each proof's existing `proofserialization.Encode` output, 8-byte aligned.
- **Pointer relocation** (`src/image_relocation.zig`, new) — factors the "encoded pointer -> mapped host pointer" arithmetic out of `main.zig`'s macOS/native mmap fallback into a small, independently tested module, so each sub-image in the pair is rebased relative to its own start address rather than the whole image's base.
- **New/expanded fixtures** — a genuinely re-proven `alt` fixture (different public-input value) and `altSameStatement` fixture (same public-input value, independent witness) for `OpenedCellPublicInput`, so `verify_pair_test.zig` exercises both the rejection path (different statements) and the acceptance path (matching, non-trivial statements) — previously only trivial empty-statement pairs were covered.
- **Makefile/CI wiring** — `build-r5-aggregator`, `run-aggregator`, `zkc-verify-aggregator` targets to build/run/interpret the aggregator through the real R5/zkc pipeline.

---

> [!NOTE]
> **Medium Risk**
> Touches core verifier aggregation and untrusted proof-image pointer rebasing; mistakes could accept mismatched statements or mis-parse pair layouts, though coverage is extensive.
> 
> **Overview**
> Adds a **two-proof aggregator** path to verifier-ray: a compile-time `-Daggregator` entry point loads a pair image (two `VerifyInput`s behind an `AggregatorInput` header), runs **`verifyPair`** (public-input consistency first, then two full `verify` calls against the same RISC-V system), and wires native/R5 smoke targets plus zkc.
> 
> **Pair image encoding** is new Go code (`EncodeAggregatorPair`): 16-byte absolute pointers plus two sub-images from existing `proofserialization.Encode`, with a generator that derives pair fixtures from the committed single proof image (not committed to git). **Loading** extends `main.zig` with pair mmap/rebase (shared `image_relocation` + header patching) and pins `AggregatorInput` in `proof_abi.zig`.
> 
> **Tests/fixtures**: `verify_pair_test.zig`, end-to-end aggregator image test, relocation tests, shared `fixture_map`; public-input fixtures gain **alt** (different statement) and **altSameStatement** proofs for pair consistency coverage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7a787bd648337112edb636f6e4b918519ed311bd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


verifier-ray/src/image_relocation.zig
```diff
@@ -0,0 +1,25 @@
...
+    if (@bitSizeOf(usize) != 64) @compileError("image_relocation assumes a 64-bit usize");
+}
+
+pub const Error = error{
+    PointerBeforeImage,
+    AddressOverflow,
+};
+
+/// Returns an encoded absolute pointer's byte offset within its image.
+pub fn imageOffset(encoded_pointer: u64, encoded_base: usize) Error!usize {
```
YaoJGalteland, line 16: @ivokub do you know if we need to provide the macOS native support? The production R5 path doesn't need it at all. the image_relocation could be removed if we dropped macOS native support, or redesigned the image format to use relative offset

PR #3894, opened 2026-08-31
feat(riscv-guest): rollup guest program stub with shared SSZ package

Filter94:
Introduces guest-common (generic SSZ primitives extracted for reuse), the rollup guest stub (echo/sentinel, no proof verification/folding), read_input in lineth-accelerators, and rewires l2-execution's SSZ codec onto guest_common.

This PR implements issue(s) #

### Checklist

  PR.


riscv-guests/guest-common/src/ssz.zig
```diff
@@ -0,0 +1,87 @@
...
+//! non-monotonic/overlapping element ordering) — a caller catches one error across an entire
+//! decode call, regardless of which layer (this module's list walking or a container's own
+//! fixed-head checks) detected the problem.
+
+const std = @import("std");
+
+// ── Primitive reads/writes (little-endian, matching SSZ) ────────────────────
+
+pub inline fn readU32(data: []const u8, off: usize) u32 {
+    return std.mem.readInt(u32, data[off..][0..4], .little);
```
gauravahuja, line 24: What is the expectation when data size is less than offset + 4? I am not sure if invalid data is read incase index is out of bounds. Same concern about readU64 and writing

riscv-guests/guest-common/src/ssz.zig
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// ── Generic "List[VariableSizeType, N]" codec ────────────────────────────────
+//
+// SSZ encodes a list of variable-size elements exactly like a container's
+// variable-field region: an offset table (4 bytes per element, each an
+// absolute offset from the start of this region) followed by the
+// concatenated element bytes, in order.
+
+pub fn decodeVariableList(alloc: std.mem.Allocator, data: []const u8, max_len: usize) ![]const []const u8 {
```
gauravahuja: I think max_items is a better name for max_len.

riscv-guests/guest-common/src/ssz.zig
```diff
@@ -0,0 +1,87 @@
...
+    if (first_off > data.len) return error.InvalidSsz;
+    const n = first_off / 4;
+    if (n > max_len) return error.BoundsViolation;
+
+    const result = try alloc.alloc([]const u8, n);
+    for (0..n) |i| {
+        const off_i = readU32(data, i * 4);
+        const end_i: u32 = if (i + 1 < n) readU32(data, (i + 1) * 4) else blk: {
+            if (data.len > std.math.maxInt(u32)) return error.InvalidSsz;
+            break :blk @intCast(data.len);
```
gauravahuja: This is difficult to read, I think it can be simplified

```
if (data.len > std.math.maxInt(u32)) return error.InvalidSsz;                                                                                                                                                                                                                                      
for (0..n) |i| {                                                                                                                                                                                                                                                                                   
      const off_i = readU32(data, i * 4);                                                                                                                                                                                                                                                            
      const end_i: u32 = if (i + 1 < n) readU32(data, (i + 1) * 4) else @intCast(data.len)
```

riscv-guests/rollup/src/rollup.zig
```diff
@@ -0,0 +1,95 @@
...
+pub fn run(alloc: std.mem.Allocator, input: rollup_ssz.RollupProofPrivateInput) !rollup_ssz.RollupOutput {
+    if (input.l2_execution_proofs.len == 0) return error.EmptyProofs;
+    const first = input.l2_execution_proofs[0];
+    const last = input.l2_execution_proofs[input.l2_execution_proofs.len - 1];
+
+    const program_vks = try dedupSortedProgramVks(alloc, input.l2_execution_proofs);
+
+    var filtered_addresses: std.ArrayListUnmanaged([20]u8) = .empty;
+    for (input.l2_execution_proofs) |p| {
+        for (p.proof.filtered_addresses) |addr| try filtered_addresses.append(alloc, addr);
```
gauravahuja: is there any bound on filtered_addresses? Looks like Number of proofs * Number of addr/proof * 20 bytes? Will this result in OOM?

riscv-guests/guest-common/src/ssz.zig
```diff
@@ -0,0 +1,87 @@
...
+    if (data.len == 0) return &.{};
+    if (data.len < 4) return error.InvalidSsz;
+
+    const first_off = readU32(data, 0);
+    if (first_off == 0 or first_off % 4 != 0) return error.InvalidSsz;
+    if (first_off > data.len) return error.InvalidSsz;
+    const n = first_off / 4;
+    if (n > max_len) return error.BoundsViolation;
+
+    const result = try alloc.alloc([]const u8, n);
```
gauravahuja, line 60: Will this result in a memory leak if an error is returned? No one frees this in case of error

riscv-guests/lineth-accelerators/src/io.zig
```diff
@@ -1,6 +1,33 @@
...
+
+/// Size of the linker-defined IN memory region (1 GiB).
+const IN_REGION_SIZE: usize = 0x40000000;
+
+/// Read the private input from the memory-mapped IN region: hands back a zero-copy slice into
+/// that region via `buf_ptr`/`buf_size`. The input slot's address is this module's own detail —
+/// a guest entry point only ever sees the slice.
+pub fn read_input(buf_ptr: *[*]const u8, buf_size: *usize) void {
+    const base = @intFromPtr(&_input_start);
+    const size_ptr: *const u64 = @ptrFromInt(base);
```
gauravahuja: This is a complex casting. input_start is u8 but you are casting it to * u64. I don't know enough zig and memory alignment issues to assertatively say if this will lead to some runtime error or not.

riscv-guests/rollup/src/rollup_ssz.zig
```diff
@@ -0,0 +1,614 @@
...
+    pos += 4;
+    writeU32(body, pos, @intCast(off_chunks));
+    pos += 4;
+    writeU32(body, pos, @intCast(off_proofs));
+    pos += 4;
+    writeU32(body, pos, @intCast(off_prefix));
+    pos += 4;
+    writeU32(body, pos, @intCast(off_suffix));
+    pos += 4;
+    writeU32(body, pos, @intCast(off_boundary));
```
gauravahuja: Do you need to assert off_conflations, off_chucks etc are less than 2^32?

riscv-guests/guest-common/src/ssz.zig
Filter94: In Zig, there are a few release modes: https://zig.guide/build-system/build-modes/ some of them are runtime safe (access to an out of bounds item causes panic) the performant ones are not.

If `off <= data.len` it's an undefined behavior, so it's a valid concern. However it's unfeasible to write every function in a safe way. To keep the high performance functions must rely on the preconditions maintained by the callers. So I think in the guest program's zig codebase we should have some functions unsafe, but check the preconditions at some point

riscv-guests/guest-common/src/ssz.zig
Filter94: FYI, no memory is freed in ZkC env at the moment, but memory leaks are still a concern, at least because of the tests and to be future proof when we might need to implement memory re-allocation.

I added the `errdefer alloc.free(result);`

riscv-guests/lineth-accelerators/src/io.zig
Filter94: This is an unobvious, but documented piece. `_input_start` variable doesn't hold any actually useful value. But the address of this variable is the beginning of the memory region which contains the input. So its type is arbitrary and irrelevant. However I'll change it to be u64, because it contains the length of the input so we can avoid 1 casting

riscv-guests/rollup/src/rollup_ssz.zig
Filter94: I wouldn't go overly defensive here

gauravahuja (approved):

Issue #3868, opened 2026-08-27
[Coordinator] Cannot finalize validium chains: read path is hardcoded to the rollup client and validium V2 is unsupported


coeniebeyers:
## Impact

__environment:__ local dev stack (validium mode); affects any chain run with `data_availability=VALIDIUM`

__user impacted:__ Y. Anyone operating a validium chain with the stock coordinator; it never finalizes.

## Describe the bug

On a validium chain the coordinator crashes at startup and finalization never happens. Two separate gaps cause this:

1. **The finalization read path ignores the DA mode.** The finalization-monitor and `L1BasedLastFinalizedBlockProvider` always construct `Web3JLinethRollupSmartContractClientReadOnly` ([CoordinatorApp.kt#L195](https://github.com/LFDT-Lineth/lineth-monorepo/blob/main/coordinator/app/src/main/kotlin/lineth/coordinator/app/CoordinatorApp.kt#L195), and again at L281, plus `ConflationAppV1`), even when `l1Submission.dataAvailability = VALIDIUM`. That client's `parseContractVersion` only accepts rollup versions, and the validium V2 contract reports `CONTRACT_VERSION() == "2.0"`, so the coordinator dies with `IllegalStateException: Unsupported contract version: 2.0`. The write path already switches by DA mode (`createLineaContractClient`); the read path doesn't.
2. **Validium V2 isn't supported even by the validium client.** `Web3JLineaValidiumSmartContractClientReadOnly.getVersion()` hardcodes `V1` and `LineaValidiumContractVersion` only defines `V1`, while `make deploy-validium` deploys the **V2** contract. So even with a DA-aware read path, `finalizeBlocks` would be encoded as V1 (`FinalizationDataV3`) and revert on-chain.

### Steps to Reproduce
1. `make start-env-with-validium-and-tracing-v2-ci` (note: the `deploy-validium` target also has an env-var naming mismatch vs `deployPlonkVerifierAndValidiumV2.ts`, e.g. `INITIAL_L2_STATE_ROOT_HASH` vs the expected `VALIDIUM_INITIAL_STATE_ROOT_HASH`, which we can file as a separate issue; with those corrected the stack deploys the validium contracts and sets `data_availability=VALIDIUM`)
2. Watch the coordinator logs.

### Actual result
Coordinator FATAL at startup: `IllegalStateException: Unsupported contract version: 2.0` (thrown by `parseContractVersion`). No conflation, no submissions, L1 finalization never advances.

[...]

PR #3910, opened 2026-09-01
feat(coordinator): support finalization on validium chains

coeniebeyers:
This PR implements issue(s) #3868

## Summary

- Lets the coordinator finalize **validium** chains. It previously crashed at startup with `IllegalStateException: Unsupported contract version: 2.0` and never finalized, for two reasons: the finalization-monitor and `L1BasedLastFinalizedBlockProvider` read paths were always wired with `Web3JLinethRollupSmartContractClientReadOnly` regardless of DA mode (its `parseContractVersion` rejects `"2.0"`), and `Web3JLineaValidiumSmartContractClientReadOnly.getVersion()` was hardcoded to `V1` while `deploy-validium` deploys the V2 contract — so once past the startup crash, `finalizeBlocks` would be built as V1 `FinalizationDataV3` and revert on-chain.
- The first commit adds validium V2 support in the client libraries: dynamic `CONTRACT_VERSION()` read (cached as in the rollup client), the generated `ValidiumV2` wrapper, and `finalizeBlocks` encoded as `FinalizationDataV4` (same tuple as rollup V8; parity tests assert the encoding is byte-identical, with distinct values in every V4-specific field). `acceptShnarfData` is unchanged between V1 and V2 (the checked-in V1 and V2 ABIs are identical for it) and reuses the V1 encoding.
- The second commit makes the L1 read path DA-aware: the finalization-monitor client is selected by `l1Submission.dataAvailability`, mirroring `createLineaContractClient`, and the `FinalizedStateDataClientReadOnly` combining interface keeps the monitor wiring cast-free.
- Forced transactions are disabled under validium; `ConflationAppHelper.forcedTransactionsEnabled` is the single source of truth for both the app guard and the DAO selection. The validium V2 contract enforces forced-transaction inclusion at finalization, so a stored forced transaction would halt finalization — `FORCED_TRANSACTION_SENDER_ROLE` should not be granted on validium until coordinator support is added (happy to file that as a follow-up issue).

## Validation

- `:coordinator:app` compiles with tests on current `main`; the `FinalizationFunctionBuildersTest` and `linea-contract-clients` suites pass; spotless is clean.
- Live-validated on a stack built entirely from current `main`, with only the coordinator image replaced by one built from this branch: `make start-env-with-validium-and-tracing-v2-ci` starts cleanly (no "Unsupported contract version" error), conflates, submits, and L1 finalization keeps pace with the L2 head continuously — observed finalized 181 vs head 184 after ~30 minutes, sweeping ~180 blocks including bursts of externally-driven application traffic on the L2.
- The `deploy-validium` target's env vars now match `deployPlonkVerifierAndValidiumV2.ts` on current `main`, so the mismatch mentioned in #3868's reproduction steps has since been fixed upstream and needs no action here.

Internally reviewed (thanks @mkrielza). The commits are structured so this can be split into (1) V2 client support and (2) the DA-aware read path if two smaller PRs are easier to review.

### Checklist

* [ ] If this change is deployed to any environment (including Devnet), E2E test coverage exists or is included in this PR. *(validium e2e is currently workflow_dispatch-only; live dev-stack validation described above)*
* [x] I have informed the team of any breaking changes if there are any. *(None for rollup deployments; validium deployments previously could not finalize at all.)*

### DIFF
--- a/coordinator/app/src/main/kotlin/lineth/coordinator/app/CoordinatorApp.kt
+++ b/coordinator/app/src/main/kotlin/lineth/coordinator/app/CoordinatorApp.kt
@@ -89,7 +89,7 @@ class CoordinatorApp(
         configs.smartContractErrors.size,
         configs.smartContractErrors,
       )
-      configs.l1Submission?.dynamicGasPriceCap?.let { dgc ->
+      configs.l1Submission.dynamicGasPriceCap.let { dgc ->
         log.trace("dynamicGasPriceCap.timeOfDayMultipliers: {}", dgc.timeOfDayMultipliers)
         log.trace(
           "dynamicGasPriceCap.gasPriceCapCalculation.timeOfTheDayMultipliers: {}",
@@ -148,7 +148,7 @@ class CoordinatorApp(
         BlobsPostgresDao(
           config =
           BlobsPostgresDao.Config(
-            maxBlobsToReturn = configs.l1Submission?.blob?.dbMaxBlobsToReturn ?: 50u,
+            maxBlobsToReturn = configs.l1Submission.blob.dbMaxBlobsToReturn,
           ),
           connection = sqlClient,
         ),
@@ -167,15 +167,11 @@ class CoordinatorApp(
       ),
     )
 
-  // The data-availability mode decides which L1 contract client the read path uses, so l1Submission is
-  // now mandatory for every coordinator (the TOML [l1-submission] section always provides it). Fail
-  // loudly naming the missing key rather than silently defaulting to ROLLUP on a validium chain.
-  private val dataAvailability: L1SubmissionConfig.DataAvailability =
-    requireNotNull(configs.l1Submission) {
-      "l1Submission config is required to determine the L1 data-availability mode (rollup vs validium)"
-    }.dataAvailability
+  // The data-availability mode decides which L1 contract client the read path uses. l1Submission is
+  // non-nullable in CoordinatorConfig (the TOML [l1-submission] section always provides it).
+  private val dataAvailability: L1SubmissionConfig.DataAvailability = configs.l1Submission.dataAvailability
 
-  // Forced transactions are unsupported under validium, so the DAO must be disabled there too — not just
+  // Forced transactions are unsupported under validium, so the DAO must be disabled there too, not just
   // the ForcedTransactionsApp. ConflationAppHelper is the single source of truth for that decision.
   private val forcedTransactionsDao = run {
     if (!ConflationAppHelper.forcedTransactionsEnabled(configs.forcedTransactions, dataAvailability)) {
@@ -280,20 +276,18 @@ class CoordinatorApp(
     // decision (shared with the forced-transactions DAO selection above).
     // KNOWN LIMITATION: the validium V2 contract itself enforces forced-transaction inclusion at
     // finalization, so storing a forced transaction on such a deployment halts finalization until
-    // coordinator support is added — do not use FORCED_TRANSACTION_SENDER_ROLE with this coordinator.
-    val ftxEnabled = ConflationAppHelper.forcedTransactionsEnabled(configs.forcedTransactions, dataAvailability)
-    if (!ftxEnabled && configs.forcedTransactions?.disabled == false) {
-      log.warn(
-        "Forced transactions are enabled in config but the coordinator does not support them on validium " +
-          "chains yet; disabling. Storing a forced transaction on a validium V2 contract halts finalization.",
-      )
-    }
-    if (!ftxEnabled) {
+    // coordinator support is added. Do not use FORCED_TRANSACTION_SENDER_ROLE with this coordinator.
+    val ftxConfig = configs.forcedTransactions
+    if (ftxConfig == null || !ConflationAppHelper.forcedTransactionsEnabled(ftxConfig, dataAvailability)) {
+      // If we get here with forced transactions enabled in config, the chain must be running in validium mode.
+      if (ftxConfig?.disabled == false) {
+        log.warn(
+          "Forced transactions are enabled in config but the coordinator does not support them on validium " +
+            "chains yet; disabling. Storing a forced transaction on a validium V2 contract halts finalization.",
+        )
+      }
       ForcedTransactionsApp.createDisabled()
     } else {
-      val ftxConfig = requireNotNull(configs.forcedTransactions) {
-        "forcedTransactions config must be present when forced transactions are enabled"
-      }
       val l1EthClient = createEthApiClient(
         rpcUrl = ftxConfig.l1Endpoint.toString(),
         log = LogManager.getLogger("clients.l1.eth.ftx"),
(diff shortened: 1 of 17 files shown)


coordinator/app/src/main/kotlin/lineth/coordinator/app/CoordinatorApp.kt
```diff
@@ -162,8 +166,19 @@ class CoordinatorApp(
         persistenceRetryer = persistenceRetryer,
       ),
     )
+
+  // The data-availability mode decides which L1 contract client the read path uses, so l1Submission is
+  // now mandatory for every coordinator (the TOML [l1-submission] section always provides it). Fail
+  // loudly naming the missing key rather than silently defaulting to ROLLUP on a validium chain.
+  private val dataAvailability: L1SubmissionConfig.DataAvailability =
```
fluentcrafter: If we make `CoordinatorConfig.l1Submission` required, we must make not nullable instead of this check here
```
val l1Submission: L1SubmissionConfig,
```

coordinator/app/src/main/kotlin/lineth/coordinator/app/CoordinatorApp.kt
```diff
@@ -244,10 +274,26 @@ class CoordinatorApp(
...
-    if (configs.forcedTransactions == null || configs.forcedTransactions.disabled) {
+    // The coordinator does not support forced transactions on validium chains yet: ForcedTransactionsApp
+    // reads the contract through the rollup client, which fails version detection against a validium
+    // contract. ConflationAppHelper.forcedTransactionsEnabled is the single source of truth for this
+    // decision (shared with the forced-transactions DAO selection above).
+    // KNOWN LIMITATION: the validium V2 contract itself enforces forced-transaction inclusion at
+    // finalization, so storing a forced transaction on such a deployment halts finalization until
+    // coordinator support is added — do not use FORCED_TRANSACTION_SENDER_ROLE with this coordinator.
+    val ftxEnabled = ConflationAppHelper.forcedTransactionsEnabled(configs.forcedTransactions, dataAvailability)
+    if (!ftxEnabled && configs.forcedTransactions?.disabled == false) {
```
fluentcrafter: this statement sounds contradicting... :( 
suggestion:
```
if (ftxEnabled && dataAvailability == VALIDIUM)
```

coordinator/app/src/main/kotlin/lineth/coordinator/app/CoordinatorApp.kt
```diff
@@ -244,10 +274,26 @@ class CoordinatorApp(
...
+        "Forced transactions are enabled in config but the coordinator does not support them on validium " +
+          "chains yet; disabling. Storing a forced transaction on a validium V2 contract halts finalization.",
+      )
+    }
+    if (!ftxEnabled) {
       ForcedTransactionsApp.createDisabled()
     } else {
-      val ftxConfig = configs.forcedTransactions
+      val ftxConfig = requireNotNull(configs.forcedTransactions) {
+        "forcedTransactions config must be present when forced transactions are enabled"
```
fluentcrafter: This an impossible code branch. `ftxConfig = true` means that `configs.forcedTransactions != null`. We shall rearange the code in a better way.

fluentcrafter (commented):
Thanks for the contribution. Overall LGTM, just left some suggestions

coordinator/app/src/main/kotlin/lineth/coordinator/app/CoordinatorApp.kt
coeniebeyers: Making l1Submission non-nullable simplifies things, excellent suggestion

coordinator/app/src/main/kotlin/lineth/coordinator/app/CoordinatorApp.kt
coeniebeyers: fixed in 3, let me know what you think

coordinator/app/src/main/kotlin/lineth/coordinator/app/CoordinatorApp.kt
coeniebeyers: I've restructured to make it easier to read and get rid of the branch that could never fail

coordinator/app/src/main/kotlin/lineth/coordinator/app/CoordinatorApp.kt
```diff
@@ -244,10 +267,18 @@ class CoordinatorApp(
...
+    // Forced transactions are rollup-only for now; see ConflationAppHelper.forcedTransactionsEnabled.
+    val ftxConfig = configs.forcedTransactions
+    if (ftxConfig == null || !ConflationAppHelper.forcedTransactionsEnabled(ftxConfig, dataAvailability)) {
+      // If we get here with forced transactions enabled in config, the chain must be running in validium mode.
+      if (ftxConfig?.disabled == false) {
+        log.warn(
+          "Forced transactions are enabled in config but the coordinator does not support them on validium " +
+            "chains yet; disabling. Storing a forced transaction on a validium V2 contract halts finalization.",
+        )
+      }
```
fluentcrafter, line 279: IMO, this branching is still overly complicated to see the true intent.  
Why not just like below

```suggestion
    if (ftxConfig.isEnabled() && dataAvailability == L1SubmissionConfig.DataAvailability.VALIDIUM) {
      // If we get here with forced transactions enabled in config, the chain must be running in validium mode.
      log.warn(
          "Forced transactions are enabled in config but the coordinator does not support them on validium " +
            "chains yet; disabling. Storing a forced transaction on a validium V2 contract halts finalization.",
        )
        ForcedTransactionsApp.createDisabled()
      }
```

fluentcrafter (dismissed):

coordinator/app/src/main/kotlin/lineth/coordinator/app/CoordinatorApp.kt
coeniebeyers: Agreed it is still hard to read.

I don't want to push another update and potentially dismiss your approval, but I tried restructuring closer to your suggestion and for us to use ftxConfig as a non-nullable object (in the last branch, where we construct the forced transactions app) we need an explicit "== null" or "!= null" check before it in the if..else statements, doing a .isEnabled() isn't good enough for the compiler.  So your suggestion works, but we would need another 
```
} else if (ftxConfig == null || ftxConfig.disabled) {
  ForcedTransactionsApp.createDisabled()
```
before the else that build the forced transactions app

fluentcrafter: >we need an explicit "== null" or "!= null" check before it in the if..else statements, doing a .isEnabled() isn't good enough for the compiler.

`.isEnabled()` is a extension function that works with nullable fields, see definition [here](https://github.com/LFDT-Lineth/lineth-monorepo/blob/main/coordinator/app/src/main/kotlin/lineth/coordinator/config/v2/FeatureToggle.kt#L9)

I just tested in CoordinatorApp this snipped and `./gradlew spotlessApply compileAll` works fine 
```
init {
      if (configs.forcedTransactions.isEnabled()) {
        log.info("Forced transactions feature enabled")
      }
  }
```

fluentcrafter (approved):

gauravahuja (approved):

PR #3916, opened 2026-09-02
feat(arithmetization): zkc reference guest output check

amkCha:
This PR implements issue(s) #

### Checklist

  PR.


riscv-guests/l2-execution/test/zkc_reference_runner.zig
```diff
@@ -1,29 +1,41 @@
...
-    \\  -x               stop on the first disagreeing block
-    \\  --report-only    print the summary but always exit 0
+    \\Corpus mode:
+    \\  zkc-reference-runner --fixtures DIR --install-prefix DIR --makefile PATH [options]
+    \\Smoke mode:
+    \\  zkc-reference-runner --input FILE.ssz --expect-guest-output HEX|FILE --install-prefix DIR --makefile PATH [options]
+    \\
+    \\  --fixtures DIR             the blockchain_tests/ JSON tree (corpus mode)
+    \\  --input FILE               already-extended SSZ input (smoke mode; skips wrap)
+    \\  --expect-guest-output X    golden guest_output hex (no 0x) or path to a file containing it
```
Filter94: Can we avoid this option and the file in favor of assertions on the output hash's preimage? Is precompile to log arbitrary data off trace ready yet?

riscv-guests/l2-execution/test/zkc_reference_runner.zig
```diff
@@ -117,37 +142,134 @@ fn runBlockUnderZkc(
...
+    const json_path = try std.fmt.allocPrint(alloc, "{s}.json", .{extended_path});
+    const flags = effectiveZkcFlags();
     const in_arg = try std.fmt.allocPrint(alloc, "IN_BYTES=@{s}", .{extended_path});
     const elf_arg = try std.fmt.allocPrint(alloc, "BIN_EXT={s}", .{elf});
     const json_arg = try std.fmt.allocPrint(alloc, "JSON_EXT={s}", .{json_path});
-    const flags_arg = try std.fmt.allocPrint(alloc, "ZKC_EXEC_FLAGS={s}", .{zkc_flags});
     const makefile_arg = try std.fmt.allocPrint(alloc, "-f{s}", .{makefile});
 
-    const res = try std.process.run(alloc, io, .{
-        .argv = &.{ "make", "-s", makefile_arg, "elf-exec", elf_arg, in_arg, json_arg, flags_arg },
```
Filter94, line 131: I didn't know `elf-trace` exists. What is the reason to support `elf-exec` at all?

riscv-guests/l2-execution/test/zkc_reference_runner.zig
```diff
@@ -117,37 +142,134 @@ fn runBlockUnderZkc(
...
 
+fn effectiveZkcFlags() []const u8 {
+    if (zkc_flags) |f| return f;
+    if (std.mem.eql(u8, zkc_target, "elf-trace")) return "--stats";
+    return "--fast";
+}
+
+/// Find `guest_output = 0x<hex>` in zkc stdout/stderr; return lowercase hex without the `0x`.
+fn parseGuestOutputHex(alloc: std.mem.Allocator, text: []const u8) ?[]const u8 {
+    const start = std.mem.indexOf(u8, text, guest_output_prefix) orelse return null;
```
Filter94, line 193: Why doesn't ZkC structure its output? This reliance on `guest_output_prefix` seems fragile. How prover host will parse ZkC's output?

riscv-guests/l2-execution/test/zkc_reference_runner.zig
amkCha: so for tracing we will have to do 2 phases : run a "fast" mode (for now it's exec) to then be able to // the trace generation per say (trace mode). We might drop exec altogether and work with flags

riscv-guests/l2-execution/test/zkc_reference_runner.zig
amkCha: We should - we haven't discussed it with the prover yet. I'll add it to the questions during the crypto meeting, let me add a TODO for now - when we have this structured, we can update the pipeline

riscv-guests/l2-execution/test/zkc_reference_runner.zig
amkCha: > Can we avoid this option and the file in favor of assertions on the output hash's preimage?
= > Yes, much better
> Is precompile to log arbitrary data off trace ready yet?
= > Didn't hear of it, was it supposed to be delivered soon ?

riscv-guests/l2-execution/test/zkc_reference_runner.zig
Filter94: Ok. Let's keep the support for both and use exec for now until arithmetization is faster. IMO, the way things are now it's prohibitively slow, WDYT?

riscv-guests/l2-execution/test/zkc_reference_runner.zig
Filter94: Please leave a TODO for this at least. I would try to starnardize the output and let it be handled by some imported lib. Json, for example which is a part of the std lib

riscv-guests/l2-execution/test/zkc_reference_runner.zig
Filter94: @OlivierBBB Do you have more info on this?

riscv-guests/l2-execution/test/zkc_reference_runner.zig
```diff
@@ -211,55 +215,59 @@ fn parseGuestOutputHex(alloc: std.mem.Allocator, text: []const u8) ?[]const u8 {
...
-    return std.mem.trim(u8, raw, " \n\r\t");
+fn decodeHex(alloc: std.mem.Allocator, hex: []const u8) ![]u8 {
+    if (hex.len % 2 != 0) return error.InvalidHex;
+    const out = try alloc.alloc(u8, hex.len / 2);
+    _ = try std.fmt.hexToBytes(out, hex);
+    return out;
+}
+
+/// Host path: decode extended input → runL2Execution → encodeOutput (schema ‖ keccak256(PI)).
+fn nativeExpectedGuestOutput(alloc: std.mem.Allocator, raw_input: []const u8) ![l2_execution_ssz.OUTPUT_SIZE]u8 {
```
Filter94, line 218: Yes, host execution is a good option too, given that tracing will take much longer than a host execution. To speed things up, we can do this in parallel to the ZkC tracing 👍

riscv-guests/l2-execution/test/testdata/stateless_input.expected_guest_output
Filter94, line 1: I believe it's not used anymore

riscv-guests/l2-execution/test/zkc_reference_runner.zig
```diff
@@ -1,29 +1,45 @@
...
-    \\  --limit N        stop after N blocks (dev speed)
-    \\  -x               stop on the first disagreeing block
-    \\  --report-only    print the summary but always exit 0
+    \\Corpus mode:
+    \\  zkc-reference-runner --fixtures DIR --install-prefix DIR --makefile PATH [options]
+    \\Smoke mode:
+    \\  zkc-reference-runner --input FILE.ssz --install-prefix DIR --makefile PATH [options]
+    \\
+    \\  --fixtures DIR             the blockchain_tests/ JSON tree (corpus mode)
+    \\  --input FILE               already-extended SSZ input (smoke mode; skips wrap)
```
Filter94: Do we really need the `--input` here? IMO we shouldn't bloat the functionality and use `match` instead or generalize this to be a glob pattern

riscv-guests/l2-execution/test/zkc_reference_runner.zig
```diff
@@ -197,6 +334,15 @@ pub fn main(init: std.process.Init) !void {
     tmp_dir = try makeTempDir(io, gpa);
     defer cleanupTempDir(io, gpa, tmp_dir);
 
+    // Smoke mode: single extended input; expected guest_output from native encodeOutput.
+    if (smoke_input) |input| {
```
Filter94: This really seems strange to me. If the input is a single file, not a directory, then we're performing a smoke test and verifying the output. Can we unify the execution path instead? Or we can make it a separate flag, whether we want to verify the output or not

riscv-guests/l2-execution/test/zkc_reference_runner.zig
amkCha: Tracing is slow yes - it's running reasonably for the test vector `stateless_input` so I've added that on the CI for instance. But for a small Mainnet block, it's taking 10min on a big xxl instance

riscv-guests/l2-execution/test/testdata/stateless_input.expected_guest_output
amkCha: Forgot to commit the deletion ! it's removed

riscv-guests/l2-execution/test/zkc_reference_runner.zig
amkCha: I had a discussion with @OlivierBBB this morning about it : the precompile is a priori ready, it's the same method, via the custom upcode we plugged the other day (write output accelerator), that will be called as the one logging `guest_output`. Now when I run the guest program, be it on the test block or a Mainnet one, I don't see the PI logging like `txFromHash` which I should no ? Are we missing some wiring ?

riscv-guests/l2-execution/test/zkc_reference_runner.zig
amkCha: Apparently it was discussed with the prover : the output should just be a hash of all the inputs (32 bytes). This will be done on our side by @OlivierBBB. So in the example above, it will be the hash(prefix+hash(inputs)). Just a question, the prefix comes from the l2-execution, and it's a standard you've agreed on with the prover ? To know if we keep it. I don't see it here, maybe I'm looking in the wrong place https://github.com/LFDT-Lineth/lineth-monorepo/blob/71008db56bdd83a4e0008fb9e5e4a4a4b3a81eb0/rollup_spec/Readme.md

riscv-guests/l2-execution/test/zkc_reference_runner.zig
amkCha: // is now done

riscv-guests/l2-execution/test/zkc_reference_runner.zig
Filter94: Output will be just a hash, but we also had a discussion about a way to output some additional info into a different output stream. That way we don't have to run the guest program on the host machine, same as you do in this PR. We could modify the Guest program to output the preimage of a hash and use this data in the prover or any where else it's propagated to

riscv-guests/l2-execution/test/zkc_reference_runner.zig
Filter94: The guest program doesn't know about this custom precompile and I don't have a spec to implement it. Do you have any pointers? Maybe an .h file with the interface we can use or the implementation calling a custom precompile?

riscv-guests/l2-execution/test/spec_runner.zig
```diff
@@ -125,6 +126,43 @@ pub fn run(comptime Adapter: type, init: std.process.Init, opts: Options) !Stats
...
+        return globMatch(rel_path, pattern) or globMatch(std.fs.path.basename(rel_path), pattern);
+    }
+    return std.mem.indexOf(u8, rel_path, pattern) != null;
+}
+
+fn isGlobPattern(pattern: []const u8) bool {
+    return std.mem.indexOfAny(u8, pattern, "*?") != null;
+}
+
+fn globMatch(s: []const u8, pattern: []const u8) bool {
```
Filter94, line 143: Is there an implementation we can reuse instead of the custom implementation?

riscv-guests/l2-execution/test/zkc_reference_runner.zig
```diff
@@ -249,53 +239,111 @@ fn nativeExpectedGuestOutputThread(raw_input: []const u8, out: *NativeResult) vo
...
+
+        stats.files += 1;
+        stats.blocks += 1;
+        const failed_before = stats.failed;
+        if (try runOneExtendedSsz(init, gpa, full)) stats.passed += 1 else stats.failed += 1;
+        if (opts.stop_on_fail and stats.failed > failed_before) break;
+    }
+}
+
+fn runOneExtendedSsz(init: std.process.Init, alloc: std.mem.Allocator, input_path: []const u8) !bool {
```
Filter94, line 299: Can we instead move this logic to [runAndCheck](https://github.com/LFDT-Lineth/lineth-monorepo/pull/3906/changes#diff-fc0fa9da910156b5f9211be0d7ff471ba4d3a20f9233c8ce878949351ad1777eR72) in `ZkcAdapter`, rather than introducing a new walker and matcher?

`spec_runner` is meant to be generic enough to support different ways of running the fixture tests

riscv-guests/l2-execution/test/spec_runner.zig
Filter94: Also, it seems to be an incomplete implementation of the glob pattern. We really have to use the existing implementations for things like this

riscv-guests/l2-execution/test/spec_runner.zig
```diff
@@ -43,8 +43,9 @@ pub const Options = struct {
     single_file: ?[]const u8 = null,
```
Filter94, line 43: I think this option was added by mistake by me. I don't see what purpose it serves. Same goes for `fixtures_dir` if the glob pattern matching is implemented properly

PR #3915, opened 2026-09-02
feat(prover-ray): export grandproduct.RowLimitAction for verifier-ray

YaoJGalteland:
## Summary
- Exports `grandproduct.RowLimitAction` (was `rowLimitAction`, unexported) by capitalizing its type and fields (`Query`, `Limit`), so out-of-package consumers can read the guarded permutation query and its row-limit — mirroring how `lookuptologderivsum.RowLimitVerifierAction` is already consumed externally.
- Updates the single call site in `compilePermutations` (permutation.go) and the existing internal test (rowlimit_internal_test.go) to the new exported names. No behavioral change — this is a pure rename/visibility change.
- Unblocks verifier-ray's codegen, which needs to enumerate `RowLimitAction` instances registered on a `wiop.System` to generate the Zig `rowlimit` sub-verifier's per-permutation checks (see the follow-up PR on `fix/grandproduct-rowlimit-allowlist`).

## Test plan
- [x] `go test ./wiop/compilers/grandproduct/...` passes (existing `RowLimitAction` prover/verifier tests updated for the rename)
- [x] `go build ./...` passes

---

> [!NOTE]
> **Low Risk**
> Rename and export-only change with no logic changes; tests still cover registration and row-limit enforcement.
> 
> **Overview**
> Exports **`grandproduct.RowLimitAction`** (previously unexported `rowLimitAction`) and its **`Query`** / **`Limit`** fields so verifier-ray codegen can inspect registered row-limit checks on a `wiop.System`, matching the existing pattern for lookup **`RowLimitVerifierAction`**.
> 
> **`compilePermutations`** and internal row-limit tests are updated to the new names only; prover panic and verifier error behavior are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2e6d9ca8705731f7672c3bcf26e74b2607d42f50. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3917, opened 2026-09-02
feat(verifier-ray): extend rowlimit codegen to grandproduct permutations

YaoJGalteland:
## Summary
- Extends `BuildRowLimitSystem` (verifier-ray/codegen) to also collect `grandproduct.RowLimitAction`, not just `lookuptologderivsum.RowLimitVerifierAction`, so the generated Zig `rowlimit` sub-verifier enforces `MaxPermutationRows` on grandproduct-compiled permutation queries' A/B sides, alongside the existing lookup row-limit checks.
- Fixes a codegen bug where the `.grandproduct = system_%d_grandproduct` field was missing from the generated `verifier.Systems{...}` literal — the const was already emitted but never wired into the struct.
- Adds a `Permutation/GrandProductPermutation` end-to-end fixture (a real `wiop.NewPermutation` reduced by `grandproduct.Compile` directly) to `testdata/generate/main.go`, exercising `RowLimitAction` for the first time via `AssertAllVerifierActionsHandled` and the full `verifier.verify` path — regenerated `vanishing.zig`/`verify.zig` accordingly.
- Bumps the `prover-ray` dependency pseudo-version in `verifier-ray/codegen` and `verifier-ray/testdata/generate` to pick up the exported `RowLimitAction` (from the companion `feat/grandproduct-rowlimit-export` PR).

## Test plan
- [x] `go test ./...` in `verifier-ray/codegen` passes, including new `TestBuildRowLimitSystemExtractsPermutationCheck`
- [x] `make test-zig` in `verifier-ray` passes (regenerated fixtures + Zig tests, including the new permutation row-limit check)
- [ ] Rebase onto `main` and bump the `prover-ray` pseudo-version once the export PR merges

Depends on #3915 being merged first (this branch's go.mod currently pins prover-ray@d4bb7a17e55e by pseudo-version; bump it to the merge commit once #3915 lands).

---

> [!NOTE]
> **Medium Risk**
> Changes proof-verification codegen and row-limit enforcement for permutations; mistakes could accept oversized traces or mis-wire sub-verifiers, though coverage is strengthened with new fixtures and pinned error tests.
> 
> **Overview**
> Extends verifier-ray **rowlimit codegen** so `BuildRowLimitSystem` collects **`grandproduct.RowLimitAction`** in addition to lookup row-limit actions, mapping permutation A/B table modules into the same `RowLimitCheck` shape and emitting each action’s **`a.Limit`** (not a hard-coded `MaxPermutationRows`). **`grandproduct.RowLimitAction`** is added to the Zig verifier action allowlist in `actions.go`.
> 
> Fixes generated **`verifier.Systems`** wiring so **`.grandproduct`** is included alongside vanishing, logderiv, rowlimit, and PCS (the grandproduct system was emitted but not hooked up).
> 
> Adds **Go tests** for permutation row-limit extraction and custom limit preservation, bumps **`prover-ray`** and CI **`ZKC_VERSION`**, and expands the fixture generator with a **`GrandProductPermutation`** end-to-end case plus **`invalidVerify`** proofs that are PCS/vanishing-valid but fail in a named sub-verifier. Zig tests assert the exact error (**`ResultMismatch`**) to prove **grandproduct** is exercised on the full verify path; regenerated **`vanishing.zig`** / **`verify.zig`** reflect the new case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9bf67a40fec9a37e924a37f312587cecd441b5a7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


verifier-ray/codegen/rowlimit.go
```diff
@@ -60,27 +79,54 @@ func BuildRowLimitSystem(sys *wiop.System) (RowLimitSystem, error) {
...
+				}
+				includingsModules, err := moduleSizes(tablesModules(a.Query.B))
+				if err != nil {
+					return RowLimitSystem{}, err
+				}
+
+				out.Checks = append(out.Checks, RowLimitCheck{
+					IncludedModules:   includedModules,
+					IncludingsModules: includingsModules,
+					Limit:             wiop.MaxPermutationRows,
```
ivokub: **[P2] Preserve the action-specific row limit**

`grandproduct.RowLimitAction.Check` validates the query against `a.Limit`, and the action exports that field specifically so consumers such as verifier-ray codegen can reproduce its semantics. Hardcoding `wiop.MaxPermutationRows` here silently changes the generated verifier whenever an action carries a different, particularly lower, effective limit.

The current grand-product compiler happens to set `a.Limit == wiop.MaxPermutationRows`, so its present standard path agrees, but a manually registered action or future compiler change would be rejected by the Go verifier while the Zig verifier could accept it. Could this use `Limit: a.Limit` and add a codegen regression with a deliberately smaller action limit?

verifier-ray/testdata/generate/main.go
```diff
@@ -761,9 +795,114 @@ func buildCompiledFixtureCases() ([]fixtureCase, []codegen.CompiledSystem, error
...
+		const resultRoundIdx, zEndpointModAIdx = 2, 1
+		if len(honestFixture.view.rounds) <= resultRoundIdx ||
+			len(honestFixture.view.rounds[resultRoundIdx].cells) <= zEndpointModAIdx {
+			return nil, nil, fmt.Errorf("Permutation/GrandProductPermutation: result round has no Z-endpoint cell at index %d", zEndpointModAIdx)
+		}
+
+		invalidFixture := honestFixture
+		invalidRounds := append([]runtimeTraceRound(nil), honestFixture.view.rounds...)
+		invalidCells := append([]runtimeTraceCell(nil), honestFixture.view.rounds[resultRoundIdx].cells...)
+		invalidCells[zEndpointModAIdx] = flippedExtTraceCell(invalidCells[zEndpointModAIdx])
```
ivokub, line 869: **[P2] Add a negative fixture that reaches the grand-product checker**

Nice to have, but not important to address right now. Only posting for completeness.

This mutation retains the original PCS opening after changing a transcript cell. Since `verifier.verify` runs PCS before `grandproduct.verify`, the proof can be rejected by PCS without exercising the newly wired `.grandproduct` system.

I confirmed the coverage gap by removing every generated `.grandproduct = system_N_grandproduct` assignment in a temporary copy: `zig build test` still passed. A stronger regression would generate a self-consistent proof for mismatched A/B witnesses; that proof should remain PCS- and vanishing-valid and fail specifically in `grandproduct.verify`. This may require separate full-verifier and standalone-vanishing invalid fixtures so the vanishing test does not require the same input to fail `QuotientIdentityMismatch`.

ivokub (dismissed):
In principle looks good. There is only a small mismatch imo with the corresponding verifier action in Go reference. Right now it works, but later on could drift away.

Approval for now and I'll re-approve later when fixed.

ivokub (approved):
LGTM! Thanks for addressing the comments.

PR #3925, opened 2026-09-03
fix(ci): retarget doc.linea publishing to Consensys-Incorporated

Othryades:
## Problem

The `doc.linea` repository was transferred to a new GitHub organisation. The config-docs publish jobs still request a GitHub App installation on the old `Consensys/doc.linea` path, which no longer resolves. Token minting fails before checkout:

```
GET /repos/Consensys/doc.linea/installation -> 404 Not Found
```

This surfaced on the `Besu plugin options docs` workflow on a push to `main`. The same hardcoded owner is present in the Prover and Postman workflows, and in the shared publish action that Coordinator and Maru consume, so those will fail on their next publish for the same reason.

## Change

Retargets owner/repository from `Consensys` to `Consensys-Incorporated`:

- `.github/workflows/besu-plugin-options-docs.yml` — `owner` + `repository`
- `.github/workflows/prover-config-docs.yml` — `owner` + `repository`
- `.github/workflows/postman-config-docs.yml` — `owner` + `repository`
- `.github/actions/publish-config-docs/action.yml` — `docs-owner` default (used by Coordinator and Maru)

Comments and `workflow_dispatch` descriptions updated to match. No change to secret names, permissions, triggers, publish gating, or the namespace boundary checks.

## Verification

Publish jobs only run on push to `main` / release tags, so this cannot be exercised from a PR. After merge, re-run `Besu plugin options docs` on `main` to confirm the token step succeeds.

If it still fails on `Consensys-Incorporated/doc.linea/installation`, the App installation did not follow the repository transfer and needs to be granted on the new organisation — that is an org-settings action, separate from this PR.

### DIFF
--- a/.github/actions/publish-config-docs/action.yml
+++ b/.github/actions/publish-config-docs/action.yml
@@ -2,7 +2,7 @@ name: 'Publish config docs to doc.linea'
 description: >-
   Reusable publishing for any component using the shared `config-docs` engine.
   Downloads the artifact produced by `validate-config-docs`, checks out
-  `Consensys/doc.linea` with a GitHub App token, atomically swaps the component's
+  `Consensys-Incorporated/doc.linea` with a GitHub App token, atomically swaps the component's
   generated namespace into place, refuses any change that escapes the
   namespace, and opens a pull request against doc.linea's main branch.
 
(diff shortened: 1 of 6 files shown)


nadeemb53 (approved):

eloi010 (approved):
LGTM

PR #3931, opened 2026-09-04
feat(riscv-guest): Adding fallback precompile implementations

Filter94:
This PR implements issue(s) #

### Checklist

  PR.


riscv-guests/l2-execution/src/zkvm_io.zig
```diff
@@ -0,0 +1,17 @@
...
+
+/// Linker-defined 1 GiB input region.
+const IN_REGION_SIZE: usize = 0x40000000;
+
+pub fn read_input(buf_ptr: *[*]const u8, buf_size: *usize) void {
+    const base: [*]const u8 = @ptrCast(&_in_start);
+    const payload_len = std.mem.readInt(u64, base[0..8], .little);
+    if (payload_len > IN_REGION_SIZE - 8) @panic("input payload_len exceeds IN region");
+    buf_ptr.* = base + 8;
+    buf_size.* = @intCast(payload_len);
```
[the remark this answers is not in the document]
Filter94: I'll address it here https://github.com/LFDT-Lineth/lineth-monorepo/pull/3894

ivokub:
### Additional Review Findings & Build Fixes

Here are a few follow-up observations and improvements from testing the build pipeline locally:

---

#### 1. `riscv-guests/Makefile`: `install-constantine-deps` fails on Ubuntu due to unversioned `llvm-ar` / `llvm-nm`

In `riscv-guests/Makefile`:
```makefile
command -v nim >/dev/null; command -v nimble >/dev/null; command -v llvm-ar >/dev/null;
```
On Ubuntu 24.04 (and Debian), `apt-get install llvm` installs versioned binaries (`/usr/bin/llvm-ar-21` and `/usr/bin/llvm-nm-21`), but does not create unversioned `/usr/bin/llvm-ar` or `/usr/bin/llvm-nm` symlinks. As a result, the check `command -v llvm-ar` fails immediately after package installation.

**Suggestion:** Auto-create symlinks in `$HOME/.local/bin/` if unversioned binaries are missing:
```bash
if ! command -v llvm-ar >/dev/null; then
    latest_ar=$(find /usr/bin -name "llvm-ar-*" 2>/dev/null | sort -V | tail -n1)
    [ -n "$latest_ar" ] && ln -sf "$latest_ar" "$HOME/.local/bin/llvm-ar"
fi
if ! command -v llvm-nm >/dev/null; then
    latest_nm=$(find /usr/bin -name "llvm-nm-*" 2>/dev/null | sort -V | tail -n1)
    [ -n "$latest_nm" ] && ln -sf "$latest_nm" "$HOME/.local/bin/llvm-nm"
fi
```

---

#### 2. `riscv-guests/guest-crypto-ctt/build.zig`: Forward `LLVM_AR` to `nimble` task

In `guest-crypto-ctt/build.zig`:
```zig
const task = b.addSystemCommand(&.{ tc.nimble, "-y", "make_lib_riscv64_freestanding" });
```
`resolveToolchain` resolves `tc.llvm_ar`, but `task` does not set it in the child process environment. Because `constantine.nimble` checks `if existsEnv"LLVM_AR"`, forwarding the resolved path ensures `nimble` respects custom/versioned tools specified by the user:
```zig
task.setEnvironmentVariable("LLVM_AR", tc.llvm_ar);
```

---

#### 3. `riscv-guests/l2-execution/build.zig`: Don't tie host CLI binaries to default `install` step

In `l2-execution/build.zig`:
```zig
b.installArtifact(l2_execution_runner_exe);
b.installArtifact(l2_execution_wrap_exe);
```
Registering `l2_execution_runner_exe` and `l2_execution_wrap_exe` on the default `install` step means running `make compile` (or `zig build` to generate the freestanding RISC-V ELF) unconditionally attempts to build those host tools as well. Because the host tools link `blst` and `mcl`, anyone wanting to compile only the guest ELF is blocked if they don't have host C crypto libraries installed.

**Suggestion:** Register host CLI tools under an explicit step (e.g. `b.step("tools", ...)` or `run`) rather than `b.installArtifact(...)` by default.

---

#### 4. `riscv-guests/guest-crypto-ctt/build.zig.zon`: Update Constantine pin for BN254 fix

`build.zig.zon` currently pins commit `fe7d318e7e4bfb1a605105a77ed0e0dfe8566296`. That commit predates the BN254 pairing check infinity bypass fix (`e25ee9f`). Once the pairing fix branch is merged into the Constantine fork, `build.zig.zon` should be bumped to the new commit hash and package fingerprint.

riscv-guests/l2-execution/src/zkvm_io.zig
```diff
@@ -0,0 +1,17 @@
...
+//! followed by the payload.
+
+const std = @import("std");
+
+extern var _in_start: u8;
+
+/// Linker-defined 1 GiB input region.
+const IN_REGION_SIZE: usize = 0x40000000;
+
+pub fn read_input(buf_ptr: *[*]const u8, buf_size: *usize) void {
```
ivokub: Why reimplement riscv-guests/lineth-accelerators/src/io.zig?

riscv-guests/l2-execution/src/zkvm_provide.zig
```diff
@@ -91,66 +67,108 @@ fn ripemd160(data: [*]const u8, len: usize, output: *[32]u8) callconv(.c) i32 {
...
     return OK;
 }
 fn secp256r1_verify(msg: *const [32]u8, sig: *const [64]u8, pubkey: *const [64]u8, verified: *bool) callconv(.c) i32 {
-    zesu_zkvm_stdlibs.secp256r1_verify(msg, sig, pubkey, verified);
+    verified.* = verifyP256(msg, sig, pubkey) catch false;
     return OK;
 }
+
+/// P-256 ECDSA verify over a pre-hashed message, compact big-endian r‖s signature, and raw x‖y key.
+fn verifyP256(msg: *const [32]u8, sig: *const [64]u8, pubkey: *const [64]u8) !bool {
```
ivokub, line 82: Is it zesu implementation copied as is? Why cannot we do verifyPrehashed from std?

riscv-guests/l2-execution/src/block_rlp_size.zig
```diff
@@ -0,0 +1,90 @@
...
+fn withdrawalsSize(withdrawals: []const input.Withdrawal) usize {
+    var payload: usize = 0;
+    for (withdrawals) |wd| {
+        payload += listSize(uintSize(wd.index) + uintSize(wd.validator_index) + ADDRESS_SIZE + uintSize(wd.amount));
+    }
+    return listSize(payload);
+}
+
+/// Returns null when decoded transactions have no matching raw transaction bytes.
+pub fn compute(ep: *const input.ExecutionPayload, spec: primitives.SpecId) ?u64 {
```
ivokub: This is hand-rolled RLP length arithmetic feeding `env.block_rlp_size` for the EIP-7934 block-size limit, and as far as I can tell nothing tests it directly — no unit test in the package, and the host reference suite would only catch an error if a corpus block happened to land within a few bytes of the limit. Since the guest computes this without re-serializing, a wrong constant or edge case here silently changes which blocks pass validation in the zkVM.

A few specific spots I'd want pinned down:

- `headerSize`: `uintSize(0)` for difficulty and the pre-merge nonce slot, and `NONCE_SIZE = 9` — worth a comment or test showing these match the actual RLP encoding of the reconstructed header (difficulty is always 0 post-merge, but the encoding of an empty-vs-zero value differs).
- `compute`: the ommers list is hardcoded as `listSize(0)` (= 1 byte, `0xc0`) — correct only because this guest rejects withdrawals-bearing and pre-merge inputs by policy; a short comment saying so would prevent a future copy-paste into a context where ommers exist.
- `transactionsSize`: the `raw[0] >= 0xc0` legacy-vs-typed discrimination depends on the raw bytes being the full signed RLP — a differential check would catch any mismatch with what `decodeTxsFromInput` accepts.

Could you add a differential unit test that RLP-encodes a few reconstructed blocks (e.g. via zesu's own RLP encoder over decoded fixtures, across fork boundaries — London/Shanghai/Cancun/Prague/Amsterdam optional-field activation is exactly where the size arithmetic branches) and asserts `compute` returns the exact encoded length? Empty and single-byte-edge (`0x7f`/`0x80`, 55/56-byte payload) cases would cover the prefix-boundary paths cheaply.

riscv-guests/l2-execution/README.md
```diff
@@ -19,7 +19,7 @@ Run from the parent directory:
...
+`make -C l2-execution compile` writes the guest as a **statically-linked rv64im ELF** to `riscv-guests/l2-execution/zig-out/bin/evm_execution_guest` — the [zkvm-standards](https://github.com/eth-act/zkvm-standards/blob/main/standards/riscv-target/target.md) artifact ("Object Format: ELF, statically linked"), linked via `build_common`'s shared `installGuestElf`. The ZKC interpreter loads the guest (via ELF→JSON); `make -C l2-execution exec` builds it and runs it there — see the [parent README](../README.md#zkc-interpreter-integration). `make test` runs the native Zig test, which requires the native crypto libraries documented in the [parent README](../README.md#native-test-dependencies).
```
ivokub: Stale after this PR: `make test` no longer links blst/mcl/secp256k1/OpenSSL — the native tests use Zesu's `extern` backend with the in-tree Constantine archive — and the referenced parent-README section `../README.md#native-test-dependencies` was deleted in the same commit, so this link now points at a section that no longer exists. Suggest:

```
`make test` runs the native Zig tests through the same Constantine-backed crypto providers as the guest (see [Crypto Backends](../README.md#crypto-backends)).
```

ivokub (commented):
Left a few comments. In principle I think would be fine as is, see if you want to address. Though we still need to update CTT dependency after PR3 is merged there.

And I don't understand the timeout in CI (for guest workflow). Is it machine stalling or performance issue? The fixtures error is inconvenient, we should store these fixtures in our caches somehow imo.

ivokub (approved):

riscv-guests/l2-execution/src/block_rlp_size.zig
Filter94: We shouldn't have this logic here in the first place. I created a PR to export in on Zesu side https://github.com/Consensys-Incorporated/zesu/pull/145

riscv-guests/l2-execution/src/zkvm_io.zig
Filter94: I think I got confused with merge conflict resolution here. This file shouldn't exist of course

Filter94:
> And I don't understand the timeout in CI (for guest workflow). Is it machine stalling or performance issue?

This PR was created a while back, I don't know if all the timeouts still make sense, but in https://github.com/LFDT-Lineth/lineth-monorepo/pull/3988 I removed the timeouts I introduced to debug the issue I had with Constantine

> The fixtures error is inconvenient, we should store these fixtures in our caches somehow imo.

Fixtures error is fixed in that other PR I have open. Instead of running many Zig builds in parallel, they are now structured into a Workflow DAG and also I removed some of the unnecessary build dependencies so less jobs trigger fixture download

riscv-guests/l2-execution/src/zkvm_provide.zig
Filter94: Std's `verifyPrehashed` rejects zero scalar while `EIP-7951` allows such signatures. WIthout this function reference test is failing. 
```
FAIL tests/osaka/eip7951_p256verify_precompiles/test_p256verify.py::test_valid[fork_Amsterdam-blockchain_test_from_state_test---hash_0][0] disagree: fixture=valid machine=invalid (EngineReject)
FAIL tests/osaka/eip7951_p256verify_precompiles/test_p256verify.py::test_valid[fork_Amsterdam-blockchain_test_from_state_test---hash_N][0] disagree: fixture=valid machine=invalid (EngineReject)
progress: for_amsterdam/osaka/eip7951_p256verify_precompiles/p256verify/valid.json files=1 cases=17 passed=15 failed=2 skipped=0
```

riscv-guests/l2-execution/src/block_rlp_size.zig
Filter94: Zesu PR was merged, I updated the dependecy

riscv-guests/l2-execution/src/zkvm_provide.zig
ivokub: Though when looking at EIP-7951, the error cases are:
* Invalid signature component bounds (r or s not in range (0, n))
* Invalid public key (point at infinity or not on curve)

This would suggest that verifyPrephashed was correct? (i.e. it rejected r or s 0, and also public key at infinity). I'll need to take a few moments to understand EIP and the edge cases.

Filter94: But the input hash is h, not r:
```
P256VERIFY call expects 160 bytes as input that is interpreted as byte concatenation of:

32 bytes: message hash h
32 bytes: signature component r
32 bytes: signature component s
32 bytes: public key x-coordinate qx
32 bytes: public key y-coordinate qy
```

https://eips.ethereum.org/EIPS/eip-7951#abi-for-p256verify-operation

ivokub: Yup, the input is already prehashed. So I guess the issue is that verifyPrehashed assumes that the prehashed input is not `0`? I'll check now exactly what is happening -- if std package assumes non-zero prehashed input (which imo is odd) then indeed we need to override the method.

ivokub: Yeah, it is odd from Zig std imo.

ivokub (approved):
Yep, your override makes sense (though I think Zig std is wrong here and we shouldn't have the override at all). Good to merge from my side, thanks!

PR #3946, opened 2026-09-09
perf(verifier-ray): Merkle Capping

Tabaie:
Verifier-side companion to https://github.com/LFDT-Lineth/lineth-monorepo/pull/3828


verifier-ray/test/pcs_test.zig
```diff
@@ -147,6 +177,42 @@ test "pcs verify cases from prover-ray vectors" {
     }
 }
 
+test "pcs rejects reordered input cap tables" {
```
ivokub, line 180: Could we add negative tests that exercise the actual cap-to-root authentication here? This test only reorders the input-cap tables and is rejected by the structural validation before the reconstructed cap is compared with its committed root.

In particular, please use the non-empty `normal_flow` fixture to independently mutate:

- an input-cap node;
- a revealed input-cap row;
- a running-cap node;
- a running-cap auxiliary digest.

Each mutation should make `pcs.verify` return `MerkleProofInvalid`. As a mutation check, I removed the root comparison from `MerkleCap.authenticate`, and the current Zig test suite still passed, so the security-critical binding is not presently protected against regression.

verifier-ray/src/crypto/merkle.zig
```diff
@@ -167,6 +229,29 @@ pub const InputTreeOpening = struct {
         return step.ancestor;
     }
 
+    /// Authenticates the portion of a sparse row-opening branch below an
+    /// already-authenticated input-tree frontier.
+    pub fn authenticateToCap(self: InputTreeOpening, idx: usize, frontier: []const poseidon2.Digest) Error!void {
+        const depth = try frontierDepth(frontier);
+        const height = self.leaves.len;
+        if (height == 0 or self.siblings.len != height - 1 - depth or depth >= height) {
```
ivokub: `height - 1 - depth` is evaluated before `depth >= height`, so a malformed direct call can underflow and trap instead of returning `InvalidFrontier`. The current `pcs.verify` path bounds the depth before reaching this helper, so I do not see a verifier soundness issue, but could we reorder this as `height == 0 or depth >= height or self.siblings.len != height - 1 - depth`?

verifier-ray/testdata/generate/fri/main.go
```diff
@@ -664,6 +784,39 @@ func writePCSFixtures() error {
...
+	fmt.Fprintln(&out, "pub const InputCapTableData = struct { size_log2: u8, rows: []const RowOpeningData };")
+	fmt.Fprintln(&out, "pub const InputCapData = struct { nodes: []const [8]u32, tables: []const InputCapTableData };")
+	fmt.Fprintln(&out, "pub const MerkleCapData = struct { nodes: []const [8]u32, aux: []const ?[8]u32 };")
+	fmt.Fprintln(&out, "pub const BranchData = struct { leaf: [8]u32, siblings: []const [8]u32 };")
+	fmt.Fprintln(&out, "pub const FriProofData = struct { round_roots: []const [8]u32, round_caps: []const MerkleCapData, final_poly: []const [6]u32, running_queries: []const []const BranchData };")
+	fmt.Fprintln(&out, "pub const OpeningProofData = struct { input_queries: []const []const InputTreeOpeningData, input_caps: []const InputCapData, fri_proof: FriProofData };")
+	fmt.Fprintln(&out, "pub const PcsCase = struct { name: []const u8, system: pcs.System, roots: []const [8]u32, entry_claims: []const []const [6]u32, zeta: [6]u32, fold_alphas: []const [6]u32, query_positions: []const usize, proof: OpeningProofData, expect_verify_error: []const u8 = \"\" };\n")
```
ivokub: This makes `go test ./...` fail in `testdata/generate` because `fmt.Fprintln` is given a string that already ends in a newline. Could we remove the trailing `\n` and use a separate `fmt.Fprintln(&out)` if the blank line is intentional?

verifier-ray/bench/bench_pcs/build.zig
```diff
@@ -0,0 +1,46 @@
...
+    r5_options.addOption(bool, "disable_accelerators", disable_accelerators);
+    verifier_mod.addOptions("r5_config", r5_options);
+    if (!disable_accelerators) verifier_mod.addImport("lineth_accelerators", accel_mod);
+    const profiling_options = b.addOptions();
+    profiling_options.addOption(bool, "is_enabled", false);
+    profiling_options.addOption(bool, "is_r5_marks", true);
+    verifier_mod.addOptions("profiling_config", profiling_options);
+
+    const fixture_mod = b.addModule("large_pcs_fixture", .{
+        .root_source_file = b.path("../../zig-out/bench/large-pcs.zig"),
```
ivokub, line 28: `LARGE_PCS_BENCH_OUT` is configurable in the Makefile, but this import always uses the default path. Consequently, `make LARGE_PCS_BENCH_OUT=/tmp/custom.zig profile-large-pcs` generates the custom file and then builds against this different path. Could we either propagate the configured path into the Zig build or remove the ineffective override?

verifier-ray/bench/bench_pcs/run.go
```diff
@@ -0,0 +1,132 @@
...
+
+var (
+	markRE  = regexp.MustCompile(`VERIFIER-MARK\s+([0-9]+)\s+([0-9]+)`)
+	cycleRE = regexp.MustCompile(`clock cycle: ([0-9]+)`)
+)
+
+type marker struct{ cycle, value uint64 }
+
+func main() {
+	zkcBin := "zkc"
```
ivokub: Could this use the module-pinned `go tool zkc` mechanism? On the current checkout, the globally installed `zkc` is older and rejects the arithmetization’s `#[global]` annotation, whereas the version pinned by `arithmetization/go.mod` compiles the benchmark guest. Using the pinned tool would make this target reproducible.

verifier-ray/bench/bench_pcs/run.go
```diff
@@ -0,0 +1,132 @@
...
+	if waitErr != nil {
+		fmt.Fprintf(os.Stderr, "zkc exec failed: %v\nlast output:\n%s\n", waitErr, strings.Join(tail, "\n"))
+		os.Exit(1)
+	}
+	start, startOK := markers[10]
+	end, endOK := markers[11]
+	if !startOK || !endOK || end.cycle < start.cycle {
+		fatal(fmt.Errorf("PCS markers missing or out of order; last output:\n%s", strings.Join(tail, "\n")))
+	}
+	fmt.Printf("pcs.verify RISC-V cycles: %d\n", end.cycle-start.cycle)
```
ivokub, line 96: Is this benchmark intended to report overall verifier cycles or the number of hashes performed? It currently reports only the cycle difference between the two markers. If hash count is the intended metric, it looks like a profiling counter still needs to be enabled or added; otherwise the PR description should describe this as a cycle benchmark.

verifier-ray/src/query/pcs.zig
```diff
@@ -502,6 +532,178 @@ pub fn deriveChallenges(
...
+    recon: Reconstructed(system),
+    info: InputCapInfo(system),
+    cap: InputCap,
+    root: poseidon2.Digest,
+    aux_storage: []?poseidon2.Digest,
+) Error![]const poseidon2.Digest {
+    const expected_nodes = @as(usize, 1) << @intCast(info.depth);
+    if (info.depth == 0) {
+        if (cap.nodes.len != 0 or cap.tables.len != 0) return Error.InvalidCap;
+        return &[_]poseidon2.Digest{root};
```
YaoJGalteland: The function returns `Error![]const poseidon2.Digest`.

If `info.depth == 0`, `root` is a runtime value, so returning a slice over an array literal built from `root` would point to temporary storage in the current call frame. That slice would dangle after the function returns.

If `info.depth != 0`, returns `cap.nodes`, which is proof-owned storage and outlives the call, so the two return paths have different lifetime semantics and only one is sound.

While this doesn’t seem to be an issue in the current call graph, because `authenticateInputCap` is only called when `info.depth != 0`.

verifier-ray/src/crypto/merkle.zig
YaoJGalteland: The three conditions are in the wrong order.

I think it should be:

```zig
if (height == 0 or depth >= height or self.siblings.len != height - 1 - depth)
```

The guards need to come first because:

* `height == 0` protects `height - 1` from underflow.
* `depth >= height` protects `height - 1 - depth` from underflow. Once `height >= 1`, `height - 1` is safe, but subtracting `depth` still underflows whenever `depth > height - 1`, which is exactly `depth >= height`.
* Only after those two checks is `height - 1 - depth` guaranteed to be well-defined.

So by the time the third condition is evaluated, both subtractions are safe.

ivokub: Duplicate of https://github.com/LFDT-Lineth/lineth-monorepo/pull/3946#discussion_r4004049925. Fixed in 6b2c610c336a8fb34cc0c67e6a32c42034e6146c

verifier-ray/bench/bench_pcs/run.go
ivokub: Let it be for now, I think we should reconsider the whole benchmarks later etc.

verifier-ray/src/query/pcs.zig
ivokub: Fixed in cb73e6543aca0ffbf38630036539a501cd43a130

verifier-ray/src/crypto/merkle.zig
YaoJGalteland: This looks like dead code, no any callers.

`InputTreeOpening.RecoverRoot` is test-only as well.

There are currently three `RecoverRoot` functions in prover-ray, which is a bit confusing, but only one seems to be live:

* `MerkleCap.RecoverRoot`, live
* `Branch.RecoverRoot`, test-only
* `InputTreeOpening.RecoverRoot`, test-only

Would it make sense to remove the two test-only ones, or at least rename/limit them so it’s clearer which path is actually used?

ivokub: Yup, lets remove the test-only paths. I'll double check if the tests are covered already by the merkle cap full test, in that case there is no need to keep the dead code around.

ivokub: I'll push a commit soon

ivokub: Pushed both on prover side and here

ivokub: Fixed as well.

YaoJGalteland (approved):

PR #3949, opened 2026-09-09
fix(coordinator): preserve Amsterdam execution payload fields

nadeemb53:
The coordinator was sending incomplete Amsterdam block data to the RISC-V prover: `slotNumber` was missing and `blockAccessList` was always empty.

This PR reads the full block and its access list from Besu and includes them in the prover request. It checks that the block, access list, and execution witness belong together, so data from different blocks cannot be mixed. It also updates the request schema to accept `slotNumber`, while keeping it optional for older blocks.

Tests cover the new fields, older blocks, and missing or mismatched data. This fixes how the coordinator builds the request; acceptance by a real prover and full E2E proving still need validation.


jvm-libs/linea/web3j-extensions/build.gradle
```diff
@@ -25,7 +25,7 @@ dependencies {
   testImplementation "com.fasterxml.jackson.core:jackson-annotations"
   testImplementation "com.fasterxml.jackson.core:jackson-databind"
   testImplementation "org.wiremock:wiremock:${libs.versions.wiremock.get()}"
-  testImplementation project(":jvm-libs:linea:besu-rlp-and-mappers")
+  implementation project(":jvm-libs:linea:besu-rlp-and-mappers")
```
jonesho: nit: please move this up at line20

jvm-libs/linea/web3j-extensions/src/test/kotlin/linea/web3j/ethapi/Web3jExecutionPayloadClientTest.kt
```diff
@@ -0,0 +1,139 @@
...
+
+  @Test
+  fun `preserves withdrawals blob gas and slot zero with a valid empty BAL`() {
+    val header = BlockHeaderBuilder.fromHeader(block.header).blockHeaderFunctions(MainnetBlockHeaderFunctions())
+      .slotNumber(0L).balHash(Hash.EMPTY_BAL_HASH).blobGasUsed(131072L).excessBlobGas(BlobGas.of(262144L))
+      .buildBlockHeader()
+    val withdrawal = Withdrawal(UInt64.valueOf(3), UInt64.valueOf(7), Address.ZERO, GWei.of(32))
+    val modified = Block(header, BlockBody(block.body.transactions, emptyList(), Optional.of(listOf(withdrawal))))
+    stub("debug_getRawBlock", RLP.encodeBlock(modified).encodeHex())
+    stub("debug_getRawBlockAccessList", "0xc0")
```
jonesho: Question: does "0xc0" mean empty BAL list in the protocol? i.e. debug_getRawBlockAccessList will return 0xc0 if BAL is empty?

jonesho (dismissed):

jvm-libs/linea/web3j-extensions/src/test/kotlin/linea/web3j/ethapi/Web3jExecutionPayloadClientTest.kt
nadeemb53: That's right, you can check the spec here: [EIP-7928](https://eips.ethereum.org/EIPS/eip-7928)

jonesho (dismissed):

coordinator/clients/prover-client/src/test/kotlin/lineth/coordinator/clients/prover/RiscvProofRequestDtoMapperTest.kt
```diff
@@ -27,6 +28,17 @@ import kotlin.time.Instant
  */
 class RiscvProofRequestDtoMapperTest {
 
+  @Test
+  fun `serialized payload includes slot zero and omits absent slot`() {
```
fluentcrafter: What's special about slot 0?

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClientFactory.kt
```diff
@@ -87,5 +93,46 @@ fun createEthApiClient(
...
+  require(
+    block.parentBeaconBlockRoot == null ||
+      block.parentBeaconBlockRoot.decodeHex().contentEquals(ByteArray(32)),
+  ) {
+    "Nonzero parent beacon block root: block=${block.number}"
+  }
+  if (block.slotNumber != null) {
+    require(
+      block.withdrawalsRoot != null && block.requestsHash != null &&
+        block.parentBeaconBlockRoot != null && block.blobGasUsedRaw != "0" && block.excessBlobGasRaw != "0",
```
jonesho, line 133: nit: Why not also required `block.blockAccessListHash != null`? and why check `block.blobGasUsedRaw and `block.excessBlobGasRaw` instead of `block.blobGasUsed` and `blob.excessBlobGas`?

jonesho (dismissed):
LGTM, but better get one more review, probably from @gauravahuja ?

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClientFactory.kt
nadeemb53: `blockAccessListHash` is already checked in `BlockData`. It requires both fields together, so checking here would be redundant.
Web3j’s numeric getters return zero even when the field is missing. The raw getters return `"0"` for missing fields versus `"0x0"` for an explicit zero, so this checks that the fields are actually present.

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jExecutionPayloadClient.kt
```diff
@@ -0,0 +1,46 @@
...
+          "Block access list hash does not match header for block ${block.number}"
+        }
+        blockAccessList
+      }
+    }
+    return blockAccessList.thenApply(block::toExecutionPayload)
+  }
+
+  private fun getBlockAccessList(blockHash: String): SafeFuture<ByteArray> =
+    Request("debug_getRawBlockAccessList", listOf(blockHash), web3jService, RawDataResponse::class.java)
```
Filter94: Can we instead use `engine_getPayloadBodiesByRangeV2`? Debug API tends to be less stable and less standardized, thus less reliable

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClient.kt
```diff
@@ -118,15 +120,21 @@ class Web3jEthApiClient(
       .requestAsync { it.transactionCount.toULong() }
 
   override fun ethFindBlockByNumberFullTxs(blockParameter: BlockParameter): SafeFuture<Block?> {
-    return web3jClient
```
Filter94, line 121: I believe this means that we should contribute to Web3j to support the new block fields

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClientFactory.kt
```diff
@@ -87,5 +93,46 @@ fun createEthApiClient(
...
+    web3jService,
+    requestRetryConfig,
+    vertx,
+    stopRetriesOnErrorPredicate,
+    blockValidator,
+  )
+}
+
+/** Validate L2 policy before dropping unsupported Ethereum fields from the shared block model. */
+fun validateLinethBlock(block: EthBlockExtended.Block) {
```
Filter94, line 107: So it's used only in the test?

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jExecutionPayloadClient.kt
nadeemb53: In that case `engine_getPayloadBodiesByHashV2`would be better since we already have the block hash and want the BAL for that exact block.

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClientFactory.kt
nadeemb53: It’s used in the coordinator too, [ConflationAppV2](https://github.com/LFDT-Lineth/lineth-monorepo/pull/3949/changes#diff-2a938086c8746beadf1333c6f342c7620dd4801f17e730c07e370c4448bddd89R82) passes it in as `blockValidator`. It checks the RPC response before we map it to our domain model, so we don’t drop unsupported values.

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClient.kt
nadeemb53: I have opened an issue for this: https://github.com/LFDT-web3j/web3j/issues/2313

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClientFactory.kt
```diff
@@ -87,5 +93,46 @@ fun createEthApiClient(
...
+  require(
+    block.requestsHash == null ||
+      block.requestsHash!!.decodeHex().contentEquals(Hash.EMPTY_REQUESTS_HASH.bytes.toArray()),
+  ) {
+    "Execution requests are not supported: block=${block.number}"
+  }
+  require(block.blobGasUsed == BigInteger.ZERO && block.excessBlobGas == BigInteger.ZERO) {
+    "Blob gas is not supported: block=${block.number}"
+  }
+  require(
```
Filter94, line 124: So when we decide to add support for the parentBeaconBlockRoot, we'll have to make changes to Coordinator?

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClientFactory.kt
Filter94, line 133: Why `block.blobGasUsedRaw != "0" && block.excessBlobGasRaw != "0"` have to be true?

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClientFactory.kt
nadeemb53: Yes. When we do add the support, we’ll need to preserve the actual root through the model and mappings, as well as relax this check.

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClientFactory.kt
nadeemb53: This is just to check that both fields are present. Web3j returns "0" from the raw getters when a field is missing, but "0x0" when it’s actually set to zero. The numeric getters return zero either way, so the earlier check wouldn’t catch missing fields.

jvm-libs/linea/web3j-extensions/src/main/kotlin/linea/web3j/ethapi/Web3jEthApiClientFactory.kt
Filter94: What I'm hinting is that maybe we can remove this check right away? What's the risk of it?

nadeemb53: I have removed the check and updated the model and mappings to preserve the root.

Filter94 (approved):

PR #3940, opened 2026-09-07
refactor(prover-ray): refactor bus message

Soleimani193:
This PR implements a refactor of the bus message to prepare for the removal of Hooking from Fiat Shamir. Instead of hooking the bus coins to a prior Fiat-Shamir state, the coins are generated directly in an early round and then used in the bus compiler. This avoids manipulating the fist-shamir state via hooking.

Hooking has caused the following bugs in the prover

 AdvanceRound samples every coin on the round from the overwritten state and never restores the original state. Two bugs follow.

1. The transcript is discarded, so Fiat-Shamir stopped binding it.

FS-state is reverted to the state at the seed round, and any transcript between the seed round and seeded Coins is discarded.

2. Seeding is per round, so any coin in the round, including those that are not related to the message bus, is seeded.

### Checklist

* [ ] I wrote new tests for my new core changes.
* [ ] I have successfully ran tests, style checker and build against my new changes locally.
* [ ] If this change is deployed to any environment (including Devnet), E2E test coverage exists or is included in this
  PR.
* [ ] I have informed the team of any breaking changes if there are any.

---

> [!NOTE]
> **High Risk**
> Changes how cross-shard bus challenges and γ/contributions bind into the Fiat–Shamir transcript; mistakes in round layout or contribution timing could desynchronize shards or weaken soundness.
> 
> **Overview**
> Refactors message-bus **α/β** challenge generation so shards no longer **override Fiat–Shamir via pre-sampling hooks**. Bus coins are declared on the round immediately after round 0; with **`CompileOptions.SharedRandomness`**, **γ** is a round-0 public input absorbed into the transcript before those coins are sampled, and every bus column must live on that coin round (compile-time panic otherwise).
> 
> Shared-randomness **contributions** move from a prover action + FS hook to **lazy cells** on the coin round that hash the round’s PCS commitment (after commit actions run), with the verifier still checking published contribution limbs. **`Compile`** now gets **α/β** from `registerSharedRandomness` and places the result round after the latest of coin and participant rounds.
> 
> Tests drop hook/drive helpers in favor of **`Prove`/`Verify`**, add round-placement and **preflight-derived γ** integration coverage, and remove the old hook-centric shared-randomness tests. **`minimal_elf`** import path is updated to `prover-ray/zkcdriver/minimal-elf`.
> 
[...]

PR #3959, opened 2026-09-10
fix(verifier-ray): remove fiat-shamir hooking from verifier

Soleimani193:
This PR replicates the prover's behavior and removes the Fiat-Shamir hook.  The coins for messageBus are generated via the standard Fiat-Shamir path. As the data in round 0 is the same across all shards, Fiat-Shamir derives the same coins to be used in the message bus.

### Checklist

  PR.

---

> [!NOTE]
> **High Risk**
> Changes how Fiat-Shamir coins and cross-shard shared-randomness contributions are replayed and checked—core proof verification behavior that must stay byte-exact with the prover.
> 
> **Overview**
> Aligns **verifier-ray** with the updated **prover-ray** message-bus shared-randomness model by dropping transcript **Fiat-Shamir hooking** and tightening the contribution check.
> 
> **Transcript / coins:** γ is no longer injected via `PreSamplingHooks` / `Transcript.setState`. Codegen stops emitting `shared_randomness_coin_round` and `shared_randomness_gamma_refs` on `protocol.Spec`, and `replayWithTranscript` derives message-bus coins on the normal path after round messages (including round-0 γ cells) are absorbed—matching the PR’s claim that identical round-0 data yields identical coins across shards.
> 
> **Shared-randomness sub-verifier:** The contribution checker no longer builds a Poseidon2 sponge over every prior committed round. It hashes **only the message-bus coin round’s PCS commitment** (or zeros when that round has no commitment), then compares the multiset hash to the public-input contribution limbs. Go codegen and Zig templates switch from a `rounds[]` list to a single `commitment_round`.
> 
> Tests and fixture builders are updated for the new round layout and expectations; **prover-ray** is bumped in `go.mod`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 00498d8006d926924e2d261f44a89a49d6731414. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


verifier-ray/codegen/shared_randomness.go
```diff
@@ -83,12 +78,9 @@ func BuildSharedRandomnessSystem(sys *wiop.System) (SharedRandomnessSystem, erro
...
 	}
 
-	out.Rounds = make([]SharedRandomnessRound, coinRound.ID)
-	for i := 0; i < coinRound.ID; i++ {
-		out.Rounds[i] = SharedRandomnessRound{
-			RoundIndex:    i,
-			HasCommitment: sys.Rounds[i].HasCommitment,
-		}
+	out.CommitmentRound = CommitmentRoundCtx{
+		RoundIndex:    coinRound.ID,
```
ivokub, line 82: Stale reference on the unchanged doc lines just above: the `BuildSharedRandomnessSystem` doc still says the round "can never drift from the round the prover's `[messagebus.SharedRandomnessContributionAssigner]` ran on," but that type was deleted in #3940 — the contribution is now resolved by lazy cells, not a prover action. The underlying guarantee (the round is read off the registered checker action) is still correct; only the named anchor is stale. Suggest:

```go
// checker was registered on is read directly off the action via
// [wiop.Round.ID], so it can never drift from the round whose PCS commitment
// the prover hashes into its contribution.
```

ivokub (commented):
One leftover outside the diff, in `verifier-ray/src/crypto/fiat_shamir.zig` (~lines 49–56, `Transcript.setState`): its doc still describes it as "matching `Runtime.SetFSState` inside a `Round.PreSamplingHooks` entry" — both deleted on the prover side by #3950, and this PR removes the gamma-splice in `replayWithTranscript` that was its only call site. I grepped the whole verifier-ray tree at this head: `Transcript.setState` has **no callers** — the only references are its own definition and the internal `self.hasher.setState(state)` delegation (the `MDHasher.setState` it forwards to is a separate, still-used primitive).

Since it's now dead code whose only purpose was the hooking path this stack removes, I'd suggest deleting the method rather than rewording the doc — keeping the transcript's mutable surface minimal for future audits:

```zig
// delete setState (fiat_shamir.zig:49-56) entirely
```

If you'd rather keep it as a general primitive, at minimum drop the `Runtime.SetFSState`/`PreSamplingHooks` references from the doc, which point at APIs that no longer exist.

gusiri (approved):

PR #3960, opened 2026-09-10
feat(arithmetization): expose guest output and its hash

### LINKED ISSUE #3958: Expose guest output and its hash
🔐 🔐 `Should this be a security advisory instead?` 🔐 🔐

## Description
Write a clear and concise description of what you want to happen. Add any considered drawbacks.

## Motivation
This section should indicate what value we are receiving from it.

## Tasks
- [ ] Medium-level task that needs to be implemented for this feature request to be considered complete.

## Acceptance criteria
- [ ] List tests that need to be implemented or checks to consider this task completed. 

## Risks
- [ ] These are risks that may prevent completion of the task as well as areas to pay special attention to

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

letypequividelespoubelles:
This PR implements issue(s) #

### Checklist

  PR.


arithmetization/src/main/riscv/main.zkc
```diff
@@ -71,13 +71,12 @@ fn main<registers, ram, keccak_state, poseidon2_state>() {
...
 
-    // -- PREDECODING PROOF SECTION -- 
+    // -- PREDECODING PROOF SECTION --
     // parameters: clock_cycle, base, pc, exec_end, output_write_address
     // interpreter!(0, base, entry_point, exec_end, 0)
-    // -- PREDECODING PROOF SECTION END-- 
+    // -- PREDECODING PROOF SECTION END--
 
-    // -- NO PREDECODING PROOF -- 
+    // -- Interpreting the riscV transpiled program --
```
amkCha: Maybe it was done automatically, can you put the old comment back ?

arithmetization/src/main/riscv/interpreter.zkc
```diff
@@ -223,15 +223,30 @@ fn interpreter<registers, ram, keccak_state, poseidon2_state>(clock_cycle:u32, b
         // interpreter!(clock_cycle + 1, base, new_pc, exec_end, new_output_write_address)
         // -- PREDECODING PROOF SECTION END--
 
-        // -- NO PREDECODING PROOF SECTION --
+        // -- Interpreting the riscV transpiled program --
```
amkCha: same here

amkCha (approved):

arithmetization/src/main/lib/io/write_output.zkc
```diff
@@ -3,10 +3,12 @@ include "../../riscv/ram/read.zkc"
 
 // Copy `size` bytes of guest RAM starting at byte address `offset` into the public
 // `guest_output` buffer at `output_write_address`, returning the advanced cursor.
-fn write_output<ram>(offset:Address, size:OutputSize, output_write_address:OutputAddress) -> (new_output_write_address:OutputAddress) {
+fn write_output<ram, guest_output_tmp>(offset:Address, size:OutputSize, output_write_address:OutputAddress) -> (new_output_write_address:OutputAddress) {
```
DavePearce, line 6: <strike>You don't need to include `guest_output_tmp` here within the `<...>`.  In fact, it should cause an compilation error and I'm surprised it doesn't.</strike>

Sorry, I see.  Its the hash which is an output.

DavePearce: Ok, so we could eliminate `guest_output_tmp` if we could read from WOMs.  We actually could enable that I think, so might be worth considering.

DavePearce: Also, there are other ways we could address this to avoid yet another scratch memory.  For example, if the `write_output` was always given in one go, you could compute the hash directly and write it.

DavePearce (approved):
Ok, I approved it.  But, overall, this is adding considerable overhead that we should optimise away.  There are various ways we could achieve that (I outlined just one in my comment).  A little bit of help from the guest program side would go a long way.

arithmetization/src/main/lib/io/write_output.zkc
letypequividelespoubelles: yes, absolutely for reading the WOM. Same, if the output is in one go, there are other imporvment (like rm `output_write_address` ...)

PR #3964, opened 2026-09-10
ci(ci): restrict dependency cache writes to main

nadeemb53:
PR builds create separate dependency caches that compete with main's Gradle cache. This PR lets other branches reuse caches while only non-PR runs on `main` save them, reducing cache eviction and repeat downloads during the Maven Central IP ban.

### Changes

- Apply this cache policy to Go, pnpm, Zig, Hardhat, native libraries, Gradle, CodeQL and golangci-lint. Keep existing cache opt-outs.
- Run Java/Kotlin CodeQL only on main, including weekly and manual main scans. Keep PR analysis for other languages.
- Let Go restore main's cache when dependency files change.
- Fix permissions so native crypto libraries can be restored into `/usr/local`.
- Add Spotless formatter dependencies to the main Gradle cache seed.


.github/actions/setup-go/action.yml
```diff
@@ -0,0 +1,48 @@
...
+          go env GOMODCACHE
+          go env GOCACHE
+          echo CACHE_PATHS
+        } >> "$GITHUB_OUTPUT"
+    - uses: actions/cache/restore@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+      id: restore
+      if: ${{ steps.cache.outcome == 'success' && hashFiles(inputs.cache-dependency-path) != '' }}
+      with:
+        path: ${{ steps.cache.outputs.paths }}
+        key: ${{ steps.cache.outputs.prefix }}${{ steps.go.outputs.go-version }}-${{ hashFiles(inputs.cache-dependency-path) }}
```
jonesho, line 48: Why not used `restore-keys` with `${{ steps.cache.outputs.prefix }}${{ steps.go.outputs.go-version }}-`?

nadeemb53: makes sense

.github/actions/cache/action.yml
```diff
@@ -0,0 +1,37 @@
+name: 'Shared cache'
+description: 'Restore caches on every ref; save only on main and when the caller permits it.'
+inputs:
+  path:
+    description: 'Paths to cache'
```
fluentcrafter: Path(s) - is it single or list of paths?

.github/actions/cache/action.yml
```diff
@@ -0,0 +1,37 @@
...
+  cache-hit:
+    description: 'Whether the exact key was restored'
+    value: ${{ steps.writable.outputs.cache-hit || steps.readonly.outputs.cache-hit }}
+runs:
+  using: composite
+  steps:
+    # Use the upstream post-job save so the cache includes the completed build.
+    - uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+      id: writable
+      if: ${{ github.ref == 'refs/heads/main' && !startsWith(github.event_name, 'pull_request') && inputs.cache-save == 'true' }}
```
fluentcrafter, line 26: `!startsWith(github.event_name, 'pull_request')` does it also exclude commits on PRs?

fluentcrafter (approved):

.github/actions/cache/action.yml
nadeemb53: Yes, new commits on a PR still run as `pull_request`, so they only restore caches. The saving happens after merge on `main`. The extra check also covers `pull_request_target`, where the ref can be `main`.

PR #3977, opened 2026-09-15
perf(verifier-ray): make vanishing bucket loop runtime to cut code size

YaoJGalteland:
## Summary

- Switch `verifyBucket`'s `bucket` parameter and the inner loops over `vanishings` and `positions` from `comptime` to runtime.
- Update `bench_compress/build.zig` to wire the required `r5_config` options so the benchmark compiles under the new setup.

## Problem

The comptime path monomorphized a distinct `verifyBucket` instantiation per bucket and `inline for` unrolled every vanishing constraint into straight-line code. On the real RISC-V arithmetization this produced ~85 instantiations totalling ~6.0 MiB of ~7.9 MiB `.text`, which:

1. Pushed the guest ELF past `elf_to_json`'s 2,000,000-record pre-decoding cap.
2. Dominated interpreted instruction-fetch cost in `zkc`.

## Results

R5 guest, `--release=small -Dstrip=true -Dr5=true`, measured on the same commit tree with only this PR's two files differing:

|              | before    | after     | change            |
| ------------ | --------- | --------- | ----------------- |
| `.text`      | 8,260,356 B (7.88 MiB) | 944,820 B (0.90 MiB)  | **-7,315,536 B (-88.6%)** |
| `.rodata`    | 19,738,516 B (18.82 MiB) | 20,199,904 B (19.26 MiB) | +461,388 B (+2.3%) |
| Total binary | 28,024,120 B (26.73 MiB) | 21,160,784 B (20.18 MiB) | **-6,863,336 B (-24.5%)** |

`.text` shrinks **8.7x**. In pre-decoding records (`.text / 4`):

| | records | vs 2,000,000 cap |
| --- | --- | --- |
| before | 2,065,089 | **over by 65,089** |
| after | 236,205 | under by 1,763,795 |

So `elf_to_json` now succeeds without needing `ELF2JSON_MAX_DECODED_RECORDS` to be raised; on the base commit it fails outright with `decoded program would have 2065089 records (cap 2000000)`.

`.rodata` grows slightly because the expression/bucket metadata is now consumed as data at runtime instead of being partly folded into unrolled code. That trade is strongly favourable: `.rodata` is not pre-decoded into instruction records, so it does not count against the cap.

### Instantiation counts

`verifyBucket` disappears as a distinct symbol entirely (it is small enough to inline into `verify`). The remaining per-module instantiations are cheap:

| function | instantiations | bytes |
| --- | --- | --- |
| `verify` | 1 | 150,028 |
| `evalExpr` | 118 | 134,856 |
| `cancellationAtPoint` | 6 | 2,428 |
| `evalLagrangeSelector` | 6 | 1,984 |
| `powModuleSize` | 6 | 1,768 |
| `staticRootPower` | 15 | 1,444 |
| **total `query.vanishing`** | **152** | **292,508 B (0.28 MiB)** |

The whole `query.vanishing` namespace is now 21x smaller than the old `verifyBucket` symbols alone. The count is still driven by `comptime module` (~100 modules), but each copy is a compact switch-driven loop (~1.1 KiB) rather than an unrolled expression tree, because `evalExpr` was already runtime-indexed.

## Why this is safe

Nothing requires `bucket` to be comptime:
- `Bucket`/`Vanishing` are plain data (ratio, slice of expression indices, claim offset).
- Expression indices are already consumed at runtime by `evalExpr`.
- `module` and `static_n` stay `comptime` — they legitimately fold static-size exponentiation and root-of-unity lookups at compile time.

This mirrors a precedent already documented in the same file: `evalExpr`/`evalOp` were previously de-comptime'd on `expr_index` for exactly this reason (see the note there describing a stack overflow caused by "the sheer number of monomorphized node-specific function bodies"). This PR applies the same treatment one level up.

## Test plan

- [x] `zig build --release=small -Dstrip=true -Dr5=true` — `.text` drops 8.7x (numbers above).
- [x] `elf_to_json` on the resulting ELF — now under the 2M-record cap, no override needed.
- [x] `zig build test` — full verifier-ray Zig test suite passes.
- [x] `zig fmt --check` clean on both changed files.

---

> [!NOTE]
> **Medium Risk**
> Changes verifier hot-path control flow from fully unrolled comptime to runtime loops over proof metadata; semantics should match but this is core PLONK vanishing/quotient checking on the zkVM guest.
> 
> **Overview**
> **Shrinks the R5 verifier guest** by stopping comptime monomorphization of vanishing-bucket verification: `bucket`, inner `vanishings`, and `cancelled_positions` loops are now **runtime** (while `module` / `static_n` stay comptime), matching the earlier `evalExpr` pattern. That removes dozens of huge unrolled `verifyBucket` copies and reportedly cuts `.text` by ~88%, bringing the ELF back under `elf_to_json`'s 2M pre-decode record cap.
> 
> **Bench build fix:** `bench_compress/build.zig` now supplies required **`r5_config`** (`is_r5_zkvm`, optional `-Ddisable-accelerators` for software Poseidon2) and only links **`lineth_accelerators`** when accelerators are enabled.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3e8a1da37590b72f8ad4fbdc8ddc0fb7f516e012. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ivokub (approved):
LGTM!

PR #3986, opened 2026-09-17
fix(jvm-libs): fix VertxHttpJsonRpcClient hangs on error paths in Vertx 5

gauravahuja:
## Context 
# Vertx 5: How context-aware futures cause vertx http client test flakiness

**TL;DR:** Vertx 5 changed how futures dispatch their completion callbacks. When a future completes
on an event-loop thread but a listener was registered from a different thread (e.g. a JUnit test
thread), Vertx 5 schedules the callback via Netty's task queue. On loaded CI Linux runners this
scheduling can stall for 10–15+ seconds, causing tests to time out. The fix: register the
completion callback from *inside* the event-loop chain, not from outside it.

---

## What changed in Vertx 5

In Vertx 4, `Future` was a simple promise object with no thread affiliation. Completion callbacks
were always called synchronously by whoever completed the future — simple, predictable.

In Vertx 5, every `Future` carries a `ContextInternal` field. When a callback is registered on a
context-aware future, Vertx checks three conditions before firing it
([`FutureBase.emitResult`, line 81](https://github.com/eclipse-vertx/vert.x/blob/5.1.5/vertx-core/src/main/java/io/vertx/core/impl/future/FutureBase.java#L81)):

```java
// FutureBase.java
if (context != null && !context.isRunningOnContext()) {
    context.execute(new EmitResultTask<>(...));   // PATH 2 — cross-context dispatch
} else if (context != null) {
    emitResult(context, result, cause, listener); // PATH 3 — inline, with context set
} else {
    signalComplete(result, cause, listener);      // PATH 1 — direct synchronous call
}
```

`isRunningOnContext()` ([`ContextInternal.java`, line 217](https://github.com/eclipse-vertx/vert.x/blob/5.1.5/vertx-core/src/main/java/io/vertx/core/internal/ContextInternal.java#L217))
requires two conditions:

```java
default boolean isRunningOnContext() {
    return current() == this  // thread-local: the context currently dispatching
        && inThread();        // eventLoop.inEventLoop()
}
```

The HTTP client assigns a `streamCtx` to each request. The connection pool creates a *separate*
context object `Y` on the same Netty thread. They share the same thread but are **different
objects**, so `current() == streamCtx` is false when `Y` is dispatching — triggering path 2.

---

## What path 2 actually does

Path 2 calls `context.execute(task)`
([`ContextBase.java`, line 94](https://github.com/eclipse-vertx/vert.x/blob/5.1.5/vertx-core/src/main/java/io/vertx/core/impl/ContextBase.java#L94)):

```java
public final void execute(Runnable task) {
    if (executor().inThread()) {
        task.run();               // already on the right thread → run INLINE
    } else {
        executor().execute(task); // wrong thread → schedule via Netty
    }
}
```

`executor().execute()` for an event-loop context (E1) delegates to
[`EventLoopExecutor.execute()`](https://github.com/eclipse-vertx/vert.x/blob/5.1.5/vertx-core/src/main/java/io/vertx/core/impl/EventLoopExecutor.java#L39),
which calls `nettyEventLoop.execute(task)` — adding the task to Netty's task queue and writing an
`eventfd` to wake the event loop.

This gives two very different outcomes depending on *who* registers the listener:

| Who registers the listener | `inThread()` in `context.execute` | Outcome |
|---|---|---|
| A callback already running on E1 (path 2 triggered from E1) | `true` → `task.run()` **inline** | Fast, no scheduling ✓ |
| The JUnit test thread | `false` → `nettyEventLoop.execute()` | Requires OS to schedule E1 ✗ |

---

## The race: when does `emitResult` fire from the JUnit thread

`FutureImpl.addListener()`
([line 82](https://github.com/eclipse-vertx/vert.x/blob/5.1.5/vertx-core/src/main/java/io/vertx/core/impl/future/FutureImpl.java#L82))
has a subtle behaviour: if the future is already complete when `.onComplete()` is called, it fires
`emitResult` **immediately on the calling thread**:

```java
public void addListener(Completable<? super T> listener) {
    Object v;
    synchronized (this) {
        v = value;
        if (v == null) { /* future pending — store listener, return */ }
    }
    // future already complete — fire emitResult on THIS thread right now
    emitResult((T) v, null, listener);
}
```

There are two cases:

**Case A — listener added before the future completes (slow requests)**

The JUnit thread wires up `.onComplete` while the HTTP request is still in flight. When E1
later completes the future, `emitResult` fires from E1. Path 2 is triggered but
`context.execute()` sees `inThread() = true` → runs the `EmitResultTask` **inline**. Fast. ✓

**Case B — future completes before the listener is added (WireMock fast path)**

WireMock runs in-process and responds in microseconds. The entire HTTP chain (request → response
→ body → chain completion) can finish on E1 *before* the JUnit thread even reaches
`.onComplete { bridge.handle(ar) }`. When the JUnit thread then calls `.onComplete`, the future
is already done, so `emitResult` fires immediately **from the JUnit thread**, hits path 2 with
`inThread() = false`, and hands off to Netty's task queue.

```
JUnit thread                          E1 (Netty event loop)
──────────────────────────────────────────────────────────────────
httpClient.request()               →  [initiates TCP connect]
  → chain created (pending)
                                      WireMock responds immediately
                                      request → response → body read
                                      → entire chain completes
                                      E1 finishes → enters epoll_wait
chain.onComplete { bridge.handle(ar) }
  chain already complete!
  addListener → emitResult from JUnit thread
    C1.context = streamCtx
    current() = JUnit thread  ← not E1
    isRunningOnContext() = false → path 2
    context.execute(task):
      executor().inThread() = false
      → nettyEventLoop.execute(task)  ← Netty schedules, writes eventfd

bridge.future().toSafeFuture()
failsWithin(14s) → [BLOCKED]
                                      E1 on CI: overloaded, not scheduled
                                      eventfd written but E1 gets no CPU
                                      ...14 seconds pass...
                                      E1 wakes, bridge.handle → too late
TEST FAILS: timeout
```

---

## Why it only failed on CI

On **macOS** (local dev): kqueue wakes E1 promptly after the eventfd write → passes.

On **CI Linux** (shared runners, many parallel jobs): E1's Netty thread competes for CPU with
dozens of other test threads. The OS can take 10–15+ seconds to schedule it → test times out.

The failure was *not* flaky in the usual sense. The timing relationship is structural: our
`makeRequest` is always called from the test thread (not the event loop), so whenever WireMock
responds fast enough, we always hit Case B, and CI Linux runners are always loaded enough to
trigger the stall.

## Summary

- In Vertx 5, the `captureTime(requestFuture.toCompletableFuture()).toVertxFuture()` round-trip in `VertxHttpJsonRpcClient.makeRequest` loses native context propagation on error paths, causing callers that observe the result via `.toSafeFuture()` to hang indefinitely. The fix registers the metrics timer as a side effect only (no `.toVertxFuture()`) and returns the original `requestFuture` directly.
- Fixes `makesRequest_connectionFailure` test by using `localhost` with a guaranteed-closed port instead of the non-routable hostname `service-not-available`, giving an immediate "connection refused" rather than a DNS resolution timeout that outlasts the JUnit `@Timeout(15s)` in Vertx 5.

## Test plan

- [ ] `./gradlew :jvm-libs:generic:json-rpc:test` — all tests pass including `makesRequest_connectionFailure`, `makesRequest_502_response`, `makesRequest_malFormattedJsonResponse`
- [ ] `./gradlew :jvm-libs:generic:json-rpc:spotlessCheck` — no formatting violations

---

> [!NOTE]
> **Medium Risk**
> Changes async completion and error propagation for all outbound JSON-RPC HTTP calls; incorrect bridging could affect metrics timing or leave futures incomplete, but scope is localized to this client.
> 
> **Overview**
> Fixes **indefinite hangs** when JSON-RPC HTTP clients observe `makeRequest` failures via `toSafeFuture()` under **Vertx 5** (notably on CI Linux/epoll).
> 
> `makeRequest` now completes a **context-free `Promise` bridge** from inside the Vert.x event-loop chain (including HTTP error and `recover` paths), instead of relying on futures that can dispatch completion through `context.execute()` when a non-event-loop thread attaches listeners. Request metrics still use `captureTime` on the bridge’s `CompletableFuture`, but the returned future is **`bridge.future()`** directly—**no `toVertxFuture()` round-trip**. Response parsing logs success/failure inline in `handleResponse` rather than in a trailing `.andThen`.
> 
> The **connection-failure** test targets **`localhost` on a closed ephemeral port** for immediate “connection refused” instead of a slow DNS failure on `service-not-available`, so it stays within the 15s JUnit timeout.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7e678eead7c53c1964f7bc4004b5a8e6028e2a0a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jvm-libs/generic/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/client/VertxHttpJsonRpcClient.kt
```diff
@@ -65,59 +65,87 @@ class VertxHttpJsonRpcClient(
   ): Future<Result<JsonRpcSuccessResponse, JsonRpcErrorResponse>> {
```
Filter94, line 65: IIUC, this fix is generic and applicable to any Vertx's `Future`. I suggest we structure the code the same way, introduce a generic function here and if we see the same issue anywhere else, we could as well move it to vertx extensions. WDYT?

jvm-libs/generic/json-rpc/src/main/kotlin/net/consensys/linea/jsonrpc/client/VertxHttpJsonRpcClient.kt
```diff
@@ -65,59 +65,87 @@ class VertxHttpJsonRpcClient(
...
-                  message =
-                  "HTTP errorCode=${response.statusCode()}, message=${response.statusMessage()}",
-                  httpStatusCode = response.statusCode(),
-                ),
+    // Bridge via a context-free Promise so any caller thread can observe the result without
+    // going through Vertx's context.execute() dispatch. Promise.promise() produces a context-free
+    // PromiseImpl whose listeners always fire via signalComplete — direct, no scheduler.
+    val bridge = Promise.promise<Result<JsonRpcSuccessResponse, JsonRpcErrorResponse>>()
+
+    // Complete the bridge from INSIDE the chain (running on E1), never from a .onComplete
```
fluentcrafter, line 73: These comment shall be adjustment. For example, it references "running on E1" what will E1 mean in 2 months from now after we merge?

Likewise, below we have references to "Path 2" ...

PR #3988, opened 2026-09-18
Ci/l2 execution guest pipeline

Filter94:
This PR implements issue(s) #

### Checklist

  PR.


.github/workflows/riscv-guests-zkc-interpreter-run.yml
```diff
@@ -105,20 +190,60 @@ jobs:
         env:
           ZKC_REF: ${{ inputs.zkc-ref }}
         run: make -C arithmetization prepare-zkc
-      # Execute the guest under zkc in fast mode (--fast): a runnability check, so it skips tracing
-      # to stay light. Tracing and constraint checking are measured weekly instead — see
-      # arithmetization-weekly-zkc-metrics.yml. Uses exec's default
-      # INPUT (the committed sample stateless input). A trap/failure still exits non-zero here.
-      # ZKC_REF is step-scoped above, so elf-exec does not apply the override a second time.
+      # The downloaded ELF was built with the production keccak configuration.
       - name: Run l2-execution guest under zkc (fast, accel keccak) and check guest_output
```
ivokub: The old step ran `make -C riscv-guests/l2-execution zkc-smoke-exec`, which goes through `test/zkc_reference_runner.zig`'s `smoke_suite` — it runs the guest under ZkC **and compares the output against the host machine**, failing on disagreement. The new step calls `make -f arithmetization/src/test/Makefile elf-exec` directly on the prebuilt ELF, which only proves the ELF runs without trapping; nothing checks `guest_output` against anything anymore. The step name still says "and check guest_output", which is now misleading.

If the intent is that this gate only proves runnability (the header comment says as much), please either drop "and check guest_output" from the step name and explicitly accept losing the per-PR host-vs-ZkC equivalence check on the committed sample input — or restore the comparison (e.g. keep `zkc-smoke-exec` here, or add a host-output diff step). Same applies to the `elf-trace` step below.

.github/workflows/riscv-guests-zkc-interpreter-run.yml
```diff
@@ -76,22 +78,105 @@ permissions:
...
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: rollup-accel-${{ github.sha }}
+          path: riscv-guests/rollup/zig-out
+          if-no-files-found: error
+          retention-days: 1
+
+  host-tests:
+    name: Guest unit tests (host)
+    needs: build-guest-artifacts
```
ivokub, line 121: `host-tests` and `reference-tests` both declare `needs: build-guest-artifacts`, but neither consumes the guest ELF — they run `make test` / `make reference-test` natively. This means the host unit tests wait for the full production riscv64 ELF build (which compiles Constantine for riscv64 and is likely the slowest step in the DAG) before starting, so per-PR wall-clock may be *longer* than the old fully-parallel layout even though total runner-minutes drop. If the fail-fast on a broken guest build isn't a deliberate goal for the host jobs, drop `needs:` from those two and keep it only on the zkc jobs.

.github/workflows/riscv-guests-zkc-interpreter-run.yml
```diff
@@ -76,22 +78,105 @@ permissions:
...
+        run: make -C riscv-guests/l2-execution compile KECCAK_ACCEL=true
+
+      - name: Build rollup guest and sample input
+        run: make -C riscv-guests/rollup compile require-input KECCAK_ACCEL=true
+
+      - name: Upload prepared l2-execution ELF
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: l2-execution-accel-${{ github.sha }}
+          path: riscv-guests/l2-execution/zig-out/bin/evm_execution_guest
```
ivokub, line 105: `l2-execution-accel-${{ github.sha }}` is stable within a run (good), but a re-run of the same SHA uploads a second copy under the same name, and re-running only a failed downstream job reuses the existing (possibly stale) artifact. Consider `overwrite: true` (upload-artifact v4+) or appending `${{ github.run_attempt }}` if you want re-runs to reflect freshly-built artifacts.

.github/workflows/riscv-guests-zkc-interpreter-run.yml
```diff
@@ -76,22 +78,105 @@ permissions:
...
+    if: github.event_name != 'pull_request' || github.event.pull_request.draft == false
+    runs-on: gha-lfdt-lineth-ss-ubuntu-24-amd64-med
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          submodules: false
+
+      - name: Setup RISC-V guests environment
+        uses: ./.github/actions/setup-riscv-guests
```
ivokub, line 131: `host-tests` and `reference-tests` each run `setup-riscv-guests` and then rebuild the host Constantine archive (`libconstantine.host.a`) that the l2-execution host tools link. You could build the host archive once in `build-guest-artifacts` (which already has the full toolchain), upload it, and have both test jobs restore it — same artifact-sharing pattern as the guest ELFs.

ivokub (commented):
CI consolidation looks directionally right — building the guest ELFs once and fanning out, plus dropping the blst/mcl build path, are real wins. A few correctness issues inline, and two optimization notes that anchor to files outside this PR's diff:

**Cache the Nim/Constantine toolchain** (`.github/actions/setup-riscv-guests/action.yml:24`, the `Install constantine build deps` step — changed in the base branch, not this PR). `install-constantine-deps` downloads choosenim + Nim 2.2.10 and installs LLVM on every job, and the `guest-crypto-ctt` build then recompiles the Constantine static libs. This now runs in 4 jobs of this pipeline plus the other workflows that use this action. `~/.choosenim`, `~/.local/bin`, `~/.nimble` and the Constantine build output are all cacheable under a key hashing `NIM_VERSION`/`CHOOSENIM_VERSION` from `riscv-guests/Makefile` and the `guest-crypto-ctt` sources. Probably the single biggest remaining repeated cost.

**zig-cache key hashes the whole Makefile** (`.github/actions/setup-riscv-guests/action.yml:45`, the `Cache Zig compile cache` step — also base branch). Adding `riscv-guests/Makefile` to the `hashFiles(...)` key means a Nim/choosenim version bump (which lives in `install-constantine-deps`) invalidates the zig compile cache even though no zig build input changed. Consider moving the dep-install recipe to its own file, or hashing only the Makefile fragments that affect the zig build.

**Stale workflow names / runner tier** (inline anchor not possible — the lines aren't in the diff): `arithmetization-guest-programs-exec.yml:1` and `arithmetization-guest-programs-trace.yml:1` are still named `... l2-execution, Add in Zig and Blake in Rust` / `... l2-execution, Add in Zkc`, but the l2-execution steps were removed in this PR. Also, `arithmetization-guest-programs-trace.yml:43` still runs on `gha-lfdt-lineth-ss-ubuntu-24-amd64-xxl` even though the l2-execution trace was its only memory-heavy step — please confirm the remaining `make -C arithmetization zkc-compile-and-check` actually needs xxl, or drop to a smaller tier.

ivokub (changes requested):
See inline comments and reply

.github/workflows/riscv-guests-zkc-interpreter-run.yml
Filter94: I leave this as out of scope for this PR. The primary goal of this PR is to make the pipeline green to merge the precompiles PR

.github/workflows/riscv-guests-zkc-interpreter-run.yml
Filter94: This dependency is set on purpose to save ci resources. Currently if there's some compilation or build error very early on, it would fail 3 parallel jobs all at once (`build-guest-artifacts`, `host-tests`, `reference-tests`) wasting quite a lot of resources

ivokub (approved):

